PR #1, opened 2017-10-12
Migrate from the php site

MilanRusev:
Currently without the scripts and fetching eth price


lib/sanbase/model/latest_btc_wallet_data.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Sanbase.Model.LatestBtcWalletData do
+  use Ecto.Schema
+  import Ecto.Changeset
+  alias Sanbase.Model.LatestBtcWalletData
+
+
+  @primary_key{:address, :string, []}
+  schema "latest_btc_wallet_data" do
+    # field :address, :string
+    field :satoshi_balance, :float
```
valo: Are you sure this should be a `float` here? I think satoshis are not divisible: https://en.bitcoin.it/wiki/Satoshi_(unit)

lib/sanbase/model/latest_eth_wallet_data.ex
```diff
@@ -0,0 +1,24 @@
...
+
+
+  @primary_key{:address, :string, []}
+  schema "latest_eth_wallet_data" do
+    # field :address, :string
+    field :balance, :float
+    field :last_incoming, :naive_datetime
+    field :last_outgoing, :naive_datetime
+    field :tx_in, :float
+    field :tx_out, :float
```
valo: Should we use `decimal` for these values? I am a bit hesitant to use `float` to represent money amounts...

lib/sanbase_web/controllers/cashflow_controller.ex
```diff
@@ -0,0 +1,43 @@
...
+  def index(conn, _params) do
+    data = Cashflow.get_project_data
+
+    data = Enum.group_by(data,
+      fn(el) ->
+        %{project: el.project, coinmarketcap: el.coinmarketcap}
+      end,
+      fn(el) ->
+        el.wallet_data
+      end)
```
valo: There is a shorthand for this:

```elixir
data = Cashflow.get_project_data
|> Enum.group_by(&Map.take(&1, [:project, :coinmarketcap]), &(&1.wallet_data))
```

I see there are other transformations on this data alter, which can be extracted into separate functions and added to the pipeline. This should make this code more readable.

lib/sanbase_web/controllers/cashflow_controller.ex
```diff
@@ -0,0 +1,43 @@
...
+
+    data = Enum.map(data,
+      fn({k, v}) ->
+        market_cap_usd = if (k.coinmarketcap !== nil), do: k.coinmarketcap.market_cap_usd, else: nil
+        wallet_data = Enum.filter(v, fn(x) -> x !== nil end)
```
valo: Are you sure there could be `nil` values in the `v` here?

lib/sanbase_web/controllers/cashflow_controller.ex
```diff
@@ -0,0 +1,43 @@
...
+        balance = Enum.reduce(wallet_data, 0, fn(x, acc) -> x.balance + acc end)
+        %{
+          name: k.project.name,
+          ticker: k.project.ticker,
+          logo_url: k.project.logo_url,
+          market_cap_usd: market_cap_usd,
+          balance: balance,
+          wallets: wallet_data
+        }
+      end)
```
valo: I would definitely try to split into a separate function. I think the `SanbaseWeb.CashflowView` makes sense to be used here, as this data is formatted for the presentation layer here. The upside of moving it to a separate function in the view is that we will be able to write unit tests for this function without touching the DB.

app/pages/cashflow/index.js
```diff
@@ -0,0 +1,198 @@
...
+    <script dangerouslySetInnerHTML={{ __html: `
+      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+      })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+      ga('create', 'UA-100571693-1', 'auto');
+      ga('send', 'pageview');
+     `}} />
+  </Head>
```
valo: We should be able to extract these headers into separate components, but this is something that can be done later. For now I think we can merge this as it is and then create separate PRs to split this UI into smaller pieces that can be reused across the views.

app/pages/cashflow/index.js
```diff
@@ -0,0 +1,198 @@
...
+                          )
+                        })}
+                          </tbody>
+                      </table>
+                  </div>
+              </div>
+          </div>
+      </div>
+  </div>
+  <script dangerouslySetInnerHTML={{ __html: `
```
valo: We should remove any uses of `dangerouslySetInnerHTML`, but that can be done in a subsequent PRs.

lib/sanbase/model/latest_btc_wallet_data.ex
MilanRusev: I agree, we may need to change the database types. Currently I made the same DB schema as the one for project transparency. We should refactor the DB datatypes for both projects.

lib/sanbase/model/latest_eth_wallet_data.ex
MilanRusev: The same thing for the DB schema being the the same as project transparency

lib/sanbase_web/controllers/cashflow_controller.ex
MilanRusev: Yes, if there is no wallet data for a project. The sql query is with a left join.

lib/sanbase/model/latest_btc_wallet_data.ex
valo: In that case we should probably keep it as it is and make subsequent PRs, which are fixing the schema

valo:
I have some more comments on this, but let's merge it as it is and fix them later.

PR #2, opened 2017-10-16
Import price data from coinmarketcap into influxDB

valo:
- [ ] API calls to coinmarketcap to fetch the data
- [ ] Throttle the calls to a given rate per minute, so that we don't get banned https://github.com/grempe/ex_rated
- [ ] Fix the price fetching, as filtering by timestamp does not work ATM
- [ ] Add a supervised process, which updates the prices on a regular basis (each 5-10 min)
- [ ] Update the readme with instructions how to install InfluxDB locally to run the scraping scripts


valo:
Seems like the 5 min resolution data will be a bit hard to scrape, as we will need to make a gazillion requests to coinmarketcap. Let's initially scrape the historical daily candles and provide API and pandas bindings for that. That should be enough for low frequency (intra-week and intra-month trading analysis)

valo:
After thinking more about this, it seems that providing intra day pricing data will be quite hard. Let's first start importing daily candles from coin market cap and expose this to the public. From there we can investigate how hard it will be to scrape intraday candles and may be sell this dataset.

PR #4, opened 2017-10-17
Replicate the spreadsheet

MilanRusev:
**!!!NOTE: Breaks compatibility with projecttransparency db schema!!!**


valo:
I think we should probably use only integer IDs as primary keys and have unique keys for the current fields used as primary keys. This is going to make things much easier! Let's discuss tomorrow how hard it will be to switch to that...

MilanRusev:
Note: breaks compatibility with projecttransparency db schema

valo:
> !!!NOTE: Breaks compatibility with projecttransparency db schema!!!

This is not a problem. When we migrate the PT to the new schema we will use an API exposed by this service to fetch the data in the format we need there.

lib/sanbase/external_services/ico_spreadsheet.ex
```diff
@@ -0,0 +1,193 @@
...
+  use Tesla
+
+  plug Tesla.Middleware.BaseUrl, "https://sheets.googleapis.com/v4/spreadsheets/"
+  plug Tesla.Middleware.JSON
+  plug Tesla.Middleware.Compression
+  plug Tesla.Middleware.Logger
+
+  # TODO: spreadsheet_id & api key should be fetched from env variables
+  @spreadsheet_id "INSERT_SPREADSHEET_ID"
+  @api_key "INSERT_API_KEY"
```
valo: These can be set as `Mix.Config` settings: https://hexdocs.pm/mix/Mix.Config.html

We should think if we want to keep this code as it is here or have a Mix task for it, as it seems we will run it from time to time. If we have a mix task for it we can also pass these as parameters. Example of a mix task: https://github.com/santiment/sanbase2/pull/10/files#diff-199e45473d5aa147353f97be387e33fbR1

MilanRusev: Yes, I created a mix task that runs it: https://github.com/santiment/sanbase2/pull/4/commits/422292ffff8e9a8d94d496f33d3c7abaab7390b1#diff-22c8b9d01749c470d5144cd2b25919cf. Good idea to pass them as parameters to the task, I didn't think about that. Currently only the project list is passed but it will be fetched from the database.

lib/mix/tasks/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  defp import(document_id, api_key, project_names) do
+    {:ok, _started} = Application.ensure_all_started(:sanbase)
+
+    res = IcoSpreadsheet.get_project_data(document_id, api_key, project_names)
+    |> Enum.reverse()
+    |> Enum.uniq_by(fn row -> row.project_name end)
+    |> Enum.reverse()
+
+    IO.inspect res
```
valo: Please remove such debug statements.

lib/mix/tasks/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,54 @@
...
+
+    Sanbase.DbScripts.ImportIcoSpreadsheet.import(res)
+
```
valo: Do we need such extra empty lines?

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,131 @@
...
+
+    fill_ico(project, ico_spreadsheet_row)
+    |> Ecto.Changeset.put_assoc(:currencies, ensure_currencies(ico_spreadsheet_row.ico_currencies))
+    |> Ecto.Changeset.put_assoc(:cap_currency, ensure_currency(ico_spreadsheet_row.cap_currency))
+    |> Repo.insert_or_update!()
+  end
+
+  defp fill_project(ico_spreadsheet_row) do
+    ensure_project(ico_spreadsheet_row.project_name)
+    |> IO.inspect()
```
valo: Please remove such print statements.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,131 @@
...
+    Repo.get_by(Project, name: project_name)
+    |> Repo.preload([:eth_addresses, :btc_addresses, :market_segment, :infrastructure, {:ico, [:currencies, :cap_currency]}])
+    |> case do
+      result = %Project{} -> result
+      nil -> %Project{}
+    end
+  end
+
+  defp ensure_market_segment(market_segment_name) do
+    if(!is_nil(market_segment_name)) do
```
valo: Can we use pattern matching in the arguments of the function to avoid the if? that should make the function more readable. The same applies to other places around the code too.

lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
```diff
@@ -0,0 +1,200 @@
...
+    value =
+    case column do
+      c when c in [:ico_start_date, :ico_end_date] ->
+        parse_date(value)
+      c when c in [:tokens_issued_at_ico] ->
+        parse_int(value)
+      c when c in [:tokens_sold_at_ico, :usd_btc_icoend, :funds_raised_btc, :usd_eth_icoend, :minimal_cap_amount, :maximal_cap_amount] ->
+        parse_decimal(value)
+      # c when c in [] ->
+      #   parse_boolean(value)
```
valo: I think we should remove commented out code. There is no real value in keep it around. but it makes reading the code a bit harder

lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
```diff
@@ -0,0 +1,200 @@
...
+          IO.write("parse_decimal error: ")
+          IO.inspect value
+          nil
+      end
+    else
+      value
+    end
+  end
+
+  defp parse_boolean(value) do
```
valo: I think this function can be removed. I don't see it used anywhere

MilanRusev: We removed the boolean columns from the import but that may change in the future, so we may keep it.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -75,34 +74,30 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
...
-      Repo.get_by(Infrastructure, code: infrastructure_code)
-      |> case do
-        result = %Infrastructure{} -> result
-        nil ->
-          %Infrastructure{}
-          |> Infrastructure.changeset(%{code: infrastructure_code})
-      end
-    else
-      nil
+  defp ensure_market_segment(_), do: nil
```
valo: There is even easier way to write this: `defp ensure_market_segment(nil), do: nil` and then have the function, which handles all non-`nil` cases ;-)

lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
```diff
@@ -84,34 +82,30 @@ defmodule Sanbase.ExternalServices.IcoSpreadsheet do
...
-    else
-      value
+  defp parse_int(value), do: value
+
+  defp parse_decimal(value) when is_binary(value) do
+    case Decimal.parse(value) do
+      {:ok, result} -> result
+      _ -> #TODO: return error
+        IO.write("parse_decimal error: ")
+        IO.inspect value
```
valo: Generally we should be using `Logger` and not `IO` for code which we deploy to production. For this one it may be OK as we will be running it as a mix task.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,124 @@
...
+      team_token_wallet: ico_spreadsheet_row.team_token_wallet
+      })
+  end
+
+  defp fill_ico(project, ico_spreadsheet_row) do
+    ensure_ico(project)
+    |> Ico.changeset(%{
+      project_id: project.id,
+      start_date: ico_spreadsheet_row.ico_start_date,
+      end_date: ico_spreadsheet_row.ico_start_date,
```
valo: I think there is a typo here and should be `ico_spreadsheet_row.ico_end_date`

lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
```diff
@@ -0,0 +1,188 @@
...
+
+    case value do
+      v when v in ["yes", "no"] ->
+        Map.put(parsed_value_row, :cap_currency, nil)
+      _ -> parsed_value_row
+    end
+  end
+
+  defp remove_nils(list) when is_list(list) do
+    Enum.filter(list, &(!is_nil(&1)))
```
valo: A bit more readable version: `Enum.reject(list, &is_nil/1)` No need to change this, as it is a minor thing.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -0,0 +1,124 @@
...
+      comments: ico_spreadsheet_row.comments,
+      })
+  end
+
+  defp ensure_project(project_name) do
+    Repo.get_by(Project, name: project_name)
+    |> Repo.preload([:eth_addresses, :btc_addresses, :market_segment, :infrastructure, {:ico, [:currencies, :cap_currency]}])
+    |> case do
+      result = %Project{} -> result
+      nil -> %Project{}
```
valo: Shouldn't we use `project_name` parameter when initializing a new project?

lib/sanbase/model/ico.ex
```diff
@@ -0,0 +1,33 @@
...
+    field :tokens_sold_at_ico, :decimal
+    field :funds_raised_btc, :decimal
+    field :usd_btc_icoend, :decimal
+    field :usd_eth_icoend, :decimal
+    field :minimal_cap_amount, :decimal
+    field :maximal_cap_amount, :decimal
+    field :main_contract_address, :string
+    field :comments, :string
+    belongs_to :project, Project
+    belongs_to :cap_currency, Currency, on_replace: :nilify
```
valo: I get an error about this column. I don't see it being added in any migration. Are you sure you haven't forget to commit some migration change?

valo:
I am pretty sure the `structure.sql` is not up to date with the migrations defined in this PR. @MilanRusev, please do this sequence of commands to verify that and fix any issues you might find:

```
mix ecto.drop
mix ecto.create
mix ecto.migrate
```

Check the changes you see in `structure.sql` and also try to rerun the import. I am pretty sure there will be some problems, like missing columns.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
MilanRusev: The `project_name` is filled with the `changeset()` function inside `fill_project()` (the same is with `ensure_ico()`).

Is it a good idea to fill it partially with `%Project{} |> Project.changeset(%{name: project_name)` (or `%Ico{} |> Ico.changeset(%{project_id: project.id)` ) - the `changeset()` may potentially fail for validating another column that hasn't been supplied?

lib/sanbase/model/ico.ex
MilanRusev: Hmm, the cap_currency_id column is defined in this migration: https://github.com/santiment/sanbase2/blob/spreadsheet-db-schema/priv/repo/migrations/20171017125741_create_icos.exs

Maybe the error is connected with the `structure.sql` comment below?

MilanRusev:
> I am pretty sure the structure.sql is not up to date with the migrations defined in this PR. @MilanRusev, please do this sequence of commands to verify that and fix any issues you might find:
> 
> mix ecto.drop
> mix ecto.create
> mix ecto.migrate
> Check the changes you see in structure.sql and also try to rerun the import. I am pretty sure there will be some problems, like missing columns.

I deleted and regenerated `structure.sql` and it generated the same file. What goes wrong when you run the same commands yourself?

valo:
May be the problem was that I haven't integrated the latest versions of the branches. Now it seems to be working better. I am merging this as it is now. We will need to make some corrections to the other branch, which scrapes data from etherscan.

PR #6, opened 2017-10-18
Migration: drop non-null constraint on project.coinmarketcap_id

tzanko-matev:


priv/repo/migrations/20171018120438_remove_nonnull_constraint_for_coinmarketcap_id.exs
```diff
@@ -0,0 +1,16 @@
...
+    alter table("project") do
+      modify :coinmarketcap_id, :text, null: true
+    end
+  end
+
+  def down do
+    alter table("project") do
+      modify :coinmarketcap_id, :text, null: false
+    end
+  end
```
valo: You just need a `change` method here. The framework can figure out how to revert the operation automatically:

```elixir
def change do
  alter table("project") do
    modify :coinmarketcap_id, :text, null: false
  end
end
```

tzanko-matev: I tried, it can't. The modify method is not-reversible: https://hexdocs.pm/ecto/Ecto.Migration.html#modify/3

valo: You are right. Didn't read this part.

priv/repo/migrations/20171018120438_remove_nonnull_constraint_for_coinmarketcap_id.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Sanbase.Repo.Migrations.RemoveNonnullConstraintForCoinmarketcapId do
+  use Ecto.Migration
+
+  def up do
+    alter table("project") do
+      modify :coinmarketcap_id, :text, null: true
```
valo: I think we want to make the column nullable, so it should be `null: true` here. `up` is called when the migrations are run forward.

PR #8, opened 2017-10-18
Redirect to cashflow when opening the index page

valo:


tzanko-matev:
Is the following needed?
```js
  <Layout>
    <p>Your current shopping list:</p>
    <TodoList items={props.items}/>
  </Layout>
```

valo:
It is not. I'll remove it

valo:
@tzanko-matev fixed. please have another look

tzanko-matev:
@valo good to go

PR #9, opened 2017-10-18
Tidy up the frontend app

valo:
DRY up the views using layouts and components.


app/components/projects-table/index.js
```diff
@@ -0,0 +1,84 @@
...
+                            { txOut.toLocaleString('en-US') }
+                          </div>
+                        )
+                      })
+                    }
+                    </td>
+                  </tr>
+                )
+              })
+            }
```
valo: Can we split these maps into separate functions, which are named properly and this way make the whole template a bit shorter? I think is going to make this template a bit more readable. What do you think?

app/components/topbar/index.js
```diff
@@ -0,0 +1,18 @@
...
+  <div className="row topbar">
+    <div className="col-lg-6">
+      <div style={{'paddingTop': '24px', 'paddingLeft': '16px'}}>
+        <i className="material-icons">search</i>
+      </div>
+    </div>
+    <div className="col-lg-6">
+      <ul className="nav-right pull-right list-unstyled">
+        <li>
+          <span style={{'display': 'inline-block', 'paddingTop': '22px', 'paddingLeft': '24px', 'paddingRight': '24px', 'fontSize': '14px'}}>12.5 Ξ</span>
```
valo: Can we use styled jsx for the CSS? This is the standard way to handle CSS in the next.js bundle: https://github.com/zeit/styled-jsx It is even possible to split the CSS into a separate file and import it in the JS file and then use `styleName` to refer to a class. The nice thing is that the CSS is being scoped to the current component and won't impact any HTML elements outside of it.

app/pages/cashflow.js
```diff
@@ -0,0 +1,89 @@
...
+        wallets: [
+          {
+            last_outgoing: null,
+            balance: null,
+            tx_out: null
+          }
+        ]
+      }
+    ],
+    eth_price: 2
```
valo: Can we keep the API call here? We will need to use the API for the code we merge into master

app/components/projects-table/index.js
bobrovova: I agree with you. Done

app/components/topbar/index.js
```diff
@@ -0,0 +1,21 @@
+import styles from './index.scss'
+
+export default (props) => (
+  <div className="row topbar">
+    <style dangerouslySetInnerHTML={{ __html: styles }}></style>
+    <div className="col-lg-6">
+      <div className="search" style={{'paddingTop': '24px', 'paddingLeft': '16px'}}>
```
valo: Can we move this in-like styles to the css file of the component?

app/pages/roadmap.js
```diff
@@ -1,287 +1,214 @@
...
-        <div class="row topbar">
-            <div class="col-lg-6">
-                <div style="padding-top: 24px; padding-left: 16px;">
-                    <i class="material-icons">search</i>
+    <MainHead />
+    <SideMenu activeItem="roadmap"/>
+    <div className="container vert-stretch" id="main">
+      <Topbar />
+      <HeaderPage name="SANbase Roadmap" description="Please see our SANbase roadmap below. It is a living document; milestones may adjust." />
+      <style dangerouslySetInnerHTML={{ __html: styles }}></style>
```
valo: Why do we need to inject the css like this? Isn’t it possible to just use the classes in the ‘style’ object?

app/components/header-page/index.js
```diff
@@ -0,0 +1,12 @@
+import styles from './roadmap.scss'
```
valo: This import can't be working as this file is located into another folder

app/pages/cashflow.js
```diff
@@ -1,197 +1,98 @@
...
-                }
-            }
+      });
+    });
+  }
 
-        });
+  static async getInitialProps(){
+    //const res = await fetch(WEBSITE_URL + '/api/cashflow')
+    //const data = await res.json()
```
valo: We should be fetching the data from the server, so please enable this.

PR #22, opened 2017-11-06
Generate notifications for significant price changes

valo:
As a first version we will post the notifications on a slack channel


valo:
Let's rock & roll!

PR #23, opened 2017-11-09
Migrate cmc script

tzanko-matev:


lib/sanbase/external_services/coinmarketcap/rate_limiter.ex
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def sleep_algorithm ({:allow, count}) do
+    Process.sleep(@time_between_requests)
+    {:ok, count}
+  end
+
+  def sleep_algorithm ({:deny, _}) do
+    {:ok, {_,_,wait_period, _, _}} = Hammer.inspect_bucket(@bucket, @scale, @limit)
+    IO.puts("Denied!" <> to_string(wait_period))
```
valo: Let's use `Logger` instead of `IO.puts`. It will allow us to control to level of logging

lib/sanbase/external_services/coinmarketcap/ticker_server.ex
```diff
@@ -0,0 +1,60 @@
...
+    Process.send_after(self(), {:"$gen_cast", :sync}, update_interval)
+
+    {:noreply, state}
+  end
+
+  defp store_ticker(ticker) do
+    case Repo.get(LatestCoinmarketcapData, ticker.id) do
+      nil  -> %LatestCoinmarketcapData{id: ticker.id}
+      entry -> entry
+    end
```
valo: May be it will be worth it to extract this in a separate function?

lib/sanbase/external_services/coinmarketcap/rate_limiter.ex
valo: Also, isn't it the case that this log message will print quite often?

tzanko-matev:
I replaced logging with Logger and refactored the function.

PR #24, opened 2017-11-12
Migrate etherscan

tzanko-matev:


config/config.exs
```diff
@@ -32,6 +32,8 @@ config :hammer,
   backend: {Hammer.Backend.ETS, [expiry_ms: 60_000 * 60 * 4,
                                  cleanup_interval_ms: 60_000 * 10]}
 
+#config :tesla, adapter: :ibrowse
```
valo: Please, remove such commented out lines. There is no point to keep them in the code

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+  @average_block_time_ms 10000 #Actually it's close to 15s
+  @default_timespan_ms 30*24*60*60*1000 # 30 days
+
+  def start_link(_state) do
+    GenServer.start_link(__MODULE__, :ok)
+  end
+
+  def init(:ok) do
+    Logger.info fn ->
+      "Starting etherscan polling client"
```
valo: There is no need to have a function here. A string will be perfectly fine

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+
+      {:ok, %{update_interval_ms: update_interval_ms}}
+    else
+      :ignore
+    end
+  end
+
+  def handle_cast(:sync, %{update_interval_ms: update_interval_ms} = state) do
+    #1. Get current block number
+    endblock = Requests.get_latest_block_number()
```
valo: I am wondering if we should take the last block. What happens if there is a fork at the moment on the chain and the chain diverges and merges later? May be we should not look at the last X blocks, essentially waiting for several confirmations before we take a block. What do yo think?

lib/sanbase/external_services/coinmarketcap/ticker_server.ex
```diff
@@ -0,0 +1,65 @@
...
+  defp store_ticker(ticker) do
+    ticker.id
+    |> get_or_create_ticker()
+    |> LatestCoinmarketcapData.changeset(
+      %{market_cap_usd: ticker.market_cap_usd,
+	name: ticker.name,
+	price_usd: ticker.price_usd,
+	symbol: ticker.symbol,
+	update_time: NaiveDateTime.add(~N[1970-01-01 00:00:00],
+	  ticker.last_updated)
```
valo: I think there is a better function for this: `DateTime.from_unix!(ticker.last_updated)`

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+    startblock = endblock - Float.ceil(@default_timespan_ms/@average_block_time_ms)
+    #startblock = 0
+    
+    TrackedEth
+    |> Repo.all
+    |> Enum.each(&(fetch_and_store_async(&1, startblock, endblock)))
```
valo: I think it is a good idea to wait for the processes to end before continuing. My concern is that a given sync process night not be finished before the next update happens. Another approach will be to have the standard supervisor as it was before and use https://hexdocs.pm/elixir/Task.Supervisor.html#async_stream/6. If one of the jobs dies, the parent process will die and it will be restarted. May be I just don't understand the current way it is intended to work?

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+
+
+  defp convert_to_eth(wei) do
+    D.div(D.new(wei), D.new(1000000000000000000))
+  end
+
+  def fetch(address, startblock, endblock) do
+    changeset = %{
+      update_time: NaiveDateTime.utc_now,
+      balance: convert_to_eth(Balance.get(address).result) |> D.to_float(),
```
valo: Can't we store the balance as a `Decimal` in the DB?

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+        %{ 
+	  last_outgoing: NaiveDateTime.add(~N[1970-01-01 00:00:00], ts),
+	  tx_out: convert_to_eth(value) |> D.to_float()
+         })
+      nil -> changeset
+    end
+  end
+
+  def fetch_and_store_async(tracked_eth, startblock, endblock) do
+    #Logger.debug("fetch_and_store_async: #{tracked_eth.address}")
```
valo: Please, let's not keep such commented code

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -0,0 +1,107 @@
...
+      balance: convert_to_eth(Balance.get(address).result) |> D.to_float(),
+    }
+    
+    case Tx.get_last_outgoing_transaction(address, startblock, endblock) do
+      %Tx{timeStamp: ts, value: value}-> Map.merge(changeset,
+        %{ 
+	  last_outgoing: NaiveDateTime.add(~N[1970-01-01 00:00:00], ts),
+	  tx_out: convert_to_eth(value) |> D.to_float()
+         })
+      nil -> changeset
```
valo: Why do we return a changeset if there are no outgoing transactions? May be it will make sense to have some tests for this code. The tricky part will be how to write tests with a minimal amount of mocks. One approach could be to split the code which fetches data from external sources and the code which processes the incoming data. This way we would be able to test the processing code in isolation.

tzanko-matev: We need it in case the balance changed

lib/sanbase/external_services/etherscan/worker.ex
tzanko-matev: I can change it to have several confirmations. Ultimately I don't think it's a big deal.

valo: I am concerned that a transaction can appear on the latest block and then disappear. This is something that Dmitry also raised as a concern. I think these blocks are called `Uncles` and you can see here that they appear quite frequently: https://etherscan.io/blocks May be we can get away with having a small number of confirms, like 4, which is around 1 min and mitigate any problems like this?

lib/sanbase/external_services/etherscan/worker.ex
valo: How the balance changes without transactions?

config/dev.exs
```diff
@@ -65,3 +65,9 @@ config :ex_admin,
 if File.exists?("config/dev.secret.exs") do
   import_config "dev.secret.exs"
 end
+
+config :sanbase, Sanbase.ExternalServices.Etherscan.RateLimiter,
+  scale: 1000,
+  limit: 5,
+  time_between_requests: 1000
```
valo: I think the build is failing because this config is missing in the `test.exs` file. BTW, you can run the tests with `mix test` and most probably reproduce the failure.

lib/sanbase/external_services/etherscan/requests.ex
```diff
@@ -0,0 +1,160 @@
...
+
+  defmodule Balance do
+    defstruct [:status, :message, :result]
+    
+    defp get_query(address) do
+      [
+	module: "account",
+	action: "balance",
+	address: address,
+	tag: "latest",
```
valo: Please, fix the indentation here and also in several other places.

lib/sanbase/external_services/etherscan/requests.ex
```diff
@@ -0,0 +1,160 @@
...
+
+  alias Sanbase.ExternalServices.Etherscan.RateLimiter
+  alias Sanbase.ExternalServices.Etherscan.Requests
+  
+  plug RateLimiter.Tesla
+  plug Tesla.Middleware.BaseUrl, "https://api.etherscan.io/api"
+  plug Tesla.Middleware.Compression
+
+  plug Tesla.Middleware.Query, [
+    apikey: Keyword.get(config(), :apikey)
```
valo: I have the doubts that this value will be evaluated on compile time, which I am not sure will work for us...

tzanko-matev: I checked and it is runtime. Actually the config() function is called before every HTTP request

lib/sanbase/external_services/etherscan/requests.ex
```diff
@@ -0,0 +1,158 @@
...
+  ]
+  plug Tesla.Middleware.Logger
+
+  defmodule Balance do
```
valo: Is it possible to split these modules in separate files? Defining several modules in the same file can make finding the module definitions very hard. Also it will make the files shorter and easier to read.

lib/sanbase/external_services/etherscan/requests.ex
valo: Very strange. I tested it too and it seems to be evaluated on runtime. Some macro magic is going on here :-)

lib/sanbase/external_services/etherscan/internaltx_requests.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Sanbase.ExternalServices.Etherscan.Requests.InternalTx do
+
+  alias Sanbase.ExternalServices.Etherscan.Requests
+  alias __MODULE__
+
+  defstruct [:blockNumber,
+             :timeStamp,
+             :hash,
+	     :from,
```
valo: Whitespace issue here

lib/sanbase/external_services/etherscan/internaltx_requests.ex
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def get(address) do
+    Requests.get("/", query: get_query(address))
+    |> case do
+         %{status: 200, body: body} ->
+           response = Poison.decode!(body, as: %{result: [%InternalTx{}]})
+           response.result
+       end
+    end
```
valo: I think the indentation is not very correct here... We should try to integrate the auto formatter soon, so that we don't need to fix problems like this...

lib/sanbase/external_services/etherscan/rate_limiter.ex
```diff
@@ -0,0 +1,83 @@
...
+
+
+  def sleep_algorithm({bucket,_,_,tbr}, {:allow, count}) do
+    # Logger.debug fn ->
+    #   "Sleeping. Rate limiter: #{bucket}, delay: #{tbr}"
+    # end
+    Process.sleep(tbr)
+    # Logger.debug fn ->
+    #   "Executing request. Rate limiter: #{bucket}, count: #{count}"
+    # end
```
valo: Can we remove these commented out lines, please?

lib/sanbase/external_services/etherscan/rate_limiter.ex
```diff
@@ -0,0 +1,83 @@
...
+    Logger.info fn ->
+      "Rate limit exceeded. bucket: #{bucket}, wait_period: #{wait_period}"
+    end
+
+    Process.sleep(wait_period)
+    sleep_algorithm(state, Hammer.check_rate(bucket, scale, limit))
+  end
+
+  def handle_call(:wait, _, state) do
+    #Logger.debug "waiting"
```
valo: Some leftover commented code

lib/sanbase/model/latest_eth_wallet_data.ex
```diff
@@ -6,18 +6,18 @@ defmodule Sanbase.Model.LatestEthWalletData do
...
   schema "latest_eth_wallet_data" do
-    field :balance, :float
-    field :last_incoming, :naive_datetime
-    field :last_outgoing, :naive_datetime
-    field :tx_in, :float
-    field :tx_out, :float
-    field :update_time, :naive_datetime
+    field :balance, :decimal
+    field :last_incoming, :utc_datetime
+    field :last_outgoing, :utc_datetime
```
valo: What is the practical difference between `naive_datetime` and `uct_datetime`? I am just curious why you had to change it.

valo:
There is a problem with the current implementation of the scraping. Some of the projects we track are way beyond the first 200 projects on CMC and because of this they don't get their data updated. Here is a screenshot. I think we should be fetching the data for all projects in the `projects` table and not for the first 200 projects on CMC:

screenshot 2017-11-15 15 13 08

tzanko-matev:
@valo I noticed that too. Right now we use the API call https://api.coinmarketcap.com/v1/ticker, which gives data for many projects, but apparently not all. I'll have it make additional calls for the projects that are not in top 200

valo:
It is possible to call the API for a specific ticker or just start paginating the tickers until all the projects in out DB are covered.

tzanko-matev:
@valo It is possible, However I just increased the limit for how many projects we fetch. Now I set it to get 1000 projects. This should be sufficient.

lib/sanbase/external_services/etherscan/internaltx_requests.ex
tzanko-matev: This is the default indentation which emacs elixir mode does

PR #28, opened 2017-11-17
Simplified the etherescan scraping

valo:
We don't need a special Task supervisor to run the tasks in the
etherscan server. It is enough to trap exists, so that the supervisor
does not die when a Task dies and use Task.async_stream to run the
tasks. This will ensure that the tasks that die will just be ignored and
won't impact the other tasks.


valo:
Let's go!

PR #33, opened 2017-11-19
Scrape BTC prices from coinmarketcap

valo:
We are getting the BTC prices of all assets from CMC when we fetch the
USD prices, so we should start collecting these.

I refactored a little bit the InfluxDB storage, as we need to start
separating the concerns a little bit there. At the end of the day the
influx store should not really care if we store prices or something
else, but be an adapter for working with the influx store.


valo:
Looks good

PR #51, opened 2017-11-25
Parse project information from the CMC project page

valo:
Parse various information about a given project from the CMC page.


lib/sanbase/external_services/coinmarketcap/scraper.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  def fetch_project_page(coinmarketcap_id) do
+    %Tesla.Env{status: 200, body: body} = get("/#{coinmarketcap_id}/")
+
+    body
+  end
+
+  def parse_project_page(html, project_info) do
+    project_info
+    |> struct!(
```
IvanIvanoff: What is the reason to use `struct!` instead of `%ProjectInfo{name: name(html) ...}` ?

valo: The idea is to have a `ProjectInfo` structure, which is passed through all the methods and "enriched" with more info at each step.

IvanIvanoff: That is also equivalent to `%ProjectInfo{project_info | name: name ...}`, updating only the name. This way the code is more explicit about the types, which in my opinion is a desirable result.

lib/sanbase/model/project.ex
```diff
@@ -37,4 +38,12 @@ defmodule Sanbase.Model.Project do
     |> validate_required([:name])
     |> unique_constraint(:name)
   end
+
+  def initial_ico(%Project{id: id}) do
+    Ico
+    |> where([i], i.project_id == ^id)
+    |> order_by(asc: :start_date)
```
IvanIvanoff: You can use https://hexdocs.pm/ecto/Ecto.Query.html#first/2

lib/sanbase/external_services/coinmarketcap/scraper.ex
valo: I agree. I’ll change it

IvanIvanoff:
As this PR is touching coinmarketcap.ex I think it's a good time to move it inside external_services/coinmarketcap directory

valo:
Actually I think the pattern to have `Sanbase.Coinmarketcap` module, which is used as a facade in front of many `Fanbase.Coinmarketcap.*` modules is a nice one. This way we can split functionality and know that the main module is a kind of a public interface.

lib/sanbase/external_services/coinmarketcap.ex
```diff
@@ -70,6 +73,38 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
     Application.get_env(:sanbase, __MODULE__)
   end
 
+  defp fetch_project_data(project) do
+    fetch_project_info(project)
```
IvanIvanoff: Let's keep it this way for now, but I think it is not necessary to try to fetch project info every 5 minutes (the default CMC update interval). Could be reworked by starting two types of tasks - :sync_prices and :sync_project_info, which are handled by different `handle_cast`s

valo: The idea is that if there is no missing data, the project info won't be updated.

PR #57, opened 2017-11-29
Default column values for project & ico tables

MilanRusev:
* [x] ico.start_date, ico.end_date - current date
* [x] ico.cap_currency - ETH
* [x] project.infrastructure - ETH
    * [x] For any existing data: if "Infrastructure" is null then set it to "ETH"


lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
```diff
@@ -7,6 +7,7 @@ defmodule Sanbase.ExternalServices.IcoSpreadsheet do
   plug Tesla.Middleware.JSON
   plug Tesla.Middleware.Compression
   plug Tesla.Middleware.Logger
+  plug Tesla.Middleware.Timeout, timeout: 100000
```
valo: What turned out about this timeout? Is it needed or something else is the problem?

lib/sanbase/model/ico.ex
```diff
@@ -31,7 +31,6 @@ defmodule Sanbase.Model.Ico do
     |> cast(attrs, [:start_date, :end_date, :tokens_issued_at_ico, :tokens_sold_at_ico, :funds_raised_btc, :funds_raised_usd, :funds_raised_eth, :usd_btc_icoend, :usd_eth_icoend, :minimal_cap_amount, :maximal_cap_amount, :main_contract_address, :comments, :project_id, :cap_currency_id])
     |> calculate_funds_raised()
     |> validate_required([:project_id])
-    |> unique_constraint(:project_id)
```
valo: Why this constraint need to be removed?

lib/sanbase/model/model.ex
```diff
@@ -0,0 +1,53 @@
...
+    {:ok, currency} = Repo.transaction(fn ->
+      get_currency(currency_code)
+      |> case do
+        nil -> insert_currency!(currency_code)
+        currency -> currency
+      end
+    end)
+
+    currency
+  end
```
valo: I think we should move these functions to the `Currency` module. I am not sure `model/model.ex` makes much sense. It feels a bit like `model/utils.ex`, but I think we can do better than that and put these functions in the models they work with.

lib/sanbase/model/model.ex
```diff
@@ -0,0 +1,53 @@
...
+    {:ok, infrastructure} = Repo.transaction(fn ->
+      get_infrastructure(infrastructure_code)
+      |> case do
+        nil -> insert_infrastructure!(infrastructure_code)
+        infrastructure -> infrastructure
+      end
+    end)
+
+    infrastructure
+  end
```
valo: And these 3 functions seem to belong to the `Infrastructure` module. This will also remove the need to have the word `infrastructure` in their name. Something like `Infrastructure.get_or_insert` will be quite understandable.

lib/sanbase_web/admin/model/ico.ex
```diff
@@ -1,7 +1,47 @@
...
+    resource = resource
+    |> set_ico_cap_currency_default()
+    |> set_ico_start_date_default()
+    |> set_ico_end_date_default()
+
+    {conn, params, resource}
+  end
+
+  defp set_ico_cap_currency_default(%Ico{cap_currency_id: nil}=ico) do
+    currency = Model.get_currency("ETH")
```
valo: I think `Currency.get` will work great here. This is connected with refactoring the `Model` module.

priv/repo/seeds.exs
```diff
@@ -10,17 +10,27 @@
...
+    nil -> infrastructure_eth
+    _ -> Model.get_or_insert_infrastructure(infrastructure_code)
+  end
+
   %Project{ name: name,
 	    ticker: ticker,
 	    logo_url: logo_url,
-	    coinmarketcap_id: coinmarkecap_id
+	    coinmarketcap_id: coinmarkecap_id,
+      infrastructure: infrastructure
```
valo: Indentation issue here

lib/sanbase/model/project.ex
```diff
@@ -27,7 +27,7 @@ defmodule Sanbase.Model.Project do
     belongs_to :market_segment, MarketSegment, on_replace: :nilify
     belongs_to :infrastructure, Infrastructure, on_replace: :nilify
     belongs_to :latest_coinmarketcap_data, LatestCoinmarketcapData, foreign_key: :coinmarketcap_id, references: :coinmarketcap_id, type: :string, on_replace: :nilify
-    has_one :ico, Ico
+    has_many :ico, Ico
```
valo: Very interesting... When it is possible that a project will have several ICOs?

lib/sanbase_web/admin/model/ico.ex
```diff
@@ -1,7 +1,47 @@
...
 
+  alias Sanbase.Model
+  alias Sanbase.Model.Ico
+  alias Sanbase.Model.Currency
+
   register_resource Sanbase.Model.Ico do
 
+    controller do
+      # doc: https://hexdocs.pm/ex_admin/ExAdmin.Register.html#after_filter/2
+      after_filter :set_defaults, only: [:new]
```
valo: Good job finding this approach for fixing the defaults! I like it :)

lib/sanbase/external_services/ico_spreadsheet/ico_spreadsheet.ex
MilanRusev: I have no idea, it started working at some point. If it happens again I'll investigate it further

lib/sanbase/model/ico.ex
MilanRusev: Because the association project-ico should be one-to-many.

lib/sanbase/model/project.ex
MilanRusev: Potentially it is possible. It is part of the requirements from Garry & Maksim.

PR #58, opened 2017-11-29
Enchance/jest

Partysun:
- JEST (polyfill, babelrc and some gotchas with next and jest versions)
 - Updated CI config.
 - Updated package.json LICENCE to private, ver to semver
 - Super simple spec for header
 - New ENV var - NODE_ENV.
 - docker-compose in web app, for local dev workflow

### DIFF
--- a/app/__tests__/header.spec.js
+++ b/app/__tests__/header.spec.js
@@ -1,0 +1,18 @@
+/* eslint-env jest */
+import React from './../node_modules/react'
+import { mount } from 'enzyme'
+import renderer from 'react-test-renderer'
+import Header from '../components/header-page'
+
+describe('Header', () => {
+  it('Header should render correct name', () => {
+    const app = mount(<Header name='check' />)
+    expect(app.find('h1').text()).toEqual('check')
+  })
+
+  it('Header should render correctly', () => {
+    const component = renderer.create(<Header />)
+    const tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
+  })
+})
--- a/app/components/main-head/index.js
+++ b/app/components/main-head/index.js
@@ -2,7 +2,7 @@ import Head from 'next/head'
 
 export default (props) => (
   <Head>
-    <meta charset='utf-8' />
+    <meta charSet='utf-8' />
     <title>SANBase</title>
     <meta name='viewport' content='initial-scale=1.0, width=device-width, shrink-to-fit=no' />
     <link rel='shortcut icon' href='/static/cashflow/img/favicon.png' />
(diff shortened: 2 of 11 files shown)


app/__tests__/header.spec.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-env jest */
```
valo: Is it possible to put the specs in the same folder as the component? I am testing this approach for one other project and I think it works very good because it is very easy to find the tests of the component. You have everything related to the component in a single folder. What do you think?

app/components/main-head/index.js
```diff
@@ -18,6 +18,7 @@ export default (props) => (
     <script src='https://www.kryogenix.org/code/browser/sorttable/sorttable.js' />
     <script src='https://cdn.datatables.net/1.10.15/js/jquery.dataTables.min.js' />
     <script src='https://cdn.datatables.net/1.10.15/js/dataTables.bootstrap4.min.js' />
+    {process.env.NODE_ENV === 'production' &&
```
valo: I think we should try to put the GA token in an environment variable and inject GA only when the token is set. This will be much more robust approach and will allow us to set a different GA account for staging.

app/tempPolyfills.js
```diff
@@ -0,0 +1,5 @@
+const raf = global.requestAnimationFrame = cb => {
+  setTimeout(cb, 0)
+}
```
valo: Why is this needed? Can we add a comment with some link explaining what is the purpose of this code.

app/__tests__/header.spec.js
```diff
@@ -0,0 +1,18 @@
...
+describe('Header', () => {
+  it('Header should render correct name', () => {
+    const app = mount(<Header name='check' />)
+    expect(app.find('h1').text()).toEqual('check')
+  })
+
+  it('Header should render correctly', () => {
+    const component = renderer.create(<Header />)
+    const tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
```
valo: I am a bit split about these snapshots. What is the general idea? To validate if the HTML structure is changing?

app/__tests__/header.spec.js
Partysun: The reason is next.js. In my projects, I put specs with component, but Next.js has some gotchas.
Example jest+next usage --> https://github.com/zeit/next.js/tree/canary/examples/with-jest
For good indexing, I will use the structure in __tests__ directory like in source.

app/components/main-head/index.js
Partysun: Yes, it's the good point. I don't love any error messages in the console, while I am doing anything and I put this check in code.

app/tempPolyfills.js
Partysun: https://dev.to/letsbsocial1/requestanimationframe--polyfill-in-react-16-2ce

app/__tests__/header.spec.js
Partysun: First of all, it's the small test for check JEST works properly or not.
But the second, snapshot testing in this approach can help me see changing in my components. I can verify by my eyes the component structure in snapshot files.

app/tempPolyfills.js
valo: Please, add this link as a comment in the file so that we don't lose it.

app/__tests__/header.spec.js
valo: OK, let's continue using them and see how it goes :)

app/__tests__/header.spec.js
valo: Are you sure that moving the tests to the component folder does not work?

PR #62, opened 2017-11-29
Add migrations for dropping TrackedEth and TrackedBtc

IvanIvanoff:
Add migrations that drop the no longer used DB tables TrackedEth and TrackedBtc


valo:
You need to update the `structure.sql`. After you run the migrations this file should be regenerated

PR #65, opened 2017-12-02
Redirect /cashflow/ to /cashflow

valo:
It is confusing to be able to open /cashflow, but /cashflow/ to show
404. Refectored the redirect logic into a separate utility function.


app/pages/cashflow/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { redirect_to } from '../../lib/nextjs_utils';
```
Partysun: `import { redirect_to } from 'lib/nextjs_utils'`

Current babel config gives this shorthand.

valo:
@Partysun we should add the linter to the jenkins build. If we want to really stick to the linter rules, they need to be enforced for each PR automatically.

PR #66, opened 2017-12-04
Use the Twitter api to fetch information about projects' twitter acco…

IvanIvanoff:


lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -0,0 +1,103 @@
...
+    |> Stream.run()
+
+    Process.send_after(self(), {:"$gen_cast", :sync}, update_interval_ms)
+    {:noreply, state}
+  end
+
+
+  defp blocking_start_twitter_fetch({twitter_accounts, index}) do
+    time_window_ms = Keyword.get(config(), :time_window, @default_time_window)
+    Process.sleep(index * time_window_ms)
```
valo: This is a kind of strange. Why do we need to sleep more as the index increases? If we need to keep a given rate limit, can we use the already existing rate limit server: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/rate_limiting/server.ex? The way it is used is to start a rate limiting servicer and then call the `wait` API with the name of the server https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/rate_limiting/server.ex#L41

lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -0,0 +1,103 @@
...
+      ordered: false,
+      timeout: 3 * 60 * 1000,
+      on_timeout: :kill_task
+    )
+    |> Stream.run()
+  end
+
+  defp fetch_and_store({project_id, "https://twitter.com/" <> twitter_name}) do
+    # Handle trailing slash
+    [twitter_name|_] = String.split(twitter_name, "/")
```
valo: I more readable way to do this is with https://hexdocs.pm/elixir/String.html#trim_trailing/1

lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  defp fetch_twitter_user_data({project_id, twitter_name}) do
+    user_data = ExTwitter.user(twitter_name, include_entities: false)
+
+    %TwitterAccountData{
+      followers_count: user_data.followers_count,
+      project_id: project_id,
+    }
+    |> TwitterAccountData.changeset()
```
valo: I would suggest keeping this data in influxDB as a time series, as we will want to see how the number of follower change over time.

valo:
I think we need to keep this data as time series, so that we can see how it changes over time. Also it will be nice to have the historical data, but I am not sure how to obtain this... I see there is a friendship/show API, but I don't see any timestamps there. Any ideas will be welcome.

lib/sanbase/external_services/twitter_data/worker.ex
valo: OK. Let's keep it like this

lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -0,0 +1,108 @@
...
+  alias Sanbase.ExternalServices.TwitterData.{Worker, Store}
+
+  @default_update_interval 1000 * 60 * 60 * 6
+
+  def start_link(_state) do
+    GenServer.start_link(__MODULE__, :ok)
+  end
+
+  def init(:ok) do
+    Process.flag(:trap_exit, true)
```
valo: Something I figure out now: if this process traps exits it should also process the `EXIT` messages, so that the mailbox is not becoming too big. I think we should add an empty `handle_info` method for the `EXIT` signals. I think we might need this in some other workers too...

IvanIvanoff: Done. When this and the github activity PR get merged  I'll submit a refactoring PR with some changes of how we use Influxdb + fixing the handling of :EXIT signals where needed.

lib/sanbase/application.ex
```diff
@@ -16,11 +16,17 @@ defmodule Sanbase.Application do
       # Start the endpoint when the application starts
       supervisor(SanbaseWeb.Endpoint, []),
 
-      # Time series DB connection
-      Sanbase.Prices.Store.child_spec,
+      # Start the Faktory supervisor
+      supervisor(Faktory.Supervisor, []),
```
valo: This supervisor should be at the end of the supervisors list, as we need the influxdb stores to be started before the workers boot.

lib/sanbase/external_services/twitter_data/store.ex
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def import(measurements) when is_list(measurements) do
+    measurements
+    |> Stream.map(&Measurement.convert_measurement_for_import/1)
+    |> Stream.map(&write_data/1)
+    |> Stream.run()
+  end
+
+  def time_series_for_measurement(measurement_name, from, to) do
```
valo: This should be a private function as far as I see

mix.exs
```diff
@@ -20,7 +20,7 @@ defmodule Sanbase.Mixfile do
   def application do
     [
       mod: {Sanbase.Application, []},
-      extra_applications: [:logger, :runtime_tools, :sasl],
+      extra_applications: [:logger, :runtime_tools, :sasl, :extwitter],
```
valo: I think `extwitter` should not be needed to be added here. The compiler should be adding these apps automatically. Can you also check if `mix release` builds without warnings about missing apps, please?

lib/sanbase/external_services/twitter_data/store.ex
IvanIvanoff: Currently not used, but this should be in the public API for providing all data for a measurement (twitter account). Also I think it's better to rename it to `all_records_for_measurement` as already have `last_record_for_measurement`

mix.exs
IvanIvanoff: The compiler doesn't complain about `:extwitter` indeed, but it needs `:oauther` to be added in `:applications` or  `:included_applications`

valo: Right now the apps that need to be included in the release are added here, so we probably need to add it there https://github.com/santiment/sanbase2/blob/master/rel/config.exs#L49

lib/sanbase/external_services/twitter_data/store.ex
valo: OK. Let's not spend time on this. When we start implementing the time series APIs we will come back and refactor it.

lib/sanbase/application.ex
IvanIvanoff: Noted and added a short comment describing why it is last.

PR #68, opened 2017-12-05
Feature/login

Partysun:
We are ready to ship this new frontend. :tada:


app/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
valo: Is this the correct name here?

app/src/App.js
```diff
@@ -0,0 +1,27 @@
...
+import Cashflow from './Cashflow'
+import SideMenu from './SideMenu'
+
+export const App = ({isDesktop}) => (
+  <div className='App'>
+    {isDesktop && <SideMenu />}
+    <Switch>
+      <Route exact path='/cashflow' component={Cashflow} />
+      <Route path={'/login'} component={Login} />
+      <Route exact path={'/'} component={Cashflow} />
```
valo: Shouldn't we have the roadmap and the signals page here?

app/src/Login.js
```diff
@@ -0,0 +1,170 @@
...
+          address: account,
+          addressHash: hashMessage}})
+        .then(({ data }) => {
+          const { token, user } = data.ethLogin
+          dispatch({
+            type: 'SUCCESS_LOGIN',
+            token,
+            username: user.username
+          })
+          // TODO:
```
valo: What is that TODO? May be it needs a description? It will also be nice to add the name of the author of the TODO and the date here, so that we know who to ask about it :)

app/src/Login.js
```diff
@@ -0,0 +1,170 @@
...
+          client.resetStore()
+        }).catch((error) => {
+          dispatch({
+            type: 'FAILED_LOGIN',
+            errorMessage: error
+          })
+          console.error(error)
+        })
+      }).catch(error => {
+        // TODO: User denied, Account, etc.
```
valo: Please add a better description here. I don't understand the above one. Also add your name and date in it, so that we know who to ask about it later. It might be the case that someone else works on this code later.

app/src/Login.js
```diff
@@ -0,0 +1,170 @@
...
+  graphql(requestAuthGQL, {
+    name: 'authWithSAN',
+    options: { fetchPolicy: 'network-only' }
+  }),
+  lifecycle({
+    componentDidMount () {
+      this.props.checkMetamask(hasMetamask())
+      setTimeout(() => {
+        this.props.appLoaded()
+      }, 1000)
```
valo: What is this timeout for and why it is 1 second?

app/src/SideMenu.js
```diff
@@ -0,0 +1,99 @@
...
+              Cash Flow
+            </Link>
+          </li>
+        </ul>
+        <li>
+          Signals
+        </li>
+        <li>
+          Roadmap
+        </li>
```
valo: We need these sections... We can't merge and deploy this without them.

app/src/SideMenu.js
```diff
@@ -0,0 +1,99 @@
...
+      </ul>
+    </div>
+
+    <br />
+
+    {user.username
+      ? <div className='user-auth-control'>
+        You are logged in!
```
valo: I think it will be nice to also show the SAN balance of the currently logged user, as a POC that we track that. Is it going to be hard to achieve that?

app/src/reducers/projects.js
```diff
@@ -0,0 +1,36 @@
...
+        error: false,
+        items: items,
+        ethPrice: action.payload.data.eth_price
+      }
+    case 'FAILED_PROJECTS':
+      return {
+        ...state,
+        isLoading: false,
+        error: true
+      }
```
valo: Just a question: when we implement a GQL API for the cashflow we won't need these redux actions, right?

app/src/reducers/user.js
```diff
@@ -0,0 +1,68 @@
+export const initialState = {
+  isLoading: true,
+  pending: false,
+  error: false,
+  data: {},
+  account: null,
```
valo: Is this an `ethAccount` object? It will be possible to have several accounts as one user. Is this going to be a problem with the current state?

app/src/reducers/user.js
```diff
@@ -0,0 +1,68 @@
...
+          username: action.username
+        }
+      }
+    case 'PENDING_LOGIN':
+      return {
+        ...state,
+        pending: true,
+        isLoading: false,
+        error: false
+      }
```
valo: I don't see this action being used anywhere. Do we need it?

app/src/App.js
Partysun: Yes. It's the place for future pages. When I will add page details, you can check it again.

app/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+ReactDOM.render(
+  <ApolloProvider client={client}>
+    <Provider store={store}>
+      <Router>
+        <Route path='/' component={App} />
+      </Router>
+    </Provider>
+  </ApolloProvider>,
+  document.getElementById('root'))
+// registerServiceWorker()
```
valo: If we don't use this service worker, may be we should remove it for now? I prefer not to have dead code in the codebase, as it only causes maintenance headaches.

app/src/web3Helpers.js
```diff
@@ -0,0 +1,34 @@
...
+export const setupWeb3 = cbk => {
+  if (!window.web3) { return }
+  const web3 = new Web3(window.web3.currentProvider) // eslint-disable-line
+  setInterval(() => {
+    const selectedAccount = web3.eth.accounts[0]
+    if (prevSelectedAccount !== selectedAccount) {
+      prevSelectedAccount = selectedAccount
+      cbk(false, selectedAccount || null)
+    }
+  }, 100)
```
valo: Why do we need to use `setInterval` here? I am generally hesitant to using `setInterval` and `setTimeout` as they are could lead to race conditions and weird behavior. Is there another way we can achieve the same functionality (may be a synchronous way)?

app/src/Cashflow.js
```diff
@@ -0,0 +1,180 @@
...
+const columns = [{
+  Header: 'Project',
+  id: 'project',
+  filterable: true,
+  sortable: true,
+  accessor: d => ({
+    name: d.name,
+    ticker: d.ticker
+  }),
+  Cell: ({value}) => <div>{value.name} ({value.ticker})</div>,
```
valo: Why did we remove the icon from the table?

app/src/reducers/projects.js
Partysun: Yes

valo:
Please add screenshots to the new look'n feel as I think this PR changes it significantly and I want to make sure it is acceptable for all before merging this to master.

app/src/App.js
valo: We can't remove the signals and the roadmap pages.

app/src/web3Helpers.js
Partysun: https://github.com/MetaMask/faq/blob/master/DEVELOPERS.md
last section

app/src/index.js
Partysun: It's right. I remove it

app/src/Cashflow.js
Partysun: I will add it later. I create table with full data and features and this is the next task with table. Bobr should add it later.

Partysun:
sanbase WIP styles.

app/src/registerServiceWorker.js
```diff
@@ -0,0 +1,109 @@
...
+    });
+}
+
+export function unregister() {
+  if ('serviceWorker' in navigator) {
+    navigator.serviceWorker.ready.then(registration => {
+      registration.unregister();
+    });
+  }
+}
```
valo: Can we remove this file?

app/src/web3Helpers.js
valo: Wow! O.o I hope there is a way to improve this in the future

PR #81, opened 2017-12-09
Fix the production build of the react app

valo:
The production build of the react app should be copied in the static folder of the phoenix app and served from there. We should make sure the production deployment is able to server the `index.html` and the react JS and CSS.


valo:
Merged this into #83, so closing this PR

PR #83, opened 2017-12-09
Feature/login

valo:
Recreating the PR so that we can fix the production deployment and then merge it into master. We can't afford to block master until the deployment is fixed.


valo:
We should not commit node_modules. Please remove it and squash the commits or something, as right now we have 50k+ added lines coming from node_modules in the history

valo:
Oh. I just saw it is removed. I'll probably merge and squash this...

app/server/index.js
```diff
@@ -0,0 +1,20 @@
...
+// react-router 4.0 need this
+app.get('*', (req, res) => {
+  res.sendFile(path.resolve(__dirname, '..', 'build', 'index.html'))
+})
+
+const PORT = process.env.PORT || 9000
+
+app.listen(PORT, () => {
+  console.log(`App listening on port ${PORT}!`)
+})
```
valo: I don't think we need such server. It won't resolve the problem we have here. What we need to do is to copy the react assets in the static folder of the Phoenix app and serve the `index.html` along with the assets around it.

PR #84, opened 2017-12-11
Some refactoring improving fault tolerance and readability of the code

IvanIvanoff:
- Add Task.Supervisor supervisor in the application
- Refactor some of the workers:  start all tasks under a task supervisor so crashes in them do not disturb the worker, improving the uptime.
- Refactor Prices.Store and its config to reduce the code size
- Run Elixir formatter on all affected files

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap.ex
+++ b/lib/sanbase/external_services/coinmarketcap.ex
@@ -62,7 +62,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
   end
 
   def handle_info(msg, state) do
-    Logger.info("Unknown message received: #{msg}")
+    Logger.warn("Unknown message received: #{msg}")
     {:noreply, state}
   end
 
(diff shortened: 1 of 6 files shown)


lib/sanbase/external_services/coinmarketcap.ex
```diff
@@ -40,21 +42,30 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
...
-    |> Stream.run
+    |> Stream.run()
 
     Process.send_after(self(), {:"$gen_cast", :sync}, update_interval)
 
     {:noreply, state}
   end
 
+  def handle_info(msg, state) do
+    Logger.info("Unknown message received: #{msg}")
```
valo: May be this should be changed to a `warn`? If this happens we should investigate what is going on for sure.

IvanIvanoff: Noted, also should be changed in the ethscan worker.

PR #85, opened 2017-12-12
Added Error boundary and raven logs

Partysun:
We need to add new env var for production. RAVEN_DSN
⚠️  Thie PR commit into revert-82-revert-68-feature/login.


valo:
I am going to register an account on sentry.io for us. We can host the server in our cluster, but let's signup for the free tier first and evaluate. We can switch later if we like the service.

PR #86, opened 2017-12-12
Feature/icons

Partysun:
Update icons.
What about the icon for Encrypgen (DNA)?
sanbase


valo:
We don't show `DNA` in production https://sanbase-low.santiment.net/cashflow so there is no need to fix this for now

PR #87, opened 2017-12-12
Use the window.location.origin if there is nothing set in the ENV

valo:
On staging and production we use the same image and the WEBSITE_URL
varialbe is used on compile time, so we can't use this variable when
building the production images. The origin should be computed on
runtime.

### DIFF
--- a/app/src/index.js
+++ b/app/src/index.js
@@ -22,7 +22,7 @@ import './index.css'
 
 Raven.config(process.env.RAVEN_DSN).install()
 
-const origin = process.env.WEBSITE_URL || 'http://localhost:4000'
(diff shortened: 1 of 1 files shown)


app/src/index.js
```diff
@@ -22,7 +22,10 @@ import './index.css'
 
 Raven.config(process.env.RAVEN_DSN).install()
 
-const origin = process.env.WEBSITE_URL || 'http://localhost:4000'
+let origin = process.env.WEBSITE_URL || ''
+if (process.env.production) {
```
valo: Let's try to keep such conditionals to a minimum. The more of these we have, the more problems we will have like the GA tracking: works locally, but fails on stg and prod

PR #89, opened 2017-12-12
Add graphiql when mix.env is dev

IvanIvanoff:


valo:
I would recommend using the graphiql app, as it allows to set headers in the requests. With the web version of graphiql you won't be able to make authenticated requests to the GQL API.

valo:
https://github.com/skevy/graphiql-app

IvanIvanoff:
That's actually `graphiql-workspace` [graphiql-workspace](https://github.com/OlegIlyenko/graphiql-workspace)
It allows adding custom headers.

valo:
Cool! Seems like I tried only the simple interface, which didn't allow that. Let's use this.

valo:
Cool! Seems like I tried only the simple interface, which didn't allow that. Let's use this.

PR #90, opened 2017-12-12
Fix/sort mutation search

Partysun:
Ready for master


lib/sanbase_web/graphql/schema.ex
```diff
@@ -17,7 +17,7 @@ defmodule SanbaseWeb.Graphql.Schema do
     field :eth_login, :login do
       arg :signature, non_null(:string)
       arg :address, non_null(:string)
-      arg :messageHash, non_null(:string)
+      arg :message_hash, non_null(:string)
```
Partysun: @valo Elixir/Absinth lib uses snake_case, but the result is camelCase.

valo: Yes. The convention in gql is camel case and the convention in elixir is snake case :)

PR #91, opened 2017-12-12
DB changes related to Project Transparency

MilanRusev:
* DB changes
* Ex_admin changes


lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -18,17 +19,42 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
     ico_spreadsheet_row = ico_spreadsheet_row
     |> set_infrastructure_default()
 
-    project = fill_project(ico_spreadsheet_row)
+    project = insert_or_update_project(ico_spreadsheet_row)
+
+    ico = insert_or_update_ico(project, ico_spreadsheet_row)
+
+    insert_or_update_ico_currencies(ico, ico_spreadsheet_row)
```
IvanIvanoff: These could be chained with the pipe operator |>

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -105,7 +131,8 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
 
   defp ensure_ico(project) do
     case project do
-      %{ico: result = %Ico{}} -> result
+      %{icos: []} -> %Ico{}
+      %{icos: result} when is_list(result) -> List.first(result)
```
IvanIvanoff: better to use pattern matching `%{icos: [result|_]} -> result`

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -18,17 +19,40 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
...
+  defp insert_or_update_ico_currencies(ico, ico_spreadsheet_row) do
+    currencies = ensure_currencies(ico_spreadsheet_row.ico_currencies)
+    |> Enum.map(&Ecto.Changeset.change(&1))
+
+    Enum.each(currencies, fn currency ->
+      currency = currency
+      |> Repo.insert_or_update!()
+
+      ensure_ico_currency(ico, currency)
+      |> Repo.insert_or_update!()
```
valo: Why do we need to have `Currency` and `IcoCurrency`? Can't we just have one model for a currency?

MilanRusev: There is a many-to-many relationship between currencies and ICOs and IcoCurrency is the intermediate table. Only Currencies are used elsewhere too.

valo: OK. We need to export our schema to a picture and start iterating on it. I have the feeling that we are complicating it too much :-)

PR #93, opened 2017-12-13
Feature/detailed page

Partysun:
WIP


app/src/components/PanelBlock.js
```diff
@@ -1,18 +1,18 @@
...
-    {content}
+    {children}
   </div>
 )
 
 PanelBlock.defaultProps = {
   classes: '',
   title: '',
-  content: ''
+  children: ''
```
Partysun: @bobrovova children should be defined as type node and default prop is equal null.

app/src/components/GeneralInfoBlock.css
```diff
@@ -0,0 +1,7 @@
+.rowInfo {
+  display: flex;
+  flex-direction: row;
```
Partysun: @bobrovova `flex-direction` default value = 'row'.

app/src/components/GeneralInfoBlock.css
```diff
@@ -0,0 +1,7 @@
+.rowInfo {
```
Partysun: @bobrovova  Why do you use `row` in the className?

Partysun: @bobrovova Why do you write the same css code in separate files?

app/src/pages/Detailed.css
```diff
@@ -1,16 +1,16 @@
 .detailed-head {
-  overflow: hidden;
   margin-top: 32px;
   height: 84px;
   margin-right: 16px;
+  display: flex;
+  flex-direction: row;
```
Partysun: @bobrovova flex-direction is row by default again. And in other places.

PR #95, opened 2017-12-13
Add a server side rendered env.js file with the runtime config

valo:
The file uses the ENV variables on the server to set the runtime config
for the react app.


app/public/index.html
```diff
@@ -15,5 +15,6 @@
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
+    <script type="text/javascript" src="%PUBLIC_URL%/env.js"></script>
```
valo: @Partysun can we somehow add a random string to this filename each time we build a production bundle? In the ideal case this file will be cached, but will change each time we push a new version of the app.

app/src/index.js
```diff
@@ -20,87 +20,95 @@ import { loadState, saveState } from './utils/localStorage'
...
 import './index.css'
 
-Raven.config(process.env.RAVEN_DSN).install()
-
-let origin = process.env.WEBSITE_URL || ''
-if (process.env.production) {
-  origin = window.location.origin
-}
+const handleLoad = () => {
+  console.log(JSON.stringify(window.env))
```
valo: Shouldn't we remove such debug logs?

app/package.json
```diff
@@ -30,9 +30,9 @@
     "semantic-ui-react": "^0.77.0"
   },
   "scripts": {
-    "start": "node node_modules/exit_on_eof/index.js 'react-scripts start'",
+    "start": "REACT_APP_VERSION=$(node -pe 'require(\"./package.json\").version')-$(git rev-parse --short HEAD) node node_modules/exit_on_eof/index.js 'react-scripts start'",
```
valo: Why can't we just use the `git rev-parse --short HEAD` here? In dev mode I think we don't care about caching, so it should be fine the use the current git head, shouldn't it?

valo: I am also concerned that `git rev-parse --short HEAD` won't work in our deployment script. Can we use an env variable here, like `GIT_HEAD`? This way we can pass the variable to docker from the script, which runs the build.

app/src/index.js
Partysun: Yes. I did it only for you, when you test it.

app/package.json
Partysun: Probably yes. I like to see package version in my env for several views. For example, I can show version of the project on the Settings page.

valo: Can we export the version of the project using something like `DefinePlugin`?

Partysun: We don't have access to change DefinePlugin options now. We need to eject create-react-app config for this. But it will create more complex app structure.

The creat-react-app authors tell clients to use .env vars or npm scripts for that changes.

valo:
Let try this out

PR #98, opened 2017-12-15
Project graphql api related to Project Transparency

MilanRusev:
Queries:
* all projects (optionally filtered by project transparency)
* single project by id

Currently the projects API can be accessed with both user bearer token and with basic auth (so it can be accessed from both the UI and other internal services).


IvanIvanoff:
Add a few tests. For reference how to test graphql: https://github.com/santiment/sanbase2/commit/ed3eb0bc03ce8b696c70268035f83a1899f1db8f#diff-358f8f9cd5fabeb1b47fa4fc23ef1222R1

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -0,0 +1,99 @@
...
+
+  alias Sanbase.Repo
+  alias Ecto.Multi
+
+  def all_projects(parent, args, context) do
+    only_project_transparency = Map.get(args, :only_project_transparency, false)
+
+    query = from p in Project,
+    where: not ^only_project_transparency or p.project_transparency,
+    select: p
```
valo: Do you really need this `select` here? I think by default the query will return all the fields of the selected table.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -0,0 +1,99 @@
...
+        Repo.one(fallback_query)
+        |> case do
+          %{funds_raised_usd: funds_raised_usd} when not is_nil(funds_raised_usd) ->
+            [%{currency_code: "USD", amount: funds_raised_usd}]
+          %{funds_raised_btc: funds_raised_btc} when not is_nil(funds_raised_btc) ->
+            [%{currency_code: "BTC", amount: funds_raised_btc}]
+          %{funds_raised_eth: funds_raised_eth} when not is_nil(funds_raised_eth) ->
+            [%{currency_code: "ETH", amount: funds_raised_eth}]
+          _ -> []
+        end
```
valo: I think the body of this case should be moved to another function with some descriptive name that explains what the fallback query does...

valo:
Regarding authentication: we should add basic authentication for internal APIs. Here is an example how the JWT authentication works: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/context_plug.ex We can have a similar middleware, which checks if the request has a valid basic auth and if it does, set a flag in the GQL context. Then we will be able to allow/forbid certain queries.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -0,0 +1,106 @@
...
+      	and not exists (select 1
+      		from ico_currencies ic
+      		where ic.ico_id = i.id
+      			and ic.amount is not null))
+      select d.currency_code, sum(d.amount) amount
+      from data d
+      where d.currency_code is not null
+      group by d.currency_code
+      order by d.currency_code
+      '''
```
valo: OMG 😨 can we simplify this query? Why is this so complex?

MilanRusev: It's a long one because it has to fetch and union data from two different sources, but actually it is not that complex.

We have to aggregate all amounts for every currency for every ICO of the given project, this is the last part:
```
select d.currency_code, sum(d.amount) amount
      from data d
      where d.currency_code is not null
      group by d.currency_code
      order by d.currency_code
```

Now the data sources:
For ICOs that have raw data entered for at least one currency we aggregate it by currency:
```
select c.code currency_code, ic.amount
      from icos i
      join ico_currencies ic
      	on ic.ico_id = i.id
      		and ic.amount is not null
      join currencies c
      	on c.id = ic.currency_id
      where i.project_id = $1
```

For ICOs that don't have that data entered (currently everything imported from the spreadsheet) we fall back to a precalculated total:
```
select case
      		when i.funds_raised_usd is not null then 'USD'
      		when i.funds_raised_btc is not null then 'BTC'
      		when i.funds_raised_eth is not null then 'ETH'
      		else null
      	end currency_code
      	, coalesce(i.funds_raised_usd, i.funds_raised_btc, i.funds_raised_eth) amount
      from icos i
      where i.project_id = $1
      	and not exists (select 1
      		from ico_currencies ic
      		where ic.ico_id = i.id
      			and ic.amount is not null)
```

I tried doing it with a combination of Ecto and Elixir but it got too complicated and Ecto doesn't support many of the sql constructs. With plain sql it's much more straight forward.

tzanko-matev: Maybe it's good to put this comment as documentation in the source code

valo: I think we should think how we can change the schema, so that we don't have 2 variants where we take the data from (raw data entered vs data imported from the ICO).

I think such queries are a symptom of a problem in the data modeling. Let's not change the schema now, but I'll create a card for improving the data model in the future, as this it going to bite us over and over again and make such queries a pain. https://favro.com/card/bdbb4f24afc48b29c74f4fa4/San-930

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -0,0 +1,115 @@
...
+    query = from p in Project,
+    where: not ^only_project_transparency or p.project_transparency
+
+    projects = Repo.all(query)
+
+    {:ok, projects}
+  end
+
+  def eth_balance(%Project{id: id}, args, context) do
+    only_project_transparency = get_parent_args(context)
```
valo: Isn't this parameter included in the `args`, like in these resolvers https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/account_resolver.ex#L15

valo:
Can we protect this API with basic auth, as it will be initially used only internally. I think we need to figure out which fields we will open to the public and which will be internal-only... Here are some docs how authentication can be added to the GQL endpoint https://hexdocs.pm/absinthe/context-and-authentication.html#content

lib/sanbase_web/graphql/resolvers/project_resolver.ex
MilanRusev: No, it is included only in the parent call, the children resolvers have their own args.

Partysun:
We can protect API by User Bearer Token in the header of the request. The user has the rule to use some query/mutations or not.
Or not?

MilanRusev:
I currently made it so that the projects API can be accessed with both user bearer token and with basic auth (so it can be accessed from both the UI and other internal services).

config/config.exs
```diff
@@ -145,6 +145,10 @@ config :faktory_worker_ex,
   ],
   start_workers: {:system, "FAKTORY_WORKERS_ENABLED", false}
 
+config :sanbase, SanbaseWeb.Graphql.ContextPlug,
+  basic_auth_username: {:system, "GRAPHQL_BASIC_AUTH_USERNAME"},
+  basic_auth_password: {:system, "GRAPHQL_BASIC_AUTH_PASSWORD"}
```
valo: Why do we have the basic auth in this PR? Can we review these separately?

valo: Actually I see that they are dependent. Can we close the other PR in that case?

MilanRusev: Initially the basic auth change was developed in this PR, but I extracted it in a separate one because @IvanIvanoff also uses it in his PR: https://github.com/santiment/sanbase2/pull/106

We can can merge the basic auth PR first in case it's needed elsewhere before we merge the other two.

valo: OK

lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -6,7 +6,7 @@ defmodule SanbaseWeb.Graphql.AccountResolver do
   alias Sanbase.Repo
   alias Ecto.Multi
 
-  def current_user(_root, _args, %{context: %{current_user: user}}) do
+  def current_user(_root, _args, %{context: %{auth: %{auth_method: :user_token, current_user: user}}}) do
```
valo: I think we can try to implement the logic around the object and field auth using some DSL, but I'll look into this in the following days. For now this approach is OK.

valo:
@MilanRusev please, resolve the conflicts so that we can merge this

PR #99, opened 2017-12-15
Add current user mutations

IvanIvanoff:
Work in progress version.
/graphiql should be removed. Maybe also remove from this PR formatting of `project.ex`


valo:
Let's look into this after #98 is merged. Then we can return project objects and fix the auth pattern matching.

lib/sanbase/model/user_followed_project.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Sanbase.Model.UserFollowedProject do
+  use Ecto.Schema
+  import Ecto.Changeset
+  alias Sanbase.Model.UserFollowedProject
+
+  schema "user_followed_project" do
+    field :project_id, :integer
```
MilanRusev: Can you make the FK on delete cascade. Otherwise we'll have the same problem with deleting projects as in https://github.com/santiment/sanbase2/pull/105

PR #102, opened 2017-12-17
Search for projects by coinmarketcap_id instead of name

valo:
Names are not a good unique identifier, as we update the name from CMC data. We should really be using something like coinmarketcap_id as a unique identifier for the projects...


valo:
We should add a unique index on the `coinmarketcap_id` column. Let's do that when the delete is fixed in the admin panel.

PR #111, opened 2017-12-19
Fix parsing events of old github events

valo:
Seems like the events which are more than 2-3 years old has a bit
different format than the curren ones. Added a more sophisticated parser
for the repository names.

Also added support for github links which contain `www.`.


valo:
We should probably add some tests for this logic. The problem is that stubbing the http streaming will be a bit tricky, so I would rather reorganize this module a bit before that. May be it will make sense to do that when we add S3 support to the worker.

PR #117, opened 2017-12-20
Twitter ignore non existent accounts

IvanIvanoff:
We chose to avoid crashing the processes when non existing account is fetched to avoid too flooding the logs.
Some  little fixes are added - there is no `Logger.msg` function - `Logger.info` should be used.

### DIFF
--- a/lib/sanbase/external_services/twitter_data/worker.ex
+++ b/lib/sanbase/external_services/twitter_data/worker.ex
@@ -26,10 +26,11 @@ defmodule Sanbase.ExternalServices.TwitterData.Worker do
   end
 
   def init(:ok) do
-    :ok = ExTwitter.configure(
-      consumer_key: get_config(:consumer_key),
-      consumer_secret: get_config(:consumer_secret),
-    )
+    :ok =
+      ExTwitter.configure(
+        consumer_key: get_config(:consumer_key),
+        consumer_secret: get_config(:consumer_secret)
+      )
 
     if get_config(:sync_enabled, false) do
       Store.create_db()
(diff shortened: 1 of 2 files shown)


lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -67,32 +68,41 @@ defmodule Sanbase.ExternalServices.TwitterData.Worker do
...
   def fetch_twitter_user_data(twitter_name) do
     Server.wait(@rate_limiter_name)
-    # GET https://api.twitter.com/1.1/users/show.json?screen_name=twitter_name
-    ExTwitter.user(twitter_name, include_entities: false)
+
+    try do
+      ExTwitter.user(twitter_name, include_entities: false)
+    rescue
+      _ -> nil
+    end
```
valo: Can we handle the different types of responses here? This approach seems to be like too of a big hammer solution...

lib/sanbase/external_services/twitter_data/worker.ex
```diff
@@ -67,32 +68,54 @@ defmodule Sanbase.ExternalServices.TwitterData.Worker do
...
+        Logger.warn("Connection error while trying to fetch twitter user data: #{e.reason}")
+        nil
+
+      e in ExTwitter.Error ->
+        Logger.warn("Error trying to fetch twitter user data for #{twitter_name}: #{e.message}")
+        nil
+
+      _ ->
+        nil
+    end
```
valo: I can't believe this `ExTwitter` library is really structured like this. It should not be using exception handling, but the `{:error, reason}` pattern... I don't think we can do anything about it right now. Let's proceed like this :-/

valo:
Looks good. Let's merge after the newsletter is out.

lib/sanbase/external_services/twitter_data/worker.ex
IvanIvanoff: https://github.com/parroty/extwitter/issues/67
Seems like they tried to migrate to error tuples but eventually reverted. I'll follow the project and rework the error handling here when possible.

valo:
Please resolve the conflict there. `deployment.yaml` should be removed. We are not holding this file in this repo any more. Just remove it and all will be fine

PR #121, opened 2017-12-21
Basic auth to graphql api

MilanRusev:
Added basic auth to graphql and changed the authentication map in graphql context

### DIFF
--- a/lib/sanbase_web/graphql/context_plug.ex
+++ b/lib/sanbase_web/graphql/context_plug.ex
@@ -3,33 +3,76 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
 
   import Plug.Conn
 
+  import Sanbase.Utils, only: [parse_config_value: 1]
+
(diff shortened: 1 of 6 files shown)


lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -3,33 +3,76 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
...
+    |> case do
+      ^auth_attempt -> {:ok, username}
+      _ ->
+        Logger.warn("Invalid basic auth credentials in request")
+        {:error, :invalid_credentials}
+    end
+  end
+
+  defp config(key) do
+    Application.get_env(:sanbase, __MODULE__)
```
IvanIvanoff: I think it's better to use fetch_env! as it will raise an error if the env variable is not set and not return a default value

valo:
Looks good. Let's merge after the newsletter is released.

PR #124, opened 2017-12-21
Fix/icons

Partysun:


valo:
Are you sure about the names? For example here https://sanbase-low.santiment.net we use "Santiment Network Token"

valo:
These names are scraped from https://coinmarketcap.com

valo:
BTW, may be we can use a more stable ID for the icons, like `coinmarketcap_id`. This won't be changing.

Partysun:
Yes, it's more better. I will add mixed stylesheets for icons in next week and add this.

PR #126, opened 2017-12-21
Added Frontend Readme, Update Readme/Changelogs

Partysun:


valo:
Thanks for adding these docs! I didn’t know about the ‘u’ trick to update the snapshots

PR #130, opened 2017-12-24
Added middlewares for defining access control

valo:
For now we have 3 middlewares:
* BasicAuth for internal APIs
* JWTAuth for frontend authenticated requests that also checks SAN
  balance
* MultipleAuth for specifying several auth methods. If one of the
  methods pass, the request is permited


lib/sanbase_web/graphql/middlewares/multiple_auth.ex
```diff
@@ -0,0 +1,26 @@
+defmodule SanbaseWeb.Graphql.Middlewares.MultipleAuth do
+  @behavior Absinthe.Middleware
```
IvanIvanoff: for consistency let's put `@behaviour` after `@moduledoc`

PR #132, opened 2017-12-25
Optimize the github activity tracking and scheduling

valo:
Download the github archives in S3, so that when they need to be
reprocessed we can download them directly from there. The bandwidth to
S3 will be much bigger than to the external service.

Also improved the scheduling to keep track which archives are processed
for each project and do not reprocess archives if they are all scraped
before, even if there is no activity for them.


IvanIvanoff:
Is it intentional that this PR target is `use-dotenv-config` but not `master`?

test/sanbase/github/scheduler_test.exs
```diff
@@ -76,10 +76,54 @@ defmodule Sanbase.Github.SchedulerTest do
...
+
+    mark_as_processed_interval(san_project.id, days_ago(5), days_ago(1))
+    mark_as_processed_interval(omg_project.id, days_ago(5), days_ago(2))
+
+    Scheduler.schedule_scrape()
+
+    assert_called SanbaseWorkers.ImportGithubActivity, :perform_async, [_], 48 # 2 days
+  end
+
   defp days_ago(days) do
```
IvanIvanoff: We now have this function in `Sanbase.DateTimeUtils`

PR #142, opened 2017-12-28
Add async resolving gql api

IvanIvanoff:
1. Fields are resolved in one process sequentially. Object fields with custom resolvers that make http request or query the DB can be started asynchronously.

2. Run elixir formatter

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/account_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/account_resolver.ex
@@ -49,16 +49,16 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
     |> User.changeset(%{email: new_email})
     |> Repo.update()
     |> case do
-         {:ok, user} ->
-           {:ok, user}
+      {:ok, user} ->
+        {:ok, user}
 
-         {:error, changeset} ->
-           {
-             :error,
-             message: "Cannot update current user's email to #{new_email}",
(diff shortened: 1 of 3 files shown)


MilanRusev:
Shouldn't we add async to every GQL query that accesses a db/service? Currently it's not added to `AccountResolver.eth_login, AccountResolver.change_email, AccountResolver.unfollow_project, AccountResolver.follow_project, GithubResolver.activity , GithubResolver.available_repos, PriceResolver.history_price, PriceResolver.current_price, PriceResolver.available_prices, ProjectResolver.all_projects, ProjectResolver.project, ProjectResolver.all_projects_with_eth_contract_info`

IvanIvanoff:
We won't get benefit adding async resolving to every db/service. First, the query must execute more than one resolver (`GithubResolver.activity`, `GithubResolver.available_repos`, etc. are only executing one) - we can fire a new process to do the db query, but in the current process we will only await for the result. Second, the resolvers should be able to run in parallel.
Resolvers by default run in one process, i.e. sequentially. Benefit is gained when we can start several at the same time

MilanRusev:
Ah I see. I thought Absinthe was using one process for all GQL requests.

valo:
Please, resolve merge conflicts

MilanRusev:
We'd better resolve them after https://github.com/santiment/sanbase2/pull/161 is merged as there are changes too.

PR #152, opened 2018-01-02
The from-to period should be divisible by the interval

IvanIvanoff:


MilanRusev:
I tested it locally and it worked

PR #161, opened 2018-01-05
Add calculations for funds raised totals for project, ico; remove total funds raised total columns from ICO

MilanRusev:
This PR does 2 things:

# Add calculations for funds raised totals for project, ico
The totals are calculated in USD, ETH, BTC. Any currency conversions are done for ICO end date. Aggregates all ICO currency amounts converting them to the appropriate currency

The values for a project are the aggregated values for its ICOs

GQL query:
```
query {
  projectFull(id:1)
  {
    name,
    fundsRaisedUsdIcoEndPrice,
    fundsRaisedEthIcoEndPrice,
    fundsRaisedBtcIcoEndPrice,
    icos {
      endDate,
      fundsRaisedUsdIcoEndPrice,
      fundsRaisedEthIcoEndPrice,
      fundsRaisedBtcIcoEndPrice
    }
  }
}
```

# Remove total funds raised total columns from ICO
Remove funds raised usd/btc/eth columns from ico table & use only the details IcoCurrencies


valo:
@MilanRusev can you propose some changes in the schema, so that these calculations will be simpler?

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -12,12 +12,17 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
 
   def import(ico_spreadsheet) when is_list(ico_spreadsheet) do
     Logger.info("Starting ICO spreadsheet import...")
+    usd = ensure_currency("USD") |> Repo.insert_or_update!()
+    eth = ensure_currency("ETH") |> Repo.insert_or_update!()
+    btc = ensure_currency("BTC") |> Repo.insert_or_update!()
```
valo: If the method `ensure_currency` becomes `find_or_insert_currency!` and does the DB operation, isn't going to simplify the usage of the method? As far as I see everywhere we insert or update the currency after calling this method.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
```diff
@@ -149,9 +150,63 @@ defmodule Sanbase.DbScripts.ImportIcoSpreadsheet do
...
     Repo.delete_all(from c in IcoCurrencies, where: c.ico_id == ^ico.id and c.currency_id not in ^currency_ids)
+
+    ico
+  end
+
+  defp move_funds_raised_to_details(ico, ico_spreadsheet_row, main_currencies) do
+    ico = Repo.preload(ico, [ico_currencies: [:currency]])
+
+    ico_currency = funds_raised_try_from_existing_ico_currencies(ico, ico_spreadsheet_row, main_currencies)
+      || funds_raised_try_from_totals(ico, ico_spreadsheet_row, main_currencies)
```
valo: Give that we already created all the `IcoCurrency` objects for all projects do we need to continuously keep checking this in the code? Can we assume that if there are no `IcoCurrency`s attached to the project, the total funds are just 0?

lib/sanbase/model/ico.ex
```diff
@@ -61,62 +51,32 @@ defmodule Sanbase.Model.Ico do
...
+    funds_raised_ico_end_price_from_currencies(ico, "BTC", end_date)
   end
 
-  defp calculate_usd_from_btc({:ok, btc}, :error, usd_btc) when not is_nil(btc) and not is_nil(usd_btc) do
-    {:ok, Decimal.mult(btc,usd_btc)}
+  defp funds_raised_ico_end_price_from_currencies(ico, target_ticker, date) do
+    timestamp = Sanbase.DateTimeUtils.ecto_date_to_datetime(date)
+
+    Repo.preload(ico, [ico_currencies: [:currency]]).ico_currencies
+    |> Enum.reduce(nil, fn(ic, acc) ->
```
valo: Can we set the initial value of the accumulator to `0` and safe one case in the reducer?

lib/sanbase/model/project.ex
```diff
@@ -119,17 +120,60 @@ defmodule Sanbase.Model.Project do
...
+    Repo.all(query)
+  end
+
+  defp add_if_not_nil(first, second) do
+    cond do
+      !is_nil(first) and !is_nil(second) -> Decimal.add(first, second)
+      !is_nil(first) -> first
+      !is_nil(second) -> second
+      true -> nil
+    end
```
valo: It is possible to have separate functions for each of these cases. Then no `case` will be needed.

lib/sanbase/model/project.ex
```diff
@@ -119,17 +120,60 @@ defmodule Sanbase.Model.Project do
...
+  def funds_raised_usd_ico_end_price(project) do
+    Repo.preload(project, :icos).icos
+    |> Enum.reduce(nil, fn(ico, acc) ->
+      add_if_not_nil(acc, Ico.funds_raised_usd_ico_end_price(ico))
+    end)
+  end
+
+  def funds_raised_eth_ico_end_price(project) do
+    Repo.preload(project, :icos).icos
+    |> Enum.reduce(nil, fn(ico, acc) ->
```
valo: Can we just do `sum` here? https://hexdocs.pm/elixir/Enum.html#sum/1

lib/sanbase/prices/utils.ex
```diff
@@ -0,0 +1,86 @@
...
+    |> case do
+      nil ->
+        with price_btc <- fetch_last_price_before(ticker_from <> "_BTC", timestamp),
+            true <- !is_nil(price_btc),
+            price_btc_usd <- fetch_last_price_before("BTC_USD", timestamp),
+            true <- !is_nil(price_btc_usd) do
+          Decimal.mult(price_btc, price_btc_usd)
+        else
+          _ -> nil
+        end
```
valo: Can we extract this logic into a separate function with a descriptive name? Seems like almost the same logic is used below too.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -30,6 +30,11 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
       _ -> Repo.all(query)
     end
 
+    projects = case funds_raised_ico_end_price_requested?(resolution) do
+      true -> Repo.preload(projects, [icos: [ico_currencies: [:currency]]])
+      _ -> projects
+    end
```
valo: If we use batching, do we need such kind of optimization?

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
valo: This is part of the import, so let's not refactor it. We should remove it at some point.

lib/sanbase/db_scripts/import_ico_spreadsheet.ex
valo: This is part of the import, so let's not refactor it. We should remove it at some point.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
MilanRusev: This way if all of the three fields `fundsRaisedUsdIcoEndPrice`, `fundsRaisedEthIcoEndPrice`, `fundsRaisedBtcIcoEndPrice` are requested we will execute the preload query one time instead of three times. The batching is only for a single field.

valo: I would vote for just preloading this data all the time, no matter which fields are fetched. I don't think the added code complexity justifies the optimization we are trying to achieve here... Do you think adding these 3 additional queries will slow down things too much?

PR #173, opened 2018-01-09
CMC service - don't update project name

MilanRusev:
Change coinmarketcap service not to overwrite project's name


lib/sanbase/external_services/project_info.ex
```diff
@@ -51,8 +51,12 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
       |> Ico.changeset(Map.from_struct(project_info))
       |> Repo.insert_or_update!
 
+      project_attrs = project_info
```
valo: Can we add a comment why we need to skip updating the name of the project?

valo:
We don't need any more, so closing for now

PR #182, opened 2018-01-11
Add tests for etherbi api - token burn rate and transaction volume

IvanIvanoff:


lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -1,19 +1,17 @@
...
     from_unix = DateTime.to_unix(from, :seconds)
     to_unix = DateTime.to_unix(to, :seconds)
 
-    case get("/burn_rate?ticker=#{ticker}&from_timestamp=#{from_unix}&to_timestamp=#{to_unix}") do
-      %Tesla.Env{status: 200, body: body} ->
-        {:ok, result} = Poison.decode(body)
+    etherbi_url = get_config(:url)
+    url = "#{etherbi_url}/burn_rate?ticker=#{ticker}&from_timestamp=#{from_unix}&to_timestamp=#{to_unix}"
 
+    options = [recv_timeout: 30_000]
```
valo: Isn't 30sec too much? Can we go with something like 10-15sec? 30sec will probably mean a timeout on the socket. In the future I think we should cache this data in our time series DB.

test/sanbase_web/graphql/etherbi_api_test.exs
```diff
@@ -0,0 +1,99 @@
+defmodule Sanbase.Github.EtherbiApiTest do
+  use SanbaseWeb.ConnCase
+  use Phoenix.ConnTest
+
+  import Sanbase.Utils, only: [parse_config_value: 1]
+  import Mockery
+  import Mock
+
+  test "fetch burn rate", context do
+    with_mock HTTPoison,
```
valo: If we use `mock`, the tests should be marked with `async: false`. To be honest I am a bigger fan of `mockery` as you can mock the whole module and run the tests concurrently. It has an additional overhead of putting an `injection` in your code, but I think it is fine the inject the modules that make external calls.

In the long term I am thinking to introduce `Gateway` modules for each external call that we make. This will allow to have a common approach for mocking the gateway modules. May be we can come up with an interface, which will allow us to easily mock gateway modules in all tests.

test/sanbase_web/graphql/etherbi_api_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      assert %{"transactionVolume" => "700000000000001803841"} in transaction_volumes
+      assert %{"transactionVolume" => "1749612781540000000000"} in transaction_volumes
+    end
+  end
+
+  defp get_config(key) do
+    Application.fetch_env!(:sanbase, SanbaseWeb.Graphql.Resolvers.EtherbiResolver)
+    |> Keyword.get(key)
+    |> parse_config_value()
+  end
```
valo: Is this function used?

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
IvanIvanoff: Let's reduce it to 15s

test/sanbase_web/graphql/etherbi_api_test.exs
IvanIvanoff: not anymore, removed

PR #183, opened 2018-01-12
Added Twitter Chart. Monkey Patch Fix Chart.js

Partysun:
Added Twitter Chart
  Added gql twitter data and history
  Monkey patch fix problem with chart.js lib
    chart.js show many points in x mode,
    when we have very small interval between points.
    We choose the first point.
  Cleaning.


app/src/index.js
```diff
@@ -33,6 +39,7 @@ const handleLoad = () => {
   if (process.env.production) {
     origin = window.location.origin
   }
+  origin = 'https://sanbase.stage.internal.santiment.net/'
```
valo: What is the purpose of this?

Partysun: Sorry. This is the terrible mistake.

PR #184, opened 2018-01-12
Refactor getting config values

IvanIvanoff:
Up until now in every module we need to fetch a config variable we had to define a function `config`, `get_config`, `get_config_value` etc. which is not consistent and not unneeded work. The current module from which the `Config.get` is called can be determined by using macros.


lib/sanbase/influxdb/store.ex
```diff
@@ -10,8 +10,9 @@ defmodule Sanbase.Influxdb.Store do
   defmacro __using__(_options \\ []) do
     quote do
       use Instream.Connection, otp_app: :sanbase
+      import Sanbase.Utils.Config
```
valo: Shouldn't we use `require Sanbase.Utils.Config` in all places?

lib/sanbase/internal_services/ethauth.ex
```diff
@@ -1,7 +1,8 @@
 defmodule Sanbase.InternalServices.Ethauth do
   use Tesla
 
-  import Sanbase.Utils, only: [parse_config_value: 1]
+  require Sanbase.Utils.Config
+  alias Sanbase.Utils.Config
```
valo: Is it possible to only do `alias` here or `require` is always needed?

IvanIvanoff: `require` is needed to introduce the macros. `import` can be used, but it directly imports the function names as if they are part of that module and there could be name colisions

valo: OK

valo:
Please, resolve merge conflicts and let me know if this is ready to be merged.

PR #185, opened 2018-01-12
Improve GitHub activity scheduler

valo:
The current GH activity scheduler is timing out when trying to schedule
jobs. This is because of a very big query, which can't be optimized
without creating a gigantic index on it.

In this PR, the computations are broken down into many smaller queries
to the DB. Given that the scheduling is much faster than the processing,
even a bit slower scheduling will fine.

Also switched to using the `default` queue, which allows to check the
size of the queue before adding more work into it.

### DIFF
--- a/lib/sanbase_workers/import_github_activity.ex
+++ b/lib/sanbase_workers/import_github_activity.ex
@@ -12,7 +12,7 @@ defmodule SanbaseWorkers.ImportGithubActivity do
   alias ExAws.S3
   @github_archive "http://data.githubarchive.org/"
 
-  faktory_options queue: "github_activity", retry: -1, reserve_for: 900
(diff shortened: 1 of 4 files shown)


lib/sanbase_workers/import_github_activity.ex
```diff
@@ -12,7 +12,7 @@ defmodule SanbaseWorkers.ImportGithubActivity do
   alias ExAws.S3
   @github_archive "http://data.githubarchive.org/"
 
-  faktory_options queue: "github_activity", retry: -1, reserve_for: 900
+  faktory_options queue: "default", retry: -1, reserve_for: 900
```
IvanIvanoff: change the queue name in the config, too

valo: Changed

PR #186, opened 2018-01-12
Ico - move funds raised totals to details

MilanRusev:
Move the values of ICO total columns "funds raised usd/btc/eth" to the details table IcoCurrencies:

**The algorithm:**
For each ICO that doesn't have filled in details currency amounts (all its IcoCurrencies are with null amounts):
1. Find the first X in ("ETH", "BTC", "USD") (in that order) for which ICO."funds raised X" is not null and there exists and IcoCurrency for X for this ICO. If found move the total to the detail table. If not found:
2. Find the first X in ("ETH", "BTC", "USD") (in that order) for which ICO."funds raised X" is not null and create an IcoCurrency for X and the total amount for this ICO


priv/repo/migrations/20180112084549_ico_move_funds_raised_totals_to_details.exs
```diff
@@ -0,0 +1,84 @@
...
+
+  import Ecto.Query, warn: false
+
+  alias Sanbase.Repo
+  alias Sanbase.Model.Ico
+  alias Sanbase.Model.IcoCurrencies
+  alias Sanbase.Model.Currency
+
+  def up do
+    Repo.transaction(fn ->
```
valo: I think that the migrations are automatically wrapped in transactions. Also do you think the whole operation will really take 10min?

priv/repo/migrations/20180112084549_ico_move_funds_raised_totals_to_details.exs
```diff
@@ -0,0 +1,84 @@
...
+      amount ->
+        %IcoCurrencies{}
+        |> IcoCurrencies.changeset(
+          %{ico_id: ico.id,
+            currency_id: currency.id,
+            amount: amount})
+    end
+  end
+
+  defp ensure_currency(currency_code) do
```
valo: I think we can improve the naming here and use something like `find_or_insert_currency`, as `ensure` does not give much information what exactly is ensured. I know that we use this in a bunch of places and I think that we can gradually improve the naming as the time goes. I just want to start the discussion now :)

valo:
If we really think that the migration will take a lot of time, may be it will make more sense to schedule it as a data migration background job?

MilanRusev:
It doesn't take a lot of time. I added the transaction because `Repo.stream()` works only in a trasaction, but now tested it without the wrapper and it also works.

PR #196, opened 2018-01-16
Project ROI uses ICO starting prices

MilanRusev:


lib/sanbase/model/project.ex
```diff
@@ -116,20 +116,29 @@ defmodule Sanbase.Model.Project do
...
+    end
+  end
 
-      Store.fetch_last_known_price_point(project.ticker <> "_USD", timestamp)
+  defp token_usd_ico_price(nil, _currency_from, _ico_start_date, _current_datetime), do: nil
+  defp token_usd_ico_price(_price_from, _currency_from, nil, _current_datetime), do: nil
+  defp token_usd_ico_price(price_from, currency_from, ico_start_date, current_datetime) do
+    with :gt <- Ecto.DateTime.compare(current_datetime, Ecto.DateTime.from_date(ico_start_date)),
+        {:ok, timestamp, _} <- Ecto.Date.to_iso8601(ico_start_date) <> "T00:00:00Z" |> DateTime.from_iso8601() do
+      Store.fetch_last_known_price_point(currency_from <> "_USD", timestamp)
```
valo: I think there is no need for a `case` here. The pattern matching can happen the `with` blocks. Then there will be a shared `else` block, which returns `nil`.

lib/sanbase/model/project.ex
```diff
@@ -102,7 +102,7 @@ defmodule Sanbase.Model.Project do
       zero = Decimal.new(0)
 
       project.icos
-      |> Enum.reduce({project, 0, zero}, &roi_usd_accumulator/2)
+      |> Enum.reduce({project, 0, zero}, &roi_usd_total_paid_accumulator/2)
```
valo: After our conversations from yesterday, I was wondering if it will be possible to convert this complex accumulation into a several step process with `map`, `filter` and `sum`? Something like `project.icos |> map |> reject(&is_nil/1)`. Then the result can be counted and summed.

lib/sanbase/model/project.ex
```diff
@@ -101,13 +101,17 @@ defmodule Sanbase.Model.Project do
...
+      paid_by_ico
+      |> Enum.reduce(zero, &Decimal.add/2)
       |> case do
-        {_, _, ^zero} -> nil
-        {_, count, total_paid} ->
+        ^zero -> nil
+        total_paid ->
           project.latest_coinmarketcap_data.price_usd
-          |> Decimal.mult(Decimal.new(count))
+          |> Decimal.mult(Decimal.new(length(paid_by_ico)))
```
valo: Why do we multiply by the number of ICOs here? Shouldn't we divide the current MCAP by the sum all the money collected by all ICOs? ROI = MCAP / initial_money_collected

MilanRusev: Yes there is a bug, the prices should be weighed by the amount of tokens sold in each ICO

valo:
Please, resolve the merge conflicts

PR #198, opened 2018-01-16
Improve the parsing of the etherscan pages

valo:
If there is an error while fetching the page, return nil and don't not
try to parse the HTML.

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap.ex
+++ b/lib/sanbase/external_services/coinmarketcap.ex
@@ -57,7 +57,17 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
     Task.Supervisor.async_stream_nolink(
       Sanbase.TaskSupervisor,
       Repo.all(query),
-      &fetch_project_data/1,
+      &fetch_project_info/1,
+      ordered: false,
+      max_concurrency: 5,
+      timeout: 60_000
+    )
+    |> Stream.run()
+
+    Task.Supervisor.async_stream_nolink(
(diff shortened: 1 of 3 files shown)


valo:
This should fix https://favro.com/card/bdbb4f24afc48b29c74f4fa4/San-1068

lib/sanbase/external_services/etherscan/scraper.ex
```diff
@@ -14,14 +14,21 @@ defmodule Sanbase.ExternalServices.Etherscan.Scraper do
   @max_redirects 10
 
   def fetch_address_page(address) do
-    %Tesla.Env{status: 200, body: body} = get("/address/#{address}")
-
-    body
+    case get("/address/#{address}") do
+      %Tesla.Env{status: 200, body: body} ->
+        body
+      response ->
```
IvanIvanoff: Here you can pattern match `%Tesla{status: status, body: body}` and write a more clear log similar to `Error with status #{status} fetching html page: #{body}`

lib/sanbase/external_services/coinmarketcap.ex
```diff
@@ -57,7 +57,17 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
     Task.Supervisor.async_stream_nolink(
       Sanbase.TaskSupervisor,
       Repo.all(query),
```
IvanIvanoff: Run the query once outside both task supervisors to avoid accessing the DB twice

PR #204, opened 2018-01-17
Record in/out transactions

IvanIvanoff:
Deployable but the influxdb db should be cleared once the etherbi transactions bug is fixed.

Fetch the transactions from etherbi and store them in time series data base.

Note: Has to be merged and deployed after https://github.com/santiment/sanbase2/pull/209


config/config.exs
```diff
@@ -58,6 +58,12 @@ config :sanbase, Sanbase.ExternalServices.TwitterData.Store,
   pool: [ max_overflow: 10, size: 20 ],
   database: "twitter_followers_data"
 
+config :sanbase, Sanbase.Etherbi.Store,
+  host: {:system, "IFNLUXDB_HOST", "localhost"},
```
valo: There is a typo in the name of the variable

lib/sanbase/etherbi/funds_movements.ex
```diff
@@ -0,0 +1,58 @@
...
+  @etherbi_url Config.module_get(Sanbase.Etherbi, :url)
+  @http_client Mockery.of("HTTPoison")
+
+  def transactions_in(wallets, from, to) do
+    from_unix = DateTime.to_unix(from, :seconds)
+    to_unix = DateTime.to_unix(to, :seconds)
+
+    url =
+      "#{@etherbi_url}/transactions_in?&from_timestamp=#{from_unix}&to_timestamp=#{to_unix}&wallets=#{
+        inspect(wallets)
```
valo: I think `Poison.encode!` will make more sense than using `inspect`. Also, passing JSON as GET parameter is a bit weird. Can't we have some escaping issues with this? Another problem with using GET parameters for encoding potentially long data is the URL length limit. I would suggest to use POST for such kind of data, which could potentially become quite long

lib/sanbase/etherbi/funds_movements.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def transactions_out(wallets, from, to) do
+    from_unix = DateTime.to_unix(from, :seconds)
+    to_unix = DateTime.to_unix(to, :seconds)
+
+    wallets = Poison.encode!(wallets)
+
+    url =
+      "#{@etherbi_url}/transactions_in?from_timestamp=#{from_unix}&to_timestamp=#{to_unix}&wallets=#{
```
valo: It is possible to pass the GET parameters as `params` to the `get` function: https://hexdocs.pm/httpoison/HTTPoison.html#request/5

lib/sanbase/github/store.ex
```diff
@@ -88,11 +88,14 @@ defmodule Sanbase.Github.Store do
...
-          |> Kernel.trunc()
-      end
+      activity =
+        if is_float(activity) do
+          activity
+            |> Float.ceil()
+            |> Kernel.trunc()
+        else
+          activity
+        end
```
valo: Why the data should be truncated here?

IvanIvanoff: This should not be in this PR

lib/sanbase/etherbi/funds_movements.ex
IvanIvanoff: For now I'm only using it with one wallet, so it won't get big. For multiple wallets passing a single from and to timestamps won't work well and has to be reworked

lib/sanbase/etherbi/store.ex
```diff
@@ -0,0 +1,80 @@
...
+    |> parse_transactions_time_series()
+  end
+
+  defp transactions_from_to_query(measurement, from, to, _resolution, transaction_type) do
+
+    ~s/SELECT volume, token FROM "#{measurement}"
+    WHERE transaction_type = '#{transaction_type}'
+    AND time >= #{DateTime.to_unix(from, :nanoseconds)}
+    AND time <= #{DateTime.to_unix(to, :nanoseconds)}/
+    # GROUP BY time(#{resolution}) fill(none)/
```
valo: Let's not keep commented out code around

lib/sanbase/etherbi/transactions.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  def init(:ok) do
+    if Config.get(:sync_enabled, false) do
+      Store.create_db()
+
+      update_interval_ms = Config.get(:update_interval, @default_update_interval)
+      token_decimals = build_token_decimals_map()
+
+      GenServer.cast(self(), :sync)
+      {:ok, %{update_interval_ms: update_interval_ms, token_decimals: token_decimals}}
```
valo: If we keep the `token_decimals` in the state of the process, what is going to happen if a new project gets added through the admin panel?

IvanIvanoff: We can easily rebuild this map in `handle_cast`. The bigger problem I see though is when we add a new project what will we do with the past transactions. We would check for transactions after `last_datetime`, but there would be transactions before that time with that new token.

lib/sanbase/etherbi/transactions.ex
```diff
@@ -0,0 +1,108 @@
...
+  # number of decimal places `nil` is written instead and it gets filtered by the Store.import()
+  defp convert_to_measurement(
+         transactions_data,
+         measurement_name,
+         transaction_type,
+         token_decimals
+       ) do
+    transactions_data
+    |> Enum.map(fn {datetime, volume, _address, token} ->
+      if decimal_places = Map.has_key?(token_decimals, token) do
```
valo: Can we filter out the tokens without decimals before fetching the transactions and avoiding fetching the transactions at all? That should remove this `if`

lib/sanbase_web/graphql/schema.ex
```diff
@@ -129,6 +129,16 @@ defmodule SanbaseWeb.Graphql.Schema do
       resolve(&EtherbiResolver.transaction_volume/3)
     end
 
+    @desc "Shows the flow of funds in an exchange wallet"
+    field :transactions, list_of(:transaction) do
+      arg(:wallet, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, :datetime, default_value: DateTime.utc_now())
+      arg(:interval, :string, default_value: "1h")
+      arg(:transaction_type, :string)
```
valo: Can we use an `enum` for this? https://hexdocs.pm/absinthe/Absinthe.Schema.Notation.html#enum/3

lib/sanbase/etherbi/store.ex
IvanIvanoff: done

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: I tried but I did not like that it has to be used without quotes and uppercase. It just differs too much from everything else. Example: `transaction_type: IN`. Though, that's only a style preference and could be changed if found to be the better option

valo:
What is remaining to be done on this one?

IvanIvanoff:
Wrong transactions will be fetched from etherbi if we deploy before the bugfix is applied there

valo:
Can we deploy this and then clear the influxDB cache after the transactions fix is in place?

IvanIvanoff:
Yes, we can

lib/sanbase/etherbi/transactions.ex
```diff
@@ -0,0 +1,131 @@
+defmodule Sanbase.Etherbi.Transactions do
+  @moduledoc ~S"""
+    This module is a GenServer that periodically sends requests to etherbi API.
+    In and out transactions are fetched and saved in a time series database for
+    easier aggregation and querying.
+  """
```
valo: Can we add a function here, which will purge the whole cache? We will need this function when we add new exchange wallets and when we fix bugs, like the invalid transaction filtering bug.

IvanIvanoff: There are `list_measurements` and `drop_measurement` in the `Store` module, but they are intended to be used in migrations. Do you mean such purging or one that can happen at runtime?

valo: The use case will be:

1. Garry adds/removes a new exchange wallet in the admin panel
2. The cache is cleared automatically and starts repopulating immediately

I am not sure how we will hook up the add/remove event from the admin to he cache purge, but in any case we will need some function to call to do it.

IvanIvanoff: There are `list_measurements` and `drop_measurement` in the `Store` module which maybe can be used in a migration. Do you mean this kind of purge or one that can happen on runtime? For the later we can develop into a complete tool the Elixir mini-project I used to import the data from staging to prod without downtime.

valo: That will work. So no need for a special function

PR #206, opened 2018-01-18
Select the first record with matching ticker and not empty twitter link

IvanIvanoff:
`Repo.get_by(Project, ticker: ticker)` gets confused when there are more than one record with matching tickers. Change the code to get the first result with matching ticker and not empty twitter link

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/twitter_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/twitter_resolver.ex
@@ -3,22 +3,31 @@ defmodule SanbaseWeb.Graphql.Resolvers.TwitterResolver do
   alias Sanbase.Model.Project
   alias Sanbase.ExternalServices.TwitterData.Store
 
+  import Ecto.Query
+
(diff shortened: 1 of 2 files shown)


lib/sanbase_web/graphql/resolvers/twitter_resolver.ex
```diff
@@ -3,22 +3,31 @@ defmodule SanbaseWeb.Graphql.Resolvers.TwitterResolver do
...
+
   def twitter_data(_root, %{ticker: ticker}, _resolution) do
-    with %Project{twitter_link: twitter_link} <- Repo.get_by(Project, ticker: ticker),
+    with query <-
+           from(
+             p in Project,
+             where: p.ticker == ^ticker and not is_nil(p.twitter_link),
+             select: p.twitter_link
+           ),
+         [twitter_link|_] <- Repo.all(query),
```
valo: May be it is worth extracting this in a separate function for reuse?

PR #210, opened 2018-01-22
Do not try to send slack notification if there is no webhook url

valo:
If the webhook URL is not set, don't send notifications to slack.


IvanIvanoff:
In other places we enable/disable the services via a special bool param in the config. It's more convinient as you have to set it to true/false and not go search for the correct link to enable it if it was disabled.

valo:
Fair point. Added a properly named flag for this.

PR #212, opened 2018-01-22
Don't filter out projects without coinmarketcap_id when querying Project Transparency

MilanRusev:


lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -25,7 +25,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     end
 
     query = from p in Project,
-    where: not is_nil(p.coinmarketcap_id)
+    where: (^only_project_transparency or not is_nil(p.coinmarketcap_id))
```
valo: Can we have 2 different queries depending on the value of `only_project_transparency`? It is quite hard to figure out this boolean statement in the query at the moment.

PR #215, opened 2018-01-23
GraphQL APIs for adding posts, voting for them and getting the current ranking

valo:
This PR adds GraphQL APIs for adding new posts and voting/unvoting for them.

Each new post will need to be approved from the admin panel first, before it can be accessed through the API.

When voting for a post, the amount of votes will be the amount of SAN tokens the user has. The amount of SAN tokens is being updated every 5 min when they are requested.


lib/sanbase/auth/user.ex
```diff
@@ -1,15 +1,21 @@
...
 
   @salt_length 64
 
+  @san_balance_cache_seconds 60 * 5 # 5 minutes
+
   schema "users" do
     field(:email, :string)
     field(:username, :string)
     field(:salt, :string)
+    field(:san_balance, :integer)
```
IvanIvanoff: Isn't `decimal` more suitable for that field?

valo: I need to double check if we are not keeping this number in raw format, which will be a big int.

lib/sanbase/voting/poll.ex
```diff
@@ -0,0 +1,55 @@
...
+    has_many(:posts, Post)
+
+    timestamps()
+  end
+
+  def changeset(%Poll{} = poll, attrs \\ %{}) do
+    poll
+    |> cast(attrs, [:start_at, :end_at])
+    |> validate_required([:start_at, :end_at])
+    |> unique_constraint(:start_at)
```
IvanIvanoff: Why do we need unique constraint on `:start_at`?

valo: It ensures that 2 concurrent calls for getting `current_poll` won't create 2 current polls

lib/sanbase/voting/post.ex
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def changeset(%Post{} = post, attrs \\ %{}) do
+    post
+    |> cast(attrs, [:title, :link])
+    |> validate_required([:poll_id, :user_id, :title, :link])
+    |> unique_constraint(:poll_id, name: :posts_poll_id_title_index)
+  end
+
+  def admin_changeset(%Post{} = post, attrs \\ %{}) do
```
IvanIvanoff: This is not used anywhere

lib/sanbase_web/graphql/schema/voting_types.ex
```diff
@@ -0,0 +1,30 @@
+defmodule SanbaseWeb.Graphql.VotingTypes do
+  use Absinthe.Schema.Notation
+
+  alias SanbaseWeb.Graphql.Resolvers.VotingResolver
+
+  object :poll do
+    field(:start_at, non_null(:datetime))
+    field(:end_at, non_null(:datetime))
+    field(:posts, list_of(:post)) do
+      resolve fn poll, _, _ ->
```
IvanIvanoff: Put that resolver in `VotingResolver` for consistency

lib/sanbase_web/graphql/resolvers/voting_resolver.ex
```diff
@@ -0,0 +1,74 @@
...
+  def total_san_votes(%Post{} = post, _args, _context) do
+    total_san_votes =
+      post
+      |> Repo.preload(votes: [:user])
+      |> Map.get(:votes)
+      |> Enum.map(&Map.get(&1, :user))
+      |> Enum.map(&User.san_balance!/1)
+      |> Enum.reduce(Decimal.new(0), &Decimal.add/2)
+
+    {:ok, total_san_votes}
```
IvanIvanoff: This will return without dividing by 10^18 to account for the decimals places. Lets discuss where it will be best to have that (frontend or backend)

lib/sanbase_web/graphql/resolvers/voting_resolver.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  def vote(_root, %{post_id: post_id}, %{
+        context: %{auth: %{current_user: user}}
+      }) do
+    %Vote{}
+    |> Vote.changeset(%{post_id: post_id, user_id: user.id})
+    |> Repo.insert()
+    |> case do
+      {:ok, _vote} -> {:ok, Repo.get(Post, post_id)}
+      {:error, error} -> {:error, error}
```
IvanIvanoff: This error is of type changeset which Absinthe cannot interpret. Fastest (and dirtiest) solution is to put it inside `inspect`.

lib/sanbase_web/graphql/resolvers/voting_resolver.ex
```diff
@@ -0,0 +1,74 @@
...
+        context: %{auth: %{current_user: user}}
+      }) do
+    %Post{user_id: user.id, poll_id: Poll.find_or_insert_current_poll!().id}
+    |> Post.changeset(post_args)
+    |> Repo.insert()
+    |> case do
+      {:ok, post} ->
+        {:ok, post |> Repo.preload([:votes, :user])}
+
+      {:error, %{errors: errors}} ->
```
IvanIvanoff: Same as above. Better handle the changeset error like: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/account_resolver.ex#L52

PR #219, opened 2018-01-25
Test drone builds

valo:


na--:
Included in https://github.com/santiment/sanbase2/pull/308

PR #221, opened 2018-01-26
Add deletePost API and fix cascading deletes

valo:
Users can delete their posts now. When a post is deleted all the
attached votes will be deleted too.

Cacading deletes will happen for posts when deleting polls and users.

Also changed the currentPoll API to include the unapproved posts, as we
will filter these out on the client side.


lib/sanbase_web/graphql/schema/voting_types.ex
```diff
@@ -6,10 +6,7 @@ defmodule SanbaseWeb.Graphql.VotingTypes do
   object :poll do
     field(:start_at, non_null(:datetime))
     field(:end_at, non_null(:datetime))
-
-    field :posts, list_of(:post) do
-      resolve(&VotingResolver.approved_posts/3)
-    end
+    field(:posts, list_of(:post))
```
IvanIvanoff: Aren't we showing not approved posts this way and allow voting for them?

valo: The posts will be filtered on the client side. This is needed, so that users can delete their posts immediately after they submitted them.

IvanIvanoff: That way it will be possible for someone to see not approved posts. Can't we just list all aproved posts + your post? Something like `where not is_nil(approved_at) OR user.id == ^current_user.id` and avoid all filtering in the frontend

valo: I think it is not a big deal that we will return all unapproved posts in the API. Even if someone inspects the network requests, what is the problem? Even if they vote for an unapproved posts, if the post is a scam it won't participate in the final poll

IvanIvanoff: There is no problem either way. It's just a small change that reduces the amount of work needed in the frontend

valo: @Partysun how do you prefer this to be done:

* To return all posts and filter the approved in the frontend

or

* To return all approved posts + all unapproved posts that belong to the current user

PR #222, opened 2018-01-26
Refactor parts of Store modules. Replace fetching last datetime from using LAST to using ORDER BY LIMIT 1

IvanIvanoff:
Refactor the `Store` modules. Make generic `first_datetime` and `last_datetime` functions to be applicable to all db layouts.

Note: https://github.com/influxdata/docs.influxdata.com/issues/1095
When using wildcard inside a selector, the time is not picked but showed as beginning of epoch (0 unix). A workaround to be able to select the last/first datetime without knowing the field/tag names is to use `ORDER BY ... LIMIT 1`

### DIFF
--- a/lib/sanbase/etherbi/store.ex
+++ b/lib/sanbase/etherbi/store.ex
@@ -4,32 +4,6 @@ defmodule Sanbase.Etherbi.Store do
   alias Sanbase.Influxdb.Measurement
   alias Sanbase.Etherbi.Store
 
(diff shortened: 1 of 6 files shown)


valo:
I am not sure I understand when we use `LAST(*)` and then `ORDER LIMIT 1`. Can you explain please?

IvanIvanoff:
Should only use `ORDER BY time LIMIT 1`. Fixed.

PR #225, opened 2018-01-28
Implement etherbi fetcher logic in a module to be resued

IvanIvanoff:
NOTE: When adding the new env variables rename `ETHERBI_TRANSACTIONS_ENABLED` to `ETHERBI_TRANSACTIONS_SYNC_ENABLED`.

This PR introduces a module `EtherbiGenServer` which injects behaviour that abstracts away the repeating code for starting and initializing the genserver, creating influxdb and repeating the work at given time intervals.


config/test.exs
```diff
@@ -50,7 +50,7 @@ config :sanbase, Sanbase.Github.Store,
 config :sanbase, Sanbase.ExternalServices.TwitterData.Store,
   database: "twitter_followers_data_test"
 
-config :sanbase, Sanbase.Etherbi.Store,
+config :sanbase, Sanbase.Etherbi.Transactions.STore,
```
valo: Typo: `STore` -> `Store`

IvanIvanoff: Fixed but still have not pushed it

lib/sanbase/etherbi/burn_rate.ex
```diff
@@ -0,0 +1,62 @@
...
+
+  alias Sanbase.Etherbi.BurnRate.Store
+  alias Sanbase.Repo
+  alias Sanbase.Model.Project
+
+  def work() do
+    # Precalculate the number by which we have to divide, that is pow(10, decimal_places)
+    token_decimals = build_token_decimals_map()
+
+    query = from(p in Project, where: not is_nil(p.ticker), select: p.ticker)
```
valo: I think we can narrow down the query here. We can try to extract this query https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/project_resolver.ex#L55 into a common place and reuse that

lib/sanbase/etherbi/transaction_volume.ex
```diff
@@ -0,0 +1,66 @@
...
+      tickers,
+      &fetch_and_store(&1, token_decimals),
+      max_concurency: 1,
+      timeout: 170_000
+    )
+    |> Stream.run()
+  end
+
+  def fetch_and_store(ticker, token_decimals) do
+    with {:ok, transaction_volumes} <- Fetcher.transaction_volume(ticker) do
```
valo: I think we can skip fetching the `transaction_volume` if we don't have the `token_decimals` for the project, right? We won't store any data in that case anyway.

IvanIvanoff: Or even better - add `not is_nil(p.token_decimals)` when fetching the tickers and avoid explicit checking here.

valo: Yep. Let's do it like this

lib/sanbase/etherbi/burn_rate.ex
```diff
@@ -0,0 +1,66 @@
...
+  # Better return no information than wrong information. If we have no data for the
+  # number of decimal places `nil` is written instead and it gets filtered by the Store.import()
+  defp convert_to_measurement(
+         burn_rates,
+         ticker,
+         token_decimals
+       ) do
+    burn_rates
+    |> Enum.map(fn {datetime, burn_rate} ->
+      if decimal_places = Map.get(token_decimals, ticker) do
```
valo: Do we need this check now?

IvanIvanoff: Not really, pushed a change.

PR #226, opened 2018-01-29
Add voted_at field for each post

valo:
The `voted_at` field gives the datetime when the current user voted for
a given post.

Also fixed the total_san_votes to preload the eth_accounts of the user,
so that the san balance can be calculated of the users.


valo:
Merging as @Partysun is waiting on the fix

PR #231, opened 2018-01-30
Login using email links

valo:
Add GraphQL APIs for requesting an email with a login link and for validating a login token and email.

The login links are valid for 24 hours.

The emails are being sent using mandrill, using a predefined template there. We need to get the mandrill transactional emails enabled and test this locally.


config/config.exs
```diff
@@ -199,6 +200,10 @@ config :ex_aws,
 
 config :sanbase, Sanbase.Etherbi, url: {:system, "ETHERBI_URL"}
 
+config :sanbase, Sanbase.MandrillApi,
+  apikey: {:system, "MANDRILL_APIKEY"},
+  from_email: {:system, "MANDRILL_FROM_EMAIL", "admin@santiment.net"}
```
IvanIvanoff: Wouldn't it be better if we had `login@santiment.net` or something like that?

valo: For now we haven't registered such email address and I haven't explored how email forwarding works exactly. Is there any particular reason we might want to change it now?

lib/sanbase/auth/user.ex
```diff
@@ -63,4 +80,58 @@ defmodule Sanbase.Auth.User do
...
+    cond do
+      user.email_token != token ->
+        false
+
+      Timex.diff(Timex.now(), user.email_token_generated_at, :minutes) >
+          @login_email_valid_minutes ->
+        false
+
+      user.email_token_validated_at == nil ->
+        true
```
IvanIvanoff: Should return `false` if the token is not validated?

valo: If the token is never validated and everything from the above is not true, we should consider it valid.

config/config.exs
IvanIvanoff: No, just aesthetic reasons

lib/sanbase/mandrill_api.ex
```diff
@@ -0,0 +1,50 @@
...
+      template_name: template,
+      template_content: [],
+      message: %{
+        to: [
+          %{
+            email: recepient
+          }
+        ],
+        global_merge_vars: build_global_merge_vars(variables),
+        from_email: Config.module_get(__MODULE__, :from_email),
```
IvanIvanoff: Fetching from the same module is shorter with `Config.get(:from_email)`.

lib/sanbase_web/endpoint.ex
```diff
@@ -58,4 +60,12 @@ defmodule SanbaseWeb.Endpoint do
       {:ok, config}
     end
   end
+
+  def website_url() do
+    Config.module_get(__MODULE__, :website_url)
```
IvanIvanoff: `Config.get(:website_url)`

lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -39,6 +39,27 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
+  end
+
+  def email_login_verify(%{token: token, email: email}, _resolution) do
+    with {:ok, user} <- User.find_or_insert_by_email(email),
+         true <- User.email_token_valid?(user, token),
+         {:ok, token, _claims} <- SanbaseWeb.Guardian.encode_and_sign(user, %{salt: user.salt}),
+         {:ok, user} <- User.mark_email_token_as_validated(user) do
+      {:ok, %{user: user, token: token}}
+    else
+      _ -> {:error, :login_failed}
```
IvanIvanoff: Unify the error messages. IMO the second argument being just a string is most readable (use `message` when you want to specify `details`, too)

test/sanbase/auth/user_test.exs
```diff
@@ -62,4 +62,71 @@ defmodule Sanbase.Auth.UserTest do
...
+
+  test "update_email_token updates the email_token and the email_token_generated_at" do
+    user =
+      %User{salt: User.generate_salt()}
+      |> Repo.insert!()
+
+    {:ok, user} = User.update_email_token(user)
+
+    assert user.email_token != nil
+    assert Timex.diff(Timex.now(), user.email_token_generated_at, :seconds) == 0
```
IvanIvanoff: Could this fail if the VM decides to switch off this process and return after more than 1 second?

valo: Switching off the process for more than 1 sec seems very unlikely to me. If we start getting random failures in the builds we can change this.

PR #243, opened 2018-02-01
Fix the emailLogin GraphQL API call

valo:
Add a special return type for the emailLogin API call. Also verify with
tests that it is working properly.

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/account_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/account_resolver.ex
@@ -45,7 +45,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
          {:ok, _user} <- User.send_login_email(user) do
       {:ok, %{success: true}}
     else
-      error -> {:error, message: "Can't login"}
+      _ -> {:error, message: "Can't login"}
     end
   end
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -39,13 +39,13 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
-    with {:ok, user} <- User.find_or_insert_by_email(email, username),
+  def email_login(%{email: email} = args, _resolution) do
+    with {:ok, user} <- User.find_or_insert_by_email(email, args[:username]),
          {:ok, user} <- User.update_email_token(user),
          {:ok, _user} <- User.send_login_email(user) do
-      {:ok, :success}
+      {:ok, %{success: true}}
     else
-      _ -> {:error, message: "Can't login"}
+      error -> {:error, message: "Can't login"}
```
IvanIvanoff: Did you intended to use this `error` as part of the message?

valo: Yeah. I was debugging something. I'll return it to the old value.

PR #245, opened 2018-02-02
Allow to set moderation comments for the posts

valo:
Change the Post to have a field `state` which holds if the post is
approved or declined. Add a moderation_comment column, which will
contain the comment from the person that declined the post.


lib/sanbase/voting/post.ex
```diff
@@ -6,28 +6,39 @@ defmodule Sanbase.Voting.Post do
...
     field(:link, :string)
-    field(:approved_at, Timex.Ecto.DateTime)
+    field(:state, :string)
+    field(:moderation_comment, :string)
 
     timestamps()
   end
 
-  def changeset(%Post{} = post, attrs \\ %{}) do
+  def create_changeset(%Post{} = post, attrs) do
```
IvanIvanoff: This is used only in the tests

valo: I thought that I changed that. Let me fix it

valo: Changed it. GQL is making sanitization of the parameters, but I decided to have 2 separate change sets just in case.

PR #253, opened 2018-02-08
Tech indicators GQL api

MilanRusev:
Devops PR for TECH_INDICATORS_URL env var: https://github.com/santiment/devops/pull/114


lib/sanbase_web/graphql/schema.ex
```diff
@@ -161,6 +163,43 @@ defmodule SanbaseWeb.Graphql.Schema do
 
       resolve(&EtherbiResolver.exchange_fund_flow/3)
     end
+
+    @desc "MACD for a ticker and given currency and time period"
+    field :macd, list_of(:macd) do
+      arg(:ticker, non_null(:string))
+      @desc "Currently supported: USD, BTC"
+      arg(:currency, non_null(:string))
```
IvanIvanoff: You can do that with enum object: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/schema/etherbi_types.ex#L20

MilanRusev: I removed the enum deliberately - it becomes too complicated - it has problems with conversions to proper cases when propagating the value to the python api; also if we allow arbitrary currencies on future we'll have to change the api

IvanIvanoff:
Please run `mix format` on this PR

lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+  def price_volume_diff(ticker, currency, from_datetime, to_datetime, aggregate_interval) do
+    from_unix = DateTime.to_unix(from_datetime, :nanoseconds)
+    to_unix = DateTime.to_unix(to_datetime, :nanoseconds)
+
+    url =
+      "#{tech_indicators_url()}/indicator/pricevolumediff?ticker=#{ticker}&currency=#{currency}&from_timestamp=#{
+        from_unix
+      }&to_timestamp=#{to_unix}&aggregate_interval=#{aggregate_interval}"
```
valo: It is possible to pass these parameters without string interpolation, but as a structure using `params` https://hexdocs.pm/httpoison/HTTPoison.html#request/5

lib/sanbase_web/admin/model/project.ex
```diff
@@ -51,7 +52,8 @@ defmodule Sanbase.ExAdmin.Model.Project do
 
       panel "BTC Addresses" do
         markup_contents do
-          a ".btn .btn-primary", href: "/admin/project_btc_addresses/new?project_id=" <> to_string(project.id) do
+          a ".btn .btn-primary",
+            href: "/admin/project_btc_addresses/new?project_id=" <> to_string(project.id) do
```
valo: I think this is part of another PR. May be merging master will remove it.

MilanRusev: This is a result of running `mix format`. In the other PR it hasn't been run.

PR #254, opened 2018-02-08
Fetch all transactions from etherscan and save them in time series. Keep saving the latest data to the pg table

IvanIvanoff:
Save all transactions for wallets with `isError == 1 && trxreception_status == 0 && value > 0` from etherscan to influxdb grouped  by the project they belong to.

Keep the old behaviour, too. Save the latest eth wallet data in the corresponding pg table until(if) its usage is removed. 
Do both things with a single query so the rate limit does not require changes.

Keep a separate measurement for tracking the last checked block number. Implement it this way because if (for example) a contract address is queried it could end up with only transactions with `value = 0` and no points will be inserted in the db. In that case we do not know which starting block to choose for the next query.

### DIFF
--- a/lib/sanbase/external_services/etherscan/requests/tx.ex
+++ b/lib/sanbase/external_services/etherscan/requests/tx.ex
@@ -30,7 +30,9 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
       address: address,
       startblock: startblock,
       endblock: endblock,
-      sort: "desc"
+      sort: "asc",
+      page: 1,
+      offset: 2500
     ]
   end
 
--- a/lib/sanbase/external_services/etherscan/requests/tx.ex
+++ b/lib/sanbase/external_services/etherscan/requests/tx.ex
@@ -1,6 +1,11 @@
 defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
+  @moduledoc ~s"""
+    Module for fetching transactions from etherscan.io for a given address.
+  """
   alias Sanbase.ExternalServices.Etherscan.Requests
   alias __MODULE__
+
+  require Logger
 
   defstruct [
     :blockNumber,
@@ -23,6 +28,32 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
     :confirmations
   ]
 
+  @doc ~s"""
+    Issues a HTTP GET request to etherscan.io api to get the transactions from
+    `startblock` to `endblock` in ascending order.
+    Returns `{:ok, list()}` on success, `{:error, String.t}` otherwise
+  """
+  @spec get_all_transactions(String.t(), Integer.t(), Integer.t()) ::
+          {:ok, list()} | {:error, String.t()}
+  def get_all_transactions(address, startblock, endblock) do
+    String.downcase(address) |> get(startblock, endblock)
+  end
+
+  # Private functions
+
+  defp get(address, startblock, endblock) do
+    Requests.get("/", query: get_query(address, startblock, endblock))
+    |> case do
+      %{status: 200, body: body} ->
+        {:ok, parse_tx_json(body)}
+
+      %{status: status, body: body} ->
+        error = "Error fetching transactions for #{address}. Status code: #{status}: #{body}"
+        Logger.warn(error)
+        {:error, error}
+    end
+  end
+
   defp get_query(address, startblock, endblock) do
     [
       module: "account",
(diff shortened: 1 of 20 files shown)


lib/sanbase/external_services/etherscan/requests/tx.ex
```diff
@@ -60,4 +60,9 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
       String.downcase(tx.from) == normalized_address
     end)
   end
+
+  def get_all_transactions(address, startblock, endblock) do
+    normalized_address = String.downcase(address)
+    {:ok, get(address, startblock, endblock)}
```
valo: BTW, we might need to do pagination here, as the API returns a max of 10000 transactions: https://etherscan.io/apis#accounts

IvanIvanoff: Hm, can we use `sort: "asc"` (combined with the check of the last block number we have record for) in the query and just let it return multiple times 10000 results until all transactions are fetched?

valo: yeah. that should work. I doubt there are more than 10000 transactions for an address in a single block

lib/sanbase/external_services/etherscan/requests/tx.ex
```diff
@@ -23,25 +28,45 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
     :confirmations
   ]
 
+  @doc ~s"""
+    Issues a HTTP GET request to etherscan.io api to get the transactions from
+    `startblock` to `endblock` in ascending order.
+    Returns `{:ok, list()}` on success, `{:error, String.t}` otherwise
+  """
+  @spec get_all_transactions(String.t(), Integer.t(), Integer.t()) ::
+          {:ok, list()} | {:error, String.t()}
```
valo: We should think if we can have a tool to check the type specs against our code and find bugs.

lib/sanbase/external_services/etherscan/requests/tx.ex
```diff
@@ -23,25 +28,45 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Tx do
...
     [
       module: "account",
       action: "txlist",
       address: address,
       startblock: startblock,
       endblock: endblock,
-      sort: "desc"
+      sort: "asc",
+      page: 1,
+      offset: 2500
```
valo: Why there is a hardcoded offset here?

lib/sanbase/external_services/etherscan/worker.ex
```diff
@@ -96,13 +121,42 @@ defmodule Sanbase.ExternalServices.Etherscan.Worker do
...
+    end
+  end
+
+  defp fetch_and_store(%{address: address, coinmarketcap_id: id}, endblock) do
+    transactions = fetch_all_transactions(address, id, endblock)
+
+    filtered_transactions =
+      transactions
+      |> Enum.reject(fn %Tx{isError: error, txreceipt_status: status, value: value} ->
+        error == "1" || status == "0" || value == "0"
```
valo: Why `error` should be 1? Doesn't that mean that the transaction was not successful?

valo: Oh. this is a reject. Just show it now.

lib/sanbase/influxdb/store.ex
```diff
@@ -61,6 +67,23 @@ defmodule Sanbase.Influxdb.Store do
...
+        database = Sanbase.Utils.Config.get(:database)
+
+        database
+        |> Sanbase.Utils.Config.get()
+        |> Instream.Admin.Database.create()
+        |> __MODULE__.execute()
+
+        Instream.Admin.RetentionPolicy.create(name, database, duration, replication, default)
+        |> __MODULE__.execute()
+      end
```
valo: Where is this function used?

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -95,6 +93,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     end)
   end
 
+  def eth_spent(%Project{id: id}, %{days: days}, _resolution) do
+    async(fn ->
+      coinmarketcap_id =
+        Repo.one(from(p in Project, where: p.id == ^id, select: p.coinmarketcap_id))
```
valo: It should be possible to get the `coinmarketcap_id` from the project passed to the `eth_spent` function.

lib/sanbase/external_services/etherscan/requests/tx.ex
IvanIvanoff: dialyzer, but we should start writing more specs for it to be effective

lib/sanbase/external_services/etherscan/requests/tx.ex
IvanIvanoff: The default (and max) is 10000 which used to timeout sometimes. On a second thought I increasted the adapter's timeout..

lib/sanbase/influxdb/store.ex
IvanIvanoff: First I thought that we can use a retention policy to keep just the last 30 days of data. But these transactions are not many so I switched to keeping them all so I can remove it

PR #255, opened 2018-02-08
Add new GQL fields for the project overview

valo:
Add these fields:
* volume change in the last 24h
* average developer activity for the last 30 days
* twitter data for each project

Also change the percent change fields to return floats, instead of
decimals encoded as strings.

Fixed some compilation warnings and removed some optimizations that are
not really necessary.


lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -261,6 +246,35 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+      [] ->
+        {:ok, nil}
+    end
+  end
+
+  def average_dev_activity(%Project{ticker: ticker}, _args, _resolution) do
+    month_ago = Timex.shift(Timex.now(), days: -30)
+
+    case Github.Store.fetch_activity_with_resolution!(ticker, month_ago, Timex.now(), "30d") do
+      [%{"SUM(activity)" => total_activity}] ->
```
IvanIvanoff: This seems a little strange. Executing locally I get:
`[
  {#DateTime<2017-12-20 00:00:00Z>, 295},
  {#DateTime<2018-01-19 00:00:00Z>, 25192}
]`
There are two problems - this query cannot guarantee that exactly 1 result is returned. And the format is list of `{datetime, activity}`

valo: This is incorrect. I need to fix it. Good catch!

IvanIvanoff: Even if you provide `from, `to` and group by the difference `from-to` you can still get more than one aggregated record due to the influxdb behaviour. Easiest way to write a new `Store` function which returns just one result is to pass `from` and `to` and do not use `GROUP BY`, just where clauses for the time.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -67,84 +61,67 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
-          value
-      end
+  def btc_balance_from_loader(loader, project) do
+    balance =
+      loader
+      |> Dataloader.get(SanbaseRepo, :btc_addresses, project)
+      |> Enum.map(& &1.latest_btc_wallet_data)
+      |> Enum.reject(&is_nil/1)
+      |> Enum.map(& &1.balance)
+      |> Enum.reduce(Decimal.new(0), &Decimal.add/2)
```
IvanIvanoff: You can replace the first 3 `Enum`s with `Stream` (same for eth_balance)

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -261,6 +250,35 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+      [] ->
+        {:ok, nil}
+    end
+  end
+
+  def average_dev_activity(%Project{ticker: ticker}, _args, _resolution) do
+    month_ago = Timex.shift(Timex.now(), days: -30)
+
+    case Github.Store.fetch_activity_with_resolution!(ticker, month_ago, Timex.now(), "30d") do
+      {_dt, total_activity} ->
```
IvanIvanoff: This function returns a list

lib/sanbase/github/store.ex
```diff
@@ -29,6 +42,14 @@ defmodule Sanbase.Github.Store do
     GROUP BY time(#{interval}) fill(0)/
   end
 
+  defp total_activity_query(ticker, from, to) do
+    ~s/SELECT SUM(activity)
+    FROM "#{ticker}"
+    WHERE time >= #{DateTime.to_unix(from, :nanoseconds)}
+    AND time <= #{DateTime.to_unix(to, :nanoseconds)}
+    fill(none)/
```
IvanIvanoff: I think you cannot use `fill` without `GROUP BY`

IvanIvanoff: I think this `fill` does nothing without `GROUP BY`

PR #267, opened 2018-02-12
Introduce a new internal DB and fix fetching of the datetime needed for the queries

IvanIvanoff:
Note: The k8s job for clearing the transactions influxdb database must be executed.

Use a separate table for saving the last datetime we have data for `address_in` and `address_out`

### DIFF
--- a/lib/sanbase/etherbi/transactions.ex
+++ b/lib/sanbase/etherbi/transactions.ex
@@ -70,12 +70,14 @@ defmodule Sanbase.Etherbi.Transactions do
        ) do
     transactions_data
     |> Enum.map(fn {datetime, volume, address, token} ->
-      %Sanbase.Influxdb.Measurement{
-        timestamp: datetime |> DateTime.to_unix(:nanoseconds),
-        fields: %{volume: volume / Map.get(token_decimals, token)},
-        tags: [transaction_type: transaction_type, address: address],
-        name: token
-      }
+      if decimals = Map.get(token_decimals, token) do
+        %Sanbase.Influxdb.Measurement{
+          timestamp: datetime |> DateTime.to_unix(:nanoseconds),
+          fields: %{volume: volume / Map.get(token_decimals, token)},
+          tags: [transaction_type: transaction_type, address: address],
+          name: token
+        }
+      end
     end)
   end
 end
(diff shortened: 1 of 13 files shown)


lib/sanbase/etherbi/transactions.ex
```diff
@@ -70,12 +70,14 @@ defmodule Sanbase.Etherbi.Transactions do
...
-      %Sanbase.Influxdb.Measurement{
-        timestamp: datetime |> DateTime.to_unix(:nanoseconds),
-        fields: %{volume: volume / Map.get(token_decimals, token)},
-        tags: [transaction_type: transaction_type, address: address],
-        name: token
-      }
+      if decimals = Map.get(token_decimals, token) do
+        %Sanbase.Influxdb.Measurement{
+          timestamp: datetime |> DateTime.to_unix(:nanoseconds),
+          fields: %{volume: volume / Map.get(token_decimals, token)},
```
valo: I think we can use `decimals` here, not get the value from the map again.

lib/sanbase/etherbi/transactions/store.ex
```diff
@@ -32,6 +66,14 @@ defmodule Sanbase.Etherbi.Transactions.Store do
 
   # Private functions
 
+  defp select_last_address_datetime(address, trx_type) do
+    address_trx_type = "#{address}_#{trx_type}"
+
+    ~s/SELECT LAST(last_datetime)
+    FROM "#{@last_address_measurement}"
+    WHERE address_trx_type='#{address_trx_type}'/
```
valo: Shouldn't this query be scoped by ticker?

IvanIvanoff: We fetch all transactions regardless of the ticker, so we do not need to scope it that way.

For an exchange address we get all data for given time period as seen here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/etherbi/etherbi_api.ex#L167

After that we write every transaction to the ticker's corresponding measurement: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/etherbi/transactions.ex#L77

That said, we are sure that up to that point we have the data for all tickers. To be able to scope by address AND ticker it will require major rewrites both in sanbase and etherbi

PR #270, opened 2018-02-13
kc login aesthetics pt1

brighteye:
This is a first pass... there is more to do, and Yura your comments are noted on the design (marketing bullets, gradient). So more changes are coming, just wanted to get this first crack into the pipeline.

For the button colors, I ended up putting semantic UI overrides in the main index.css file - that seemed the only place they were working while I was figuring out the design, but I have a feeling this is the wrong way to do it. 

So Yura, if we need to move them, no problem. I'll post some questions to you in Discord around how we should approach the semantic UI customization...


valo:
@brighteye please, take a look at the failing build: https://jenkins.internal.santiment.net/blue/organizations/jenkins/Santiment%2Fsanbase2/detail/PR-270/1/pipeline/#step-17-log-11

Seems like there is a problem with the indentation of the code. This is because we have a checker for code style and I think it is complaining for incorrect indentation in one place.

PR #274, opened 2018-02-14
Fetch all normal and internal transactions from etherscan

IvanIvanoff:
Rework the etherscan worker. Now fetch all transactions and save them in time series db

Note: Revisit the code once the new overview page is deployed and remove all stuff related to `LatestEthWalletData`


valo:
Let's merge and deploy tomorrow in case something goes foobar

PR #279, opened 2018-02-15
Add migration for removing duplicate wallet data. Migrate all address columns to case insensitive format.

IvanIvanoff:
This migration:

1. Converts ProjectEthAddress' address to case insensitive.
2. Removes all duplicates from LatestEthWalletData that were introduced with a previous PR. In this previous PR we started downcasing the address before importing it, so it was doubled. Remove all records for which the address is seen more than once (compared case insensitive) and leave the one which is lowercased (for consistency).

3. Do the same with the corresponding BTC tables.


priv/repo/migrations/20180215105804_remove_duplicates.exs
```diff
@@ -0,0 +1,84 @@
...
+  import Ecto.Query
+  alias Sanbase.Repo
+
+  alias Sanbase.Model.{
+    LatestEthWalletData,
+    LatestBtcWalletData
+  }
+
+  def up do
+    execute("CREATE EXTENSION IF NOT EXISTS citext WITH SCHEMA public;")
```
valo: This extension seems to be supported by RDS, but still we should be aware that extensions could be problematic in RDS.

priv/repo/migrations/20180215105804_remove_duplicates.exs
```diff
@@ -0,0 +1,84 @@
...
+    Repo.insert_all(LatestBtcWalletData, filtered_btc_data)
+  end
+
+  defp count_address(map, addr) do
+    Enum.count(map, fn %LatestEthWalletData{address: address} ->
+      String.downcase(addr) == String.downcase(address)
+    end)
+  end
+
+  defp count_address(map, addr) do
```
valo: This function will be shadowed by the previous one

valo:
Can we test this on a dump of staging first?

priv/repo/migrations/20180215105804_remove_duplicates.exs
IvanIvanoff: Yeah, the pattern match should be in function declaration. We should update our `seeds.exs` to be able to catch search errors earlier.

priv/repo/migrations/20180215105804_remove_duplicates.exs
```diff
@@ -0,0 +1,84 @@
...
+
+    migrate_btc()
+  end
+
+  def down do
+    # We delete duplicated data - that cannot be rollbacked
+  end
+
+  defp migrate_eth() do
+    alter table(:project_eth_address) do
```
MilanRusev: Isn't it possible that in project_eth_address we also have duplicates with lower and upper cases? If yes this statement will probably fail if we don't handele them first. The same for `migrate_eth()`

priv/repo/migrations/20180215105804_remove_duplicates.exs
```diff
@@ -0,0 +1,84 @@
...
+      modify(:address, :citext, null: false)
+    end
+
+    latest_eth_data = Repo.all(LatestEthWalletData)
+
+    filtered_eth_data =
```
MilanRusev: What is the benefit of fetching all data in elixir and reinserting it again to just using a single delete sql statement?

priv/repo/migrations/20180215105804_remove_duplicates.exs
IvanIvanoff: I manually checked both prod and stage for such cases

priv/repo/migrations/20180215105804_remove_duplicates.exs
IvanIvanoff: None I am aware of, it's just easier to implement for me

PR #283, opened 2018-02-16
Add migration for removing duplicate wallet data. Migrate all address columns to case insensitive format.

IvanIvanoff:
Note: The k8s job for creating the `citext` extension is executed only on staging


valo:
Did we manage to resolve the `citext` extension ops problem?

IvanIvanoff:
On staging I checked and `\dx` now liststs `citext` as installed. Previous to executing the k8s job it was listed only in the internal table for available extensions

PR #285, opened 2018-02-16
Price API and collection improvements and fixes

valo:
Use Dataloader to efficiently get the historical prices from influxDB.

Use success/error tuples in the GQL api, so that we don't have any
exceptions happening while a GQL query is resolved.

Removed some price queries which are not used.

Also fixed a regression in the price collection. The price import have
stopped working after a guard clause was introduced.

### DIFF
--- a/lib/sanbase_web/controllers/daily_prices_controller.ex
+++ b/lib/sanbase_web/controllers/daily_prices_controller.ex
@@ -19,7 +19,7 @@ defmodule SanbaseWeb.DailyPricesController do
         acc
         |> Map.put(
           pair,
-          Store.fetch_prices_with_resolution(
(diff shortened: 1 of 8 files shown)


lib/sanbase_web/controllers/daily_prices_controller.ex
```diff
@@ -19,7 +19,7 @@ defmodule SanbaseWeb.DailyPricesController do
         acc
         |> Map.put(
           pair,
-          Store.fetch_prices_with_resolution(
+          Store.fetch_prices_with_resolution!(
```
valo: This is a regular REST call, so we can use a bang method and fail the whole request here.

PR #293, opened 2018-02-19
Price-volume diff discord notification

MilanRusev:
Left to do:
* Merge & deploy tech_indicators PR: https://github.com/santiment/tech_indicators/pull/6
* Merge & deploy devops PR: https://github.com/santiment/devops/pull/125


lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -0,0 +1,95 @@
...
+
+  import Sanbase.DateTimeUtils, only: [seconds_ago: 1]
+
+  require Sanbase.Utils.Config
+
+  @http_service Mockery.of("HTTPoison")
+
+  @notification_type_name "price_volume_diff"
+  # roughly 3 months
+  @calculation_interval_in_sec 3 * 30 * 24 * 60 * 60
```
IvanIvanoff: Instead of subtracting seconds you can use `Timex.shift` with granularity `:months`

config/config.exs
```diff
@@ -165,9 +165,13 @@ config :sanbase, Sanbase.Etherbi.TransactionVolume,
   sync_enabled: {:system, "ETHERBI_TRANSACTION_VOLUME_SYNC_ENABLED", false}
 
 config :sanbase, Sanbase.Notifications.CheckPrices,
-  webhook_url: {:system, "NOTIFICATIONS_WEBHOOK_URL"},
-  notification_channel: {:system, "NOTIFICATIONS_CHANNEL", "#signals-stage"},
-  slack_notifications_enabled: {:system, "SLACK_NOTIFICATIONS_ENABLED", false}
+  webhook_url: {:system, "CHECK_PRICES_WEBHOOK_URL"},
+  notification_channel: {:system, "CHECK_PRICES_CHANNEL", "#signals-stage"},
+  slack_notifications_enabled: {:system, "CHECK_PRICES_NOTIFICATIONS_ENABLED", false}
```
IvanIvanoff: will this still send notifications to Slack?

MilanRusev: If it is enabled, yes. I haven't changed anything about this indicator, only renamed the environment vars so they are consistent with the new ones. If we decide to keep this indicator we'll need to change it to post in discord too.

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -0,0 +1,93 @@
...
+
+  @notification_type_name "price_volume_diff"
+  # roughly 3 months
+  @calculation_interval_in_days 3 * 30
+  # 60 minutes
+  @cooldown_period_in_sec 60 * 60
+  @price_volume_diff_threshold 0.1
+
+  def exec(project, currency) do
+    unless not notifications_enabled?() or
```
valo: Using `unless` and `not` could be very confusing. Can we switch this to `if`? I think it will make the code more readable.

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -0,0 +1,93 @@
...
+          {:ok, [%{price_volume_diff: nil}]} -> Decimal.new(0)
+          {:ok, [%{price_volume_diff: price_volume_diff}]} -> price_volume_diff
+          _ -> Decimal.new(0)
+        end
+
+      Decimal.cmp(price_volume_diff, Decimal.new(@price_volume_diff_threshold))
+      |> case do
+        :lt -> nil
+        _ -> send_notification(project, currency, price_volume_diff)
+      end
```
valo: Can we extract the body of this `if` into a separate function? Right now this function is quite long and nested

lib/sanbase_web/graphql/schema.ex
```diff
@@ -170,6 +170,7 @@ defmodule SanbaseWeb.Graphql.Schema do
       arg(:from, non_null(:datetime))
       arg(:to, :datetime, default_value: DateTime.utc_now())
       arg(:interval, :string, default_value: "1d")
+      arg(:last_positions_count, :integer, default_value: 0)
```
valo: What is the use of this argument?

MilanRusev: It designates how many values will be returned by the tech indicators service, it's for optimizing the size of the response. If it's 1 then only the indicator value for the last day is returned.

Maybe we can call it `response_last_positions_count` to be clearer.

IvanIvanoff: Something like `response_size_in_days` should be more understandable

MilanRusev: renamed it to `result_tail_size`

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -0,0 +1,132 @@
...
+    |> case do
+      :lt -> false
+      _ -> true
+    end
+  end
+
+  defp notification_type_name(currency), do: @notification_type_name <> "_" <> currency
+
+  defp get_calculation_interval() do
+    to_datetime = DateTime.utc_now() |> Timex.beginning_of_day()
```
IvanIvanoff: If we use `beginning_of_day` we won't be able to get the information for the last hours for most parts of the day.

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -0,0 +1,132 @@
...
+    from_datetime = Timex.shift(to_datetime, days: -@approximation_window - @comparison_window)
+
+    %{from_datetime: from_datetime, to_datetime: to_datetime}
+  end
+
+  defp send_notification(project, currency, indicator) do
+    {:ok, %HTTPoison.Response{status_code: 204}} =
```
IvanIvanoff: why we explicitly pattern match for only 204 status code?

MilanRusev: I tested it and discord returns 204 on success. Generally it would be better to match for all success status codes in [200, 300) but I doubt it's worth the effort since Discord didn't return anything but 204 so far

lib/sanbase/notifications/price_volume_diff.ex
MilanRusev: Influxdb's intervals (and thus the returned values) are split on beginnings of days. This way we're aligning with them.

Otherwise we will get incomplete data for the first and last days (which may or may not be desired?)

MilanRusev: Actually the purpose of this notification is to alert you for a potential signal as soon as possible so it may be better to work even with incomplete data.

config/config.exs
valo: The slack notifications are disabled for the moment. We can enable them and put a discord web hook, which is compatible with slack: https://discordapp.com/developers/docs/resources/webhook#execute-slackcompatible-webhook

PR #297, opened 2018-02-21
Make the number of top projects to follow configurable

valo:
Allow to configure the number of top CMC projects to follow.

Removed the unique index on name and added one on CMC id. This should resolve some issues when scraping data.


config/config.exs
```diff
@@ -124,7 +124,8 @@ config :sanbase, Sanbase.ExternalServices.Coinmarketcap,
 
 config :sanbase, Sanbase.ExternalServices.Coinmarketcap.TickerFetcher,
   update_interval: 5 * 1000 * 60,
-  sync_enabled: {:system, "COINMARKETCAP_TICKERS_ENABLED", false}
+  sync_enabled: {:system, "COINMARKETCAP_TICKERS_ENABLED", false},
+  top_projects_to_follow: {:system, "TOP_PROJECTS_TO_FOLLOW", "25"}
```
IvanIvanoff: Shouldn't that be 200 as the PRs name is 'follow top 200' ?

valo: The default is 25. I am going to change it in the config to 200 in the devops repo.

PR #300, opened 2018-02-22
Extend the functionality of the posts/insigths 

IvanIvanoff:
- Add some fields to the post type
- Allow updating of a post
- Allow uploading files via the GQL api

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -219,6 +219,12 @@ config :sanbase, Sanbase.MandrillApi,
 
 config :sanbase, Sanbase.TechIndicators, url: {:system, "TECH_INDICATORS_URL"}
 
+config :arc,
+  storage: Arc.Storage.S3,
+  # To support AWS regions other than US Standard
+  virtual_host: true,
+  bucket: {:system, "POSTS_IMAGE_BUCKET", "sanbase-images"}
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{Mix.env()}.exs"
--- a/lib/sanbase/voting/post.ex
+++ b/lib/sanbase/voting/post.ex
@@ -1,9 +1,13 @@
 defmodule Sanbase.Voting.Post do
   use Ecto.Schema
+
   import Ecto.Changeset
+  import Ecto.Query
+
   use Timex.Ecto.Timestamps
 
-  alias Sanbase.Voting.{Poll, Post, Vote}
+  alias Sanbase.Model.Project
+  alias Sanbase.Voting.{Poll, Post, Vote, PostImage}
   alias Sanbase.Auth.User
 
   @approved "approved"
@@ -15,25 +19,31 @@ defmodule Sanbase.Voting.Post do
     has_many(:votes, Vote, on_delete: :delete_all)
 
     field(:title, :string)
+    field(:short_desc, :string)
     field(:link, :string)
+    field(:text, :string)
     field(:state, :string)
     field(:moderation_comment, :string)
+
+    has_many(:images, PostImage, on_delete: :delete_all)
+
+    many_to_many(
+      :related_projects,
+      Project,
+      join_through: "posts_projects",
+      on_replace: :delete,
+      on_delete: :delete_all
+    )
 
     timestamps()
   end
 
   def create_changeset(%Post{} = post, attrs) do
     post
-    |> cast(attrs, [:title, :link])
-    |> validate_required([:poll_id, :user_id, :title, :link])
-    |> validate_length(:title, max: 140)
-    |> unique_constraint(:poll_id, name: :posts_poll_id_title_index)
-  end
-
-  def changeset(%Post{} = post, attrs \\ %{}) do
-    post
-    |> cast(attrs, [:title, :link, :state, :moderation_comment])
-    |> validate_required([:poll_id, :user_id, :title, :link])
+    |> cast(attrs, [:title, :short_desc, :link, :text])
+    |> related_projects_cast(attrs)
+    |> images_cast(attrs)
+    |> validate_required([:poll_id, :user_id, :title])
     |> validate_length(:title, max: 140)
     |> unique_constraint(:poll_id, name: :posts_poll_id_title_index)
   end
@@ -41,4 +51,32 @@ defmodule Sanbase.Voting.Post do
   def approved_state(), do: @approved
 
   def declined_state(), do: @declined
+
+  # Helper functions
+
+  defp related_projects_cast(changeset, %{related_projects: related_projects}) do
+    projects = Project |> where([p], p.id in ^related_projects) |> Sanbase.Repo.all()
+
+    changeset
+    |> put_assoc(:related_projects, projects)
+  end
+
+  defp related_projects_cast(changeset, _), do: changeset
+
+  defp images_cast(changeset, %{image_urls: image_urls}) do
+    images = PostImage |> where([i], i.image_url in ^image_urls) |> Sanbase.Repo.all()
+
+    if Enum.any?(images, fn %{post_id: post_id} -> not is_nil(post_id) end) do
+      changeset
+      |> Ecto.Changeset.add_error(
+        :images,
+        "The images you are trying to use are already used in another post"
+      )
+    else
+      changeset
+      |> put_assoc(:images, images)
+    end
+  end
+
+  defp images_cast(changeset, _), do: changeset
 end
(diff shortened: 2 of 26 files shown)


lib/sanbase/voting/post.ex
```diff
@@ -15,30 +19,88 @@ defmodule Sanbase.Voting.Post do
...
     field(:moderation_comment, :string)
 
+    has_many(:images, PostImage, on_delete: :delete_all)
+
+    many_to_many(
+      :related_projects,
+      Project,
+      join_through: "posts_projects",
+      on_replace: :delete,
+      on_delete: :delete_all
```
valo: Does this mean that the projects are removed when the post is deleted?

lib/sanbase/voting/post.ex
```diff
@@ -15,30 +19,88 @@ defmodule Sanbase.Voting.Post do
...
+    post_images_urls = Map.get(attrs, :image_urls, [])
+    post_images = PostImage |> where([i], i.image_url in ^post_images_urls) |> Sanbase.Repo.all()
+
     post
-    |> cast(attrs, [:title, :link])
-    |> validate_required([:poll_id, :user_id, :title, :link])
+    |> cast(attrs, [:title, :short_desc, :link, :text])
+    |> related_projects_cast(attrs)
+    |> create_images_cast(attrs)
+    |> put_assoc(:images, post_images)
```
valo: Isn’t this done in the ‘create_images_cast’?

lib/sanbase_web/graphql/resolvers/file_resolver.ex
```diff
@@ -0,0 +1,54 @@
...
+  """
+  def upload_image(_root, %{images: images}, _resolution) do
+    # In S3 there are no folders so the file name just contains some random text
+    # and a slash in it. Locally (in test and dev mode) the files are treaded as if
+    # they are located in a folder called `scope`
+    hash_algorithm = :md5
+
+    image_data =
+      for arg <- images, into: [] do
+        content_hash = image_content_hash!(arg, hash_algorithm)
```
valo: I think using a bang method in the resolver could be dangerous

lib/sanbase_web/graphql/resolvers/voting_resolver.ex
```diff
@@ -86,16 +100,44 @@ defmodule SanbaseWeb.Graphql.Resolvers.VotingResolver do
...
+
+  def delete_post(_root, %{id: post_id}, %{
+        context: %{auth: %{current_user: %User{id: user_id}}}
+      }) do
+    case Repo.get(Post, post_id) |> Repo.preload([:images]) do
+      %Post{user_id: ^user_id} = post ->
+        extract_image_url_from_post(post)
+        |> Enum.map(&Sanbase.FileStore.delete/1)
+
+        Repo.delete!(post)
```
valo: We should not use bang methods in the resolvers...

lib/sanbase_web/graphql/schema.ex
```diff
@@ -275,19 +285,47 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+      arg(:text, :string)
+      arg(:related_projects, list_of(:integer))
+      arg(:image_urls, list_of(:string))
 
       middleware(JWTAuth)
       resolve(&VotingResolver.create_post/3)
     end
 
+    @desc "Mutation used for updating an existing post owned by the currently logged in used"
+    field :update_post, :post do
```
valo: Why do we need to update posts? Isn’t the idea to make the posts immutable by the users?

IvanIvanoff: If we disallow updating it will be much better implementation wise. On the other hand won't that force the users to use an external editor to write their posts as I doubt that a long post is written in one sitting and once finished to click "Publish", "Finish" or whatever will put it on the blockchain.

config/config.exs
```diff
@@ -219,6 +219,12 @@ config :sanbase, Sanbase.MandrillApi,
 
 config :sanbase, Sanbase.TechIndicators, url: {:system, "TECH_INDICATORS_URL"}
 
+config :arc,
+  storage: Arc.Storage.S3,
+  # To support AWS regions other than US Standard
+  virtual_host: true,
+  bucker: {:system, "POSTS_IMAGE_BUCKET", "sanbase-images"}
```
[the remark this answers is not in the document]
IvanIvanoff: arc uses the `:ex_aws` config we already have:
`config :ex_aws,
  access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :instance_role],
  secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, :instance_role],
  region: "eu-central-1"`

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: or should I remove anything update related and simplify the process from
`create-update-update-post on blockhain` to just `create-post on blockhain` ?

config/config.exs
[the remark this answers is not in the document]
valo: Cool

lib/sanbase/voting/post.ex
valo: 👍

lib/sanbase_web/graphql/schema.ex
valo: I don't think we have posts in draft mode for now. If we start having such posts we can decide if we want to have a flag "state: draft" or have a separate model with only draft posts.

lib/sanbase/voting/post.ex
```diff
@@ -15,30 +19,70 @@ defmodule Sanbase.Voting.Post do
...
 
   def create_changeset(%Post{} = post, attrs) do
-    post
-    |> cast(attrs, [:title, :link])
-    |> validate_required([:poll_id, :user_id, :title, :link])
-    |> validate_length(:title, max: 140)
-    |> unique_constraint(:poll_id, name: :posts_poll_id_title_index)
-  end
+    project_ids = Map.get(attrs, :related_projects, [])
+    projects = Project |> where([p], p.id in ^project_ids) |> Sanbase.Repo.all()
```
valo: This is not used

lib/sanbase_web/graphql/resolvers/file_resolver.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias Sanbase.Voting.PostImage
+
+  @doc ~s"""
+    Uploads the image to S3
+  """
+  def upload_image(_root, %{images: images}, _resolution) do
+    # In S3 there are no folders so the file name just contains some random text
+    # and a slash in it. Locally (in test and dev mode) the files are treaded as if
+    # they are located in a folder called `scope`
+    hash_algorithm = :md5
```
valo: If there is no some concrete reason to use `md5` I think we should use something like `sha256`

lib/sanbase_web/graphql/resolvers/voting_resolver.ex
```diff
@@ -76,7 +90,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.VotingResolver do
     |> Repo.insert()
     |> case do
       {:ok, post} ->
-        {:ok, post |> Repo.preload([:votes, :user])}
+        {:ok, post |> Repo.preload([:votes, :user, :images])}
```
valo: It is possible to use `Dataloader` instead of preloading?

IvanIvanoff: Will it be beneficial as it's made for a single post?

lib/sanbase_web/graphql/schema/voting_types.ex
```diff
@@ -14,9 +14,18 @@ defmodule SanbaseWeb.Graphql.VotingTypes do
     field(:user, non_null(:user))
     field(:poll, non_null(:poll))
     field(:title, non_null(:string))
-    field(:link, non_null(:string))
+    field(:short_desc, :string)
+    field(:link, :string)
+    field(:text, :string)
     field(:state, :string)
     field(:moderation_comment, :string)
+    field(:images, list_of(:image_data))
```
valo: I meant adding data loader for these fields. That should allow to remove the preloads from the ecto queries when fetching the post

IvanIvanoff: Done. Working a little with the dataloader it looks cleaner with all the ecto logic living outside the resolver, so I will keep the dataloader in the resolver too.

PR #310, opened 2018-03-01
Price-volume diff notification - increase threshold

MilanRusev:
Hopefully this will decrease the spam on `#signals-stage`
Also added env var for the cooldown period

Also added debug info for stage: redirects to a jupyter notebook with the appropriate params

Threshold & debug url added to env vars: https://github.com/santiment/devops/pull/141

### DIFF
--- a/lib/sanbase/notifications/price_volume_diff.ex
+++ b/lib/sanbase/notifications/price_volume_diff.ex
@@ -16,7 +16,7 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
 
   @approximation_window 14
   @comparison_window 7
-  @price_volume_diff_threshold 0.1
+  @price_volume_diff_threshold 0.2
 
   def exec(project, currency) do
     currency = String.upcase(currency)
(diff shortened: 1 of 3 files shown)


lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -16,7 +16,7 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
 
   @approximation_window 14
   @comparison_window 7
-  @price_volume_diff_threshold 0.1
+  @price_volume_diff_threshold 0.2
```
IvanIvanoff: can we make that configurable via an ENV var?

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -16,12 +16,11 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
 
   @approximation_window 14
   @comparison_window 7
-  @price_volume_diff_threshold 0.1
 
   def exec(project, currency) do
     currency = String.upcase(currency)
 
-    if notifications_enabled?() &&
+    if notifications_enabled?() && notification_threshold() &&
```
IvanIvanoff: I'd rather only check for `notifications_enabled?()` and put a default value to the threshold (`0.2`)

IvanIvanoff:
If you are leaving the default to 0.1 (you changed that to 0.2 in your initial commit) make a PR in devops for changing it to 0.2

MilanRusev:
It is there: https://github.com/santiment/devops/pull/141

MilanRusev:
Also added debug info for stage: redirects to a jupyter notebook with the appropriate params

PR #315, opened 2018-03-06
Add phoenix live code reloading dev dependency

IvanIvanoff:
To test easily and faster (without stopping and starting your phoenix server to get the changes) add [phoenix live reloading](https://github.com/phoenixframework/phoenix_live_reload).

On Linux it requires you to manually install `inotify` ( `yum install inotify-tools`, `apt-get install inotify-tools`, etc.)


valo:
Can we add instructions in the `README` that some packages need to be installed on Linux?

PR #316, opened 2018-03-06
Add project gql fetch by coinmarketcap

IvanIvanoff:


lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -55,6 +53,23 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+      Project
+      |> Repo.get_by(coinmarketcap_id: slug)
+      |> case do
+        nil ->
+          {:error, "Project with slug '#{slug}' not found."}
+
+        project ->
+          project =
+            project
+            |> Repo.preload([:latest_coinmarketcap_data, icos: [ico_currencies: [:currency]]])
```
valo: It will be so nice to remove these preloads, but I guess this is outside of the scope of this PR

IvanIvanoff: It's exactly like the project fetching by id. Should be rewritten to use the dataloader in another PR

PR #318, opened 2018-03-06
Add uploaded file size validation and file upload tests

IvanIvanoff:


lib/sanbase_web/graphql/resolvers/file_resolver.ex
```diff
@@ -15,47 +17,77 @@ defmodule SanbaseWeb.Graphql.Resolvers.FileResolver do
 
     image_data =
       for %{filename: file_name} = arg <- images, into: [] do
```
valo: Isn't this a simple map with pattern matching? Something line 

```elixir
images
|> Enum.map(fn %{filename: file_name} = arg ->
  arg = %{arg | filename: milliseconds_str() <> "_" <> file_name}
  save_image_content(arg, hash_algorithm)
end)
```

What is the benefit of using for?

lib/sanbase_web/graphql/resolvers/file_resolver.ex
```diff
@@ -15,47 +17,77 @@ defmodule SanbaseWeb.Graphql.Resolvers.FileResolver do
...
+      {:ok, hash}
+    rescue
+      error in File.Error ->
+        %{reason: reason} = error
+        {:error, "Error calculating file's content hash. Reason: #{reason}"}
+    end
   end
 
-  defp save_images_pg(images) do
+  defp save_image_content(%Plug.Upload{filename: file_name} = arg, hash_algorithm)
```
valo: Seems to me that this `hash_algorithm` parameter is not really needed. May be a module attribute will do the trick?

lib/sanbase_web/graphql/resolvers/file_resolver.ex
IvanIvanoff: I was using for at some point to do some filtering but it's no longer needed, will change to map

lib/sanbase_web/graphql/resolvers/file_resolver.ex
IvanIvanoff: Will be cleaner with module attribute, agreed

PR #319, opened 2018-03-06
Serve blockchain analytics data from the new Flink pipeline

valo:
Removed all the code around fetching the caching the data from the old
Etherbi into influxDB. The data will come directly from influxDB from
now on and will be kept up-to-date by Flink pipelines.


config/config.exs
```diff
@@ -58,22 +58,22 @@ config :sanbase, Sanbase.ExternalServices.TwitterData.Store,
   database: "twitter_followers_data"
 
 config :sanbase, Sanbase.Etherbi.Transactions.Store,
-  host: {:system, "INFLUXDB_HOST", "localhost"},
-  port: {:system, "INFLUXDB_PORT", 8086},
+  host: {:system, "ETHERBI_INFLUXDB_HOST", "localhost"},
+  port: {:system, "ETHERBI_INFLUXDB_HOST", 8086},
   pool: [max_overflow: 10, size: 20],
-  database: "transactions"
+  database: "erc20_exchange_transactions"
```
IvanIvanoff: with these changes we should plan on how to drop the old databases

valo: I think we can manually drop them. It will be a one time thing, which won't need to happen again.

valo:
This PR also fixes https://sentry.stage.internal.santiment.net/sentry/sanbase/issues/258/

IvanIvanoff:
This PR removes default `to` datetime so this may require some changes from @Partysun

valo:
I don't think the default `to` is used by the frontend: https://github.com/santiment/sanbase2/blob/master/app/src/pages/Detailed/DetailedGQL.js#L96
https://github.com/santiment/sanbase2/blob/master/app/src/pages/Detailed/Detailed.js#L308

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -1,47 +1,103 @@
...
+      _ -> {:error, "Can't fetch the exchange funds for #{ticker}"}
     end
   end
 
-  # Private functions
-
-  defp use_cache?() do
-    Config.module_get(Sanbase.Etherbi, :use_cache)
+  defp ticker_to_contract_address(ticker) do
+    with project when not is_nil(project) <- Repo.get_by(Project, ticker: ticker),
```
IvanIvanoff: I'm concerned because that depends on the order of inserting the projects with duplicated (last time I checked in prod there are 100+ duplicated) tickers in the database. The one with smaller id will be returned and it may not be the one that's needed

valo: I downloaded the list of all projects on prod as CSV and from all projects with CMC_ID there were none with duplicate tickers. May be adding a filtering on CMC_ID will be enough?

IvanIvanoff: That should be enough, yes

PR #320, opened 2018-03-07
Address compile time warnings

IvanIvanoff:
Address some compile time warnings.

NOTE: This PR is **potentially dangerous**. Should be tested really well because of previous attempts to replace Ecto.Date and Ecto.DateTime led to timezone issues.

### DIFF
--- a/lib/sanbase/model/ico.ex
+++ b/lib/sanbase/model/ico.ex
@@ -80,11 +80,7 @@ defmodule Sanbase.Model.Ico do
   end
 
   defp funds_raised_ico_end_price_from_currencies(ico, target_ticker, date) do
-    erl_date = date |> Date.to_erl()
-
-    datetime =
-      NaiveDateTime.from_erl!({erl_date, {0, 0, 0}})
-      |> DateTime.from_naive!("Etc/UTC")
+    datetime = Sanbase.DateTimeUtils.datetime_from_date(date)
 
     Repo.preload(ico, ico_currencies: [:currency]).ico_currencies
     |> Enum.map(fn ic ->
--- a/lib/sanbase/model/ico.ex
+++ b/lib/sanbase/model/ico.ex
@@ -10,8 +10,8 @@ defmodule Sanbase.Model.Ico do
 
   schema "icos" do
     belongs_to(:project, Project)
-    field(:start_date, Ecto.Date)
-    field(:end_date, Ecto.Date)
+    field(:start_date, :date)
+    field(:end_date, :date)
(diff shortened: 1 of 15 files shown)


lib/sanbase/model/ico.ex
```diff
@@ -80,11 +80,15 @@ defmodule Sanbase.Model.Ico do
   end
 
   defp funds_raised_ico_end_price_from_currencies(ico, target_ticker, date) do
-    timestamp = Sanbase.DateTimeUtils.ecto_date_to_datetime(date)
+    erl_date = date |> Date.to_erl()
+
+    datetime =
+      NaiveDateTime.from_erl!({erl_date, {0, 0, 0}})
+      |> DateTime.from_naive!("Etc/UTC")
```
valo: Why it is needed to go through an `erl` date? May be directly creating a `NaiveDateTime` will work? https://hexdocs.pm/elixir/NaiveDateTime.html#new/2

IvanIvanoff: I searched for such function but only by keywords 'from_date' etc. Will change it

MilanRusev:
So far I found a couple of issues about ex_admin and the datetimes:
* With naive_datetime & date there is no more a datetime picker for the fields, but only a text field. One has to type in the exact format `yyyy-MM-dd` for date and `yyyy-MM-dd hh:mm:ss` for naive_datetime or they'll get an error. Here Garry may say whether this inconvenience is bad or not (for reference - see ICO.start_date/end_date, LatestEthWalletdata.Last_in/last_out/update_time)
* With date the filter in the list page is lost - e.g. on https://sanbase-low.santiment.net/admin/icos the `start_date` & `end_date` filters on the right are missing when changing ecto.date to date

IvanIvanoff:
Any idea why changing the type removes the datetime pickers :?

valo:
Something in here may be? https://github.com/smpallen99/ex_admin/blob/master/lib/ex_admin/form.ex#L1010 There is also an protocol for rendering different types, but I doubt it is that: https://github.com/smpallen99/ex_admin/blob/master/lib/ex_admin/render.ex

IvanIvanoff:
From the first link just below `Ecto.Datetime` are the `build_control` definitions for `DateTime` and `NaiveDateTime`, so it is not that.

valo:
It was just a suggestion... I am not sure why it might not work. Seems like we need to dig through the ex_admin code. Unfortunately some of it seems to be quite sophisticated :(

PR #322, opened 2018-03-09
Use floats for the etherbi data

valo:
This makes the GQL serializer render the numbers properly and not use
scientific notation.


IvanIvanoff:
There are `Decimal.new()` left in the resolver:
https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex#L20
https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex#L47

valo:
This changes the format of the request a bit. Let me try to test this locally against the JS client

PR #327, opened 2018-03-16
Price-volume diff debug info - add threshold

MilanRusev:


lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -146,7 +147,8 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
         aggregate_interval,
         window_type,
         approximation_window,
-        comparison_window
+        comparison_window,
+        notification_threshold
```
IvanIvanoff: let's use parentheses for the functions invocation here, too.

PR #331, opened 2018-03-19
Project GQL api - eth spent over time

MilanRusev:
Added field ethSpentOverTime to the project object returning a list eth spent grouped by time intervals

### DIFF
--- a/lib/sanbase/external_services/etherscan/store.ex
+++ b/lib/sanbase/external_services/etherscan/store.ex
@@ -64,7 +64,8 @@ defmodule Sanbase.ExternalServices.Etherscan.Store do
     |> Store.query()
     |> parse_trx_sum_time_series()
     |> case do
-      {:ok, [%{datetime: _datetime, amount: amount}]} -> {:ok, amount}
+      {:ok, [{_datetime, amount}]} -> {:ok, amount}
+      {:ok, []} -> {:ok, nil}
       res -> res
     end
   end
@@ -205,7 +206,7 @@ defmodule Sanbase.ExternalServices.Etherscan.Store do
       transactions
       |> Enum.map(fn [iso8601_datetime, amount] ->
         {:ok, datetime, _} = DateTime.from_iso8601(iso8601_datetime)
-        %{datetime: datetime, amount: amount}
+        {datetime, amount}
       end)
 
     {:ok, result}
--- a/lib/sanbase/external_services/etherscan/store.ex
+++ b/lib/sanbase/external_services/etherscan/store.ex
@@ -63,6 +63,10 @@ defmodule Sanbase.ExternalServices.Etherscan.Store do
     sum_from_to_query(measurement, from, to, transaction_type)
     |> Store.query()
     |> parse_trx_sum_time_series()
+    |> case do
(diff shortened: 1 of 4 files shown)


lib/sanbase/external_services/etherscan/store.ex
```diff
@@ -117,6 +158,15 @@ defmodule Sanbase.ExternalServices.Etherscan.Store do
     AND time <= #{DateTime.to_unix(to, :nanoseconds)}/
   end
 
+  defp sum_over_time_from_to_query(measurement, from, to, resolution, transaction_type) do
+    ~s/SELECT time, SUM(trx_value)
+    FROM "#{measurement}"
+    WHERE transaction_type = '#{transaction_type}'
+    AND time >= #{DateTime.to_unix(from, :nanoseconds)}
+    AND time <= #{DateTime.to_unix(to, :nanoseconds)}
+    GROUP BY TIME(#{resolution}) fill(none)/
```
valo: Shouldn't we use `fill(0)` here?

valo:
I think it should be relatively easy to copy/paste this test and adjust it to test this GQL query: https://github.com/santiment/sanbase2/blob/master/test/sanbase_web/graphql/project_api_wallet_transactions_test.exs#L4

Just to have a safety net against regressions.

valo:
And some merge conflicts needs to be resolved

lib/sanbase/external_services/etherscan/store.ex
valo: @Partysun do you think that filling with missing timestamps with `0` will be easier on the frontend? My guess is `yes`, but your input will also be valuable.

lib/sanbase_web/graphql/schema/project_types.ex
```diff
@@ -160,6 +160,14 @@ defmodule SanbaseWeb.Graphql.ProjectTypes do
       resolve(&ProjectResolver.eth_spent/3)
     end
 
+    field :eth_spent_over_time, list_of(:eth_spent_data) do
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
```
IvanIvanoff: `eth_spent` uses a single argument for specifying the last N days you want to query. would be better for those two queries to be consistent

lib/sanbase/external_services/etherscan/store.ex
MilanRusev: Currently all similar queries aren't filled with 0 but have missing timestamps: `burn_rate`, `transaction_volume`, `twitter_followers`, `github_activity` etc. If we fill the missing values with 0 we'd better do that for these queries as well.

lib/sanbase_web/graphql/schema/project_types.ex
MilanRusev: Depends on how it will be used by the frontend, with from/to it can also be plotted for different time intervals. @Partysun & @brighteye?

MilanRusev: Talked with Yura and we'd better leave it with from/to for more flexibility

PR #335, opened 2018-03-20
GraphQL query caching

valo:


lib/sanbase_web/graphql/middlewares/memory_query_cache.ex
```diff
@@ -25,44 +30,55 @@ defmodule SanbaseWeb.Graphql.Middlewares.MemoryQueryCache do
...
-      |> Enum.map(&generate_query/1)
+      |> Enum.map(&generate_query(&1, ttl))
 
     Map.put(result, :children, children)
   end
+
+  # Reduce the resolution of the datetime arguments, so that they don't change
+  # too often and invalidate the cache
+  defp convert_value(%DateTime{} = datetime, ttl) do
+    div(DateTime.to_unix(datetime, :millisecond), ttl)
```
IvanIvanoff: If divide the current timestamp `1521524946000` by 5000 we're getting a date with year 1970. Is this intentional? I think what we are aiming for is to transform `#DateTime<2018-03-20 05:54:27Z>` to `#DateTime<2018-03-20 05:50:00Z>` or something like that.

IvanIvanoff: Possible solution would be to:
`datetime = %DateTime{datetime | minute: div(minute, 5) * 5, second: 0}`
This will round the datetime as described in the previous comment

valo:
In the current state, the `allProjects` query used on the front page is mostly cached and that should relieve the influxDB instance. Any suggestions for simplifications are welcome

lib/sanbase_web/graphql/resolvers/twitter_resolver.ex
```diff
@@ -2,10 +2,18 @@ defmodule SanbaseWeb.Graphql.Resolvers.TwitterResolver do
...
+  alias SanbaseWeb.Graphql.Helpers.Cache
 
   import Ecto.Query
 
   def twitter_data(_root, %{ticker: ticker}, _resolution) do
+    Cache.func(
+      fn -> calculate_twitter_data(ticker) end,
+      {:twitter_data, ticker}
+    ).()
+  end
```
valo: This looks a bit strange to me. We should be able to plug the caching in the schema or the type...

IvanIvanoff: I think we can hide it behind a macro which can be used in the resolver and it will expand to this by using the `__ENV__.function` to populate the function name + constructing the parameters list should be possible

valo: Let's add this in the next iteration as getting it right might be tricky and I feel that we need this caching soon as the traffic to the site will be increasing.

PR #336, opened 2018-03-20
Feature/show coins

Partysun:


app/src/pages/Projects/projectSelectors.js
```diff
@@ -1,7 +1,12 @@
+export const isERC20 = project => (
```
valo: I am wondering if we should rename this to something like `hasTeamWallets`?

PR #337, opened 2018-03-20
Fix tesla adapter timeout error

tspenov:
And pass the error downwards.

This is supposed to fix such errors like:
https://sentry.stage.internal.santiment.net/sentry/sanbase/issues/208/
https://sentry.stage.internal.santiment.net/sentry/sanbase/issues/10/

image

### DIFF
--- a/lib/sanbase/external_services/etherscan/requests/internal_tx.ex
+++ b/lib/sanbase/external_services/etherscan/requests/internal_tx.ex
@@ -25,8 +25,8 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.InternalTx do
         error = "Error fetching transactions for #{address}. Status code: #{status}: #{body}"
         Logger.warn(error)
         {:error, error}
-      error ->
-        {:error, error}
+      %Tesla.Error{message: error_msg} ->
+        {:error, error_msg}
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/sanbase/external_services/etherscan/requests/internal_tx.ex
```diff
@@ -25,6 +25,8 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.InternalTx do
         error = "Error fetching transactions for #{address}. Status code: #{status}: #{body}"
         Logger.warn(error)
         {:error, error}
+      error ->
```
IvanIvanoff: To match exactly the tesla error you can pattern match on this `%Tesla.Error{} = error ->` and stringify the error, as otherwise the recipient will not be able to print it. Other types of results are not expected (maybe not possible, too) and can be left to crash the process and generate a sentry log.

tspenov: 👍 The other solution may be to just return the `message` from `%Tesla.Error{}` here: https://github.com/santiment/sanbase2/pull/337/files#diff-34347235b130678fdce1211703e8ead9R8 to become: `e in Tesla.Error -> e.message`
@IvanIvanoff what do you think ?

IvanIvanoff: Just `error` will match everything that didn't match the first two results and we won't be sure if that is some correct case that we missed or an error. Thus, it's not clear how it should be treated. My approach would be to treat the errors explicitly and either have a last case that catches everything and logs it as unrecognized message or just let it crash.

tspenov: Yes, sound good. Implemented it that way

PR #339, opened 2018-03-20
Implement eth_spent_by_all_projects GQL query

IvanIvanoff:


lib/sanbase/internal_services/parity.ex
```diff
@@ -46,7 +46,7 @@ defmodule Sanbase.InternalServices.Parity do
 
     Tesla.build_client([
       {Tesla.Middleware.BaseUrl, parity_url},
-      {Tesla.Middleware.BasicAuth, username: basic_auth_username, password: basic_auth_password},
+      # {Tesla.Middleware.BasicAuth, username: basic_auth_username, password: basic_auth_password},
```
valo: I think this should be not be commented out

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -96,6 +96,15 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     end)
   end
 
+  def eth_spent_by_all_projects(_, %{from: from, to: to}, _resolution) do
+    with {:ok, measurements} <- Etherscan.Store.public_measurements(),
```
valo: May be we can ask the DB for the projects we show on the sanbase page and compile the list of CMC_IDs? That will ensure that the projects we aggregate for is the same as the list we show on the `Projects` page.

valo: I think we should expose the `infrastructure` field for each project and for eth_spent filter only projects from the `ETH` infrastructure

IvanIvanoff: With infrastructure = `ETH` we have much much more records than we have with project wallets, for which we store transactions, so that way we'll get errors for trying to query non existing measurements.

When we clean our influxdb just asking for `public_measurements` will get us exactly what we are needing - no more and no less.

PR #342, opened 2018-03-21
Catch get balance errors

tspenov:
#### Summary

Remove `get_balance!` bang function and return nil when we can't get balance. Insert/update only when we have balance.
This should fix: https://sentry.stage.internal.santiment.net/sentry/sanbase/issues/210/

card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/external_services/etherscan/requests/balance.ex
+++ b/lib/sanbase/external_services/etherscan/requests/balance.ex
@@ -6,28 +6,21 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Balance do
 
   defstruct [:status, :message, :result]
 
-  def get_balance!(address) do
-    case get_balance(address) do
-      {:ok, result} -> result
-      {:error, error} -> raise(error)
-    end
-  end
-
   def get_balance(address) do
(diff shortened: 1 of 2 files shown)


lib/sanbase/external_services/etherscan/requests/balance.ex
```diff
@@ -6,28 +6,21 @@ defmodule Sanbase.ExternalServices.Etherscan.Requests.Balance do
...
     |> case do
       %{status: 200, body: body} ->
-        {:ok, Poison.Decode.decode(body, as: %Balance{})}
+        Poison.Decode.decode(body, as: %Balance{})
 
       %{status: status, body: body} ->
         error = "Error fetching transactions for #{address}. Status code: #{status}: #{body}"
         Logger.warn(error)
-        {:error, error}
+        nil
```
valo: May be we can use the standard `{:error, error}` tuple here? That makes it more clear that something bad happens.

IvanIvanoff:
you can run `mix format` before merging

PR #344, opened 2018-03-22
Rewrite exchange fund flow to match the new database layout

IvanIvanoff:
#### Summary
It happened that the schema of the database measurements also changed, so this PR changes the way sanbase interacts with influxdb

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
@@ -61,31 +61,30 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
     and time period.
     Uses the influxdb cached values instead of issuing a GET request to etherbi
   """
-  def exchange_fund_flow(
+  def exchange_funds_flow(
         _root,
         %{
           ticker: ticker,
           from: from,
           to: to,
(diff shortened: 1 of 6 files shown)


valo:
I am wondering if we should change this query to return the IN - OUT transaction volume. Is that going to make the thing simpler? IMO, this is the stat we really care about and we won't be showing the IN and OUT volumes separately.

IvanIvanoff:
You mean the difference between `SUM(in_amount)` and `SUM(out_amount)` for every of the `GROUP BY` intervals? This actually will catch and not display movements between team wallets as ETH Spent most of the time. Though there are some edge cases like if the in and out transactions are in different intervals

IvanIvanoff:
@valo done

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -120,4 +119,36 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
...
+        amount: in_amount,
+        transaction_type: "in"
+      },
+      %{
+        datetime: datetime,
+        amount: out_amount,
+        transaction_type: "out"
+      }
+    ]
+  end
```
valo: I think these 3 functions are not used any more

PR #348, opened 2018-03-22
Twitter mentions gql api

MilanRusev:


lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -177,6 +177,65 @@ defmodule Sanbase.InternalServices.TechIndicators do
     end
   end
 
+  def twitter_mention_count(
```
IvanIvanoff: this is one big function, can you think some way of dividing it into smaller parts?

lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -137,44 +256,74 @@ defmodule Sanbase.InternalServices.TechIndicators do
...
+                       "timestamp" => timestamp,
+                       "price_volume_diff" => price_volume_diff,
+                       "price_change" => price_change,
+                       "volume_change" => volume_change
+                     } ->
+        %{
+          datetime: DateTime.from_unix!(timestamp),
+          price_volume_diff: decimal_or_nil(price_volume_diff),
+          price_change: decimal_or_nil(price_change),
+          volume_change: decimal_or_nil(volume_change)
```
valo: Do we really want to convert everything to decimals? Do we care about precision so much? I feel like using `float` will be enough and it will make calculations simpler and the API faster

PR #355, opened 2018-03-23
Increase timeout for http requests to tech-indicators service

MilanRusev:
Hopefully this will resolve that in prod `sanbase-scrapers` fail with a :connect_timeout error when trying to access `tech-indicators` for the price-volume diff notification:
image
(for two hours between 05:20 and 07:00 it worked and before & after that it failed)

The funny thing is the GQL api on prod works well (`sanbase-web`), and also on staging there are no timeout problems whatsoever.


valo:
I am looking into adding https://prometheus.io to our cluster

PR #358, opened 2018-03-23
Add main contract address in project

tspenov:
#### Summary
card: https://favro.com/card/bdbb4f24afc48b29c74f4fa4/San-1223

* Adds `main_contract_address` in project

A migration to remove the column from `icos` table will be in subsequent PR


tspenov:
@valo Do we need to migrate the values from the icos `main_contract_address` table into project table  or we they repopulate themselves by the running scripts ?

lib/sanbase/model/project.ex
```diff
@@ -177,8 +178,8 @@ defmodule Sanbase.Model.Project do
         inner_join: p in Project,
         on: p.id == d.project_id,
         where:
-          not is_nil(p.coinmarketcap_id) and d.rank == 1 and not is_nil(d.main_contract_address) and
-            not is_nil(d.contract_block_number) and not is_nil(d.contract_abi),
+          not is_nil(p.coinmarketcap_id) and d.rank == 1 and not is_nil(d.contract_block_number) and
+            not is_nil(d.contract_abi),
```
valo: Should we keep the `not is_nil(p.main_contract_address)` here?

valo:
Let's also write a migration, because the moment we deploy all the blockchain info will disappear from the website until the addresses are populated. With a migration we won't have down time. Also, can we make this a 2 phase migration? If we do it like here we will have a bit of downtime until the old code stops and the new one starts. When we do deployments we have some time when the old and the new code are running at the same time. In case the new code fails the old code will stay around, so it is best not to break the old code for a while.

I would propose to first add the new column, start using that in the code and when all looks good in production, make a PR with cleanup of the old DB column.

IvanIvanoff:
Follow this PR: https://github.com/santiment/sanbase2/pull/362
If it is merged before this one some additional changes have to be made

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -61,7 +61,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
         inner_join: infr in Infrastructure,
         on: p.infrastructure_id == infr.id,
         where:
-          not is_nil(p.coinmarketcap_id) and not is_nil(ico.main_contract_address) and
```
IvanIvanoff: after this change do we need `inner_join` with the ico table?

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -84,7 +84,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
         p in Project,
         inner_join: ico in Ico,
         on: p.id == ico.project_id,
-        where: not is_nil(p.coinmarketcap_id) and is_nil(ico.main_contract_address),
```
IvanIvanoff: again, i think we do not need `inner_join` with the Ico table here.

priv/repo/migrations/20180418141807_copy_icos_main_contract_address_in_projects.exs
```diff
@@ -0,0 +1,24 @@
...
+      from(
+        i in Ico,
+        where: not is_nil(i.main_contract_address),
+        select: [i.project_id, i.main_contract_address]
+      )
+
+    Repo.all(query)
+    |> Enum.map(fn [project_id, main_contract_address] ->
+      Repo.get!(Project, project_id)
+      |> Project.changeset(%{main_contract_address: main_contract_address})
```
IvanIvanoff: This will crawl the `{project_id, main_contract_address}` in increasing order of the ico ids.
For Aeternity we have two Ico records with two different(!) contracts:

The 'older' one (correct): https://etherscan.io/token/0x5ca9a71b1d01849c0a95490cc00559717fcf0d1d
The 'newer' one: https://etherscan.io/address/0xbec591de75b8699a3ba52f073428822d0bfc0d7e

The currently used one is the older one, which won't be the `main_contract_address` of the project after the migration.

PR #366, opened 2018-03-26
Filter transactions between project own addresses

tspenov:
#### Summary

We should filter transactions between project eth own addresses from aggregation queries

TODO:
- [x]  probably should be added also here: 
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/etherscan/store.ex#L258 and here
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/etherscan/store.ex#L258 The problem is that the select is from multiple measurements and it is hard to build a proper `where` clause.
- [x] probably should be added here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/etherscan/store.ex#L277 but it returned some errors while testing with staging data.
- [x] test whether queries make things slower
- [x] fix tests

Ivan: Making the influxdb queries async + the in-memory caching solves the problem with the slower queries. Maybe for this data we can even increase the cache from 5 to 10 minutes and it won't matter that much. The query before caching takes around 6 seconds through kubectl proxy

Ivan2: This PR contains some cleaning of the Store module and changing the project resolver to match them. This is part of the process of unifying all store modules

[//]: # (#### Screenshots)

image 3
image 2

### DIFF
--- a/lib/sanbase/etherbi/transactions/store.ex
+++ b/lib/sanbase/etherbi/transactions/store.ex
@@ -71,10 +71,6 @@ defmodule Sanbase.Etherbi.Transactions.Store do
     GROUP BY time(#{interval}) fill(0)/
   end
 
-  defp parse_transactions_time_series(%{results: [%{error: error}]}) do
-    {:error, error}
-  end
-
   defp parse_in_out_diff_time_series(%{
          results: [
            %{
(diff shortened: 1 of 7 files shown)


tspenov:
Awesome work @IvanIvanoff . Thanks for taking that! Cannot approve but LGTM 👍

PR #370, opened 2018-03-27
Price volume diff notification change datetime

MilanRusev:
Change the datetime reported in the price-volume diff notification - now it reports the timestamp for the last datapoint used in the calculation (not the timestamp of the notification itself)

### DIFF
--- a/lib/sanbase/notifications/price_volume_diff.ex
+++ b/lib/sanbase/notifications/price_volume_diff.ex
@@ -27,7 +27,7 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
         get_indicator(project.ticker, currency, from_datetime, to_datetime)
 
       if check_notification(indicator) do
-        send_notification(project, currency, to_datetime, indicator, notification_log)
+        send_notification(project, currency, indicator, notification_log)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


valo:
Can you please provide a description of this change, as it is now clear what exactly is changing here.

lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -119,18 +121,20 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
...
-         notification_date,
-         %{price_change: price_change, volume_change: volume_change},
+         %{datetime: datetime, price_change: price_change, volume_change: volume_change},
          debug_info
        ) do
+    # Timex.shift(days: 1) is because the returned datetime is the beginning of the day
     {:ok, notification_date_string} =
-      Timex.format(notification_date, "{YYYY}-{0M}-{0D} {h24}:{m}:{s}")
+      datetime
+      |> Timex.shift(days: 1)
```
IvanIvanoff: Doesn't this combined with the comment above mean that all notifications will have `00:00:00` as time?

MilanRusev: Yes, but I decided to leave it with the time so people will know this is the end timestamp and not the last day included

PR #372, opened 2018-03-28
Reduce the boilerplate of using the GQL cache and add caching to more queries

IvanIvanoff:
#### Summary
This PR aims to reduce the boilerplate when using our GQL cache. The ideal goal is to have something like `cache_resolve(captured_mfa)` instead of `resolve(captured_mfa)` (may not be achieved in this PR, but that's one step closer)

1. Due to the way Dataloader works with middlewares we put the caching inside the dataloader functions.

2. Now we cache the PDP page

3. Caching every gql query except `current_user` (Note: never cache current_user)

4. Add cache clear before every test. Otherwise tests like 'test exchangeWallets returns []' executed could hit the cache after exchangeWallets returns non-empty test.

### DIFF
--- a/lib/sanbase_web/graphql/price_store.ex
+++ b/lib/sanbase_web/graphql/price_store.ex
@@ -1,5 +1,6 @@
 defmodule SanbaseWeb.Graphql.PriceStore do
   alias Sanbase.Prices
+  alias SanbaseWeb.Graphql.Helpers.Cache
 
   def data() do
     Dataloader.KV.new(&query/2)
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/graphql/price_store.ex
```diff
@@ -16,15 +17,26 @@ defmodule SanbaseWeb.Graphql.PriceStore do
...
-    else
-      _error -> nil
-    end
+    Cache.func(
+      fn ->
+        with {:ok, {_dt, price, _mcap, _volume}} <- Prices.Store.last_record(pair) do
+          Decimal.new(price)
+        else
+          _error -> nil
+        end
```
valo: I feel like we should split into a separate functions. It is a bit hard to read right now...

PR #377, opened 2018-03-29
Add daily active addresses store module and gql api.

IvanIvanoff:
#### Summary
Add the DAT GQL. Also include some improvements on the influxdb store, which is the start of rewriting all store modules (to return unified lists instead of tuples)

### DIFF
--- a/lib/sanbase/etherbi/daily_active_addresses/store.ex
+++ b/lib/sanbase/etherbi/daily_active_addresses/store.ex
@@ -1,0 +1,29 @@
+defmodule Sanbase.Etherbi.DailyActiveAddresses.Store do
+  @moduledoc ~S"""
+    Module with functions for working with Daily Active Users
+  """
+
+  use Sanbase.Influxdb.Store
+  alias __MODULE__
+
+  def daily_active_addresses(measurement, from, to, interval) do
+    daily_active_addresses_query(measurement, from, to, interval)
+    |> Store.query()
+    |> parse_time_series()
+  end
+
+  # Private functions
+
+  defp daily_active_addresses_query(measurement, from, to, "1d") do
+    ~s/SELECT active_addresses FROM "#{measurement}"
+    WHERE time >= #{influx_time(from)}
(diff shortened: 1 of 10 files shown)


lib/sanbase/etherbi/daily_active_addresses/store.ex
```diff
@@ -0,0 +1,29 @@
...
+  # Private functions
+
+  defp daily_active_addresses_query(measurement, from, to, "1d") do
+    ~s/SELECT active_addresses FROM "#{measurement}"
+    WHERE time >= #{influx_time(from)}
+    AND time <= #{influx_time(to)}/
+  end
+
+  defp daily_active_addresses_query(measurement, from, to, interval) do
+    ~s/SELECT SUM(active_addresses) FROM "#{measurement}"
```
valo: We should be using average here, as summing is not a valid operation in this context. If we take the sum of DAA for 2 days, this is not correct data, as there could be addresses used in the both days. A meaningful aggregation we can do is average, as that will allow to show the average unique addresses used in a given time range.

lib/sanbase/etherbi/daily_active_addresses/store.ex
```diff
@@ -12,6 +12,12 @@ defmodule Sanbase.Etherbi.DailyActiveAddresses.Store do
     |> parse_time_series()
   end
 
+  def average_daily_active_addresses(measurement, from, to) do
+    average_daily_active_addresses_query(measurement, from, to)
+    |> Store.query()
+    |> parse_time_series()
```
IvanIvanoff: This will return an empty list if there are no active addresses for this interval. We can introduce a new helper function that will map `[]` to `nil`, `0` or whatever passed as argument a this case will be common with all queries that return just 1 aggregated number instead of list of results.

valo: This is not a problem because I use `List.first` which will return `nil` if the list is empty

PR #384, opened 2018-03-31
Ex_admin - in projects index page put actions column (view/edit/delete) in front

MilanRusev:


valo:
Can you please add a screenshot of the new index page?

PR #386, opened 2018-04-03
Fix duplicate erc20 projects

tspenov:
#### Summary

Project joins only with the initial ico

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

**The ERROR:**
image

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/project_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/project_resolver.ex
@@ -54,13 +54,16 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     query =
       from(
         p in Project,
-        inner_join: ico in Ico,
-        on: p.id == ico.project_id,
         inner_join: infr in Infrastructure,
         on: p.infrastructure_id == infr.id,
-        where:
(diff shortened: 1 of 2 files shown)


tspenov:
Also here we have the same join: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/project_resolver.ex#L82 . We don't have any duplication there cause it is looking for icos with: `is_nil(ico.main_contract_address)`

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -54,13 +54,16 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
         on: p.infrastructure_id == infr.id,
-        where:
-          not is_nil(p.coinmarketcap_id) and not is_nil(ico.main_contract_address) and
-            infr.code == "ETH",
+        where: not is_nil(p.coinmarketcap_id) and infr.code == "ETH" and fragment("? = (
+              select project_id
+              from icos where
+              icos.project_id = ? and
+              icos.main_contract_address is not null
+              order by start_date
```
IvanIvanoff: I'm note sure if it's possible to have different contract addresses for two ico records. If it is it'll be better to sorted by descending end_date and thus take the 'latest' one.

PR #389, opened 2018-04-05
Add total market tracking

LucoLe:
#### Summary

Adds a scraper for the historical data on the graph2 `global/marketcap-total`
endpoint.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -52,20 +85,31 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
...
-      update_last_cmc_history_datetime(project, measurement_points)
+      update_last_cmc_history_datetime(project.coinmarketcap_id, measurement_points)
     end)
     |> Stream.run()
   end
 
   def update_last_cmc_history_datetime(_project, []), do: :ok
 
-  def update_last_cmc_history_datetime(%Project{coinmarketcap_id: coinmarketcap_id}, points) do
+  def update_last_cmc_history_datetime(ticker, points) do
```
IvanIvanoff: That's really `coinmarketcap_id`(the db field name has wrong name). Also I think it's better to pass the whole project struct here and use `%Project{coinmarketcap_id: coinmarketcap_id}`. The reason I like it is that the user of the function does not have to know what is used - he'll just pass the project struct.

lib/sanbase/external_services/coinmarketcap.ex
```diff
@@ -48,6 +48,11 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
       |> where([p], not is_nil(p.coinmarketcap_id))
       |> Repo.all()
 
+    Task.Supervisor.async_nolink(
+      Sanbase.TaskSupervisor,
+      &fetch_and_process_marketcap_total_data/0
+    )
```
valo: Shouldn't we wait for the async task to finish? Otherwise we can end up with many instances of the same async task.

lib/sanbase/external_services/coinmarketcap/graph_data.ex
valo: I think the problem is that we don't have a `Project` for the total market...

valo: But may be we should keep the name `coinmarketcap_id` as this is more correct than `ticker` in this context.

lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -52,20 +85,31 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
...
     last_price_datetime_updated =
       points
       |> Enum.max_by(&Measurement.get_timestamp/1)
       |> Measurement.get_datetime()
 
-    Store.update_last_history_datetime_cmc(coinmarketcap_id, last_price_datetime_updated)
+    Store.update_last_history_datetime_cmc(ticker, last_price_datetime_updated)
+  end
+
+  defp price_points_to_measurements(%PricePoint{} = price_point) do
```
IvanIvanoff: These functions for price points will fit better the `PricePoint` module

lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -149,11 +231,19 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
     end)
   end
 
-  defp graph_data_all_time_url(ticker) do
+  defp graph_data_currencies_all_time_url(ticker) do
     "/currencies/#{ticker}/"
```
valo: Here I think we should also use `coinmarketcap_id`, not `ticker`.

lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -118,8 +187,21 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
...
 
+  defp extract_price_points_for_interval({start_interval_sec, end_interval_sec}) do
+    graph_data_marketcap_total_interval_url(start_interval_sec * 1000, end_interval_sec * 1000)
+    |> get()
+    |> case do
+      %{status: 200, body: body} ->
+        body |> json_to_price_points()
+
+      _ ->
+        Logger.error("Failed to fetch graph data for interval")
```
IvanIvanoff: We can improve the error logs here so we can distinguish them. Here it could be logged that it's for the whole marketcap and in the other `extract_price_points_for_interval` the `coinmarketcap_id` can be added to the log.

lib/sanbase/external_services/coinmarketcap.ex
IvanIvanoff: After that one we'll start many other async operations with `max_concurency: 5` and they'll get rate limited before this one many times. On the other hand there is alot we can and should improve around these Task.Supervisors here.

lib/sanbase/external_services/coinmarketcap/graph_data.ex
LucoLe: yes sorry for the misunderstanding I'll change it to `coinmarketcap_id`

lib/sanbase/external_services/coinmarketcap.ex
```diff
@@ -48,6 +48,12 @@ defmodule Sanbase.ExternalServices.Coinmarketcap do
       |> where([p], not is_nil(p.coinmarketcap_id))
       |> Repo.all()
 
+    Task.Supervisor.async_nolink(
+      Sanbase.TaskSupervisor,
+      &fetch_and_process_marketcap_total_data/0,
+      timeout: :infinity
```
valo: There is no such option of the `async_nolink`. Timeout can be set on the `await` function

valo: Please read the docs

valo:
Please add an example GQL query to fetch the data

PR #394, opened 2018-04-09
Handle error tuples in Store.import/1 and log them as warnings

IvanIvanoff:
When there's no check from the module calling `Store.import/1` it often happens that we try to `map` and error tuple `{:error, reason}` which results in sentry errors


valo:
When does this happen? Link to sentry?

IvanIvanoff:
https://sentry.stage.internal.santiment.net/sentry/sanbase/issues/335/

IvanIvanoff:
Done

PR #397, opened 2018-04-11
Use tags for insights

tspenov:
#### Summary
Switch from related projects to tags for the insights

card: https://favro.com/card/bdbb4f24afc48b29c74f4fa4/San-1469

- [x] Migration to add Tags and PostsTags tables
- [x] Seed tags - "Crypto market" (for posts concerning all projects) and projects.ticker for post concerning particular project
- [x] create_post graphql mutation -> handle array of tags as arg, add test
- [x] Add`realated_posts` in project resolver
- [x] Add a test for this for the above
- [x] Insert new tag in tags when new project is added
- [x] Tested migrations on staging dump

Example query:
```
    mutation {
      createPost(title: "Awesome post2", text: "Example body", tags:["SAN", "ETH"]) {
				id,
        relatedProjects {
          ticker,
          name
        }
      }
    }
```
response:
```
{
  "data": {
    "createPost": {
      "relatedProjects": [
        {
          "ticker": "ETH",
          "name": "Ethereum"
        },
        {
          "ticker": "SAN",
          "name": "Santiment Network Token"
        }
      ],
      "id": "100"
    }
  }
}
```

```
{
  allProjects {
    relatedPosts {
      id,
      text
    }
 }
}
```

```
{
  "data": {
    "allProjects": [
      {
        "relatedPosts": [
          {
            "text": "Example body",
            "id": "99"
          },
          {
            "text": "Example body",
            "id": "100"
          }
        ]
      },
  ]
 }
}
```
**allTags**

```
{
  allTags {
    name
  }
}
```

```
{
  "data": {
    "allTags": [
      {
        "name": "Crypto Market"
      },
      {
        "name": "STX"
      },
      {
        "name": "EOS"
      },
      {
        "name": "SRN"
      }, 
    ...
```


lib/sanbase_web/admin/model/project.ex
```diff
@@ -122,6 +125,7 @@ defmodule Sanbase.ExAdmin.Model.Project do
     controller do
       # doc: https://hexdocs.pm/ex_admin/ExAdmin.Register.html#after_filter/2
       after_filter(:set_defaults, only: [:new])
+      after_filter(:add_tag, only: [:create])
```
tspenov: @valo Is it the right to insert in tags when new project is added from admin ? Is the admin adding it with `ticker` and `coinmarketcap_id` if he knows them or they are scraped later ?

IvanIvanoff: We need the `coinmarketcap_id` to be inserted in order to scraper other things, the ticker included. You can check: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/coinmarketcap/scraper.ex

lib/sanbase_web/admin/model/project.ex
```diff
@@ -149,6 +153,13 @@ defmodule Sanbase.ExAdmin.Model.Project do
     {conn, params, resource}
   end
 
+  def add_tag(conn, params, resource, :create) do
```
IvanIvanoff: In a non-bang function we call bang function (`add_tag!`). Can we improve the naming, or better - the handling by not calling `Repo.insert!` and use `with` here?

lib/sanbase_web/graphql/schema/project_types.ex
```diff
@@ -32,7 +32,10 @@ defmodule SanbaseWeb.Graphql.ProjectTypes do
     field(:description, :string)
     field(:token_decimals, :integer)
     field(:eth_addresses, list_of(:eth_address), resolve: dataloader(SanbaseRepo))
-    field(:related_posts, list_of(:post), resolve: dataloader(SanbaseRepo))
+
+    field(:related_posts, list_of(:post)) do
+      resolve(&ProjectResolver.posts/3)
```
IvanIvanoff: This is a heavy query - it has two `join`s. We should cache it. Note that this is not a first level resolver, so we cannot currently use `cache_resolve` as it has a parent object. We have example for this - `eth_spent_over_time` in `project_resolver.ex`. We directly use `Cache.func` to achieve this.

Another thing is the question do we want to cache this for 5 minutes or is it data we should cache for shorter time?

priv/repo/migrations/20180412083038_drop_posts_projects_table.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Sanbase.Repo.Migrations.DropPostsProjectsTable do
```
IvanIvanoff: We'll have some time that the old code will run with the migrated database, i.e. we'll have code using `posts_projects` but not such table in the db. In this case I'm not sure if it will cause any problems as it's not actively used, but in general we want to make this migration in a new PR.

lib/sanbase_web/graphql/schema/voting_types.ex
```diff
@@ -23,7 +27,7 @@ defmodule SanbaseWeb.Graphql.VotingTypes do
     field(:state, :string)
     field(:moderation_comment, :string)
     field(:images, list_of(:image_data), resolve: dataloader(SanbaseRepo))
-    field(:related_projects, list_of(:project), resolve: dataloader(SanbaseRepo))
```
IvanIvanoff: I'm wondering if we should keep this field, too. Without it we cannot link a post to a project directly, it has to be done with some filtering + queries from the frontend (maybe, not sure?). 
It'll be using a subset of the `tags` as the field below, but it will link to `:project` type

Another thing I was thinking - is it a good idea in practice to 'enrich' our tags? For example we can have 'financial' tags, 'project' tags, 'prediction' tags, etc. This way we can also display them in different colors in the frontend and a quick look will tell you what the whole post is about.
@tspenov @valo what do you think?

lib/sanbase_web/graphql/schema/project_types.ex
tspenov: Added caching of the function. Lets try with the default 5 minutes

lib/sanbase_web/graphql/schema/voting_types.ex
tspenov: @IvanIvanoff  Added the `related_projects` fetched by the tags.

lib/sanbase_web/graphql/resolvers/project_resolver.ex
```diff
@@ -688,4 +690,25 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
       end
     end)
   end
+
+  def posts(%Project{ticker: ticker} = project, _args, _resolution) when is_nil(ticker),
+    do: {:ok, []}
+
+  def posts(%Project{ticker: ticker} = project, _args, _resolution) do
+    async(Cache.func(fn -> fetch_posts_by_ticker(ticker) end, {:posts, ticker}))
```
IvanIvanoff: The fields in a query are resolved sequentially. `posts` is the single field of a query  - you do not need `async` here. Also `posts` is the resolver for `all_insights` and it'll be better to keep the names the same

lib/sanbase_web/admin/model/project.ex
tspenov: fixed - use `Repo.insert` ignoring the result.

lib/sanbase/external_services/project_info.ex
```diff
@@ -84,8 +85,25 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
...
 
+  defp insert_tag({:ok, project}, project_info) do
+    do_insert_tag(project, project_info)
+    {:ok, project}
+  end
+
+  defp insert_tag({:error, reason}, _), do: {:error, reason}
+
+  defp do_insert_tag(%Project{coinmarketcap_id: coinmarketcap_id}, %ProjectInfo{ticker: ticker})
+       when not is_nil(ticker) and not is_nil(coinmarketcap_id) do
```
IvanIvanoff: why do we need the not-nil coinmarketcap?

lib/sanbase_web/admin/model/project.ex
tspenov: Added - Try to insert tags when updating the project with new ticker. https://github.com/santiment/sanbase2/pull/397/files#diff-23381d0b63fcb34f5d10f1a46d8dd8deR88

tspenov:
@IvanIvanoff addressed the comments

lib/sanbase_web/graphql/schema/voting_types.ex
Partysun: @IvanIvanoff It's a good idea. Now we can make simple text tags. I can make difference between tags on frontend, it will be flexible now.

Partysun: @IvanIvanoff And about related_projects, if we can sort insights by tags, why do we really need this?
We need the only list of project tags.

tspenov: @Partysun all the tags are: project tickers and "Crypto Market" which is like tagging all projects.
Do you need a API query which returns a list of all tags ?

tspenov: @Partysun Added `allTags` query which will list all the tags.

lib/sanbase/external_services/project_info.ex
```diff
@@ -84,8 +85,25 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
       |> Project.changeset(Map.from_struct(project_info))
       |> Repo.update!()
     end)
+    |> insert_tag(project_info)
   end
 
+  defp insert_tag({:ok, project}, project_info) do
+    do_insert_tag(project, project_info)
```
IvanIvanoff: `case do_insert_tag(project, project_info) do
  {:ok, _} -> :ok
  {:error, changeset} -> IO.warn("Cannot insert tag on project creation. Reason: #{inspect(changeset)}")
end`

PR #410, opened 2018-04-19
Upgrade hackney to version 1.12.1

valo:
Seems like we are experiencing issue
https://github.com/benoitc/hackney/issues/420


greysteil:
Would you be up for using [Dependabot](https://dependabot.com) to create dependency update PRs like this one automatically? It just creates PRs whenever one of your dependencies can be updated. I built it and it's free - I'm just trying to get more people using it. :octocat:

valo:
Thanks for sharing. We will check it out.

greysteil:
Awesome. Let me know if you have any feedback when you do - always trying to make it better 🙂

PR #413, opened 2018-04-20
Add dataloader caching

IvanIvanoff:
#### Summary

Also separated the project resolver into two resolvers. Ultimately there should be more resolvers, separated logically 

From the screenshot we can see that the second request's value is returned from the cache

### DIFF
--- a/lib/sanbase_web/graphql/helpers/cache.ex
+++ b/lib/sanbase_web/graphql/helpers/cache.ex
@@ -15,14 +15,27 @@ defmodule SanbaseWeb.Graphql.Helpers.Cache do
     end
   end
 
+  defmacro cache_resolve_dataloader(captured_mfa_ast) do
+    quote do
+      middleware(
+        Absinthe.Resolution,
+        CacheMod.dataloader_from(unquote(captured_mfa_ast))
+      )
+    end
+  end
+
   def from(captured_mfa) when is_function(captured_mfa) do
-    fun_name =
-      captured_mfa
-      |> :erlang.fun_info()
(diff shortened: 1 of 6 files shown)


lib/sanbase_web/graphql/helpers/cache.ex
```diff
@@ -15,64 +26,211 @@ defmodule SanbaseWeb.Graphql.Helpers.Cache do
...
+
+          # Wrap in a tuple to distinguish value = nil from not having a record
+          {:ok, value} ->
+            value
+        end
+
+      %{}, args, resolution ->
+        cache_key = cache_key(name, args)
+
+        case ConCache.get(@cache_name, cache_key) do
```
tspenov: This can be extracted into a function cause its the same code above - just the `cache_key ` is different

IvanIvanoff: done

lib/sanbase_web/graphql/schema/project_types.ex
```diff
@@ -136,25 +141,25 @@ defmodule SanbaseWeb.Graphql.ProjectTypes do
...
     end
 
     field(:icos, list_of(:ico), resolve: assoc(:icos))
 
     field :signals, list_of(:signal) do
-      resolve(&ProjectResolver.signals/3)
+      cache_resolve_dataloader(&ProjectResolver.signals/3)
     end
 
     field :price_to_book_ratio, :decimal do
```
valo: Let's cache this resolver too

PR #419, opened 2018-04-23
Restore missing project data

LucoLe:
#### Summary

Removes the `bitcointalk_link` since it's duplicating `btt_link`.
Makes `email` field editable in ExAdmin.
Restores missing project data from the backup.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/migrations/20180423115739_remove_bitcointalk_link_from_projects.exs
```diff
@@ -0,0 +1,13 @@
...
+  use Ecto.Migration
+
+  def up do
+    alter table(:project) do
+      remove(:bitcointalk_link)
+    end
+  end
+
+  def down do
+    raise Ecto.MigrationError, "Irreversible migration!"
```
IvanIvanoff: I think it's better to just leave it empty

priv/repo/migrations/20180423130032_restore_lost_project_data.exs
```diff
@@ -0,0 +1,47 @@
...
+      from(
+        p in Project,
+        order_by: [asc: :id]
+      )
+
+    Repo.all(query)
+    |> Enum.map(fn project ->
+      backup_record =
+        backup_records
+        |> Enum.find(fn row -> row["coinmarketcap_id"] == project.coinmarketcap_id end)
```
IvanIvanoff: `Enum.find` returns `nil` if the record is not present. Let's add a `case` here:
```elixir
case backup_records |> Enum.find(...) do
  nil -> :ok
  backup_record -> ....
end
```

priv/repo/migrations/20180423130032_restore_lost_project_data.exs
```diff
@@ -0,0 +1,47 @@
...
+        slack_link: project.slack_link || backup_record["slack_link"],
+        facebook_link: project.facebook_link || backup_record["facebook_link"],
+        whitepaper_link: project.whitepaper_link || backup_record["whitepaper_link"]
+      })
+      |> Repo.update!()
+    end)
+  end
+
+  def down do
+    raise Ecto.MigrationError, "Irreversible migration!"
```
IvanIvanoff: again, I do not think it's good to raise here

priv/repo/structure.sql
```diff
@@ -2,14 +2,15 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.6.3
```
tspenov: @valo @IvanIvanoff @LucoLe One thing to note that does not have much in common with this PR except that the changes in the dumped structure are visible. (mostly tables are prefixed with `public` schema which is the default).  We have `psql (9.6.3, server 9.6.6)` on staging, probably on production is the same.  OSX `brew` has already `postgresql 10.3` packaged. Here are the feature differencies between versions: https://www.postgresql.org/about/featurematrix/.  We should probably try to upgrade at some point to 10.3 ?

tspenov: Also using different versions in development will result in this big diff ...

priv/repo/migrations/20180423115739_remove_bitcointalk_link_from_projects.exs
LucoLe: IMO it's a good practice to make a migration that deletes data irreversible but if you insist I'll remove that line.

priv/repo/migrations/20180423130032_restore_lost_project_data.exs
IvanIvanoff: PS: You can also use `change` instead of `up` and `down` and ecto will figure it out how to make up/down (iff possible)

priv/repo/structure.sql
valo: Seems like AWS RDS supports Postgresql 10.3 so we can definitely try to upgrade the DB. I think that might be one of the first tasks for the devops. I added a card https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1546

priv/repo/migrations/20180423130032_restore_lost_project_data.exs
valo: I don't think we ever run migrations backwards, so such method could be useful only if you want to commit a reverse migration and you want some reference how to write it...

PR #428, opened 2018-04-25
Remove main contract address from icos

tspenov:
#### Summary

A migration to remove `main_contract_address` column from `icos` table because it is already in `project`.


priv/repo/structure.sql
```diff
@@ -673,7 +672,6 @@ CREATE TABLE project (
     total_supply numeric,
     description text,
     email character varying(255),
-    bitcointalk_link character varying(255),
```
tspenov: seems also that `bitcointalk_link ` is not removed from structure.sql after being dropped

PR #433, opened 2018-04-27
Refactor ProjectCard component

LucoLe:
#### Summary

Changes the ProjectCard component to show a small card in mobile view
initially and after the user clicks on the card it expands to show more
data.

Updates the ProjectsMobile component so the size of each card is
calculated automatically. This allows for having cards with different
sizes and also when the size of the card changes (when the user clicks
on the card) everything updates properly.

[//]: # (#### Related PRs)

#### Screenshots

image

### DIFF
--- a/app/src/pages/Projects/ProjectCard.css
+++ b/app/src/pages/Projects/ProjectCard.css
@@ -3,8 +3,8 @@
   color: #d3d3d3 !important;
 }
 
-.project-card {
-  margin-bottom: 10px;
+.project-card > :first-child {
+  border-radius: 0 !important;
 }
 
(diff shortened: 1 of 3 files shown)


valo:
Does this also work for the currencies?

LucoLe:
yes

Partysun:
@LucoLe Can we remove 'more' from a small card. All card should be clickable. And remove margin between cards, make them, as a list.

LucoLe:
ok

LucoLe:
but I don't think it's a good idea to remove the margin between the cards. Are you sure we should show it as a list?

PR #441, opened 2018-05-02
Fetch the contract info by slug instead of ticker to avoid duplicates

IvanIvanoff:
#### Summary
We now have duplicate tickers that both have coinmarketcap id - that causes troubles.

#### Screenshots
Support both `ticker` and `slug` arguments for smoother transition.

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
@@ -10,6 +10,32 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
   require Logger
 
   @doc ~S"""
+    Return the token burn rate for the given ticker and time period.
+    Uses the influxdb cached values instead of issuing a GET request to etherbi
+  """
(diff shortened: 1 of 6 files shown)


IvanIvanoff:
I see no problem that the old code will continue to try tickers - we do not have changes in the DB or anything that will make the old code invalid. Or am I missing someting?

valo:
When we deploy the changes the old and the new code will be running at the same time for a while, so we will continue to get the old GQL requests for a while and they will fail because we search by slug and not by ticker in the new code.

IvanIvanoff:
In that case the question is when we can afford that downtime. The problem that this PR solves is observed in 1 or 2 small projects and does not disrupt the work of sanbase - we'll only see the error alerts

valo:
When will the frontend changes happen for this? Shouldn't we add them in this PR so that we ship both frontend and API changes together to production?

IvanIvanoff:
I've updated the PR with support for both old and new behaviour. It should be safe to merge it and make changes to the frontend without having downtime due to caching.

PR #445, opened 2018-05-04
Store insight as html

LucoLe:
#### Summary

Stores the insight as html instead of markdown.
Uses the medium-draft editor to display the stored insights so both have
the same look.
Removes unused libraries.

[//]: # (#### Related PRs)

#### Screenshots

image


Partysun:
1. Does it work with `edit draft`?

2. Script can be in onClick handler of any tag. Does it saniteze right?

3. After remove `export` notation, does everything work?

app/src/pages/EventVotesNew/CreateInsight.js
```diff
@@ -1,25 +1,29 @@
-import React from 'react'
-import 'medium-draft/lib/index.css'
+import React, {Component} from 'react'
```
Partysun: Consistent spaces

LucoLe:
1. Is not related to the changes in this PR.
2. Yes, I added a test and the scripts are stripped.
3. Everything should work exactly in the same way as before.

PR #453, opened 2018-05-08
Add oauth endpoint

tspenov:
#### Summary

card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1473
Integrates `Hydra` oauth2 server with current sanbase login methods - metamask and email.

Integrate with local `discourse` configuration:
Using this oauth2 discourse plugin: https://github.com/discourse/discourse-oauth2-basic
`Admin -> Settings -> Login`
Create client in `Hydra` - `discourse`, `discourse-secret` and add the same policy as grafana.
Configure discourse:

image


IvanIvanoff:
Can we have a folder in sanbase with `.md` files with the given setup how-tos and link them in the `README.md`

lib/sanbase/oauth2/hydra.ex
```diff
@@ -0,0 +1,74 @@
...
+        :error
+    end
+  end
+
+  defp get_config(key) do
+    Application.fetch_env!(:sanbase, Sanbase.Hydra)
+    |> IO.inspect()
+    |> Keyword.fetch!(key)
+    |> parse_config_value()
+  end
```
IvanIvanoff: Use the `Config` module for fetching the config. Also it would be better use in the config the same name as the module here, so you can just write `Config.get(key)`

lib/sanbase/oauth2/hydra.ex
```diff
@@ -0,0 +1,74 @@
...
+      %HTTPoison.Response{body: body, status_code: 200} ->
+        {:ok, Poison.decode!(body)["access_token"]}
+
+      _ ->
+        {:error, "Cannot get access token"}
+    end
+  end
+
+  def get_consent_data(consent, access_token) do
+    HTTPoison.get!(consent_url() <> "/#{consent}", [
```
IvanIvanoff: If you want to return `:error` tuples better use `HTTPoison.get` (without `!`)

lib/sanbase/oauth2/hydra.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  def accept_consent(
+        consent,
+        access_token,
+        %User{username: username, id: id, email: email} = _user
+      ) do
+    data = %{
+      "grantScopes" => ["openid", "offline", "hydra.clients"],
+      "accessTokenExtra" => %{},
+      "idTokenExtra" => %{name: username, email: email || username},
```
valo: What could be a possible mechanism to change the access of the user as his SAN token balance is changing? The use case:

1. User login and has 100 SAN on his address
2. User login into grafana and we authorize the access because he has 100 SAN tokens
3. User moves his 100 tokens to another address
4. We need to revoke the grafana access

lib/sanbase/oauth2/hydra.ex
```diff
@@ -0,0 +1,74 @@
...
+
+      _ ->
+        {:error, "Cannot get request url"}
+    end
+  end
+
+  def accept_consent(
+        consent,
+        access_token,
+        %User{username: username, id: id, email: email} = _user
```
IvanIvanoff: You do not need the `_user` here

lib/sanbase/oauth2/hydra.ex
```diff
@@ -0,0 +1,74 @@
...
+      ) do
+    data = %{
+      "grantScopes" => ["openid", "offline", "hydra.clients"],
+      "accessTokenExtra" => %{},
+      "idTokenExtra" => %{name: username, email: email || username},
+      "subject" => "user:#{id}:#{username}"
+    }
+
+    HTTPoison.patch!(consent_url() <> "/#{consent}/accept", Poison.encode!(data), [
```
IvanIvanoff: To indicate errors with `:ok/:error` returns better use the `HTTPoison.patch` function

lib/sanbase_web/controllers/root_controller.ex
```diff
@@ -1,13 +1,32 @@
...
+        %{
+          "consent" => consent,
+          "token" => token
+        } = _params
+      ) do
+    {:ok, user} = bearer_authorize(token)
+    {:ok, access_token} = Hydra.get_access_token()
+    {:ok, redirect_url} = Hydra.get_consent_data(consent, access_token)
+    :ok = Hydra.accept_consent(consent, access_token, user)
+    redirect(conn, external: redirect_url)
```
IvanIvanoff: If a match fails here the process will crash. I think a `with` construct with a predefined redirect/message describing that an error here occurred will be better

priv/repo/migrations/20180504071348_add-consent_id-users.exs
```diff
@@ -0,0 +1,9 @@
+defmodule :"Elixir.Sanbase.Repo.Migrations.Add-consentId-users" do
```
IvanIvanoff: If you use snake case with `mix ecto.gen.migration migration_name` the module name will be in a nice format

lib/sanbase/oauth2/hydra.ex
tspenov: `Grafana` doesn't save the `access_token` so it basically doesn't take into account the token `expires_in` - so once the user is logged in he stays that way based on the cookies. One think we can do to support this scheme is to put `session_life_time = 60 (1 min)
login_remember_days = 0 (no remember cookie)` - 1 min session without remembering which will force logging in. And in the consent app (sanbase) we can `accept/reject` based on the SANs in the account. @valo what do you think - does this works for our case ?

valo: What will be the UX? Does it mean you will be logged out every minute? Can we test such setup locally first?

tspenov: I tested it locally - if you close and doesn't have open `grafana` tabs next time you open it - you will have to log in. Otherwise if you have a tab with grafana open - you will stay logged in. Have to check twice but I think it works like this

tspenov: Actually the session is refreshing only if the open tab is refreshing (if there is a dashboard that is reloading) So it maybe it will make more sense to be set to 30min or so ?

lib/sanbase/oauth2/hydra.ex
```diff
@@ -5,6 +5,9 @@ defmodule Sanbase.Oauth2.Hydra do
   alias Sanbase.Utils.Config
   alias Sanbase.Auth.User
 
+  @clients_require_san_tokens ["grafana"]
+  @required_san_tokens 100
```
IvanIvanoff: Can we make it configurable via ENV var?

PR #454, opened 2018-05-09
Rework cmc price fetching

IvanIvanoff:
#### Summary

Currently there is a serious issue with fetching prices. The prices measurement consists solely of the ticker, which is not unique. We also scrape the whole CMC list of prices, so these duplicates happen often. Example is the KNC ticker.

This PR reworks the fetching so it now uses a single measurement `SAN_santiment` (ticker_cmc_id) instead of two measurements `SAN_USD` and `SAN_BTC`. The structure of the code is mostly the same as the old, hence the big number of LOC added.

As a first step this new scrapers should be enabled so they can catch up with the historical data. On a second step in a different PR the places where these prices are used (GQL, dataloaders, etc.) should be changed to use the new measurements.

Also at some point **before** switching off the old scrapers Grafana dashboards should be reworked, too. Only after that we can delete the old code and rename all modules ending with **2** to the original name.

### DIFF
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -65,8 +65,8 @@ defmodule Sanbase.Application do
         Sanbase.ExternalServices.RateLimiting.Server.child_spec(
           :graph_coinmarketcap_rate_limiter,
           scale: 60_000,
-          limit: 20,
-          time_between_requests: 2000
+          limit: 30,
+          time_between_requests: 1000
         ),
 
         # Coinmarketcap api rate limiter
@@ -81,8 +81,8 @@ defmodule Sanbase.Application do
         Sanbase.ExternalServices.RateLimiting.Server.child_spec(
           :http_coinmarketcap_rate_limiter,
           scale: 60_000,
-          limit: 20,
-          time_between_requests: 2000
+          limit: 30,
+          time_between_requests: 1000
         ),
 
         # Twitter API rate limiter
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -102,10 +102,12 @@ defmodule Sanbase.Application do
         ),
 
         # Price fetcher
+        # TODO: Change after switching over to only this cmc
         Sanbase.ExternalServices.Coinmarketcap.child_spec(%{}),
         Sanbase.ExternalServices.Coinmarketcap2.child_spec(%{}),
 
         # Current marketcap fetcher
+        # TODO: Change after switching over to only this cmc
         Sanbase.ExternalServices.Coinmarketcap.TickerFetcher.child_spec(%{}),
         Sanbase.ExternalServices.Coinmarketcap.TickerFetcher2.child_spec(%{}),
 
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -103,9 +103,11 @@ defmodule Sanbase.Application do
 
         # Price fetcher
         Sanbase.ExternalServices.Coinmarketcap.child_spec(%{}),
+        Sanbase.ExternalServices.Coinmarketcap2.child_spec(%{}),
 
         # Current marketcap fetcher
         Sanbase.ExternalServices.Coinmarketcap.TickerFetcher.child_spec(%{}),
+        Sanbase.ExternalServices.Coinmarketcap.TickerFetcher2.child_spec(%{}),
 
         # Etherscan wallet tracking worker
         Sanbase.ExternalServices.Etherscan.Worker.child_spec(%{}),
--- a/lib/sanbase/external_services/coinmarketcap2.ex
+++ b/lib/sanbase/external_services/coinmarketcap2.ex
@@ -45,7 +45,8 @@ defmodule Sanbase.ExternalServices.Coinmarketcap2 do
        %{
          missing_info_update_interval: update_interval * 10,
          total_market_update_interval: update_interval,
-         prices_update_interval: update_interval
+         prices_update_interval: update_interval,
+         total_market_task_pid: nil
        }}
     else
       :ignore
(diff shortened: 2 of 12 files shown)


lib/sanbase/external_services/coinmarketcap2.ex
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  def handle_info(:fetch_total_market, %{total_market_update_interval: update_interval} = state) do
+    # Start the task under the supervisor in a way that does not need await.
+    # As there is only one data record to be fetched we fire and forget about it,
+    # so the work can continue to scraping the projects' prices in parallel.
+    Task.Supervisor.start_child(
+      Sanbase.TaskSupervisor,
+      &fetch_and_process_marketcap_total_data/0
+    )
```
valo: What will happen if we need to scrape the total market data from the beginning? If the whole scrape is too long we might end up with running it twice. May be we should keep the scraping synchronous or wait for it to finish before scheduling it to run again?

lib/sanbase/external_services/coinmarketcap2.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  def init(_arg) do
+    if Config.get(:sync_enabled, false) do
+      # Create an influxdb if it does not exists, no-op if it exists
+      Store.create_db()
+
+      # Start scraping immediately
+      Process.send(self(), :fetch_missing_info, [:noconnect])
+      Process.send(self(), :fetch_total_market, [:noconnect])
+      Process.send(self(), :fetch_prices, [:noconnect])
```
valo: Why don't we use `GenServer.cast` here, like in https://github.com/santiment/sanbase2/pull/454/files#diff-bf6d74d6c68ad3ea1513465a6009eb66R32?

lib/sanbase/external_services/coinmarketcap2/price_point.ex
```diff
@@ -0,0 +1,63 @@
...
+         volume_usd: volume_usd,
+         price_btc: price_btc,
+         price_usd: price_usd
+       }) do
+    %{
+      price_usd: price_usd,
+      price_btc: price_btc,
+      volume_usd: volume_usd,
+      marketcap_usd: marketcap_usd
+    }
```
valo: If we keep the data in this format will it be possible to switch between USD and BTC prices in Grafana as it is done now?

lib/sanbase/external_services/coinmarketcap2.ex
IvanIvanoff: Because it's not nice for using when we have to send a message after a period of time. If we want to do that we have to use some erlang internal knowledge of how `cast` message is represented and use
`Process.send_after(self(), {:"$gen_cast", :sync}, update_interval)`. We do not have such issue if we use plain `send` and handle the message in `handle_info` instead of `handle_cast`

lib/sanbase/external_services/coinmarketcap2.ex
IvanIvanoff: The easiest way is to make it blocking until it finishes. 

Another solution is to continue running it in parallel in a new process under a supervisor, but start it as MFA (that will be a new module) as a named process. Then we'll be able here to match against `{:ok, ...}` and `{:error, {:already_started, ...}}`. That does not add much benefit for the work. It may also not be possible to start it under the Task.Supervisor in that case anymore..

lib/sanbase/external_services/coinmarketcap2/price_point.ex
IvanIvanoff: When I asked @MilanRusev he said it should not be hard.

lib/sanbase/external_services/coinmarketcap2.ex
IvanIvanoff: @valo  I've addressed this issue by keeping the Task pid in the GenServer's state and run it only if the previous task has been completed (or has not yet been started).

lib/sanbase/external_services/coinmarketcap2/ticker.ex
```diff
@@ -0,0 +1,102 @@
...
+    |> Stream.filter(fn %Ticker{last_updated: last_updated} -> last_updated end)
+    |> Enum.map(&make_timestamp_integer/1)
+  end
+
+  def convert_for_importing(
+        %Ticker{
+          last_updated: last_updated,
+          price_btc: price_btc,
+          price_usd: price_usd,
+          "24h_volume_usd": volume_usd,
```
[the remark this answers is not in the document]
IvanIvanoff: It's an atom. The `""` syntax allows it to start with number

lib/sanbase/external_services/coinmarketcap2/scraper.ex
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+
+  def parse_project_page(html, project_info) do
+    %ProjectInfo{
+      project_info
+      | name: name(html),
+        ticker: ticker(html),
+        main_contract_address: main_contract_address(html),
+        website_link: website_link(html),
```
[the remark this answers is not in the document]
IvanIvanoff: Yes, I've fixed these in the master branch but have not made the changes here. Good catch.

lib/sanbase/external_services/coinmarketcap2/scraper.ex
```diff
@@ -0,0 +1,94 @@
...
+    |> List.first()
+  end
+
+  defp github_link(html) do
+    Floki.attribute(html, "a:fl-contains('Source Code')", "href")
+    |> List.first()
+  end
+
+  defp etherscan_token_name(html) do
+    Floki.attribute(html, "a:fl-contains('Explorer')", "href")
```
IvanIvanoff: There are cases with `Explorer 2` and so on. I think it'll be good to get them all and inspect them if they contain `etherscan.io`

lib/sanbase/external_services/coinmarketcap2/ticker.ex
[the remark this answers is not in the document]
tspenov: 👍

lib/sanbase/external_services/coinmarketcap2/ticker_fetcher.ex
```diff
@@ -0,0 +1,118 @@
...
+    # Store the data in LatestCoinmarketcapData in postgres
+    tickers
+    |> Enum.each(&store_latest_coinmarketcap_data/1)
+
+    # Store the data in Influxdb
+    tickers
+    |> Enum.map(&Ticker.convert_for_importing/1)
+    |> Store.import()
+
+    Process.send_after(self(), {:"$gen_cast", :sync}, update_interval)
```
valo: May be it is worth to use `handle_info` here too so that all the messages handling is consistent?

lib/sanbase/application.ex
```diff
@@ -81,8 +81,8 @@ defmodule Sanbase.Application do
         Sanbase.ExternalServices.RateLimiting.Server.child_spec(
           :http_coinmarketcap_rate_limiter,
           scale: 60_000,
-          limit: 20,
-          time_between_requests: 2000
+          limit: 30,
```
IvanIvanoff: According to https://coinmarketcap.com/api/:
```
Limits

    Please limit requests to no more than 30 per minute.
```

PR #459, opened 2018-05-15
1586 navigation changes

LucoLe:
#### Summary

Adds new navigation.
Adds Favorites page.
Adds filtering by category. 

[//]: # (#### Related PRs)

#### Screenshots

image
image


app/src/__tests__/__snapshots__/cashflow.spec.js.snap
```diff
@@ -1,329 +1,36 @@
...
-        sortable={false}
-        style={Object {}}
-        subRowsKey="_subRows"
-      />
-    </Panel>
-  </FadeIn>
-  <Component />
-</div>
-`;
+<ProjectsTabel
```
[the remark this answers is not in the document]
LucoLe: https://forums.oneplus.com/attachments/becf79994a4fe514aa44b4c2b627609afeffcdf2eb8764be889bfab34aaa9eff-jpg.460525/

app/src/pages/Detailed/Detailed.js
```diff
@@ -213,6 +220,9 @@ export const Detailed = ({
       }}
       ticker={project.ticker} />
 
+  const isLoggedIn = Object.keys(user).length > 0 && user.constructor === Object
```
Partysun: @LucoLe Do we can use simpler '!!user.token' or '.account', as you did in ProjectNavigation.js?

Partysun: We can also move it in mapStateProps

Partysun:
screenshot 2018-05-16 18 27 30

Partysun:
screenshot 2018-05-16 18 29 56

app/src/components/ProjectsNavigation.js
```diff
@@ -16,44 +16,58 @@ const ProjectsNavigation = ({
...
               <Checkbox
                 key={key}
                 id={key}
                 label={value || 'Unknown'}
                 onChange={handleSetCategory}
                 checked={categories[key]}
-              />
-            )
+              />)
+            : 'Categories not founded'
```
LucoLe: I think this is not correct because we always have at least one category - `Unknown`. Please remove this ternary because it's redundant.

app/src/pages/Cashflow.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import 'react-table/react-table.css'
-import ProjectsTable from 'pages/Projects/ProjectsTable'
-import withProjectsData from 'pages/Projects/withProjectsData'
+import ProjectsTable from './Projects/ProjectsTable'
```
LucoLe: Please use global paths in imports. This makes moving components around much easier.

app/src/pages/Currencies.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import 'react-table/react-table.css'
-import ProjectsTable from 'pages/Projects/ProjectsTable'
-import withProjectsData from 'pages/Projects/withProjectsData'
+import ProjectsTable from './Projects/ProjectsTable'
```
LucoLe: Same here.

app/src/pages/Detailed/Detailed.js
```diff
@@ -221,7 +221,8 @@ export const Detailed = ({
       }}
       ticker={project.ticker} />
 
-  const isFavorite = () => isLoggedIn && project && user.followedProjects.includes(project.id)
+  const isFavorite = () => isLoggedIn && project &&
+    user.followedProjects && user.followedProjects.includes(project.id)
```
LucoLe: I think `user.followedProjects` is not necessary because we always return an empty array if there are no followed projects.

app/src/pages/Detailed/EthereumBlock.js
```diff
@@ -9,9 +9,9 @@ import PanelBlock from './../../components/PanelBlock'
 import { simpleSort } from './../../utils/sortMethods'
 import { formatNumber } from './../../utils/formatting'
 import { allErc20ShortProjectsGQL } from './../Projects/allProjectsGQL'
-import { CustomThComponent, CustomHeadComponent } from 'pages/Projects/ProjectsTable'
+import { CustomThComponent, CustomHeadComponent } from './../Projects/ProjectsTable'
```
LucoLe: same here

app/src/pages/Favorites.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import 'react-table/react-table.css'
-import ProjectsTable from 'pages/Projects/ProjectsTable'
-import withProjectsData from 'pages/Projects/withProjectsData'
+import ProjectsTable from './Projects/ProjectsTable'
```
LucoLe: and here

app/src/pages/Favorites.js
```diff
@@ -16,11 +16,17 @@ export const Favorites = ({
   preload,
   user
 }) => {
-  if (Projects.projects.length > 0 && user.followedProjects.length > 0) {
+  if (Projects.projects.length > 0 &&
+    user.followedProjects && user.followedProjects.length > 0) {
```
LucoLe: `user.followedProjects` is redundant

app/src/pages/Projects/ProjectsTable.js
```diff
@@ -7,13 +7,13 @@ import { Icon, Popup, Message, Loader } from 'semantic-ui-react'
...
 import { FadeIn } from 'animate-components'
 import Sticky from 'react-stickynode'
-import { formatNumber, millify } from 'utils/formatting'
-import { getOrigin, filterProjectsByMarketSegment } from 'utils/utils'
-import { simpleSort } from 'utils/sortMethods'
-import ProjectIcon from 'components/ProjectIcon'
-import Panel from 'components/Panel'
-import PercentChanges from 'components/PercentChanges'
-import ProjectsNavigation from 'components/ProjectsNavigation'
+import { formatNumber, millify } from '../../utils/formatting'
```
LucoLe: and here

app/src/components/AppMenu.js
```diff
@@ -37,13 +38,34 @@ const AppMenu = ({
...
-      <li
-        className={isPageFromLocation(location, 'signals') ? 'active' : ''}
-        onClick={() => handleNavigation('signals')}>
-        Signals
-      </li>
+        : <Fragment>
+          <Link
+            className='app-menu__page-link'
+            to={'/projects'}>
+            Markets
```
LucoLe: Why do we have a link to projects? Isn't the logo a link? If we need an explicit button why is it called Markets when in the projects navigation it's called ERC20 Projects? Anyway I think that this button should be removed and maybe the projects navigation should be visible on every page.

Partysun:
About why I added checks for categories and isFavorite. Very simple. I lauched only frontend server and got 400 from API, u know. And I got undefined errors in that places. We need to define defaults value. After my changes frontend app works stable without crashes in that places. If we define good defaults we can remove some tern. Op. But without it, we need to have that checks.

Partysun:
And about global path. I will use this in that project, only relative path for internal components. It s a standart pattern in cra community and my preference.

app/src/components/AppMenu.js
Partysun: Logo link will have another behavior. Tables with projects and currencies will be part of market analysing section. 
And we need to keep the link to 'projecs' table in the common place for old users.
That was the idea behind that changes.

PR #469, opened 2018-05-21
Use stored total_supply

LucoLe:
#### Summary

Removes the multiplication to decimal places so we'll use the
total_supply in the base token instead of the smallest amount.

Fixes the scraper's selector for total_supply to be closer to the real
css selector.

Uses the total_supply stored in the Project as a fallback if
LatestCoinmarketcapData is missing the total_supply.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -474,11 +474,14 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
-        %Project{latest_coinmarketcap_data: %LatestCoinmarketcapData{total_supply: total_supply}},
+        %Project{
+          total_supply: total_supply,
+          latest_coinmarketcap_data: %LatestCoinmarketcapData{total_supply: cmc_total_supply}
+        },
         _args,
         _resolution
       ) do
-    {:ok, total_supply}
+    {:ok, cmc_total_supply || total_supply}
```
IvanIvanoff: I think etherscan is more reliable source so we should use it as a first choice.

lib/sanbase/external_services/etherscan/scraper.ex
```diff
@@ -65,7 +65,7 @@ defmodule Sanbase.ExternalServices.Etherscan.Scraper do
   def parse_token_page!(html, project_info) do
     %ProjectInfo{
       project_info
-      | total_supply: project_info.total_supply || total_supply(html),
+      | total_supply:  total_supply(html) || project_info.total_supply,
```
IvanIvanoff: Maybe leave a comment that it's intended to overwrite if it exists as opposed to all other fields.

PR #486, opened 2018-06-01
Add testing san balance through admin

tspenov:
#### Summary
Add testing san_balance through admin panel so internal users can test services that require SAN

https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1688


lib/sanbase/auth/user.ex
```diff
@@ -68,7 +69,12 @@ defmodule Sanbase.Auth.User do
     |> change(san_balance: san_balance, san_balance_updated_at: Timex.now())
   end
 
-  def san_balance!(%User{san_balance: san_balance} = user) do
+  def san_balance!(%User{test_san_balance: test_san_balance} = _user)
+      when not is_nil(test_san_balance) do
+    test_san_balance
+  end
+
+  def san_balance!(%User{san_balance: san_balance, test_san_balance: test_san_balance} = user) do
```
IvanIvanoff: here I think you can just go with `%User{san_balance: san_balance} = user`

tspenov: 👍  leftover

app/src/pages/UserSelectors.js
```diff
@@ -1,21 +1,9 @@
...
 
 export const checkHasPremium = state => {
-  const ethAccounts = state.user.data.ethAccounts
-  if (ethAccounts) {
-    return state.user.data.ethAccounts.length > 0
-      ? state.user.data.ethAccounts[0].sanBalance >= 1000
-      : false
-  }
-  return false
+  return state.user.data.sanBalance >= 1000
```
tspenov: I don't think its a good idea to check this in the frontend. Have to change this to use some staked API endpoint

lib/sanbase/auth/user.ex
```diff
@@ -33,6 +33,7 @@ defmodule Sanbase.Auth.User do
     field(:email_token_generated_at, Timex.Ecto.DateTime)
     field(:email_token_validated_at, Timex.Ecto.DateTime)
     field(:consent_id, :string)
+    field(:test_san_balance, :decimal)
```
IvanIvanoff: Overall I do not like having this kind of fields. For the purpose of testing let's go with it, but have in mind that we should ultimately come up with a cleaner way

app/src/index.js
```diff
@@ -210,6 +210,7 @@ const handleLoad = () => {
           id,
           email,
           username,
+          sanBalance,
```
tspenov: @Partysun added sanBalance in the currentUser api call which we calculate at the backend. I think we can use it in the frontend instead of `ethAccounts[0].sanBalance`

lib/sanbase/auth/user.ex
tspenov: I agree that adding such kind of field is not optimal. But currently cannot think of a better way. If you have some suggestions we can change it ?

app/src/pages/UserSelectors.js
```diff
@@ -1,21 +1,10 @@
...
```
IvanIvanoff: Not good to be hardcoded IMO. If the user has not enough SAN tokens, the GQL query returns `{:error, "Insufficient SAN balance"}`. Can we use it somehow?

tspenov: I share your concerns but I think this is out of the scope of this PR cause it will require quite a refactoring in the frontend. The goal here is to use the `san_balance` from the `currentUser` and to simplify these functions by using it. Also the hardcoded value gives flexibility in the tooltip/error message and saves one backend call in the price to not use the backend as single source of truth.

valo: Seems like this hardcoded value is here before this PR. @Partysun do you have ideas how we can improve that? I would suggest to schedule a task to make this amount easily configurable.

tspenov: Added this card to the frontend board: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1764

tspenov:
@IvanIvanoff Can we proceed with merging this one ? Added a card for @Partysun for configuring this premium amount.

PR #487, opened 2018-06-08
Add ratelimit api

tspenov:
#### Summary
Add ratelimit to graphql API.

Current default limit is set to 20 requests / 10 secs because our frontend sometimes makes a couple of queries at the same time.

card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1758

### DIFF
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -100,6 +100,10 @@ defmodule Sanbase.Application do
           limit: 100,
           time_between_requests: 100
         ),
+        worker(PlugAttack.Storage.Ets, [
+          SanbaseWeb.Graphql.PlugAttack.Storage,
+          [clean_period: 60_000]
+        ]),
 
         # Price fetcher
         # TODO: Change after switching over to only this cmc
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -100,7 +100,7 @@ defmodule Sanbase.Application do
           limit: 100,
           time_between_requests: 100
         ),
-        supervisor(PlugAttack.Storage.Ets, [
+        worker(PlugAttack.Storage.Ets, [
           SanbaseWeb.Graphql.PlugAttack.Storage,
           [clean_period: 60_000]
         ]),
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -100,7 +100,7 @@ defmodule Sanbase.Application do
           limit: 100,
           time_between_requests: 100
         ),
-        worker(PlugAttack.Storage.Ets, [
+        supervisor(PlugAttack.Storage.Ets, [
           SanbaseWeb.Graphql.PlugAttack.Storage,
           [clean_period: 60_000]
         ]),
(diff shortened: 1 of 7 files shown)


IvanIvanoff:
Am I counting wrong, or 20 requests per 10 seconds won't work as the frontend makes 12(twelve) GQL requests for PDP. That means we cannot open more than 1 PDP every 10 seconds.
Also opening PDP + changing time range generates over 20 GQL requests

lib/sanbase/application.ex
```diff
@@ -100,6 +100,10 @@ defmodule Sanbase.Application do
           limit: 100,
           time_between_requests: 100
         ),
+        supervisor(PlugAttack.Storage.Ets, [
```
IvanIvanoff: `PlugAttack.Storage.Ets` does no implement supervisor behaviour, it's a worker. Seems like the library has not been updated and could not be started via `child_spec` so you have to use `worker`

lib/sanbase/application.ex
```diff
@@ -100,6 +100,10 @@ defmodule Sanbase.Application do
...
+          SanbaseWeb.Graphql.PlugAttack.Storage,
+          [clean_period: 60_000]
+        ]),
```
IvanIvanoff: `PlugAttack.Storage.Ets` is a `GenServer` and does not implement the supervisor behaviour.
This should be started via the `child_spec` - something like this:
```elixir
{PlugAttack.Storage.Ets, [
          SanbaseWeb.Graphql.PlugAttack.Storage,
          [clean_period: 60_000]
        ]
}
```

tspenov: going back to using worker because the current PlugAttack.Storage.Ets work properly with child spec.  It works with this: `PlugAttack.Storage.Ets.child_spec(SanbaseWeb.Graphql.PlugAttack.Storage)` but cannot pass the `clean_period` opt.

tspenov:
@IvanIvanoff yes we should decide whats the best ratelimit. Now updated to 40reqs/10secs.

test/sanbase_web/graphql/plug_attack_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    conn = %{conn | remote_ip: {192, 168, 0, 1}}
+
+    for _n <- 1..get_max_requests(), do: conn |> get("/graphql")
+
+    response = conn |> get("/graphql")
+
+    assert response.status == 403
+  end
+
+  defp get_max_requests() do
```
IvanIvanoff: Use the `Sanbase.Utils.Config` and `module_get` module:
```elixir
Config.module_get(SanbaseWeb.Graphql.PlugAttack, SanbaseWeb.Graphql.PlugAttack) |> String.to_integer()
```

tspenov: 👍

IvanIvanoff:
@tspenov is anything stopping us from merging this?

PR #488, opened 2018-06-11
Fix ico line

tspenov:
#### Summary
Use the max value from all icos for a project.
card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1762
image

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
@@ -563,10 +563,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
       |> Repo.get(id)
       |> Repo.preload([:icos])
       |> Map.get(:icos)
+      |> Enum.reject(fn ico -> is_nil(ico.token_usd_ico_price) end)
       |> Enum.map(fn ico ->
         %Ico{ico | token_usd_ico_price: Decimal.to_float(ico.token_usd_ico_price)}
       end)
-      |> Enum.reject(fn ico -> is_nil(ico.token_usd_ico_price) end)
       |> Enum.max_by(fn ico -> ico.token_usd_ico_price end, fn -> nil end)
       |> Map.get(:token_usd_ico_price)
 
--- a/lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
@@ -554,6 +554,24 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     {:ok, ico}
   end
 
+  def ico_price(%Project{id: id} = project, _args, _resolution) do
+    project
+
+    Project
+    |> Repo.get(id)
+    |> Repo.preload([:icos])
+
+    ico_price =
+      project.icos
+      |> Enum.map(fn ico ->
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -554,6 +554,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     {:ok, ico}
   end
 
+  def ico_price(%Project{id: id} = project, _args, _resolution) do
```
IvanIvanoff: Can you add a short comment what we do here and why we use max price?

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -554,6 +554,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+    ico_price =
+      Project
+      |> Repo.get(id)
+      |> Repo.preload([:icos])
+      |> Map.get(:icos)
+      |> Enum.map(fn ico ->
+        Map.merge(ico, %{token_usd_ico_price: Decimal.to_float(ico.token_usd_ico_price)})
```
IvanIvanoff: Won't this overwrite and we'll have one `token_usd_ico_price`?

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -554,6 +554,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+      end)
+      |> Enum.max_by(fn ico -> ico.token_usd_ico_price end, fn -> nil end)
```
IvanIvanoff: atoms are bigger than numbers. this will possibly cause problems
```elixir
iex(4)> nil > 5.5
true
iex(5)> Enum.max([1,2,3,nil])   
nil
```

tspenov: `Enum.max` uses the erlang term order. Enum.max_by shouldn't

IvanIvanoff: I took a fast look at the DB and I found such records

tspenov: @IvanIvanoff should be fixed now - filtered those

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -554,6 +554,25 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
...
+  Return the main sale price, which is the maximum from all icos of a project
+  """
+  def ico_price(%Project{id: id} = project, _args, _resolution) do
+    ico_price =
+      Project
+      |> Repo.get(id)
+      |> Repo.preload([:icos])
+      |> Map.get(:icos)
+      |> Enum.map(fn ico ->
+        Map.merge(ico, %{token_usd_ico_price: Decimal.to_float(ico.token_usd_ico_price)})
```
IvanIvanoff: We can go just with
```elixir
Enum.map(fn ico -> %Sanbase.Model.Ico{ico | token_usd_ico_price: ... } end)
```

tspenov: ok. done

IvanIvanoff:
If you have tested it let's merge

PR #497, opened 2018-06-18
Refactoring/the first june

Partysun:
#### Summary
Move Side effects to RxJS and redux-observables
Rename strange things


tspenov:
@Partysun ok, I'll check it

tspenov:
@Partysun Tested, after a small fix everything works fine

PR #502, opened 2018-06-21
Add api docs and explorer

tspenov:
#### Summary
* Allow `graphiql` API explorer on stage & prod
* Add a simple page with API examples

the url on PROD  will be  : `https://sanbase-low.santiment.net/apidocs`

[//]: # (#### Screenshots)
image
image


lib/sanbase_web/router.ex
```diff
@@ -38,21 +38,21 @@ defmodule SanbaseWeb.Router do
...
-        max_complexity: 5000
-      )
-    end
+    forward(
+      "/graphiql",
+      Absinthe.Plug.GraphiQL,
+      schema: SanbaseWeb.Graphql.Schema,
+      analyze_complexity: true,
+      max_complexity: 5000,
+      interface: :simple
```
tspenov: Only the `simple` interface handles correctly the url params which we use in `Run in explorer` and It misses the option to add headers. The advanced interface I think has some problems with the `localStorage` and does not show correctly the url params

lib/sanbase_web/router.ex
```diff
@@ -38,21 +38,21 @@ defmodule SanbaseWeb.Router do
...
+    )
   end
 
   scope "/", SanbaseWeb do
     pipe_through(:browser)
 
     get("/consent", RootController, :consent)
+    get("/apidocs", ApiDocsController, :api_examples)
```
IvanIvanoff: shouldn't this be `apiexamples`?  Also in the previous PR there was `/apiexplorer` if I'm not mistaken?

tspenov:
@IvanIvanoff 
the url when deployed to production will be:
`https://sanbase-low.santiment.net/apidocs` - for the docs and examples
`https://sanbase-low.santiment.net/apiexplorer` - for the graphiql explorer

lib/sanbase_web/router.ex
tspenov: updated things a bit - should be good now

IvanIvanoff:
One more thing - We should add links to the apidocs somewhere (in the footer most likely)

PR #504, opened 2018-06-21
Add 1day delay api

tspenov:
#### Summary
Add delay for anon users or users without `1000 SAN` stake

https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-1823


IvanIvanoff:
We need to add text somewhere to tell the user about this delay

tspenov:
@IvanIvanoff @valo where do you think we should add this info about the delay: on every query documentation that has a delay or somewhere in the `apidocs` or both ?

lib/sanbase_web/graphql/middlewares/api_delay.ex
```diff
@@ -0,0 +1,61 @@
...
+  alias Sanbase.Auth.User
+
+  @required_san_stake_realtime_api 1
+
+  def call(
+        %Resolution{
+          context: %{
+            auth: %{auth_method: :user_token, current_user: current_user}
+          },
+          arguments: %{to: _to}
```
IvanIvanoff: What's the behaviour with queries that have default `to` value of `DateTime.utc_now()`? I believe it's set before this middleware is executed, but I'm not sure.

lib/sanbase_web/graphql/middlewares/api_delay.ex
```diff
@@ -0,0 +1,61 @@
...
+        } = resolution,
+        _
+      ) do
+    unless has_enough_san_tokens?(current_user) do
```
IvanIvanoff: I've always found it more readable to have `if` instead of `unless`

lib/sanbase_web/graphql/middlewares/api_delay.ex
```diff
@@ -0,0 +1,61 @@
...
+  Currently the delay is one day and is achieved by transforming the `to` datetime
+  argument, taking the minimum between the original `to` and now() - 1day.
+  """
+
+  @behaviour Absinthe.Middleware
+
+  alias Absinthe.Resolution
+  alias Sanbase.Auth.User
+
+  @required_san_stake_realtime_api 1
```
IvanIvanoff: This should be a function that reads an ENV variable

lib/sanbase_web/graphql/middlewares/api_delay.ex
tspenov: The default value is set before this middleware and then changed. But will add a test to cover this.

lib/sanbase_web/graphql/middlewares/api_delay.ex
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  defp delay_1day(to_datetime) do
+    yesterday = yesterday()
+
+    case DateTime.compare(to_datetime, yesterday) do
+      :gt -> yesterday
+      _ -> to_datetime
+    end
```
IvanIvanoff: You can replace this function with
```
yesterday = Timex.shift(Timex.now, days: -1)
Enum.min_by([to_datetime, yesterday], &DateTime.to_unix/1)
```

lib/sanbase_web/graphql/middlewares/api_delay.ex
tspenov: Yeah adding new test will require using something other than `burn_rate` cause it does not have default `to` which will require more time to set it up.
Here is some output with added ApiDelay middleware and missing to in the query.
Logging the args it seems that to is set to `to: #DateTime<2018-06-22 14:33:38.773694Z>`
which is `DateTime.utc_now()`
```
[debug] ABSINTHE schema=SanbaseWeb.Graphql.Schema variables=%{}
---
query {
  githubActivity(
    ticker: "SAN",
    from: "2018-02-13 16:00:00Z",
    interval: "24h") {
      activity,
    	datetime
    }
}
---
[debug] %{from: #DateTime<2018-02-13 16:00:00Z>, interval: "24h", moving_average_interval: 10, ticker: "SAN", to: #DateTime<2018-06-22 14:33:38.773694Z>
```

valo:
Let’s add a comment in the API docs explaining this behavior. I guess we will need to hardcore the SAN amount, unless there is a way we can use a variable for it and make it easier to change. 

An some point we should make it easier to make authenticated API requests (using API keys?), so that users can get real time data. I think we have some work around that, like creating api keys, verifying them and adding docs...

tspenov:
@IvanIvanoff @valo Added a message in the '/apidocs' explaining the api behaviour.

lib/sanbase_web/graphql/middlewares/api_delay.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  defp has_enough_san_tokens?(current_user) do
+    if Decimal.cmp(
+         User.san_balance!(current_user),
+         Decimal.new(required_san_stake_realtime_api())
+       ) != :lt do
+      true
+    else
+      false
+    end
```
IvanIvanoff: If you remove the `if-else` the behaviour is the same, operator `!=` returns true/false

PR #506, opened 2018-06-25
Update Erlang/OTP to 21 and Elixir to 1.6.6

IvanIvanoff:
Erlang/OTP 21 ships by default without tuple calls. `Mockery.of` uses
tuple calls, so update the version of `Mockery` and start using the new
`Mockery.Macro.mockable`. It cannot be used via module attribute.

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ COPY ./app /app
 RUN yarn build
 
 # Elixir and phoenix assets build image
-FROM elixir:1.6.5-alpine as code_builder
+FROM elixir:1.6.6-alpine as code_builder
 
 ENV MIX_ENV prod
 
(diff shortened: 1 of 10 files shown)


valo:
Are you sure the new API of mockery should be used? Seems like the old API is still in the docs https://github.com/appunite/mockery#getting-started

IvanIvanoff:
It's because of this issue: https://github.com/appunite/mockery/issues/22

`Mockery.of` continues to works but it uses tuple calls, so there are no problems with it and OTP < 21. OTP21 deprecates tuple calls and the code should be compiled with special flags in order to active them again.

valo:
I am wondering if we can build our own macro which simplifies the mocking. Something like:

```elixir
defmockable ethauth do: Sanbase.InternalServices.Ethauth
```

IvanIvanoff:
It's possible, but I see it as a separate task

PR #509, opened 2018-06-25
Add discourse integration

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

#### Additional deploy notes

This devops pr needs to be merged: https://github.com/santiment/devops/pull/273

[//]: # (#### Screenshots)


config/config.exs
```diff
@@ -8,6 +8,8 @@ use Mix.Config
 # General application configuration
 config :sanbase, ecto_repos: [Sanbase.Repo]
 
+config :sanbase, Sanbase, url: {:system, "SANBASE_URL", "https://sanbase-low-stage.santiment.net"}
```
tspenov: can't we use `website_url` which is further down in config

lib/sanbase/discourse/api.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Sanbase.Discourse.Api do
+  require Logger
+  require Mockery.Macro
+
+  alias Sanbase.Discourse.Config
+
+  def publish(title, text, url \\ Config.discourse_url()) do
+    handle_request(
```
tspenov: Do you think it is more readable like this:
```elixir
url
|> publish_topic(title, text)
|> handle_request()
```

lib/sanbase/voting/post.ex
```diff
@@ -52,7 +53,7 @@ defmodule Sanbase.Voting.Post do
 
   def update_changeset(%Post{} = post, attrs) do
     post
-    |> cast(attrs, [:title, :short_desc, :link, :text])
+    |> cast(attrs, [:title, :short_desc, :link, :text, :discourse_topic_url])
```
tspenov: Do we need to cast `:discourse_topic_url` in create and update changesets ?

IvanIvanoff: Not really.

PR #511, opened 2018-06-26
Recalculate intervals based on first_datetime

LucoLe:
Recalculates the intervals based on the first_datetime for each
measurement.

#### Summary

[//]: # (#### Related PRs)

#### Screenshots

image
image
image
image


lib/sanbase_web/graphql/helpers/utils.ex
```diff
@@ -0,0 +1,19 @@
+defmodule SanbaseWeb.Graphql.Helpers.Utils do
```
IvanIvanoff: no need for a separate module, we already have resolver helpers

LucoLe: Well, the other module is called cache and most of the functions are cache related so I didn't want to mess up with that. I'd like to keep the new module.

IvanIvanoff: I'm referring to this module: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/resolver_helpers.ex

Ideally it can be moved to another directory

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -121,6 +129,81 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
...
+        error_msg = "Can't fetch daily active addresses for #{ticker}"
+        Logger.warn(error_msg <> "Reason: #{inspect(error)}")
+        {:error, error_msg}
+    end
+  end
+
+  @doc ~S"""
+    Return the number of daily active addresses for a given slug
+  """
+  def daily_active_addresses(
```
tspenov: There are 4 `daily_active_addresses` funcs which are almost similar. Can we extract the common parts ?

IvanIvanoff: @LucoLe they can be reduced to 1, that accepts slug and interval. The others can transform their arguments to match this one.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -131,7 +131,7 @@ defmodule SanbaseWeb.Graphql.Schema do
       arg(:ticker, non_null(:string))
       arg(:from, non_null(:datetime))
       arg(:to, :datetime, default_value: DateTime.utc_now())
-      arg(:interval, :string, default_value: "1h")
+      arg(:interval, :string, default_value: "")
```
IvanIvanoff: For consistency, if value is missing use `nil`. Otherwise you'll have wildly different default values for strings, integers, floats, decimals, etc.

lib/sanbase_web/graphql/complexity/price_complexity.ex
```diff
@@ -36,6 +36,7 @@ defmodule SanbaseWeb.Graphql.Complexity.PriceComplexity do
   def history_price(%{from: from, to: to, interval: interval}, child_complexity, _) do
     from_unix = DateTime.to_unix(from, :second)
     to_unix = DateTime.to_unix(to, :second)
+    interval = if interval == "", do: "1h", else: interval
```
IvanIvanoff: I thing that this will yield different result than the real one. If the interval is missing it most probably will be different than "1h"

lib/sanbase_web/graphql/schema.ex
LucoLe: If I set it to nil here interval is not passed at all. Try it with the current tests.

lib/sanbase_web/graphql/complexity/price_complexity.ex
LucoLe: yes, but I couldn't find a workaround. Any suggestions?

LucoLe: but unfortunately, this is not working so we need to figure out a way to send the interval to this method

lib/sanbase_web/graphql/helpers/utils.ex
```diff
@@ -48,4 +66,18 @@ defmodule SanbaseWeb.Graphql.Helpers.Utils do
       String.replace(acc, "%{#{key}}", to_string(inspect(value)))
     end)
   end
+
+  defp compound_duration_to_seconds(interval) do
```
IvanIvanoff: You can put that into datetime utils. Seems to me that we have places where we can reuse it

tspenov:
@LucoLe I think this is still valid https://github.com/santiment/sanbase2/pull/511#pullrequestreview-132317266 ?

LucoLe:
@tspenov I think @IvanIvanoff is removing the redundant code in this PR https://github.com/santiment/sanbase2/pull/518

PR #520, opened 2018-06-28
Implement apikey

IvanIvanoff:
#### Summary

That's how I would approach the apikey implementation given that we want to implement authentication and not authorization at this moment. Check the Hmac module doc for more info

Quick guide to how it works:
0. Have a secret key, that will be a k8s/tf secret.
1. Generate an user token - random string and save it plain text in the db
2. Generate the apikey as follows - `apikey = concat(user_token, "_" , hmac(secret_key, user_token))`

[//]: # (#### Related PRs)

#### Additional deploy notes
The https://github.com/santiment/devops/pull/281 PR that adds the secret key should be merged first

[//]: # (#### Screenshots)


valo:
Can you point to an explanation somewhere about the methodology for generating API keys? I want to look into it a bit more and hopefully look at some examples

IvanIvanoff:
I cannot point you to a specific link. My main idea came from the desire to be able to show the users the apikey in plaintext but not store them in plaintext and  [this](https://stackoverflow.com/a/37953361) stackoverflow topic
[Here](https://tools.ietf.org/html/rfc2104) is some more technical details.
From the comments in the first link you read `HAMC is a one-way encryption commonly used in these types of scenarios. Consider it a type of encryption except that it's one way so you cannot decrypt the data.`

Since hmac is implemented in `:crypto` (it uses openssl) it was fairly easy and quick to test it.

valo:
Makes sense. This way if the DB leaks the secret key is in another place, so the API keys can't be used and vice versa. Also it supports having multiple keys per user, so I think it is good to go.

lib/sanbase/auth/apikey/apikey.ex
```diff
@@ -0,0 +1,96 @@
...
+  @doc ~s"""
+  Returns the User struct connected to the given apikey.
+  Split the apikey by "_" and use the first part as a user id. Look for all
+  user tokens for that ID and check if the apikey is generated from any of the
+  tokens
+  """
+  @spec apikey_to_user(String.t()) :: {:ok, %User{}} | {:error, String.t()}
+  def apikey_to_user(apikey) do
+    with {:ok, {token, _rest}} <- Hmac.split_apikey(apikey),
+         true <- UserApikeyToken.has_token?(token),
```
tspenov: I think on the next row `Hmac.apikey_valid?` is checking for `UserApikeyToken.has_token?` ?

lib/sanbase_web/graphql/resolvers/apikey_resolver.ex
```diff
@@ -0,0 +1,47 @@
...
+  Generates an apikey for the given user and returns the `user` struct.
+  To fetch all apikeys use the `apikeys` field of the `user` GQL type.
+  """
+  def generate_apikey(_root, _args, %{
+        context: %{auth: %{auth_method: :user_token, current_user: user}}
+      }) do
+    with {:ok, _apikey} <- Apikey.generate_apikey(user) do
+      {:ok, user}
+    else
+      error -> {:error, "Failed to generate apikey. Inspecting error: #{inspect(error)}"}
```
tspenov: Should we show the inspect as an API error ? Probably just log the inspect and show only the message ?

lib/sanbase/auth/apikey/apikey.ex
IvanIvanoff: Also it helped me realise that I must call `apikey_valid?` when revoking apikey, otherwise user can revoke malformed apikey

PR #536, opened 2018-07-04
Make history_price and github_activity work with slug

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/github_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/github_resolver.ex
@@ -4,6 +4,14 @@ defmodule SanbaseWeb.Graphql.Resolvers.GithubResolver do
   alias SanbaseWeb.Graphql.Helpers.Utils
   alias Sanbase.Github.Store
 
+  def activity(root, %{slug: slug} = args, resolution) do
+    # Temporary solution while all frontend queries migrate to using slug. After that
+    # only the slug query will remain
+    ticker = Utils.ticker_by_slug(slug)
+    args = args |> Map.delete(:slug) |> Map.put(:ticker, ticker)
+    activity(root, args, resolution)
+  end
+
   def activity(
         _root,
         %{ticker: ticker, from: from, to: to, interval: interval, transform: "None"},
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/resolvers/github_resolver.ex
```diff
@@ -4,6 +4,14 @@ defmodule SanbaseWeb.Graphql.Resolvers.GithubResolver do
   alias SanbaseWeb.Graphql.Helpers.Utils
   alias Sanbase.Github.Store
 
+  def activity(root, %{slug: slug} = args, resolution) do
+    # Temporary solution while all frontend queries migrate to using slug. After that
+    # only the slug query will remain
+    ticker = Utils.slug_by_ticker(slug)
```
tspenov: I think here should be the opposite  - `ticker_by_slug` ?

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -4,11 +4,20 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
...
 
   @doc """
-  Returns a list of price points for the given ticker. Optimizes the number of queries
+  Returns a list of price points for the given ticker or slug. Optimizes the number of queries
   to the DB by inspecting the requested fields.
   """
+  def history_price(root, %{slug: slug} = args, resolution) do
+    # Temporary solution while all frontend queries migrate to using slug. After that
+    # only the slug query will remain
+    ticker = Utils.slug_by_ticker(slug)
```
tspenov: I think here is the same as :arrow_up:

lib/sanbase_web/graphql/schema.ex
```diff
@@ -129,7 +129,8 @@ defmodule SanbaseWeb.Graphql.Schema do
 
     @desc "Fetch price history for a given ticker and time interval."
```
tspenov: Should't we change the `@desc` to use `slug` instead of ticker ?

tspenov:
Can we change also the example here ?: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/views/api_examples_view.ex#L98 and

PR #551, opened 2018-07-10
Fix: insight draft accessible to everyone

DmitriVanGuard:
#### Summary
Only author of the draft can access it _(before insight was published)_


IvanIvanoff:
It's still possible to fetch the draft of a different user. I think it'll be better to implement this in the backend as otherwise it can happen that user B fetches the draft of user A and publishes it as if it was written by him/herself.

DmitriVanGuard:
@IvanIvanoff Very good observation! 👍

PR #552, opened 2018-07-10
Feature: "New Insight" step back button

DmitriVanGuard:
#### Summary
While creating new insight the user can go back to the previous tab.
For example:
- From "Short Description" to "Your Insight's Text"
- From "Confirm" to "Short description"

#### Screenshots
image


Partysun:
@DmitriVanGuard Can you make this button as a text link? Without any background and borders.
It will look better. What do you think?

DmitriVanGuard:
@Partysun Yes! I agree with you! Will push new commit in a few minutes

PR #555, opened 2018-07-11
Using <Link/> instead of <div/>

cherurg:
#### Summary
The commit replaces divs to links on https://sanbase-low.santiment.net/projects
Now it's possible to use right click to open the context menu and then open the project in a new tab, for example. 

[//]: # (#### Screenshots)
These elements are links now.
screen shot 2018-07-11 at 15 20 12

### DIFF
--- a/app/src/pages/Projects/ProjectsTable.css
+++ b/app/src/pages/Projects/ProjectsTable.css
@@ -93,7 +93,7 @@
   white-space: normal;
 }
 
-.ReactTable .rt-td > div {
+.ReactTable .rt-td > a,div {
   font-size: 18px;
   line-height: 1.2;
 }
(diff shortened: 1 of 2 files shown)


cherurg:
@Partysun ah, sorry, didn't check it with stylelint.
Do you have plans to integrate the linters as pre-commit hooks?

Partysun:
Yes. it is what I am thinking before reading your comment)
I will change to prettier and add pre-commit for pp. But after some changes in the deploy process only.

cherurg:
@Partysun nice. I had a question about your opinion on prettier but now I know it. By the way, recently I've created a tool that helps me to add prettier to all my projects. You may find it interesting https://github.com/recogizer/prettierify

PR #557, opened 2018-07-12
Clickhouse integration

IvanIvanoff:
#### Summary

Update1: Still WIP as Clickhouse needs to be deployed to prod.

Update2: CH deployed to prod. Now I need to rewrite some tests and remove unused code.

Update3: As there is an Etherscan scraper, all related code to `Etherscan.Store` will be removed only after this new code is deployed and working.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -0,0 +1,288 @@
...
+  """
+  @spec eth_spent_by_projects(list(), %DateTime{}, %DateTime{}) ::
+          {:ok, number()} | {:ok, nil} | {:error, String.t()}
+  def eth_spent_by_projects([], _from_datetime, _to_datetime), do: {:ok, nil}
+
+  def eth_spent_by_projects(projects, from_datetime, to_datetime) do
+    total_eth_spent =
+      projects
+      |> Enum.map(&Task.async(fn -> eth_spent(&1, from_datetime, to_datetime) end))
+      |> Stream.map(&Task.await(&1, 25_000))
```
tspenov: I think this should be a constant/module attribute ?

lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -0,0 +1,288 @@
...
+    {:ok, eth_addresses} = Project.eth_addresses(projects)
+
+    total_eth_spent_over_time =
+      eth_addresses
+      |> Enum.map(
+        &Task.async(fn ->
+          eth_spent_over_time(&1, from_datetime, to_datetime, interval)
+        end)
+      )
+      |> Stream.map(&Task.await(&1, 25_000))
```
tspenov: also here constant ?

config/test.exs
```diff
@@ -15,13 +15,22 @@ config :logger, level: :warn
...
 
-# Configure your database
+# Configure postgres database
 config :sanbase, Sanbase.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   database: "sanbase_test",
-  # closest to real case as there are 3 pods with 10 connections each
   pool_size: 30
 
+# Configure ClickHouseRepo for testing. It will be tested by using Postgres
```
tspenov: Do we use this one since we're mocking in tests ?

IvanIvanoff: I went into the direction of mocks but I do not really like it. 
A better approach to the issue is to make tests that are excluded by default and will run only in CI (or  locally if you pass `--include clickhouse`) where the jenkins will spin up a clickhouse instance.

For now I'll remove this comment

Jenkinsfile
```diff
@@ -14,7 +14,14 @@ podTemplate(label: 'sanbase-builder', containers: [
         sh "docker run --rm --name test_postgres_${scmVars.GIT_COMMIT} -d postgres:9.6-alpine"
         sh "docker run --rm --name test_influxdb_${scmVars.GIT_COMMIT} -d influxdb:1.4-alpine"
         try {
-          sh "docker run --rm --link test_postgres_${scmVars.GIT_COMMIT}:test_db --link test_influxdb_${scmVars.GIT_COMMIT}:test_influxdb --env DATABASE_URL=postgres://postgres:password@test_db:5432/postgres --env INFLUXDB_HOST=test_influxdb --env ETHERBI_INFLUXDB_HOST=test_influxdb -t sanbase-test:${scmVars.GIT_COMMIT}"
+          sh "docker run --rm \
+            --link test_postgres_${scmVars.GIT_COMMIT}:test_db \
+            --link test_influxdb_${scmVars.GIT_COMMIT}:test_influxdb \
+            --env DATABASE_URL=postgres://postgres:password@test_db:5432/postgres \
+            --env CLICKHOUSE_DATABASE_URL=postgres://postgres:password@test_db:5432/postgres \
```
IvanIvanoff: can be removed

PR #560, opened 2018-07-13
Restrict historical data access

tspenov:
Restrict historical data access for anon and not staked users to interval `[now() - 3months,  now() - 1day]`

card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-1941

*all*:
image

*3 months:*
image


tspenov:
We should test on staging how the frontend behaves.

lib/sanbase_web/graphql/middlewares/api_timeframe_restriction.ex
```diff
@@ -0,0 +1,71 @@
...
+            auth: %{auth_method: method, current_user: current_user}
+          },
+          arguments: %{from: _from, to: _to}
+        } = resolution,
+        _
+      )
+      when method in [:user_token, :apikey] do
+    if !has_enough_san_tokens?(current_user) do
+      resolution = update_in(resolution.arguments.to, &restrict_to(&1))
+      update_in(resolution.arguments.from, &restrict_from(&1))
```
IvanIvanoff: With (or similar) this one you'll be able to update the parameters in one pass and avoid unnecessary copying due to immutability:
`%Resolution{resolution | arguments: %{from: restrict_from(from), to: restrict_to(to)}}`

IvanIvanoff:
I like the approach with ex_machina. You can also check this I have opened a week ago: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-1918
ex_machina + some randomness can solve a lot of issues

lib/sanbase_web/templates/api_examples/examples.html.eex
```diff
@@ -11,8 +11,8 @@ You can:
 * Run sample queries on the left pane
 * Read the compiled documentation of our GraphQL schema on the right sidebar
 
-**Note:** Some of the queries require authorization and a stake of <%= @required_san_stake_realtime_api %>`SAN`
-to return realtime data. Otherwise they show data upto the day before.
+**Note:** Some of the queries require authorization and a stake of <%= @required_san_stake_full_access %>`SAN`
```
IvanIvanoff: There is a missing interval before `SAN`

tspenov: 👍  yes it should be better explained

tspenov:
@IvanIvanoff  updated the comments

IvanIvanoff:
We should think about the `all` button. In the ideal case it should have the same behaviour as `Sentiment` - on hover show a message that SAN staking is needed

tspenov:
@Partysun can you check  ⬆️

PR #585, opened 2018-07-27
Market segments queries return proper JSON object

LucoLe:
#### Summary

Market segments queries return a JSON object with name and count instead
of the string representation of the object.

The front end changes the list of categories based on the type of
projects that are selected and when the user changes the type of the
project the categories filter is cleared.

[//]: # (#### Related PRs)

#### Screenshots

image

### DIFF
--- a/app/src/pages/Projects/withProjectsData.js
+++ b/app/src/pages/Projects/withProjectsData.js
@@ -6,7 +6,9 @@ import {
   allProjectsGQL,
   allErc20ProjectsGQL,
   currenciesGQL,
-  allMarketSegmentsGQL
+  allMarketSegmentsGQL,
+  currenciesMarketSegmentsGQL,
+  erc20MarketSegmentsGQL
 } from './allProjectsGQL'
 
 const mapStateToProps = state => {
@@ -39,7 +41,7 @@ const mapDispatchToProps = dispatch => {
   }
 }
 
-const mapDataToProps = type => ({ Projects }) => {
+const mapProjectsToProps = type => ({ Projects }) => {
   const loading = Projects.loading
   const isError = !!Projects.error
   const errorMessage = Projects.error ? Projects.error.message : ''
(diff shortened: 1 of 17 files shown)


IvanIvanoff:
Needs elixir tests

lib/sanbase_web/graphql/schema.ex
```diff
@@ -74,11 +75,23 @@ defmodule SanbaseWeb.Graphql.Schema do
...
 
+    @desc "Fetch ERC20 projects' market segments."
+    field :erc20_market_segments, list_of(:market_segment) do
+      middleware(ProjectPermissions)
+      cache_resolve(&MarketSegmentResolver.erc20_market_segments/3)
+    end
+
+    @desc "Fetch currency projects' market segments."
+    field :currencies_market_segments, list_of(:market_segment) do
+      middleware(ProjectPermissions)
```
IvanIvanoff: You do not need this middleware in all market segment places.

lib/sanbase_web/graphql/resolvers/market_segment_resolver.ex
```diff
@@ -10,19 +10,55 @@ defmodule SanbaseWeb.Graphql.Resolvers.MarketSegmentResolver do
...
 
     {:ok, market_segments}
   end
 
-  defp to_snake_case(string) do
-    string
-    |> String.downcase()
-    |> String.replace(~r/'|`|"/, "")
-    |> String.replace(~r/[^A-za-z0-9]/, "_")
+  def erc20_market_segments(_parent, _args, _resolution) do
```
IvanIvanoff: In both functions you can reuse all code except for the function that's used for filtering. You can make a new function that accepts one additional argument `filter_fun`

Partysun:
Moved in design sprint.

app/src/pages/Projects/withProjectsData.js
```diff
@@ -39,7 +41,7 @@ const mapDispatchToProps = dispatch => {
   }
 }
 
-const mapDataToProps = type => ({ Projects }) => {
+const mapProjectsToProps = type => ({ Projects }) => {
```
Partysun: mapDataToProps is the convention of apollo community. @LucoLe 
What do you think?

LucoLe: Yeah I was wondering if I should change it but that means that `mapDataToProps` should return an object that has some methods for each type of prop that we call it or something like that. Let's have a quick call tomorrow and we'll figure it out.

PR #586, opened 2018-07-28
Feature/insights backtest price chart

DmitriVanGuard:
#### Summary
Posts now have a mini-chart of the corresponding token/project. 
The horizontal line (its color depends on the _price change_) shows a moment in time when the user has created the post.

#### Additional deploy notes
The _annotation_ and _price change_ are calculated from the first data on the post's creation date.

#### Screenshots
image
image


IvanIvanoff:
I haven't looked into the changes deeply, but I doubt that `created` time is what you need. More likely you need `updated` time that will correctly get the time the insight was published. Otherwise it can easily be exploited by making several insights and publish only the correct ones as `created` will not change.

DmitriVanGuard:
@Partysun 
image

DmitriVanGuard:
@Partysun I've made a search time comparison. Here are the results:

**Native findIndex search time**
image
In total it is => **147.685546875ms**

**Custom binarySearch search time**
image
In total it is => **11.094482421875ms**

As you can see, the native search is **10x slower** and it's only for 11 components on the page. Imagine what will be the results if the page will contain more than 20 BacktestChart components + update lifecycle.

PR #590, opened 2018-07-31
Integrate timescaledb

IvanIvanoff:
#### Summary
Note that in contrast of ClickHouse where we have tables `erc20_daily_active_addresses` and `eth_daily_active_addresses` here in TimescaleDB we have a single table.
The `contract_info_by_slug/1` function has a special case for `ethereum` that returns `ETH` as contract (it's used internally this way) 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/timescale_repo/migrations/20180802142827_add_transaction_volume.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Sanbase.TimescaleRepo.Migrations.AddTransactionVolume do
+  use Ecto.Migration
+
+  def up do
+    create table(:eth_transaction_volume, primary_key: false) do
+      add(:timestamp, :naive_datetime, primary_key: true)
+      add(:contract_address, :string, primary_key: true)
+      add(:transaction_volume, :float)
+    end
+  end
```
valo: These migrations are needed so that we can test using only a local PG database?

IvanIvanoff: Correct, it's not sanbase's job to create create the tables. I still think about the best way to approach this problem so these migrations are not run in prod.
- One possible way is to try the way I approach this issue in ClickHouse: https://github.com/santiment/sanbase2/pull/557/files#diff-6023be6004fce4718dad3dafb576d258R13

lib/mix/seeds/daily_active_addresses_seed.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Sanbase.Seeds.DailyActiveAddressesSeed do
+  import Sanbase.Seeds.Helpers
+
+  def populate() do
+    changesets = [
+      make_changeset("0x123123", random_date(), random_non_neg_integer()),
```
tspenov: since we're using only these 2 `0x123123`,  `0x543210` - we can also add them to seeds helpers as `contract_address1` and `contract_address2`

priv/timescale_repo/migrations/20180802142827_add_transaction_volume.exs
IvanIvanoff: I addressed the issue. Furthermore, I made the tests run only in the CI (or locally if you add `--include timescaledb`) so people do not need to have the timescaledb plugin to run the test suite.

lib/sanbase/utils/math.ex
```diff
@@ -0,0 +1,62 @@
...
+      -8
+      iex> Sanbase.Utils.Math.ipow(1231232,0)
+      1
+      iex> Sanbase.Utils.Math.ipow(2,500)
+      3273390607896141870013189696827599152216642046043064789483291368096133796404674554883270092325904157150886684127560071009217256545885393053328527589376
+      iex> Sanbase.Utils.Math.ipow(10,18)
+      1_000_000_000_000_000_000
+
+  """
+  def ipow(base, exp) when is_integer(base) and is_integer(exp) and exp >= 0 do
```
tspenov: can't we just use `:math.pow(base, exp) |> to_integer()`

mix.exs
```diff
@@ -96,19 +96,66 @@ defmodule Sanbase.Mixfile do
...
+      test: [
+        "load_dotenv",
+        "ecto.create -r Sanbase.Repo --quiet",
+        "ecto.load -r Sanbase.Repo",
+        "test"
+      ],
+
+      # Append `_all` so the Ecto commands apply to all repos.
+      # and run all tests
+      "ecto.setup_all": [
```
tspenov: Can we add a line in the README with the steps to setup and test timescale locally ?

IvanIvanoff: We could, be it's different for different operating systems. I could add some docs for Linux

lib/sanbase/utils/math.ex
IvanIvanoff: 1. This is now not used now, though I planned to use it where needed.
2. There is no `to_integer`. You need to do `round() |> trunc()` to achieve it

mix.exs
tspenov: you can just add a link for installing and setting locally timescale and add only `elixir` specific steps for setting timescale locally - like what mix tasks should be run (from the added new)

lib/sanbase/utils/math.ex
tspenov: there is `to_integer()` below - https://github.com/santiment/sanbase2/pull/590/files#diff-f75e4e85cadbe561d637928692ed10c4R44 which is exactly `round() |> trunc()`

lib/mix/seeds/daily_active_addresses_seed.ex
IvanIvanoff: Done

mix.exs
IvanIvanoff: Done

Dockerfile-test
```diff
@@ -17,4 +17,4 @@ RUN mix deps.compile
 
 COPY . /app
 
-CMD mix test
+CMD mix test_all --trace
```
tspenov: newline needed here

README.md
```diff
@@ -24,6 +25,12 @@ This setup is going to start 2 processes:
   * A Phoenix server, which is routing the traffic and responding to API requests
   * A node server, which is doing the server side rendering
 
+  Additional setup:
+
+  * To run features requiring TimescaleDB you have to manually install it.
+   Here the the instructions for [Mac](http://docs.timescale.com/v0.10/getting-started/installation/mac/installation-homebrew) and [Linux](http://docs.timescale.com/v0.10/getting-started/installation/linux/installation-apt-ubuntu). With TimescaleDB you need to the ecto commands with `_all` suffix - `mix ecto.setup_all`, `mix ecto.migrate_all`, etc.
```
tspenov: typo here: `With TimescaleDB you need to the ecto commands with ...`

tspenov:
there are some conflicts that needs to be resolved

PR #591, opened 2018-08-02
Add GQL API for social volume

LucoLe:
#### Summary

Adds a GQL endpoint for social volume.
Updates the apidocs to document the endpoint use.
Adds tests for the new functions in internal_services/tech_indicators.ex

[//]: # (#### Related PRs)

#### Screenshots

image

image

image

### DIFF
--- a/lib/sanbase/internal_services/tech_indicators.ex
+++ b/lib/sanbase/internal_services/tech_indicators.ex
@@ -175,14 +175,14 @@ defmodule Sanbase.InternalServices.TechIndicators do
   end
 
   def social_volume(
-        ticker,
+        project,
         datetime_from,
         datetime_to,
         interval,
         social_volume_type
       ) do
     social_volume_request(
-      ticker,
+      project,
       datetime_from,
       datetime_to,
       interval,
@@ -195,27 +195,31 @@ defmodule Sanbase.InternalServices.TechIndicators do
 
       {:ok, %HTTPoison.Response{status_code: status, body: body}} ->
         error_result(
-          "Error status #{status} fetching social volume for ticker #{ticker}: #{body}"
+          "Error status #{status} fetching social volume for project #{project}: #{body}"
         )
 
       {:error, %HTTPoison.Error{} = error} ->
         error_result(
-          "Cannot fetch social volume data for ticker #{ticker}: #{HTTPoison.Error.message(error)}"
+          "Cannot fetch social volume data for project #{project}: #{
+            HTTPoison.Error.message(error)
+          }"
         )
     end
   end
 
-  def social_volume_tickers() do
-    social_volume_tickers_request()
+  def social_volume_projects() do
+    social_volume_projects_request()
     |> case do
       {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
         Poison.decode(body)
 
       {:ok, %HTTPoison.Response{status_code: status, body: body}} ->
-        error_result("Error status #{status} fetching social volume tickers: #{body}")
+        error_result("Error status #{status} fetching social volume projects: #{body}")
 
       {:error, %HTTPoison.Error{} = error} ->
-        error_result("Cannot fetch social volume tickers data: #{HTTPoison.Error.message(error)}")
+        error_result(
+          "Cannot fetch social volume projects data: #{HTTPoison.Error.message(error)}"
+        )
     end
   end
 
@@ -434,7 +438,7 @@ defmodule Sanbase.InternalServices.TechIndicators do
   end
 
   defp social_volume_request(
-         ticker,
+         project,
          datetime_from,
          datetime_to,
          interval,
@@ -448,7 +452,7 @@ defmodule Sanbase.InternalServices.TechIndicators do
     options = [
       recv_timeout: @recv_timeout,
       params: [
-        {"ticker", ticker},
+        {"project", project},
         {"datetime_from", from_unix},
         {"datetime_to", to_unix},
         {"interval", interval}
@@ -474,8 +478,8 @@ defmodule Sanbase.InternalServices.TechIndicators do
     {:ok, result}
   end
 
-  defp social_volume_tickers_request() do
-    url = "#{tech_indicators_url()}/indicator/social_volume_tickers"
+  defp social_volume_projects_request() do
+    url = "#{tech_indicators_url()}/indicator/social_volume_projects"
 
     options = [recv_timeout: @recv_timeout]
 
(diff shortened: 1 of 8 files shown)


test/sanbase/internal_services/tech_indicators_test.exs
```diff
@@ -226,4 +226,101 @@ defmodule Sanbase.InternalServices.TechIndicatorsTest do
...
+                    datetime: DateTime.from_unix!(to)
+                  }
+                ]}
+    end
+
+    test "response: 404" do
+      from = 1_523_876_400
+      to = 1_523_880_000
+
+      mock(
```
LucoLe: Unfortunately, this generates an error that is logged when the tests are run. Do you think that is too messy or it's ok @valo @IvanIvanoff.

IvanIvanoff: Yes, it's an issue to leave error logs in the tests. For this case you should use `capture_log` from the `ExUnit.CaptureLog` module

test/sanbase/internal_services/tech_indicators_test.exs
```diff
@@ -226,4 +226,101 @@ defmodule Sanbase.InternalServices.TechIndicatorsTest do
...
+      result =
+        TechIndicators.social_volume(
+          "Foo",
+          DateTime.from_unix!(from),
+          DateTime.from_unix!(to),
+          "1h",
+          :telegram_discussion_overview
+        )
+
+      assert result == {:error, "[#{log_id}] Error executing query. See logs for details."}
```
IvanIvanoff: Mocking the `Ecto.UUID` complicates the code without any benefits. For here you should not mock and instead:
```elixir
assert {:error, err_msg} = result
assert err_msg =~ "Error executing query. See logs for details."
```

valo:
Now that I am looking into this I realize that we should be fetching these social volumes by slug, not by ticker. We are trying to deprecate the tickers as identifiers and use the slugs. This will make integration of the APIs in the google spreadsheet easier too.

LucoLe:
yeah, I was afraid you'd say that :) but this will require a change in tech indicators and I'd rather do that in a separate PR/task. Let's finalise this PR and I'll create a new task in Favro and take right after this one.

valo:
I think it will be better if we make this API public only after we use slugs. Otherwise we are releasing an API and then immediately introduce a breaking change. May be remove the official docs for now and add them later?

LucoLe:
ok let me see how much work it will be to make the changes in tech indicators

lib/sanbase_web/graphql/schema.ex
```diff
@@ -431,7 +431,7 @@ defmodule SanbaseWeb.Graphql.Schema do
         It is used to select the source of the mentions count.
     """
     field :social_volume, list_of(:social_volume) do
-      arg(:ticker, non_null(:string))
+      arg(:project, non_null(:string))
```
IvanIvanoff: let's not use `project` as it's reserved for the `project` GQL type. Something like `:ticker_slug` should be better here.

IvanIvanoff: The price measurement name is `SAN_santiment` not just `santiment`, hence the `ticker` in the `ticker_slug`

valo:
Why do we need to use `SAN_santiment` instead of `santiment`?

IvanIvanoff:
@valo when deciding how to name the new price measurements we aggreed on `ticker_slug` as many people recognize the projects mainly by tickers and by their coinmarketcap ids.

The main question is why not use just `slug` here. I guess it's momentum аfter the work on tech_indicators.
@LucoLe it's only needed to query `slug` here. The `ticker` we can get from the DB - check how it's made in all other queries.

valo:
Do we use ticker_slug construct for other metrics? I think for the others we use only slug. That’s why I am asking how is this one different

LucoLe:
We have a dictionary in tech_indicators that keeps the search strings. The keys in this dictionary need to be in the form `ticker_slug` because we use this key in Grafana to make a query for the price in influxDB we use either `ticker` or `ticker_slug` we can't use only `slug`. So this is the reason why we have the projects in this format. If you think that's too cumbersome we have only one option I can see to serve the prices from tech indicators to Grafana. This way we can change the keys to whatever we want.

IvanIvanoff:
@LucoLe Does grafana/tech_indicators make GQL query to this endpoint? If no - then better follow the convention here and not mimic the grafana/tech_indicators behaviour (which is such because it's a workaround)

lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -447,6 +447,10 @@ defmodule Sanbase.InternalServices.TechIndicators do
     from_unix = DateTime.to_unix(datetime_from)
     to_unix = DateTime.to_unix(datetime_to)
 
+    ticker_slug =
+      Repo.get_by(Project, coinmarketcap_id: slug)
+      |> Measurement.name_from()
```
IvanIvanoff: In that case it's an overhead to fetch the whole project just to use the ticker. I suggest using `ticker_by_slug` for this case. I would rather:
```
ticker = Project.ticker_by_slug(slug)
ticker_cmc_id = "#{ticker}_#{slug}"

File: lib/sanbase_web/graphql/schema.ex [hunk 5]
```diff
@@ -416,6 +416,43 @@ defmodule SanbaseWeb.Graphql.Schema do
       resolve(&TechIndicatorsResolver.emojis_sentiment/3)
     end
 
+    @desc ~s"""
+    Returns a list of mentions count for a given project and time interval.
+
+    Arguments description:
+      * slug - a string
```
IvanIvanoff: `a string uniquely identifying a project`

lib/sanbase_web/graphql/schema.ex
```diff
@@ -416,6 +416,43 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+      * interval - an integer followed by one of: `m`, `h`, `d`
```
IvanIvanoff: also supported `w` for weeks

lib/sanbase_web/graphql/schema.ex
```diff
@@ -416,6 +416,43 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+      * from - a string representation of datetime value, e.g. "2018-04-16T10:02:19Z"
+      * to - a string representation of datetime value, e.g. "2018-04-16T10:02:19Z"
```
IvanIvanoff: We can just say that datetimes are following the iso8601 standard, so it's more clear what inputs we accept

PR #594, opened 2018-08-07
Add ethaccount admin

tspenov:
#### Summary
Add eth accounts in admin and link them to the users for better debugging

card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-2063

#### Screenshots
image

### DIFF
--- a/lib/sanbase_web/admin/auth/user.ex
+++ b/lib/sanbase_web/admin/auth/user.ex
@@ -1,7 +1,6 @@
 defmodule Sanbase.ExAdmin.Auth.User do
   use ExAdmin.Register
 
-  import Ecto.Query, warn: false
   alias Sanbase.Auth.EthAccount
 
   register_resource Sanbase.Auth.User do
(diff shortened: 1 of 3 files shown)


lib/sanbase_web/admin/auth/user.ex
```diff
@@ -1,7 +1,26 @@
...
   register_resource Sanbase.Auth.User do
+    show user do
+      attributes_table
+
+      panel "Eth accounts" do
+        table_for user.eth_accounts do
+          column(:id, link: true)
+          column(:address)
+
+          unless Mix.env() == :dev do
```
IvanIvanoff: This will fail in prod. In OTP Releases there is no `Mix`.
You can fetch an ENV var set at compile time: https://github.com/santiment/sanbase2/blob/master/config/config.exs#L11
or you can make it with a module attribute (they are evaluated compile time): https://github.com/santiment/sanbase2/blob/master/lib/sanbase/mandrill_api.ex#L6

Also why in `dev` we want a different behaviour?

PR #595, opened 2018-08-07
Backend changes - migration to app.santiment.net

tspenov:
Backend changes for migration to app.santiment.net
ENV vars pr in devops: https://github.com/santiment/devops/pull/323


lib/sanbase_web/views/api_examples_view.ex
```diff
@@ -105,7 +105,7 @@ defmodule SanbaseWeb.ApiExamplesView do
   end
 
   defp explorer_url() do
-    SanbaseWeb.Endpoint.website_url() <> "/apiexplorer"
+    SanbaseWeb.Endpoint.backend_url() <> "/graphiql"
```
IvanIvanoff: why change `apiexplorer` to `graphiql` here?

tspenov: cause it won't go through the react world and be in iframe - we'll just show the plain explorer

PR #597, opened 2018-08-08
Add erc20_exchange_funds_flow GQL endpoint

LucoLe:
#### Summary

Adds a GQL endpoint for erc20_exchange_funds_flow.

Adds tests for the new endpoint.

Adds documentation for the new endpoint.

[//]: # (#### Related PRs)

#### Screenshots

image

image

### DIFF
--- a/lib/sanbase/internal_services/tech_indicators.ex
+++ b/lib/sanbase/internal_services/tech_indicators.ex
@@ -174,6 +174,29 @@ defmodule Sanbase.InternalServices.TechIndicators do
     end
   end
 
+  def erc20_exchange_funds_flow(
+        from_datetime,
+        to_datetime
+      ) do
+    erc20_exchange_funds_flow_request(
+      from_datetime,
+      to_datetime
+    )
+    |> case do
+      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
+        {:ok, result} = Poison.decode(body)
+        erc20_exchange_funds_flow_result(result)
+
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -333,6 +333,43 @@ defmodule SanbaseWeb.Graphql.Schema do
       cache_resolve(&EtherbiResolver.exchange_funds_flow/3)
     end
 
+    @desc ~s"""
```
IvanIvanoff: This description should list the arguments in camelCase as this is how they are used in the clients

lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -388,6 +411,36 @@ defmodule Sanbase.InternalServices.TechIndicators do
...
+    ]
+
+    http_client().get(url, [], options)
+  end
+
+  defp erc20_exchange_funds_flow_result(result) do
+    result =
+      result
+      |> Enum.map(fn item ->
+        for {key, val} <- item, into: %{}, do: {String.to_atom(key), val}
```
IvanIvanoff: Is `String.to_atom/1` needed here? I see that it's called on the result of calling an internal service, but still. Be noted that atoms are never garbage collected so having `String.to_atom` is potentially dangerous.

A safe alternative that should work (but needs to be tested) as the atoms are defined in the type/schema is `String.to_existing_atom/1`

PR #601, opened 2018-08-10
Fix fetching of price for projects with duplicated tickers that both have slug

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


IvanIvanoff:
I'm still not convinced that's a good change or not.
With it requiring price by ticker won't fail because of `Repo.one` though it sometimes can return wrong data.
cc @valo @tspenov @Partysun @LucoLe

valo:
Which APIs on the prod frontend are tickers at the moment?

IvanIvanoff:
This PR was made before Yura switched `historyPrice` to use `slug` instead of `ticker` and when I look now I think there are no other places where it is used. In that case I think I can move on to removing this functionality and the ticker API.

PR #608, opened 2018-08-14
Double timeout time

LucoLe:
#### Summary

It seems that this request takes much more time to resolve so doubling
the timeout time should fix the problem.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/internal_services/tech_indicators.ex
+++ b/lib/sanbase/internal_services/tech_indicators.ex
@@ -468,7 +468,7 @@ defmodule Sanbase.InternalServices.TechIndicators do
     url = "#{tech_indicators_url()}/indicator/erc20_tokens_exchange_flow"
 
     options = [
-      recv_timeout: @recv_timeout,
+      recv_timeout: 2 * @recv_timeout,
       params: [
         {"from_timestamp", from_unix},
         {"to_timestamp", to_unix}


lib/sanbase/internal_services/tech_indicators.ex
```diff
@@ -468,7 +468,7 @@ defmodule Sanbase.InternalServices.TechIndicators do
     url = "#{tech_indicators_url()}/indicator/erc20_tokens_exchange_flow"
 
     options = [
-      recv_timeout: @recv_timeout,
+      recv_timeout: 2 * @recv_timeout,
```
valo: So this request might take more than 15sec to complete?

LucoLe: in grafana it takes about 20+ sec so I guess that's the reason. Especially since it's working on STA but it's not working on PROD.

valo: We are going to cache this request, right?

LucoLe: yes

IvanIvanoff: I would also look into tech_indicators and try to improve the performance of the query itself.

PR #632, opened 2018-08-28
Jenkinsfile improvement

azhelev:
Randomize container names during test phase execution.


IvanIvanoff:
`CHANGE_ID` is `null` in the passed builds.

azhelev:
It will be populated when a PR is created from a branch, but yeah it will stay null otherwise
https://jenkins.internal.santiment.net/job/Santiment/job/sanbase2/view/change-requests/job/PR-632/1/console

tspenov:
some conflicts needs to be resolved. Otherwise if `CHANGE_ID` being sometimes `null` doesn't create a problem - lets merge it 👍

azhelev:
Merge conflict is resolved. CHANGE_ID being null does not cause issues. It should help when a branch is pushed and then immediately a PR is made. In this case Jenkins starts two builds simultaneously - build of the branch itself and a build of the pr merged with master. Both builds have the same git sha and same BUILD_ID. That's why i included CHANGE_ID as well.

PR #644, opened 2018-08-29
Set the last scraped day to 2nd of august of all projects that have data after 2nd of august

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/prices/store.ex
```diff
@@ -16,6 +16,9 @@ defmodule Sanbase.Prices.Store do
   alias Sanbase.Influxdb.Measurement
 
   @last_history_price_cmc_measurement "sanbase-internal-last-history-price-cmc"
+  def last_history_price_cmc_measurement() do
```
tspenov: Do we use this func somewhere ?

IvanIvanoff: Yes, in the migration itself we use it because you cannot access module attributes otherwise.
https://github.com/santiment/sanbase2/pull/644/files#diff-429e2aa5c6dbede984c0e584a9b74893R27

PR #651, opened 2018-08-30
Sanitize usernames

tspenov:
#### Summary
Sanitize usernames
- [ ] trim whitespace of usernames. If trimmed username is existing - update to `nil`
- [ ] update usernames containing non ascii chars to `nil`
- [ ] return proper message when username contains non-ascii


lib/sanbase/auth/user.ex
```diff
@@ -75,6 +75,28 @@ defmodule Sanbase.Auth.User do
...
+    put_change(changeset, :username, String.trim(username))
+  end
+
+  defp normalize_username(changeset, _), do: changeset
+
+  defp validate_username_change(_, username) do
+    if ascii_username?(username) do
+      []
+    else
+      [username: "Username contains non ascii characters"]
```
valo: May be change this to something more user friendly, like `Username can contain only latin letters and numbers`

priv/repo/migrations/20180831094245_sanitize_usernames.exs
```diff
@@ -0,0 +1,91 @@
...
+    Application.ensure_all_started(:tzdata)
+
+    work()
+  end
+
+  def work() do
+    Logger.debug("trim_whitespace")
+
+    all_users_non_nil_username()
+    |> Enum.reject(&(&1.username == nil))
```
valo: I think this `reject` is duplicated and not necessary here

PR #660, opened 2018-09-03
Mark exchange wallets

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/mark_exchanges.ex
```diff
@@ -0,0 +1,94 @@
...
+      },
+      ...
+    }
+  ]`
+
+  This module is tightly coupled with the format of the input.
+  The user of this module is tightly coupled with the output of the function.
+  """
+
+  use GenServer
```
tspenov: What is the purpose of the GenServer here ? I think you can just make a db query (on prod are 167 recs) and make the required transformation. Also if you want to cache why not use 
the ets with ConCache ?

IvanIvanoff: The GenServer solution is < 100 LOC, so reworking it this way won't yield simpler code.
As I want to make this a separate module in sanbase (and not sanbase web) I prefer not to use the cache, as it's graphql centric. I have given the thought in the past to spin off a second cache that will not be used for storing graphql responses, though.

tspenov: Ok, lets try it out

PR #633, opened 2018-08-28
Feature/trends

DmitriVanGuard:
#### Summary
Route for displaying trends.

[//]: # (#### Screenshots)

PR #661, opened 2018-09-04
Add topic_search_overview endpoint

LucoLe:
#### Summary

Changes `topic_search` implementation to be `topic_search_overview`
because it is not used at the moment and there is no need to deprecate
the method and introduce a new one.

#### Related PRs

Depends on: https://github.com/santiment/tech_indicators/pull/88

#### Screenshots

image

image

image

### DIFF
--- a/lib/sanbase_web/graphql/schema.ex
+++ b/lib/sanbase_web/graphql/schema.ex
@@ -487,6 +487,8 @@ defmodule SanbaseWeb.Graphql.Schema do
     end
 
     @desc ~s"""
+    This endpoint is deprecated and soon won't be available! Please use `topic_search_overview` instead.
+
     Returns lists with the mentions of the search phrase grouped by source. The results are in two formats - the messages themselves and the data for building graph representation of the result.
 
     Arguments description:
@@ -500,6 +502,7 @@ defmodule SanbaseWeb.Graphql.Schema do
       * to - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
     """
     field :topic_search, :topic_search do
+      deprecate("Use topic_search_overview")
       arg(:sources, non_null(list_of(:topic_search_sources)))
       arg(:search_text, non_null(:string))
       arg(:from, non_null(:datetime))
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -500,6 +502,7 @@ defmodule SanbaseWeb.Graphql.Schema do
       * to - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
     """
     field :topic_search, :topic_search do
+      deprecate("Use topic_search_overview")
```
IvanIvanoff: This is not used currently on sanbase. Why cannot we change the resolver function instead of making a new query? Or just rename this one?

LucoLe: I thought it was used in the frontend

LucoLe: So it turns out it's not used at the moment so I'll just change the existing endpoint.

IvanIvanoff: @LucoLe the usage of this has been merged after I wrote my comment - https://github.com/santiment/sanbase2/pull/633/files#diff-c3b1655e93fcf8cee2cda1b5e57b939dR4

lib/sanbase_web/graphql/schema.ex
```diff
@@ -500,7 +500,7 @@ defmodule SanbaseWeb.Graphql.Schema do
       * to - a string representation of datetime value according to the iso8601 standard, e.g. "2018-04-16T10:02:19Z"
     """
     field :topic_search, :topic_search do
-      arg(:sources, non_null(list_of(:topic_search_sources)))
+      arg(:source, non_null(:topic_search_sources))
       arg(:search_text, non_null(:string))
       arg(:from, non_null(:datetime))
       arg(:to, :datetime, default_value: DateTime.utc_now())
```
IvanIvanoff: Can you change `:to` to not have a default value. It looks useful, so we'll have [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency).
We have it for older queries, but not for newer.

tspenov:
@LucoLe I think the screenshots are outdated cause I don't see adding things to the docs ?

LucoLe:
The screens are not old there is nothing new to the docs only changes

PR #664, opened 2018-09-04
Create discord notification insight published

tspenov:
#### Summary
card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-2232
image

### DIFF
--- a/lib/sanbase/notifications/insight.ex
+++ b/lib/sanbase/notifications/insight.ex
@@ -7,7 +7,6 @@ defmodule Sanbase.Notifications.Insight do
 
   def publish_in_discord(post) do
     post
-    |> new_insight_discord_content()
     |> create_discord_payload()
     |> publish()
     |> case do
@@ -19,20 +18,18 @@ defmodule Sanbase.Notifications.Insight do
     end
   end
 
-  defp new_insight_discord_content(%Post{id: id, title: title} = _post) do
+  defp create_discord_payload(%Post{id: id, title: title} = _post) do
     link = posts_url(id)
 
-    ~s"""
+    content = ~s"""
     New insight published: #{title} [#{link}]
     """
+
+    Poison.encode!(%{content: content, username: insights_discord_publish_user()})
   end
 
   defp publish(payload) do
     http_client().post(discord_webhook_url(), payload, [{"Content-Type", "application/json"}])
-  end
-
-  defp create_discord_payload(content) do
-    Poison.encode!(%{content: content, username: insights_discord_publish_user()})
   end
 
   defp discord_webhook_url do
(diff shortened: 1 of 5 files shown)


lib/sanbase/notifications/insight.ex
```diff
@@ -0,0 +1,26 @@
...
+
+    http_client().post(discord_webhook_url(), payload, [{"Content-Type", "application/json"}])
+  end
+
+  defp create_discord_payload(content) do
+    Poison.encode!(%{content: content, username: insights_discord_publish_user()})
+  end
+
+  defp discord_webhook_url do
+    Config.module_get(PriceVolumeDiff, :webhook_url)
```
IvanIvanoff: Let's have a config for `Sanbase.Notifications` and there we'll have `discord_webhook_url`.

lib/sanbase_web/graphql/resolvers/post_resolver.ex
```diff
@@ -196,6 +197,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.PostResolver do
           {:ok, post} ->
             {:ok, post} = create_discourse_topic(post)
 
+            publish_insight_in_discord(post)
```
IvanIvanoff: Here we'll need a slight rework. Instead of throwing, this should return tuples and here check it with `with`

lib/sanbase_web/graphql/resolvers/post_resolver.ex
```diff
@@ -257,6 +260,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.PostResolver do
...
+  end
+
+  defp new_insight_discord_content(%Post{id: id, title: title} = post) do
+    link = posts_url(id)
+
+    content = ~s"""
+    New insight published: #{title} [#{link}]
+    """
+  end
+
```
IvanIvanoff: Let's move this logic out of the resolver. Best will be to have a modules similar to the ones we use for Discourse.

lib/sanbase/notifications/insight.ex
tspenov: changed the webhook - cause we need another channel

lib/sanbase_web/graphql/resolvers/post_resolver.ex
tspenov: I think we should only log when error. The notification should not be tied with the publishing of insight

lib/sanbase_web/graphql/resolvers/post_resolver.ex
tspenov: moved away from resolver - used the existing module

lib/sanbase_web/graphql/resolvers/post_resolver.ex
```diff
@@ -196,6 +197,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.PostResolver do
           {:ok, post} ->
             {:ok, post} = create_discourse_topic(post)
 
+            mockable(Notifications.Insight).publish_in_discord(post)
```
IvanIvanoff: can we make it the same as in other places where this is used?

PR #671, opened 2018-09-07
Improve calculations of ETH spent metrics

IvanIvanoff:
#### Summary
Still in work.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/utils/application_utils.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Sanbase.ApplicationUtils do
+  @doc ~s"""
+  Start a worker/supervisor only in particular environment(s).
+  Example: Not startuing `MySupervisor` in tests can now be done by replacing
+  `{MySupervisor, []}` in the supervisor children by
+  `run_in({MySupervisor, []}, [:dev, :prod])`
```
tspenov: `start_in` instead of `run_in` ?

lib/sanbase/utils/application_utils.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  INPORTANT NOTE: If you use it, you must use `normalize_children` on the children list.
+  """
+  defmacro start_in(expr, environments) do
```
tspenov: why it should be macro ? Could it be a regular public function ?

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -114,6 +112,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
     {:ok, projects}
   end
 
+  # Helper functions
```
tspenov: Why is the comment moved here ? I think right below is a public resolver function.

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
```diff
@@ -138,22 +141,40 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectTransactionsResolver do
 
         {:ok, []}
     end
+  rescue
```
tspenov: Is it necessary to use rescue. Isn't connection errors catched by clickhousex or clickhouse_ecto ?

lib/sanbase/utils/application_utils.ex
IvanIvanoff: it does not to be a macro

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
IvanIvanoff: Will rearrange them

lib/sanbase/utils/application_utils.ex
tspenov: ok, so then can we simplify it by using a function

lib/sanbase/utils/application_utils.ex
```diff
@@ -0,0 +1,27 @@
...
+  @doc ~s"""
+  Start a worker/supervisor only in particular environment(s).
+  Example: Not startuing `MySupervisor` in tests can now be done by replacing
+  `{MySupervisor, []}` in the supervisor children by
+  `start_in({MySupervisor, []}, [:dev, :prod])`
+
+  INPORTANT NOTE: If you use it, you must use `normalize_children` on the children list.
+  """
+  def start_in(expr, environments) do
+    require Sanbase.Utils.Config, as: Config
```
tspenov: Can we move this require out of the function ?

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
```diff
@@ -51,127 +47,85 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectTransactionsResolver do
...
   """
   def eth_spent_by_erc20_projects(_, %{from: from, to: to}, _resolution) do
-    with projects when is_list(projects) <- Project.erc20_projects(),
-         {:ok, total_eth_spent} <-
-           Clickhouse.EthTransfers.eth_spent_by_projects(projects, from, to) do
+    with projects when is_list(projects) <- Project.erc20_projects() do
+      total_eth_spent =
+        projects
+        |> Enum.map(&Task.async(fn -> calculate_eth_spent_cached(&1, from, to).() end))
+        |> Enum.map(&Task.await/1)
```
tspenov: we could probably extract this into some parallel map func into utils that takes collection and function ?

PR #681, opened 2018-09-12
Rewrite ETH spent to remove duplicates

IvanIvanoff:
#### Summary
With the duplicates on staging it now returns the same result as on prod (where there are no known duplicates)

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)
(Change executed agains stage)
image

(Prod)
image

(Stage - wrong result, same data as first screenshot)
image

### DIFF
--- a/lib/sanbase/clickhouse/eth_transfers.ex
+++ b/lib/sanbase/clickhouse/eth_transfers.ex
@@ -239,7 +239,6 @@ defmodule Sanbase.Clickhouse.EthTransfers do
             GROUP BY from, type, to, dt, transactionHash
           )
         GROUP BY time
-        ORDER BY time
       )
     GROUP BY time
     ORDER BY time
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -201,30 +186,61 @@ defmodule Sanbase.Clickhouse.EthTransfers do
...
+            SELECT any(value) as value, dt
+            FROM #{@table}
+            PREWHERE from IN (?3) AND NOT to IN (?3)
+            AND dt >= toDateTime(?4)
+            AND dt <= toDateTime(?5)
+            AND type == 'call'
+            GROUP BY from, type, to, dt, transactionHash
+          )
+        GROUP BY time
+        ORDER BY time
```
MilanRusev: Do we need the `order by` here, since there's also `order by` in the end

IvanIvanoff: Not needed, indeed. `sum` does not care about the order.

PR #685, opened 2018-09-13
Refactor EthTransfers: Deduplicate transactions

IvanIvanoff:
Also add specs and docs, removes unused functions and API endpoints
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/erc20_transfers.ex
+++ b/lib/sanbase/clickhouse/erc20_transfers.ex
@@ -67,7 +67,7 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
     to_datetime_unix = DateTime.to_unix(to_datetime)
 
     query = """
-    SELECT from, to, dt, transactionHash, any(value) / ?1 as value
+    SELECT any(contract), from, to, dt, transactionHash, any(value) / ?1 as value
     FROM #{@table}
     PREWHERE contract = ?2
     AND dt >= toDateTime(?3)
--- a/lib/sanbase/clickhouse/erc20_transfers.ex
+++ b/lib/sanbase/clickhouse/erc20_transfers.ex
@@ -67,12 +67,12 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
     to_datetime_unix = DateTime.to_unix(to_datetime)
 
     query = """
-    SELECT any(contract), from, to, dt, transactionHash, any(value) / ?1 as value
+    SELECT contract, from, to, dt, transactionHash, any(value) / ?1 as value
     FROM #{@table}
     PREWHERE contract = ?2
     AND dt >= toDateTime(?3)
     AND dt <= toDateTime(?4)
-    GROUP BY from, to, dt, transactionHash
+    GROUP BY contract, from, to, dt, transactionHash, logIndex
     ORDER BY value desc
     LIMIT ?5
     """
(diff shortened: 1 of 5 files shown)


lib/sanbase/clickhouse/erc20_transfers.ex
```diff
@@ -67,7 +67,7 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
     to_datetime_unix = DateTime.to_unix(to_datetime)
 
     query = """
-    SELECT from, to, dt, transactionHash, any(value) / ?1 as value
+    SELECT any(contract), from, to, dt, transactionHash, any(value) / ?1 as value
```
MilanRusev: In tech indicators there was also an issue with eth_transfers and the `type` column: `sanbase/clickhouse/eth_transfers.ex`

MilanRusev: Also for `erc20_transfers` the primary key is `contract, from, to, dt, transactionHash, logIndex` (contract may not be needed but `logIndex` is needed; it may be a good idea to add the contract as well for performance reasons)

MilanRusev: And for eth_transfers is `from, type, to, dt, transactionHash`

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
```diff
@@ -45,38 +45,6 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectTransactionsResolver do
     end
   end
 
-  def last_wallet_transfers(
```
tspenov: why do you remove this - don't we need last/top transfers for a set of wallets ?

lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -138,52 +139,92 @@ defmodule Sanbase.Clickhouse.EthTransfers do
...
-    |> ClickhouseRepo.all_prewhere()
-    |> divide_by_eth_decimals()
+  defp wallet_transactions_query(wallets, from_datetime, to_datetime, limit, :in) do
+    from_datetime_unix = DateTime.to_unix(from_datetime)
+    to_datetime_unix = DateTime.to_unix(to_datetime)
+
+    query = """
+    SELECT from, type, to, dt, transactionHash, any(value) / #{@eth_decimals} as value
+    FROM #{@table}
+    PREWHERE NOT from IN (?1) AND NOT from IN (?1)
```
tspenov: is the upper correct ? `NOT from IN (?1) AND NOT from IN (?1)`

IvanIvanoff: will fix it.

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
IvanIvanoff: we do not use it really

lib/sanbase/clickhouse/erc20_transfers.ex
```diff
@@ -72,7 +72,7 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
     PREWHERE contract = ?2
     AND dt >= toDateTime(?3)
     AND dt <= toDateTime(?4)
-    GROUP BY from, to, dt, transactionHash
+    GROUP BY ontract, from, to, dt, transactionHash, logIndex
```
MilanRusev: there is a typo, should be `contract`

MilanRusev: Also since `contract` is in the group by it shouldn't be with aggregation in the select - `any(contract)`

PR #688, opened 2018-09-17
Migrate Tesla HTTP Client from 0.x to 1.x

IvanIvanoff:
#### Summary
https://github.com/teamon/tesla/wiki/0.x-to-1.0-Migration-Guide


config/config.exs
```diff
@@ -65,7 +65,7 @@ config :hammer,
 
 config :xain, :after_callback, {Phoenix.HTML, :raw}
 
-config :tesla, adapter: :hackney, recv_timeout: 30_000
```
tspenov: Is the `30_000` timeout working now ?

IvanIvanoff: It never worked before, for some reason. Now the timeouts are explicitly set in the get/post request

PR #689, opened 2018-09-17
Change Poison JSON library to Jason

IvanIvanoff:
#### Summary
Why? Because https://github.com/phoenixframework/phoenix/issues/2693#issuecomment-388881650

`Poison` is left as a decoder in a few places, because `Jason` does not support decoding into structs (the `:as` option)

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -8,6 +8,8 @@ use Mix.Config
 # General application configuration
 config :sanbase, ecto_repos: [Sanbase.Repo, Sanbase.TimescaleRepo]
 
+config :ecto, json_library: Jason
+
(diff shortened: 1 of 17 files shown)


tspenov:
we can also change ecto to use it - which will be the biggest benefit since db does a lot serialize/deserialize : `config :ecto, json_library: Jason
 `

IvanIvanoff:
Good catch, I thought the latest versions defaults to Jason

IvanIvanoff:
Absinthe already uses it, the change is in the `router.ex` file

PR #693, opened 2018-09-18
Add open/close high/low  price

tspenov:
#### Summary
card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-2337

It can be used only from `sanpy`, or we can switch frontend to it.

```graphql
{
  ohlc(slug:"santiment", from: "2018-09-18 06:20:00Z", to: "2018-09-18 08:20:00Z", interval: "10m") {
    datetime,
    openPriceUsd,
    lowPriceUsd,
    highPriceUsd,
    closePriceUsd
  }
}
```


valo:
I think the API should have a bit different shape. When we fetch OHLCV for a given time interval and given resolution we want to get a time series where for each interval of time we have the open, high, low, close and volume. So if the interval is 6h we need these values for each 6h interval in the given time range.

test/sanbase_web/graphql/prices_api_test.exs
```diff
@@ -330,6 +330,70 @@ defmodule SanbaseWeb.Graphql.PricesApiTest do
...
+            priceBtc
+          }
+          open_price_usd
+          high_price_usd
+          low_price_usd
+          close_price_usd
+          open_price_btc
+          high_price_btc
+          low_price_btc
+          close_price_btc
```
IvanIvanoff: let's make them consistent and use camelCase in documentation and tests.

tspenov:
@valo so we start from the `from` datetime, split it to for ex: 6h intervals and the last is incomplete ?

lib/sanbase_web/graphql/schema.ex
```diff
@@ -155,6 +155,17 @@ defmodule SanbaseWeb.Graphql.Schema do
...
 
+    @desc "Fetch open/high/low/close price values for a given slug and time interval."
+    field :ohlcv, :ohlcv do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, :datetime, default_value: DateTime.utc_now())
+      arg(:interval, :string, default_value: "")
+
+      complexity(&PriceComplexity.history_price/3)
+      cache_resolve(&PriceResolver.ohlcv/3)
```
IvanIvanoff: This should be `cache_resolve_dataloader` I think?

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -31,6 +31,12 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
     |> on_load(&total_market_history_price_on_load(&1, args))
   end
 
+  def ohlcv(_root, %{slug: @total_market} = args, %{context: %{loader: loader}}) do
+    loader
+    |> Dataloader.load(PriceStore, @total_market_measurement, args)
+    |> on_load(&total_market_history_price_on_load(&1, args))
```
IvanIvanoff: Why is this needed? This `total_market_history_price_on_load` won't return the correct data. Better return an `:error` tuple for the total market, indicating that there is no `open`, `close`, etc. or make a custom `ohlcv_total_market_on_load` function

tspenov: well, that was for compatability. But will remove if used only in sanpy

IvanIvanoff: But the response format is in the wrong format, so it will crash every time

tspenov: yeah, you're right

lib/sanbase/prices/store.ex
```diff
@@ -125,6 +131,19 @@ defmodule Sanbase.Prices.Store do
     AND time <= #{DateTime.to_unix(to, :nanoseconds)}/
   end
 
+  defp fetch_query_ohlcv(measurement, from, to, interval) do
```
IvanIvanoff: `fetch_ohlcv_query` will be the more consistent name

lib/sanbase_web/graphql/schema/price_types.ex
```diff
@@ -9,4 +9,12 @@ defmodule SanbaseWeb.Graphql.PriceTypes do
     field(:volume, :float)
     field(:ticker, :string)
   end
+
+  object :ohlcv do
```
valo: Looks much better now. Just one small remark: we don't return the volume now, so a more proper name will be `ohlc`.

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -58,15 +58,16 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
...
-      # interval_secs = interval_secs <> "s"
-      {:ok, prices} =
-        Sanbase.Prices.Store.fetch_ohlcv(ticker_cmc_id, args.from, args.to, args.interval)
-
+  def ohlcv(_root, %{slug: slug} = args, _context) do
+    with ticker when not is_nil(ticker) <- Project.ticker_by_slug(slug),
+         true <- Regex.match?(~r/^\d+[smhdw]{1}$/, args.interval),
+         {:ok, prices} <-
+           Sanbase.Prices.Store.fetch_ohlcv(
+             ticker <> "_" <> slug,
```
IvanIvanoff: You can introduce a new function in measurement.ex similar to `name_from` where you pass a slug and it returns you the measurement name.
Further, you can add 2 cases, so it returns `nil` if either slug or ticker is missing and just test `with measurement_name when not is_nil(measurement_name) <- Measurement.name_from(slug)`

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -58,6 +59,36 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
     end
   end
 
+  def ohlc(_root, %{slug: slug} = args, _context) do
+    with measurement when not is_nil(measurement) <-
+           Measurement.name_from_slug(slug) |> IO.inspect(),
```
IvanIvanoff: IO.inspect()

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -58,6 +59,36 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
...
+         true <- Regex.match?(~r/^\d+[smhdw]{1}$/, args.interval),
```
IvanIvanoff: Do we really need that? In case of wrong interval the query will fail. If you decide to leave it let's move it in a separate module, as it can be reused.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -155,6 +155,19 @@ defmodule SanbaseWeb.Graphql.Schema do
...
     end
 
+    @desc ~s"""
+    Fetch open, high, low close price values for a given slug and every time interval between from-to.
+    """
+
+    field :ohlc, list_of(:ohlc) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, :datetime, default_value: DateTime.utc_now())
```
IvanIvanoff: Lately we aim to not have such dynamic default values.

lib/sanbase_web/graphql/resolvers/price_resolver.ex
tspenov: yeah but it goes to the database - so it must be checked

lib/sanbase_web/graphql/schema.ex
```diff
@@ -155,6 +155,19 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+      arg(:interval, :string, default_value: "5m")
```
IvanIvanoff: "5m" interval is not really meaningful. We fetch the prices every 5 minutes, so all price points - open, close, high and low in 5m interval should be equal (or different due to the internals of influxdb). `6h` or something like this will be more meaningful default value. Or not have a default value at all.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -155,6 +155,19 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+
+      resolve(&PriceResolver.ohlc/3)
```
IvanIvanoff: should be cached

lib/sanbase_web/graphql/schema.ex
valo: Let's put the default to `1d` just in case. We don't want too small resolution by default.

lib/sanbase_web/graphql/schema.ex
valo: I am also wondering if we should put a complexity limitation on the query too...

lib/sanbase_web/graphql/resolvers/price_resolver.ex
tspenov: added in utils

lib/sanbase_web/graphql/schema.ex
tspenov: removed

lib/sanbase_web/graphql/schema.ex
tspenov: done

lib/sanbase_web/graphql/schema.ex
tspenov: cached and uses the same complexity calc as history price

IvanIvanoff: @valo note that every query has complexity, adding the `complexity` macro only overrides the default way of calculating it.

lib/sanbase/utils/date_time_utils.ex
```diff
@@ -87,4 +87,10 @@ defmodule Sanbase.DateTimeUtils do
     {:ok, datetime, _} = DateTime.from_iso8601(datetime_str)
     datetime
   end
+
+  def valid_interval_string?(interval_string) when not is_binary(interval_string), do: false
+
+  def valid_interval_string?(interval_string) when is_binary(interval_string) do
+    Regex.match?(~r/^\d+[smhdw]{1}$/, interval_string)
```
IvanIvanoff: let's do this:
```elixir
if regex do
  true
else
  {:error, "The provided string #{interval_string} is not a valid string interval"}
```

Otherwise when it does not match `true` in `with`, the error log will be something like: `Reason: false` which is not informative at all.

tspenov: looks good - will add it

PR #696, opened 2018-09-20
1863 change of email verification

LucoLe:
#### Summary

Sends a verification email when a user wants to change his email.

Once this PR is merged to STA the text in the email template in mandril should be updated to state that a verification email is sent. And the front end should be fixed so the email candidate shouldn't be put in the place of the current email and also the flash message should say that a verification email is sent rather than the email was updated successfully.

[//]: # (#### Related PRs)

#### Screenshots

image


config/test.exs
```diff
@@ -21,12 +21,14 @@ config :tesla, adapter: Tesla.Mock
 config :sanbase, Sanbase.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   database: "sanbase_test",
+  ownership_timeout: 3_600_000,
```
IvanIvanoff: Let's not do this. This is masking the problem

lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -57,40 +57,33 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
-
-      {:error, changeset} ->
-        {
-          :error,
-          message: "Cannot update current user's email to #{new_email}",
-          details: Utils.error_details(changeset)
-        }
+    with {:ok, user} <- User.set_email_candidate(user, email_candidate),
+         {:ok, user} <- User.update_email_token(user),
+         {:ok, _user} <- User.send_verification_email(user) do
```
IvanIvanoff: Above this `_user` is used, so you do not get warnings for unused variables. Here you better make it `user` as the previous `user` is used below.

lib/sanbase/auth/user.ex
```diff
@@ -158,14 +170,25 @@ defmodule Sanbase.Auth.User do
   def find_or_insert_by_email(email, username \\ nil) do
```
IvanIvanoff: This is now actually `find_or_insert_by_email_candidate`

LucoLe: not exactly its `find_by_email_or_insert_by_email_candidate` but I guess `find_or_insert_by_email` is better 😃

lib/sanbase_web/graphql/resolvers/account_resolver.ex
LucoLe: The service does not return a user rather a response which we don't use for the moment so I'll just rename the variable to `_response`.

lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -57,40 +57,33 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
       {:ok, %{success: true}}
     else
       _ -> {:error, message: "Can't login"}
     end
   end
 
-  def email_login_verify(%{token: token, email: email}, _resolution) do
-    with {:ok, user} <- User.find_or_insert_by_email(email),
+  def email_login_verify(%{token: token, email: email_candidate}, _resolution) do
```
tspenov: should this method be called only `email_verify` cause it is verifying on both email change and on login with email ?

LucoLe: you're right I was thinking about logging in through the email link sent but that's some mess in my head. Thanks for noticing.

tspenov:
@LucoLe @IvanIvanoff @valo  One general question: If we use only one template for both login and email verification are we gonna be able to rephrase it so it is clear to the use. The first case user is not logged it and by the link she logs in (and also we check the email). The second case he is already logged (metamask/email) and needs to verify ownership of email. Are we gonna be able to go with the same text for both cases ?
My opinion is that we have to use 2 templates and different graphql queries se we're not changing login with email flow and just add second flow for verification.

LucoLe:
I'd really like to keep it simple for the moment because if we have to use two separate emails this will need a lot more additions to the code and IMO the email that is sent in both cases just verifies that you're the owner of the email. So I was thinking about something like this:
```
You are requested to verify that you are the owner of this email in order to use it with your sanbase account. Click the button below to verify.

<button> Verify email </button>
```

IvanIvanoff:
Agree with going forward with one, general email

lib/sanbase_web/graphql/resolvers/account_resolver.ex
```diff
@@ -57,40 +57,33 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
-    |> Repo.update()
-    |> case do
-      {:ok, user} ->
-        {:ok, user}
-
-      {:error, changeset} ->
-        {
-          :error,
-          message: "Cannot update current user's email to #{new_email}",
-          details: Utils.error_details(changeset)
```
IvanIvanoff: Let's at least log somewhere this error, so we know why it happened.

LucoLe:
You're right that's not so hard to implement but IMO it's not necessary at the moment. Anyway I call for a vote and let's pick an option @valo @IvanIvanoff @tspenov 
I vote to use generic email for the moment.

IvanIvanoff:
Let's not override the old `send_login_email` function and use it here: https://github.com/santiment/sanbase2/pull/696/files#diff-2eb55e00e70623f880749741d2a5501aR60

The `send_verification_email` will be a new one, using a new template and leave the call to it form the `change_email` resolver.

Furthermore, now the `verification` template is called `login`, which will cause only confusion

valo:
I think we should have a separate email template for the verification. The login email doesn't really make sense when you verify your email.

lib/sanbase/auth/user.ex
```diff
@@ -158,14 +171,25 @@ defmodule Sanbase.Auth.User do
...
 
       user ->
         {:ok, user}
     end
   end
 
+  def find_by_email_candidate(email_candidate, email_token) do
+    case Repo.get_by(User, email_candidate: email_candidate, email_token: email_token) do
+      nil ->
+        {:error, "Can't find user"}
```
IvanIvanoff: `"Can'tfind user with email candidate #{email_candidate}"`

lib/sanbase/auth/user.ex
```diff
@@ -204,9 +238,15 @@ defmodule Sanbase.Auth.User do
...
-  def send_login_email(user) do
-    mandrill_api().send(@login_email_template, user.email, %{
-      LOGIN_LINK: SanbaseWeb.Endpoint.login_url(user.email_token, user.email)
+  def send_verification_email(user, :login) do
+    mandrill_api().send(@login_email_template, user.email_candidate, %{
+      LOGIN_LINK: SanbaseWeb.Endpoint.login_url(user.email_token, user.email_candidate)
+    })
+  end
+
+  def send_verification_email(user, :verify_email) do
```
IvanIvanoff: won't it be more clear to have `send_login_email` and `send_verification_email`?

LucoLe: I'll change it but I still insist that in both cases that is a verification email.

IvanIvanoff:
Let's thoroughly test on stage.

valo:
Just to make something clear here. When a user registers, we can think of the first email sent to login as verification email, but any subsequent login emails are exactly that: login emails.

In that sense from business perspective login and verification are 2 different emails and this is why it is good to distinguish between them in the code. They definitely look very similar right now and probably can be abstracted and DRYed very nicely, but the benefit of the abstraction will be minimal and the cost of changing the abstraction later will be higher.

As a simple example I think that it is very possible that we introduce a third email soon: registration email, which includes a login link with additional info how to use the website and is sent only on the first login.

Just my 2 cents :)

valo:
Very interesting case indeed. Almost the same code and email handling mechanism, but two different business cases. Need to remember this and include it in some tech talk some day :)

PR #704, opened 2018-09-25
Add backend pagination for all projects queries

IvanIvanoff:
#### Summary
WIP as it will break the sorting across multiple pages

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

image
image

### DIFF
--- a/lib/sanbase/model/project.ex
+++ b/lib/sanbase/model/project.ex
@@ -111,6 +111,9 @@ defmodule Sanbase.Model.Project do
     |> Repo.one()
   end
 
+  @doc ~s"""
+  Return all erc20 projects
+  """
   def erc20_projects() do
     query =
       from(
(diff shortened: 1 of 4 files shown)


lib/sanbase/model/project.ex
```diff
@@ -146,6 +176,31 @@ defmodule Sanbase.Model.Project do
...
+    query =
+      from(
+        p in Project,
+        join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
+        inner_join: infr in Infrastructure,
+        on: p.infrastructure_id == infr.id,
+        where:
+          not is_nil(p.coinmarketcap_id) and
+            (is_nil(p.main_contract_address) or infr.code != "ETH"),
+        order_by: latest_cmc.rank,
```
tspenov: Where do we plan to use this pagination ? If its from frontend - we need to be able to specify the order field ?
Also we probably need - page_count, first, last, previous and next pages .

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
```diff
@@ -26,18 +26,36 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectResolver do
   alias SanbaseWeb.Graphql.Helpers.Cache
   alias SanbaseWeb.Graphql.Resolvers.ProjectBalanceResolver
 
-  def all_projects(_parent, args, _resolution, only_project_transparency \\ nil) do
-    only_project_transparency =
-      case only_project_transparency do
-        nil -> Map.get(args, :only_project_transparency, false)
-        value -> value
-      end
+  def all_projects_project_transparency(_parent, _args, _resolution) do
```
tspenov: This will probably break the frontend . Lets notify @Partysun before it is merged

lib/sanbase/model/project.ex
```diff
@@ -146,6 +176,31 @@ defmodule Sanbase.Model.Project do
...
+        limit: ^page_size,
+        offset: ^((page - 1) * page_size),
+        preload: [latest_coinmarketcap_data: latest_cmc]
+      )
```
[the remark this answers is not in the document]
IvanIvanoff: Good idea, will change

lib/sanbase_web/graphql/resolvers/project/project_resolver.ex
IvanIvanoff: It's only used in project transparency

PR #707, opened 2018-09-28
Revert "1863 change of email verification"

LucoLe:
Reverts santiment/sanbase2#696


valo:
when we add this back we need to make sure the migration has the same ID so that it gets skipped

LucoLe:
I have a branch with these changes so I think it should be ok but will check again.

PR #714, opened 2018-10-01
Add logging hydra

tspenov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/oauth2/hydra.ex
+++ b/lib/sanbase/oauth2/hydra.ex
@@ -25,15 +25,15 @@ defmodule Sanbase.Oauth2.Hydra do
   end
 
   def manage_consent(consent, access_token, user, client_id) do
-    user_san_balance = san_balance(user)
+    user_san_balance = san_balance!(user)
 
     if has_enough_san_tokens?(
          user_san_balance,
-         required_san_tokens_by_client(client_id)
+         required_san_tokens_by_client(:clients_that_require_san_tokens, client_id)
        ) do
       accept_consent(consent, access_token, user)
     else
-      Logger.warn(
+      Logger.info(
         "#{user.email || user.username} doesn't have enough SAN tokens" <>
           inspect(user_san_balance)
       )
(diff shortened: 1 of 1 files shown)


lib/sanbase/oauth2/hydra.ex
```diff
@@ -25,13 +25,20 @@ defmodule Sanbase.Oauth2.Hydra do
...
+
+    if has_enough_san_tokens?(
+         user_san_balance,
+         required_san_tokens_by_client(client_id)
        ) do
-      reject_consent(consent, access_token, user)
-    else
       accept_consent(consent, access_token, user)
+    else
+      Logger.warn(
```
IvanIvanoff: I prefer to use error/warn for when our code does not work as expected. Here `info` should be enough

lib/sanbase/oauth2/hydra.ex
```diff
@@ -105,12 +112,21 @@ defmodule Sanbase.Oauth2.Hydra do
...
-  defp has_enough_san_tokens?(%User{} = user, san_tokens) when not is_nil(san_tokens) do
-    Decimal.cmp(User.san_balance!(user), Decimal.new(san_tokens)) != :lt
+  defp has_enough_san_tokens?(san_balance, required_san_tokens)
+       when not is_nil(required_san_tokens) do
+    Decimal.cmp(san_balance, Decimal.new(required_san_tokens)) != :lt
   end
 
   defp has_enough_san_tokens?(%User{} = _user, _), do: true
 
+  defp san_balance(user) do
```
IvanIvanoff: should be with `!` as it's calling such function

lib/sanbase/oauth2/hydra.ex
```diff
@@ -105,12 +112,21 @@ defmodule Sanbase.Oauth2.Hydra do
...
+    User.san_balance!(user)
+  end
+
+  defp required_san_tokens_by_client(client_id) do
+    json_config_value(:clients_that_require_san_tokens)[client_id]
```
IvanIvanoff: can we merge this with the function bellow and make it:
```elixir
config = Config.get(key) |> Jason.decode!()
config[client_id]
```

PR #720, opened 2018-10-04
Add email verification when a user changes email

LucoLe:
#### Summary

When a user tries to change his email a verification email is sent to
him.

Email and email_candidate are converted to lower case before they are
stored and a check is made for the email_candidate if there is a
registered user with this email.

[//]: # (#### Related PRs)

#### Screenshots

image


app/src/pages/Account/Account.js
```diff
@@ -14,7 +14,7 @@ import * as actions from '../../actions/types'
 import './Account.css'
 const validate = require('validate.js')
 
-export const TAKEN_MSG = 'has already been taken'
+export const TAKEN_MSG = "Can't change current user's email to"
```
IvanIvanoff: Why the more descriptive message that the email has already been taken is not used?

lib/sanbase/auth/user.ex
```diff
@@ -37,13 +38,17 @@ defmodule Sanbase.Auth.User do
...
     field(:username, :string)
     field(:salt, :string)
     field(:san_balance, :decimal)
     field(:san_balance_updated_at, Timex.Ecto.DateTime)
     field(:email_token, :string)
     field(:email_token_generated_at, Timex.Ecto.DateTime)
     field(:email_token_validated_at, Timex.Ecto.DateTime)
+    field(:email_candidate_token, :string)
+    field(:email_candidate_token_generated_at, Timex.Ecto.DateTime)
+    field(:email_candidate_token_validated_at, Timex.Ecto.DateTime)
```
IvanIvanoff: `Ecto.DateTime` is to be deprecated. I've tried in the past to remove current usage but other problems arose. We need to use the built-in types (and refactor the old code)

lib/sanbase/auth/user.ex
```diff
@@ -72,14 +77,21 @@ defmodule Sanbase.Auth.User do
...
+      :email_candidate_token_validated_at,
       :username,
       :salt,
       :test_san_balance,
       :privacy_policy_accepted,
       :marketing_accepted
     ])
     |> normalize_username(attrs)
+    |> email_to_lower_case(attrs[:email], :email)
+    |> email_to_lower_case(attrs[:email_candidate], :email_candidate)
```
IvanIvanoff: Use this function: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/model/project_eth_address.ex#L23
instead of defining your own

test/sanbase/auth/user_test.exs
```diff
@@ -196,6 +196,132 @@ defmodule Sanbase.Auth.UserTest do
...
+  test "update_email_from_email_candidate updates the email with the email_candidate" do
+    email_candidate = "test+foo@santiment.net"
+
+    {:ok, user} =
+      %User{
+        email: "test@example.com",
+        salt: User.generate_salt(),
+        privacy_policy_accepted: true
+      }
+      |> Repo.insert!()
```
tspenov: I think using some user factories will make tests less verbose: https://github.com/santiment/sanbase2/blob/master/test/support/sanbase_factory.ex

lib/sanbase/auth/user.ex
```diff
@@ -72,14 +77,21 @@ defmodule Sanbase.Auth.User do
...
+      :email_candidate_token_validated_at,
       :username,
       :salt,
       :test_san_balance,
       :privacy_policy_accepted,
       :marketing_accepted
     ])
     |> normalize_username(attrs)
+    |> update_change(:email, &(&1 && String.downcase(&1)))
+    |> update_change(:email_candidate, &(&1 && String.downcase(&1)))
```
valo: Let's include the email normalization also `strip`. It is quite common when writing on mobile device to get some spaces at the end of the email. May be we should have a separate functions `normalize_email` which does `downcase` and `strip`?

app/src/App.js
```diff
@@ -18,6 +18,7 @@ import Account from './pages/Account/Account'
 import PrivacyPolicyPage from './pages/PrivacyPolicyPage'
 import BuildChallenge from './pages/BuildChallenge'
 import EmailLoginVerification from './pages/EmailLoginVerification'
+import EmailChangeVerification from './pages/EmailChangeVerification'
```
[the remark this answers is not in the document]
LucoLe: I removed this from the code so I don't know how you got a reference to that line. Probably you had to refresh the page. Anyway, that's no longer in the code.

test/support/sanbase_factory.ex
```diff
@@ -6,7 +6,20 @@ defmodule Sanbase.Factory do
...
+      san_balance: nil,
+      san_balance_updated_at: nil,
+      email_token: nil,
+      email_token_generated_at: nil,
+      email_token_validated_at: nil,
+      email_candidate_token: nil,
+      email_candidate_token_generated_at: nil,
+      email_candidate_token_validated_at: nil,
+      consent_id: nil,
+      test_san_balance: nil
```
IvanIvanoff: In a struct if you do not define a field it will be automatically set to nil

LucoLe: Good point

PR #721, opened 2018-10-04
Enhance/scss vars and mixins

DmitriVanGuard:
#### Summary
This PR enables to themify any element with sass `themify` mixin.

#### How to use it
For now, there are 2 theme options: `normal (day)` and `night`theme. The corresponding colors are stored in sass map named `theme-colors`. In order to theme an element, it's necessary to include `themify` mixin inside CSS selector body and then write properties that are to be themed. Sass function named `themed` will return the value for the specific theme (that value is stored in `theme-colors`).
```scss
a {
  @include themify() {
    color: themed('primary-link');
  }
}
```
Will be compiled into:
```scss
a {
  color: #3d4450;
}

.night-mode a {
  color: #a0a0a0;
}
```


Partysun:
Looks interesting. And how to work with this pattern, when we need to make specific night-mode css changes for an element. Right now it's ok for general variables, as the primary color, etc..
@DmitriVanGuard

PR #740, opened 2018-10-15
Setup docker-compose for local development

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


Dockerfile-dev
```diff
@@ -0,0 +1,20 @@
+
+FROM elixir:1.7.3
+
+# Install debian packages
+RUN apt-get update
+RUN apt-get install --yes build-essential inotify-tools postgresql-client
+RUN echo "nameserver 8.8.8.8" >> /etc/resolv.conf
```
valo: Why is this needed?

IvanIvanoff: The `mix deps.get` couldn't resolve the hex links for some reason, though now I tested and it seems to work without this change

README.md
```diff
@@ -36,11 +36,23 @@ This setup is going to start 2 processes:
 If you have docker you can run the app simply by running:
 
 ```bash
-$ docker-compose up
+$ bin/start.sh
```
IvanIvanoff: @valo There is no such file, is it not pushed or is a typo?

docker-compose.yaml
```diff
@@ -1,39 +1,46 @@
...
+      dockerfile: Dockerfile-dev
+
+    volumes:
+      - ./:/app:delegated
+
+    hostname: sanbase_host
+
+    command: sh -c "elixir --sname sanbase --cookie sanbase -S mix phx.server"
+    expose:
+      - 3000
```
tspenov: I think 3000 is not needed now. We should run the frontend service there at some point though.

README.md
```diff
@@ -36,11 +36,23 @@ This setup is going to start 2 processes:
...
 ```
 
 This is going to run the app on port 4000, so [`localhost:4000`](http://localhost:4000) should be accessible.
 
+There will no data in the database, so run this command to fill some information in the database:
```
tspenov: type: `There will be no data`

docker-compose.yaml
```diff
@@ -1,39 +1,46 @@
...
+
+    depends_on:
+      - postgres
       - influxdb
-      - faktory
+
     environment:
-      PORT: 4000
-      DATABASE_URL: postgres://postgres:password@db:5432/sanbase-db
+      DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment
```
tspenov: I guess here should be `0.0.0.0` for host instead of `postgres` unless you have it in `/etc/hosts` ?

docker-compose.yaml
```diff
@@ -1,39 +1,46 @@
-version: '2'
+version: '3.4'
 services:
-  db:
-    image: postgres
+  postgres:
+    image: timescale/timescaledb:0.10.1-pg10
     ports:
-      - "5432"
+      - "5432:5432"
```
tspenov: `6432:5432` -> so we don't have to stop local postgres ?

tspenov:
@valo  @IvanIvanoff  works for me locally

valo:
I think are missing a script for running the tests. I'll add that

PR #743, opened 2018-10-19
Add combined vol mcap

tspenov:
#### Summary
API endpoint : combined volume and mcap for a list of slugs
card: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-2456

```graphql
{
  projectsGroupStats(
    slugs:["bitcoin", "ethereum", "santiment"],
    from:"2018-08-01 16:00:00Z",
    to:"2018-08-02 16:00:00Z"
  ) {
    volume,
    marketcap
  }
}
```

```graphql
{
  "data": {
    "projectsGroupStats": [
      {
        "marketcap": 98705005393020,
        "volume": 3598670966330
      }
    ]
  }
}
```

### DIFF
--- a/lib/sanbase/influxdb/measurement.ex
+++ b/lib/sanbase/influxdb/measurement.ex
@@ -68,6 +68,9 @@ defmodule Sanbase.Influxdb.Measurement do
     end
   end
 
+  @doc ~s"""
+    convert a list of slugs to comma separated measurements string 
+  """
   def measurement_str_from_slugs(slugs) do
     measurements_list =
       slugs
--- a/lib/sanbase/influxdb/measurement.ex
+++ b/lib/sanbase/influxdb/measurement.ex
@@ -68,6 +68,19 @@ defmodule Sanbase.Influxdb.Measurement do
     end
   end
 
+  def measurement_str_from_slugs(slugs) do
+    measurements_list =
+      slugs
+      |> Enum.map(&Measurement.name_from_slug/1)
+      |> Enum.reject(&is_nil/1)
+
+    if Enum.count(measurements_list) > 0 do
+      {:ok, measurements_list |> Enum.join(", ")}
+    else
+      nil
+    end
+  end
+
   # Private functions
 
   defp format_timestamp(%DateTime{} = datetime) do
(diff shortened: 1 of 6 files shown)


Partysun:
Maybe better to choose the simpler name for "combinedVolumeMcap", as stats(slugs, from, to).
@tspenov ?

tspenov:
@Partysun stats seems to me too generic. Maybe `prices_stats` although it does nothing with the prices. What @IvanIvanoff and @valo think ?

valo:
`stats` is too generic IMO. May be something like `projectsGroupStats`? I would like to avoid putting the exact stats in the name as we may add more than market cap and volume this this in the future

tspenov:
`projectsGroupStats` sounds good to me

Partysun:
@tspenov AssetsGroupStats. We are using asset instead of a project now.

tspenov:
@Partysun yes but everywhere in the API we're naming it `project`. It is mentioned at least 30 times in the API schema: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/schema.ex, whereas `assets` none.

valo:
There is an additional request for this API: compute a field, which gives the % of each project in the total market cap of the group. For example:

```js
{
  "data": {
    "projectsGroupStats": [
      {
        "marketcap": 98705005393020,
        "volume": 3598670966330,
        "marketcap_percent": [0.70, 0.20, 0.10]
      }
    ]
  }
}
```

The above example says that from the total `98705005393020` market cap, 70% is BTC, 20% is ETH and 10% is SAN.

Is it going to be possible to add this?

lib/sanbase/influxdb/measurement.ex
```diff
@@ -68,6 +68,22 @@ defmodule Sanbase.Influxdb.Measurement do
     end
   end
 
+  @doc ~s"""
+    convert a list of slugs to comma separated measurements string 
+  """
+  def measurement_str_from_slugs(slugs) do
+    measurements_list =
+      slugs
+      |> Enum.map(&Measurement.name_from_slug/1)
```
IvanIvanoff: This is N queries to the database. You can optimize it to use 1 query.

PR #744, opened 2018-10-19
Clickhouse GitHub activity

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


config/dev.exs
```diff
@@ -68,14 +68,14 @@ config :sanbase, Sanbase.TimescaleRepo,
...
   loggers: [Ecto.LogEntry],
-  hostname: "clickhouse",
+  hostname: "clickhouse.default.svc.cluster.local",
   port: 8123,
   database: "default",
   username: "default",
   password: "",
   pool_timeout: 60_000,
   timeout: 60_000,
-  pool_size: {:system, "CLICKHOUSE_POOL_SIZE", "30"}
```
tspenov: why the pool size is down to 3 ?

IvanIvanoff: This is only for dev. This pool size means that all 30 connections are opened all the time. Usually you connect to staging and when 2-3 people use it, stage has 100 additional connections.

tspenov: ups didn't see that its dev

lib/sanbase/clickhouse/github.ex
```diff
@@ -0,0 +1,96 @@
...
+        UNION ALL
+
+        SELECT toDateTime(intDiv(toUInt32(dt), ?1) * ?1) as time, count(e) as event
+          FROM (
+            SELECT any(event) as e, dt
+            FROM #{@table}
+            PREWHERE owner = ?3
+            AND dt >= toDateTime(?4)
+            AND dt <= toDateTime(?5)
+            AND event NOT in (?6)
```
valo: Do we really want to change the GH activity metric directly? I am bit cautious to change the calculations without examining some comparison between the 2 metrics. May be we can add a new query for the new calculations and provide a side by side comparison? We can also extend the medium article we posted with the extended research.

valo: Oh, I see now. We are actually adding a new query :)

lib/sanbase_web/graphql/schema.ex
```diff
@@ -201,6 +201,22 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+    @desc ~s"""
+    test
+    """
+    field :github_activity2, list_of(:activity_point) do
+      arg(:slug, :string)
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string)
+      arg(:transform, :string, default_value: "None")
+      arg(:moving_average_interval_base, :string, default_value: "1w")
```
valo: I think this argument is not used right now.

lib/sanbase/clickhouse/github.ex
IvanIvanoff: Yes, it's a new query. I think it's worth renaming it to `devActivity`. This has two advantages:
- Implies that it's for development and excluding talking
- Abstracts away from Github. Who knows, we may add gitlab at some time.

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: I plan to implement in Elixir as it's not present in ClickHouse.

lib/sanbase/clickhouse/github.ex
```diff
@@ -111,4 +127,32 @@ defmodule Sanbase.Clickhouse.Github do
...
+      end)
+
+    {:ok, result}
+  end
+
+  defp average([]), do: nil
+
+  defp average(l) when is_list(l) do
+    values = Enum.map(l, fn %{activity: da} -> da end)
+    %{datetime: datetime} = List.last(l)
```
valo: Should we take the first or the last datetime of the group for each SMA group?

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: It's now implemented.

lib/sanbase/clickhouse/github.ex
valo: Makes sense

PR #750, opened 2018-10-22
Add daa signal

tspenov:
#### Summary

https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/9ff267872fcd8b7925f1d7c7?card=San-2556


config/config.exs
```diff
@@ -50,7 +50,8 @@ config :logger, :console,
   format: {Sanbase.Utils.JsonLogger, :format},
   metadata: [:request_id],
   handle_otp_reports: true,
-  handle_sasl_reports: true
+  handle_sasl_reports: true,
+  level: :debug
```
IvanIvanoff: should not be :debug in prod?

IvanIvanoff: should not be :debug in prod?

tspenov: yeah, leftover

config/config.exs
```diff
@@ -137,6 +138,15 @@ config :sanbase, Sanbase.Discourse,
   api_key: {:system, "DISCOURSE_API_KEY"},
   insights_category: {:system, "DISCOURSE_INSIGHTS_CATEGORY", "sanbaseinsights"}
 
+config :sanbase, Sanbase.Scheduler,
+  global: true,
+  jobs: [
+    daa_signal: [
+      schedule: "35 10 * * *",
```
IvanIvanoff: Why exactly  at 10:35? Is it because of the shortcomings of TimescaleDB (data written once per day).
We can also try to use ClickHouse here

tspenov: haha, this is leftover from testing. I think it should be at `noon` like 12:00 smth so the yesterday's daa is calculated. But also maybe `@daily` can work - which starts at `00:00`

config/config.exs
```diff
@@ -137,6 +137,15 @@ config :sanbase, Sanbase.Discourse,
   api_key: {:system, "DISCOURSE_API_KEY"},
   insights_category: {:system, "DISCOURSE_INSIGHTS_CATEGORY", "sanbaseinsights"}
 
+config :sanbase, Sanbase.Scheduler,
+  global: true,
+  jobs: [
+    daa_signal: [
+      schedule: "00 12 * * *",
```
IvanIvanoff: in that case you can execute it with '@daily' though I'm not sure if you'll have the data available in the DB at this time

config/config.exs
```diff
@@ -137,6 +137,15 @@ config :sanbase, Sanbase.Discourse,
   api_key: {:system, "DISCOURSE_API_KEY"},
   insights_category: {:system, "DISCOURSE_INSIGHTS_CATEGORY", "sanbaseinsights"}
 
+config :sanbase, Sanbase.Scheduler,
+  global: true,
+  jobs: [
```
IvanIvanoff: Add here `timeout: 30_000` (or bigger :?). By default the timeout is 5s. On prod/stage where it will run for ~1000 projects it can easily run for longer

config/notifications_config.exs
```diff
@@ -20,3 +20,11 @@ config :sanbase, Sanbase.Notifications.PriceVolumeDiff,
...
   webhook_url: {:system, "INSIGHTS_DISCORD_WEBHOOK_URL"},
   insights_discord_publish_user: {:system, "INSIGHTS_DISCORD_PUBLISH_USER", "New Insight"}
+
+config :sanbase, Sanbase.Notifications.Discord.DaaSignal,
+  webhook_url: {:system, "DAA_SIGNAL_DISCORD_WEBHOOK_URL"},
+  daa_signal_discord_publish_user: {:system, "DAA_SIGNAL_DISCORD_PUBLISH_USER", "DAA GOING UP"},
+  daa_signal_threshold: {:system, "DAA_SIGNAL_THRESHOLD", "100"},
+  daa_signal_timeframe_from: {:system, "DAA_SIGNAL_TIMEFRAME_FROM", "30"},
+  daa_signal_timeframe_to: {:system, "DAA_SIGNAL_TIMEFRAME_TO", "2"},
+  daa_signal_change: {:system, "DAA_SIGNAL_CHANGE", "3"}
```
IvanIvanoff: These fields are already under the `DaaSignal` module, you do not need to prepend them with `daa_signal`.
`Config.module_get(DaaSignal, :publish_user)` is expressive enough

config/config.exs
tspenov: 12 is at noon, 00:00 is daily

config/config.exs
tspenov: That is the timeout for some GenServer calls not for the job execution. But yeah we can increase it to be on the safe side

IvanIvanoff: From the documentation: 
```
    * `:timeout` - Sometimes, you may come across GenServer
      timeout errors esp. when you have too many jobs or high
      load. The default GenServer.call timeout is 5000.
```
I think we can be categorized as too many jobs as we'll be executing 1000 requests to the db

tspenov: added it to 30_000

config/notifications_config.exs
```diff
@@ -20,3 +20,11 @@ config :sanbase, Sanbase.Notifications.PriceVolumeDiff,
 config :sanbase, Sanbase.Notifications.Insight,
   webhook_url: {:system, "INSIGHTS_DISCORD_WEBHOOK_URL"},
   insights_discord_publish_user: {:system, "INSIGHTS_DISCORD_PUBLISH_USER", "New Insight"}
+
+config :sanbase, Sanbase.Notifications.Discord.DaaSignal,
+  webhook_url: {:system, "DAA_SIGNAL_DISCORD_WEBHOOK_URL"},
+  publish_user: {:system, "DAA_SIGNAL_DISCORD_PUBLISH_USER", "DAA GOING UP"},
```
IvanIvanoff: `Daily Active Addresses Going Up` would better match the price-volume diff user name

lib/sanbase/application.ex
```diff
@@ -19,6 +19,7 @@ defmodule Sanbase.Application do
 
         # Start the Postgres Ecto repository
         Sanbase.Repo,
+        Sanbase.Scheduler,
```
IvanIvanoff: New line and a comment

PR #763, opened 2018-10-25
Don't cache san_balance when Ethauth call times out

LucoLe:
#### Summary

Doesn't cache san_balance when the call to Ethauth times out.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/auth/eth_account.ex
+++ b/lib/sanbase/auth/eth_account.ex
@@ -31,7 +31,7 @@ defmodule Sanbase.Auth.EthAccount do
       {:error, error} ->
         Logger.error(error)
 
-        nil
+        :error
     end
   end
 end
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -145,7 +145,7 @@ defmodule Sanbase.Auth.User do
 
   def update_san_balance_changeset(user) do
     user = Repo.preload(user, :eth_accounts)
-    san_balance = san_balance_for_eth_accounts(user.eth_accounts)
+    san_balance = san_balance_for_eth_accounts(user)
 
     user
     |> change(san_balance: san_balance, san_balance_updated_at: Timex.now())
@@ -179,11 +179,16 @@ defmodule Sanbase.Auth.User do
     end
   end
 
-  defp san_balance_for_eth_accounts(eth_accounts) do
-    eth_accounts
-    |> Enum.map(&EthAccount.san_balance/1)
-    |> Enum.reject(&is_nil/1)
-    |> Enum.reduce(Decimal.new(0), &Decimal.add/2)
+  defp san_balance_for_eth_accounts(%User{eth_accounts: eth_accounts, san_balance: san_balance}) do
+    eth_accounts_balances =
+      eth_accounts
+      |> Enum.map(&EthAccount.san_balance/1)
+      |> Enum.reject(&is_nil/1)
+
+    case Enum.member?(eth_accounts_balances, :error) do
+      true -> san_balance
+      _ -> Enum.reduce(eth_accounts_balances, Decimal.new(0), &Decimal.add/2)
+    end
   end
 
   def find_or_insert_by_email(email, username \\ nil) do
--- a/test/sanbase/auth/user_test.exs
+++ b/test/sanbase/auth/user_test.exs
@@ -3,6 +3,7 @@ defmodule Sanbase.Auth.UserTest do
 
   import Mockery
   import Sanbase.Factory
+  import ExUnit.CaptureLog
 
   alias Sanbase.Auth.{User, EthAccount}
   alias Sanbase.Repo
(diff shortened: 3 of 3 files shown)


lib/sanbase/auth/user.ex
```diff
@@ -147,8 +147,10 @@ defmodule Sanbase.Auth.User do
     user = Repo.preload(user, :eth_accounts)
     san_balance = san_balance_for_eth_accounts(user.eth_accounts)
 
-    user
-    |> change(san_balance: san_balance, san_balance_updated_at: Timex.now())
+    case san_balance do
+      :error -> change(user, %{san_balance: 0, san_balance_updated_at: nil})
```
IvanIvanoff: The idea is in this case to not change the balance. It has happened that a user has 1000 tokens but due to one such error `0` is written for the balance and the access is lost. In this case we need to keep the last valid san_balance

lib/sanbase/auth/eth_account.ex
```diff
@@ -31,7 +31,7 @@ defmodule Sanbase.Auth.EthAccount do
       {:error, error} ->
         Logger.error(error)
 
-        nil
+        :error
```
LucoLe: This method is called in the `EthAccountResolver`, so I'm wondering if we should do some changes to the `AccountTypes`'s `eth_account` object. Maybe I'm missing something but what will happen when the ethauth call times out? Please advise.

lib/sanbase/auth/user.ex
```diff
@@ -147,8 +147,16 @@ defmodule Sanbase.Auth.User do
...
     san_balance = san_balance_for_eth_accounts(user.eth_accounts)
 
-    user
-    |> change(san_balance: san_balance, san_balance_updated_at: Timex.now())
+    case san_balance do
+      :error ->
+        change(user, %{
+          san_balance: Decimal.to_integer(user.san_balance),
+          san_balance_updated_at: nil
+        })
```
IvanIvanoff: I think here `san_balance_updated_at` should not be changed, too.
Can we not change anything in that case? This is equivalent to an empty changes

lib/sanbase/auth/user.ex
```diff
@@ -147,8 +147,10 @@ defmodule Sanbase.Auth.User do
     user = Repo.preload(user, :eth_accounts)
     san_balance = san_balance_for_eth_accounts(user.eth_accounts)
 
-    user
-    |> change(san_balance: san_balance, san_balance_updated_at: Timex.now())
+    case san_balance do
+      :error -> changeset(user)
+      _ -> change(user, %{san_balance: san_balance, san_balance_updated_at: Timex.now()})
+    end
```
IvanIvanoff: I am still confused with this code. It is a duplication of the calculation of the san balance inside the user module, though it has a different behaviour - on an error instead of getting the old balance it throws an error. I think this should not calculate but use the user module directly

test/sanbase/auth/user_test.exs
```diff
@@ -55,6 +56,21 @@ defmodule Sanbase.Auth.UserTest do
...
+
+    user =
+      insert(:user,
+        san_balance: Decimal.new(100),
+        eth_accounts: [%EthAccount{address: "0x000000000001"}],
+        privacy_policy_accepted: true
+      )
+
+    capture_log(fn ->
+      assert User.san_balance!(user) == Decimal.new(100)
```
IvanIvanoff: The title does not state what is tested, better call it - `san_balance! returns cached result when EthAccount.san_balance fails`

IvanIvanoff: Also, as this is a bang function when does it raise an exception? If it does not it probably should be just `san_balance`

lib/sanbase/auth/user.ex
```diff
@@ -145,7 +145,7 @@ defmodule Sanbase.Auth.User do
 
   def update_san_balance_changeset(user) do
     user = Repo.preload(user, :eth_accounts)
-    san_balance = san_balance_for_eth_accounts(user.eth_accounts)
+    san_balance = san_balance_for_eth_accounts(user.eth_accounts, user.san_balance)
```
IvanIvanoff: You can pass around the `%User{}` struct and pattern match in the function header the needed fields

test/sanbase/auth/user_test.exs
LucoLe: That's a good question. I guess it will fail only if the update fails for some reason. But I think that was the behaviour before as well. I think the bang is more because it's returning a value instead of a tuple it the happy path. What do you think?

IvanIvanoff: Throughout sanbase we are not as strict about using `!` as I wish. The`!` is added when the function raises, so the one without `!` could return either `{:error, "some reason"}` or just `:ok`/`:error`

test/sanbase/auth/user_test.exs
```diff
@@ -55,6 +56,21 @@ defmodule Sanbase.Auth.UserTest do
...
+    mock(Sanbase.InternalServices.Ethauth, :san_balance, {:error, "foo"})
+
+    user =
+      insert(:user,
+        san_balance: Decimal.new(100),
+        eth_accounts: [%EthAccount{address: "0x000000000001"}],
+        privacy_policy_accepted: true
+      )
+
+    capture_log(fn ->
```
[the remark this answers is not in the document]
LucoLe: I know but I don't think that's the important thing in this test. So that's why I'm not asserting on the error message.

PR #774, opened 2018-10-29
Add dex and infrastructure columns to wallets

LucoLe:
#### Summary

Adds dex and infrastructure fields to the ExchangeEthAddress model. Adds
a relation to the Infrastructure model.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/model/exchange_eth_address.ex
```diff
@@ -4,25 +4,30 @@ defmodule Sanbase.Model.ExchangeEthAddress do
...
 
   schema "exchange_eth_addresses" do
     field(:address, :string)
     field(:name, :string)
     field(:source, :string)
     field(:comments, :string)
     field(:csv, :string, virtual: true)
+    field(:dex, :boolean)
+
+    belongs_to(:infrastructure, Infrastructure)
```
LucoLe: In the Favro card, it's written that there should be a field called network. Since we already have the Infrastructure model maybe it's better to have it as a relation. Should I add a function to this model called `network` that returns the `Infrastructure.code` or that's not needed at the moment. Also naming the foreign key `network` in this model didn't feel proper.

LucoLe:
How should we do the data migration? Should it be in the migration or we should do something else so that the migration can't be run more than once?

tspenov:
@LucoLe currently you can import exchange addresses through by pasting csv in the admin.
You can ask Garry if that should continue to work. If, yes, then these new fields should be handled. 
Do we have a new spreadsheet from @alexander-gr with the new columns filled. If we do the data migration will be just pasting the csv in the admin.

lib/sanbase/model/exchange_eth_address.ex
tspenov: looks good to me with the infrastructure assoc. In admin table we should show the infrastructure code.

LucoLe:
I don't know anything about such spreadsheets but will ask @alexander-gr

lib/sanbase/model/exchange_eth_address.ex
valo: The current solution looks good. Let's use consistent naming. `Infrastructure` is good

lib/sanbase/model/exchange_eth_address.ex
```diff
@@ -4,25 +4,30 @@ defmodule Sanbase.Model.ExchangeEthAddress do
   import Ecto.Query
 
   alias __MODULE__
+  alias Sanbase.Repo
+  alias Sanbase.Model.Infrastructure
 
   schema "exchange_eth_addresses" do
```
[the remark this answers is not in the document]
LucoLe: I think I'll add a card to rename the model, I don't want to do this in this card because it might mess up the migrations.

tspenov:
Is inserting/updating through admin panel working after the change ?

IvanIvanoff:
It will be good to link all currently present exchanges to their corresponding infrastructure

LucoLe:
You can edit the dex and the infrastructure_id through ExAdmin. As for the update it will be done with @alexander-gr 's spreadsheet when we have this PR on PROD.

IvanIvanoff:
Can a single ex_admin update change all dex/infrastructure_ids? In case it can - it's fine then.

lib/sanbase/model/exchange_eth_address.ex
```diff
@@ -4,25 +4,30 @@ defmodule Sanbase.Model.ExchangeEthAddress do
...
     field(:address, :string)
     field(:name, :string)
     field(:source, :string)
     field(:comments, :string)
     field(:csv, :string, virtual: true)
+    field(:dex, :boolean)
```
IvanIvanoff: one minor thing - this will be better called `is_dex`

PR #779, opened 2018-10-31
Automate clickhouse migration

LucoLe:
#### Summary

Adds a mix task that generates an empty clickhouse migration.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/mix/tasks/gen_clickhouse_migration.ex
+++ b/lib/mix/tasks/gen_clickhouse_migration.ex
@@ -1,0 +1,26 @@
+defmodule Mix.Tasks.GenClickhouseMigration do
+  use Mix.Task
+
+  import Macro, only: [underscore: 1]
+  import Mix.Generator
+
+  @shortdoc "Generates an empty clickhouse migration"
+
+  @moduledoc """
+  Generates an empty clickhouse migration file.
+  """
+
+  def run(args) do
+    [name] = args
+    file = "db/#{timestamp()}_#{underscore(name)}.sql"
+    create_file(file, "")
+  end
+
+  defp timestamp do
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/gen_clickhouse_migration.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  def run(args) do
+    [name] = args
+    file = "db/#{timestamp()}_#{underscore(name)}.sql"
+    create_file(file, "")
+  end
+
+  defp timestamp do
+    {{y, m, d}, {hh, mm, ss}} = :calendar.universal_time()
+    "#{y}#{pad(m)}#{pad(d)}#{pad(hh)}#{pad(mm)}#{pad(ss)}"
```
IvanIvanoff: Use https://hexdocs.pm/timex/Timex.Format.DateTime.Formatter.html#c:format/2

IvanIvanoff:
1. You won't need this task. All you need to do is `mix ecto.gen.migration -r Sanbase.ClickhouseRepo example_migration`

2. In order to do CH migrations you need to have the repo in `ecto_repos` otherwise the migrations won't be run at all. Also have in mind that the CH structure is not managed by sanbase, so the setup should be similar to Timescale - these migrations must not run in prod.

LucoLe:
I'm confused is this task still relevant or not? If it's relevant I don't understand what should be done because in the description sits something completely different.

IvanIvanoff:
@LucoLe https://github.com/santiment/clickhouse-tables is managing the ClickHouse database. There we roll out an in-house migration system that is lacking this functionality. It should be implemented there as a bash/python script.

PR #782, opened 2018-11-02
Rename exchange_eth_address model

LucoLe:
#### Summary

Renames `exchange_eth_address` model to `exchange_address`. Updates the code
to catch up with this change.

#### Related PRs

Depends on: https://github.com/santiment/sanbase2/pull/774

[//]: # (#### Screenshots)


priv/repo/migrations/20181102114904_rename_exchange_eth_addresses.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.RenameExchangeEthAddresses do
+  use Ecto.Migration
+
+  def up do
+    rename(table(:exchange_eth_addresses), to: table(:exchange_addresses))
+  end
+
+  def down, do: :ok
```
IvanIvanoff: You can use `change` because this allows a rollback (or manually write down)

priv/repo/structure.sql
```diff
@@ -1600,6 +1602,14 @@ ALTER TABLE ONLY public.eth_accounts
     ADD CONSTRAINT eth_accounts_user_id_fkey FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE;
 
 
+--
+-- Name: exchange_addresses exchange_eth_addresses_infrastructure_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.exchange_addresses
+    ADD CONSTRAINT exchange_eth_addresses_infrastructure_id_fkey FOREIGN KEY (infrastructure_id) REFERENCES public.infrastructures(id);
```
IvanIvanoff: this still has `exchange_eth` in the name

priv/repo/migrations/20181102114904_rename_exchange_eth_addresses.exs
LucoLe: I'm not sure that we want this migration to be reversible. On the other hand, I don't know if we want to make this migration blocking on rollback.

priv/repo/structure.sql
LucoLe: Yeah looks like this method is not doing its job very good. Probably will have to manually do some changes in the structure.sql.

valo:
The problem is that during such a migration, the old and the new code will be running at the same time for a while, which will make the old code fail after the migration is done. Can we indeed just create a new table and copy the data over? That will be code only in the data migrations code. After the app is deployed and we see everything works properly we can submit a separate PR with a migration to delete the old table.

This approach will allow to very easily do a rollback in case something goes wrong. We will one able to just revert the PR, merge and deploy the revert.

priv/repo/structure.sql
IvanIvanoff: It's kind of expected. You're renaming the table but not recreating the index

IvanIvanoff: @LucoLe https://dba.stackexchange.com/questions/98875/postgresql-rename-a-foreignkey-after-renaming-a-table

priv/repo/structure.sql
```diff
@@ -132,9 +132,39 @@ CREATE TABLE public.exchange_addresses (
     address character varying(255) NOT NULL,
     name character varying(255) NOT NULL,
     comments text,
-    source text,
-    dex boolean,
-    infrastructure_id bigint
+    source text
```
IvanIvanoff: Why is it removing 2 columns? Probably you need to rebase your PR on master as the branch this is based on is some feature branch that is not merged

priv/repo/structure.sql
```diff
@@ -132,9 +132,39 @@ CREATE TABLE public.exchange_addresses (
...
+--
+-- Name: exchange_eth_addresses; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.exchange_eth_addresses (
+    id bigint NOT NULL,
+    address character varying(255) NOT NULL,
+    name character varying(255) NOT NULL,
+    comments text,
+    source text
```
IvanIvanoff: This create table is also strange. This PR renames this existing table to something else but the structure.sql shows that it is creating this table

priv/repo/structure.sql
```diff
@@ -1602,14 +1654,6 @@ ALTER TABLE ONLY public.eth_accounts
     ADD CONSTRAINT eth_accounts_user_id_fkey FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE;
 
 
---
--- Name: exchange_addresses exchange_eth_addresses_infrastructure_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
---
-
-ALTER TABLE ONLY public.exchange_addresses
-    ADD CONSTRAINT exchange_eth_addresses_infrastructure_id_fkey FOREIGN KEY (infrastructure_id) REFERENCES public.infrastructures(id);
```
IvanIvanoff: How is this removing from master branch `alter` on a table that does not exist? Ping me when you rebase against master so the changes are the correct one

PR #785, opened 2018-11-05
Add iex helpers

tspenov:
When running `iex -S mix` - `.iex.exs` is executed. Added some aliases so we easily work with modules in the dev console.

### DIFF
--- a/.iex.exs
+++ b/.iex.exs
@@ -1,3 +1,5 @@
+import Ecto.Query, only: [from: 2]
+
 alias Sanbase.Model.{
   Infrastructure,
   MarketSegment,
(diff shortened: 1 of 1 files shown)


.iex.exs
```diff
@@ -0,0 +1,146 @@
...
+  ProjectResolver,
+  ProjectTransactionsResolver,
+  TechIndicatorsResolver,
+  TwitterResolver,
+  UserListResolver,
+  VotingResolver
+}
+
+alias SanbaseWeb.Graphql.Helpers.{
+  Async,
```
IvanIvanoff: That one won't be needed

IvanIvanoff:
I think it will be nice to have some helper functions such as:
```elixir
now = fn -> Timex.now() end
days_ago = fn days -> Timex.shift(Timex.now(), days: -days) end
project_slugs = fn -> Sanbase.Repo.all(...) end
project_contracts = ...
```

tspenov:
Yes anything that we find that we repeat a lot and will ease the dev process we can add in the file. But especially the aliases are always so tedious to type

PR #790, opened 2018-11-07
Change CMC api from v1 to pro version

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/external_services/coinmarketcap2/ticker.ex
```diff
@@ -77,13 +82,64 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.Ticker2 do
...
-  def parse_json(json) do
-    json
-    |> Poison.decode!(as: [%Ticker{}])
-    |> Stream.filter(fn %Ticker{last_updated: last_updated} -> last_updated end)
-    |> Enum.map(&make_timestamp_integer/1)
+  @spec parse_json(String.t()) :: [%Ticker{}] | no_return
+  defp parse_json(json) do
+    %{"data" => data} =
+      json
+      |> Poison.decode!()
```
tspenov: we can use `Jason` since we are not using `as` opt ?

IvanIvanoff:
@tspenov it's two different modules because when we migrate the prices I'll just delete all the old modules.

tspenov:
There is some random failing test `Sanbase.Github.SchedulerTest` - its happening more than once, although replay fixes it

tspenov:
> @tspenov it's two different modules because when we migrate the prices I'll just delete all the old modules.

Just seems to me that the difference here is very small related to the duplicated code and I needed to compare both side to side to see what is the different thing. But, ok, lets keep it as it is

tspenov:
> @tspenov it's two different modules because when we migrate the prices I'll just delete all the old modules.

Just seems to me that the difference here is very small related to the duplicated code and I needed to compare both side to side to see what is the different thing. But, ok, lets keep it as it is

PR #791, opened 2018-11-07
Refactor daa signal

tspenov:
Discord allows max 2000 chars per message. Split into smaller messages.
Refactor using the behaviour.

### DIFF
--- a/lib/sanbase/notifications/discord/discord.ex
+++ b/lib/sanbase/notifications/discord/discord.ex
@@ -28,16 +28,18 @@ defmodule Sanbase.Notifications.Discord do
       {:ok, %HTTPoison.Response{status_code: 429, body: body} = resp} ->
         body = body |> Jason.decode!()
 
-        Logger.warn(
+        Logger.info(
           "Cannot publish #{signal_name} signal in Discord: HTTP Response: #{inspect(resp)}"
         )
 
         Process.sleep(body["retry_after"] + 1000)
 
-        send_notification(webhook, signal_name, payload, %{
-          opts
-          | retry_count: Map.get(opts, :retry_count, 0) + 1
-        })
+        send_notification(
+          webhook,
+          signal_name,
+          payload,
+          Map.update(opts, :retry_count, 0, &(&1 + 1))
+        )
 
       {:ok, %HTTPoison.Response{} = resp} ->
         Logger.error(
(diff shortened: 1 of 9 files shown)


lib/sanbase/notifications/discord/exchange_inflow.ex
```diff
@@ -19,7 +19,10 @@ defmodule Sanbase.Notifications.Discord.ExchangeInflow do
   def run() do
     projects =
       projects()
-      |> filter_projects()
+      |> Project.projects_over_volume_threshold(
+        Config.get(:trading_volume_threshold)
+        |> String.to_integer()
+      )
```
IvanIvanoff: Can you put this in a separate variable defined before `projects = ... `

lib/sanbase/notifications/discord/discord.ex
```diff
@@ -5,17 +5,40 @@ defmodule Sanbase.Notifications.Discord do
...
     case http_client().post(webhook, payload, [{"Content-Type", "application/json"}]) do
       {:ok, %HTTPoison.Response{status_code: code}} when code in 200..299 ->
         :ok
 
+      {:ok, %HTTPoison.Response{status_code: 429, body: body} = resp} ->
+        body = body |> Jason.decode!()
+
+        Logger.warn(
+          "Cannot publish #{signal_name} signal in Discord: HTTP Response: #{inspect(resp)}"
+        )
```
IvanIvanoff: This case is expected so it probably should not be logged at a `warn` level

lib/sanbase/notifications/discord/discord.ex
```diff
@@ -5,17 +5,40 @@ defmodule Sanbase.Notifications.Discord do
...
+
+        Process.sleep(body["retry_after"] + 1000)
+
+        send_notification(webhook, signal_name, payload, %{
+          opts
+          | retry_count: Map.get(opts, :retry_count, 0) + 1
+        })
```
IvanIvanoff: You can simplify this with:
```elixir
send_notification(webhook, signal_name, payload, Map.update(opts, :retry_count, 0, & &1+1))

lib/sanbase/notifications/discord/discord.ex
tspenov: lets have at least info

PR #801, opened 2018-11-13
Add absinthe prometheus metrics. Refactor the application supervisor

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -9,13 +9,14 @@ defmodule Sanbase.Application do
       Envy.auto_load()
     end
 
-    common_init()
-    common_children = common_children()
+    container_type = System.get_env("CONTAINER_TYPE") || "all"
+
+    init(container_type)
 
     {children, opts} =
-      case System.get_env("CONTAINER_TYPE") || "all" do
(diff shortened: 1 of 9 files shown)


lib/sanbase/application.ex
```diff
@@ -45,21 +46,40 @@ defmodule Sanbase.Application do
...
     # Prometheus related
     Sanbase.Prometheus.EctoInstrumenter.setup()
 
     Sanbase.Prometheus.PipelineInstrumenter.setup()
 
     Sanbase.Prometheus.Exporter.setup()
+
+    # Container specific init
+    case container_type do
+      "all" ->
```
tspenov: will we have the same function executed more than once when we run all inits ?

IvanIvanoff: Yes, but have in mind the following:
1. In prod we never run `all` - it's only for test/dev purposes
2. If there is an issue it will get caught in the tests
3. We'll remove the `workers` type soon, so we'll have only 2 types. If a function has to be executed in both contexts then it will be in the main `init`

PR #825, opened 2018-11-22
Add metrics to api

LucoLe:
#### Summary

Adds average token age consumed in days and token circulation GQL endpoints.

[//]: # (#### Related PRs)

#### Screenshots

image

image

### DIFF
--- a/lib/sanbase/blockchain/token_circulation.ex
+++ b/lib/sanbase/blockchain/token_circulation.ex
@@ -1,0 +1,50 @@
+defmodule Sanbase.Blockchain.TokenCirculation do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+  alias Sanbase.Timescaledb
+
+  @table Timescaledb.table_name("eth_coin_circulation")
+  @interval_mapping %{
+    less_than_a_day: "_-1d",
+    between_a_day_and_a_week: "1d-1w",
+    between_a_week_and_a_month: "1w-1m",
+    between_a_month_and_three_weeks: "1m-3m",
+    between_three_months_and_six_months: "3m-6m",
+    between_six_months_and_a_year: "6m-12m",
+    between_a_year_and_eighteen_months: "12m-18m",
+    between_eighteen_months_and_two_years: "18m-24m",
+    between_two_years_and_three_years: "2y-3y",
+    between_three_years_and_five_years: "3y-5y",
+    more_than_five_years: "5y-_"
+  }
+
+  def token_circulation(tc_interval, contract, from, to, interval, token_decimals \\ 0) do
+    args = [from, to, contract]
+
+    """
+    SELECT sum("#{@interval_mapping[tc_interval]}") AS value
+    FROM #{@table}
+    WHERE timestamp >= $1 AND timestamp <= $2 AND contract_address = $3
+    """
+    |> Timescaledb.bucket_by_interval(args, interval)
+    |> Timescaledb.timescaledb_execute(fn [datetime, token_circulation] ->
+      %{
+        datetime: Timescaledb.timestamp_to_datetime(datetime),
+        token_circulation: token_circulation / :math.pow(10, token_decimals)
+      }
+    end)
+  end
+
+  def token_circulation!(tc_interval, contract, from, to, interval, token_decimals \\ 0) do
+    case token_circulation(tc_interval, contract, from, to, interval, token_decimals) do
+      {:ok, result} -> result
+      {:error, error} -> raise(error)
+    end
+  end
+
+  def first_datetime(contract) do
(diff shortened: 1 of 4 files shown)


lib/sanbase/blockchain/token_circulation.ex
```diff
@@ -0,0 +1,50 @@
...
+    between_two_years_and_three_years: "2y-3y",
+    between_three_years_and_five_years: "3y-5y",
+    more_than_five_years: "5y-_"
+  }
+
+  def token_circulation(tc_interval, contract, from, to, interval, token_decimals \\ 0) do
+    args = [from, to, contract]
+
+    """
+    SELECT sum("#{@interval_mapping[tc_interval]}") AS value
```
LucoLe: I'm not sure which aggregation function I should use on the column `sum` or `average`. Please advise! @IvanIvanoff

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -10,13 +10,20 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
...
-  @doc ~S"""
-    Return the token burn rate for the given slug and time period.
-    Uses the influxdb cached values instead of issuing a GET request to etherbi
-  """
-  def burn_rate(_root, %{slug: slug, from: from, to: to, interval: interval} = args, _resolution) do
-    with {:ok, contract_address, token_decimals} <- Project.contract_info_by_slug(slug),
-         {:ok, from, to, interval} <-
+  def burn_rate(_root, %{slug: slug} = args, _resolution) do
+    with {:ok, contract_address, token_decimals} <- Project.contract_info_by_slug(slug) do
+      Cache.func(
```
tspenov: I think we're caching twice here - once with `cache_resolve` and second time here. Wouldn't cache_resolve be sufficient ?

lib/sanbase/blockchain/token_circulation.ex
IvanIvanoff: The current `sum` is correct.
One thing you might need to guard against is intervals less than 1 day, as we have the metric per day.

LucoLe: I've set the intervals in the `@interval_mapping` constant and don't allow the user to chose an arbitrary interval (at the moment the user can't choose the interval at all, but even if we allow that we'll give him the options of set intervals). So I guess that's covered.

IvanIvanoff: As far as I see the interval that the user passes is not checked and is used here. What is the behavior if it is `1h`?: https://github.com/santiment/sanbase2/pull/825/files#diff-6c0195a22adab7da76c46b6f4287f669R30

lib/sanbase_web/graphql/schema.ex
```diff
@@ -284,7 +285,40 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+
+    @desc ~s"""
+    Fetch token circulation for a project, grouped by interval.
+    Projects are referred to by a unique identifier (slug).
+    """
+    field :token_circulation, list_of(:token_circulation_data) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "1d")
```
IvanIvanoff:
```suggestion
      @desc "The interval should represent whole days, i.e. `1d`, `48h`, `1w`, etc."
      arg(:interval, :string, default_value: "1d")
```

lib/sanbase_web/graphql/schema/etherbi_types.ex
```diff
@@ -11,6 +11,11 @@ defmodule SanbaseWeb.Graphql.EtherbiTypes do
     field(:transaction_volume, :float)
   end
 
+  object :average_token_age_data do
+    field(:datetime, non_null(:datetime))
+    field(:average_token_age, :float)
```
IvanIvanoff: I think we should not use `average` here, the same as with the `active_addresses` below. It is just token age and should be agnostic of the aggregation function

lib/sanbase_web/graphql/schema.ex
```diff
@@ -284,7 +285,40 @@ defmodule SanbaseWeb.Graphql.Schema do
 
       middleware(ApiTimeframeRestriction)
       complexity(&Complexity.from_to_interval/3)
-      cache_resolve(&EtherbiResolver.transaction_volume/3)
+      # this is cached in the resolver
+      resolve(&EtherbiResolver.transaction_volume/3)
```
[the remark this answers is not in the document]
LucoLe: Do you think I should revert the changes and cache everything in the schema?

PR #827, opened 2018-11-23
enhance/mergetimeseries-optimization

DmitriVanGuard:
#### Summary
This PR increases calculation speed of the `mergeTimeseriesByKey` algorithm.

### Setup
- **Specs**: Core i5 2.9GHz
- **Data**: 4 arrays with 90 elements
- Throttling was achieved by using Google Chrome Dev Tools

### Benchmark results
| | Old algorithm | New algorithm |
| :---         |     :---:      |          :---: |
| No throttling   | ~4.4ms     | ~0.3ms    |
| 4x slowdown     | ~13.2ms       | ~1.3ms      |
| 6x slowdown     | ~19.7ms      | ~2.1ms      |


app/src/utils/utils.js
```diff
@@ -116,23 +116,49 @@ const getStartOfTheDay = () => {
...
+    for (
+      let timeserieRightIndex = timeserie.length - 1;
+      timeserieRightIndex > -1;
+      timeserieRightIndex--
+    ) {
+      while (longestTSRightIndexBoundary > -1) {
+        const longestTSData = longestTS[longestTSRightIndexBoundary]
+        const timeserieData = timeserie[timeserieRightIndex]
+        if (longestTSData[mergeKey] === timeserieData[mergeKey]) {
+          longestTS[longestTSRightIndexBoundary] = Object.assign(
```
Partysun: {...} instead of assign @DmitriVanGuard

DmitriVanGuard: @Partysun I have tried it with object spread and it slow down total time by almost 1ms. `Object.assign` does not significantly decrease readability in this part of code, so i guess we will be fine with it 👍

PR #837, opened 2018-11-26
Add watchlist admin

tspenov:
#### Summary
Add ability to moderate watchlists in admin panel

image
image

### DIFF
--- a/lib/sanbase/user_lists/user_list.ex
+++ b/lib/sanbase/user_lists/user_list.ex
@@ -19,6 +19,11 @@ defmodule Sanbase.UserLists.UserList do
     timestamps()
   end
 
+  # ex_admin needs changeset function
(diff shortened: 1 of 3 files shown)


lib/sanbase/user_lists/user_list.ex
```diff
@@ -19,6 +19,11 @@ defmodule Sanbase.UserLists.UserList do
     timestamps()
   end
 
+  # ex_admin needs changeset function
+  def changeset(user_list, attrs \\ %{}) do
+    update_changeset(user_list, attrs)
```
IvanIvanoff: This means that through the admin panel we could attempt to create a watchlist but it will fail?

tspenov: removed the create action

PR #843, opened 2018-11-27
Create exchanges table

tspenov:
* create exchanges table and fill it up
In subsequent PRs will link the table to the `exchange_addresses` and show in admin panel
After all is migrated will make an API endpoint to list all exchanges and then use it for the exchanges volume API endpoint.


tspenov:
Lets leave this for now

PR #846, opened 2018-11-27
Add exchange inflow chart to the discord signals

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/notifications/discord/discord.ex
```diff
@@ -123,39 +124,44 @@ defmodule Sanbase.Notifications.Discord do
...
       prices =
         prices
         |> Enum.map(fn list -> list |> Enum.filter(fn el -> el != 0 end) end)
         |> Enum.map(fn list ->
           list
-          |> Enum.map(&(&1 * 1.0))
+          |> Enum.map(&Math.to_float/1)
           |> Enum.map(fn num ->
-            if num > 100, do: Float.round(num, 2), else: Float.round(num, 4)
+            if num > 1, do: Float.round(num, 2), else: Float.round(num, 6)
```
tspenov: That can be  `Float.round(num, 6)`. The if was needed so the URL does not become very long due to some limitations. But since we're saving to s3 - that's not a prob

lib/sanbase/notifications/discord/discord.ex
```diff
@@ -123,39 +124,44 @@ defmodule Sanbase.Notifications.Discord do
...
           end)
         end)
 
       [_, high_values, low_values, _, _] = prices
-      min = low_values |> Enum.min() |> Float.floor(2)
-      max = high_values |> Enum.max() |> Float.ceil(2)
```
tspenov: Ceil and floor are so the scale is not exactly cut at the last decimal  place of highest/lowest value but has some room up/down

lib/sanbase/notifications/discord/discord.ex
```diff
@@ -195,6 +201,25 @@ defmodule Sanbase.Notifications.Discord do
...
   end
 
+  defp exchange_inflow_chart_values(slug, _from, to, size) do
+    from = Timex.shift(to, days: -size + 1)
+
+    with {:ok, contract, token_decimals} <- Project.contract_info_by_slug(slug),
+         {:ok, daa} <-
+           ExchangeFundsFlow.transactions_in_over_time(contract, from, to, "1d", token_decimals) do
+      exchange_inflow_values = daa |> Enum.map(fn %{active_addresses: value} -> value end)
+      max = daa_values |> Enum.max()
```
tspenov: `daa_values`

lib/sanbase/notifications/discord/exchange_inflow.ex
```diff
@@ -33,10 +33,10 @@ defmodule Sanbase.Notifications.Discord.ExchangeInflow do
     |> case do
       {:ok, list} ->
         build_payload(projects, list)
-        |> Discord.group_messages()
-        |> Enum.each(fn payload ->
+        |> Enum.each(fn {payload, embeds} ->
           payload
-          |> Discord.encode!(publish_user())
+          |> Discord.encode!(publish_user(), embeds)
+          |> IO.inspect()
```
tspenov: inspect here

tspenov:
We should move chart generation related funcs into a separate module - but that can be another iteration

IvanIvanoff:
@tspenov agree on the new module. We also should try to abstract even more the common logic between the signals' modules

PR #848, opened 2018-11-28
Add exchange volume api

tspenov:
List all exchanges:
```
{ allExchanges }
```

```
{
  "data": {
    "allExchanges": [
      "StarBit",
      "DDEX",
      "ShapeShift",
      "Bilaxy",
      "Yobit",
      "COSS",
      "Tidex tokens",
      "Indodax",
      "CoinBene",
      "Kraken",
      "YoBit",
     ...
]
}
}
```

Get exchange in/out volume for a given exchange:
```graphql
{
  exchangeVolume(
    exchange:"Bitfinex",
    from:"2018-11-20T00:00:00Z",
    to:"2018-11-28T00:00:00Z") {
    
    datetime,
    exchangeInflow,
    exchangeOutflow
  }
}
```

```
{
  "data": {
    "exchangeVolume": [
      {
        "datetime": "2018-11-20T00:00:00Z",
        "exchangeInflow": 31302985.286791388,
        "exchangeOutflow": 21375196.928917754
      },
      {
        "datetime": "2018-11-21T00:00:00Z",
        "exchangeInflow": 9305794.544514699,
        "exchangeOutflow": 11865416.470193278
      },
      {
        "datetime": "2018-11-22T00:00:00Z",
        "exchangeInflow": 1869233.3175404922,
        "exchangeOutflow": 10149200.337983582
      },
      {
        "datetime": "2018-11-23T00:00:00Z",
        "exchangeInflow": 16151397.920319388,
        "exchangeOutflow": 8036358.857381285
      },
      {
        "datetime": "2018-11-24T00:00:00Z",
        "exchangeInflow": 7501403.44171947,
        "exchangeOutflow": 5815005.249321977
      },
      {
        "datetime": "2018-11-25T00:00:00Z",
        "exchangeInflow": 28467808.17673707,
        "exchangeOutflow": 7974032.029678876
      },
      {
        "datetime": "2018-11-26T00:00:00Z",
        "exchangeInflow": 9129825.742878968,
        "exchangeOutflow": 6761136.759500962
      },
      {
        "datetime": "2018-11-27T00:00:00Z",
        "exchangeInflow": 6362990.9915780965,
        "exchangeOutflow": 10673524.057566423
      },
      {
        "datetime": "2018-11-28T00:00:00Z",
        "exchangeInflow": 6524534.175788566,
        "exchangeOutflow": 1135402.0117541507
      }
    ]
  }
}
```


lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -148,6 +148,29 @@ defmodule Sanbase.Clickhouse.EthTransfers do
...
+  @doc ~s"""
+  Returns the inflow and outflow volume for a list of exchange_addresses between two datetimes
+  """
+  def exchange_volume(exchange_addresses, from_datetime, to_datetime) do
+    {query, args} = exchange_volume_query(exchange_addresses, from_datetime, to_datetime)
+
+    query
+    |> ClickhouseRepo.query_transform(args, fn [dt, inflow, outflow] ->
+      %{
+        datetime: DateTime.from_unix!(dt),
```
IvanIvanoff: Here we are handling the date time in a different way than the other clickhousex modules. I think it makes sense to handle them the same way: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/eth_transfers.ex#L110

tspenov: Here it returns string if `toUnixTimestamp` is not used not erlang tuple probably because datetimes are taken with `toStartOfDay`.  In the other cases it returns erlang tuple so using from_erl! makes sense there.

lib/sanbase_web/graphql/resolvers/exchange_resolver.ex
```diff
@@ -0,0 +1,22 @@
...
+  alias Sanbase.Model.ExchangeAddress
+  alias Sanbase.Clickhouse.EthTransfers
+
+  def all_exchanges(_root, _args, _resolution) do
+    {:ok, ExchangeAddress.list_all_exchanges()}
+  end
+
+  def exchange_volume(_root, %{exchange: exchange, from: from, to: to}, _resolution) do
+    with {:ok, addresses} <- ExchangeAddress.addresses_for_exchange(exchange),
+         addresses <- addresses |> Enum.map(&String.downcase/1),
```
IvanIvanoff: This will lead to a compile time warning because it always matches. Let's move this transformation inside `exchange_volume/3` function as the outer world does not need to know for this requirement.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -680,6 +682,22 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+    @desc "List all exchanges"
+    field :all_exchanges, list_of(:string) do
+      cache_resolve(&ExchangeResolver.all_exchanges/3)
+    end
+
+    @desc ~s"""
+    Calculates the exchange inflow and outflow volume in usd for a given exchange in a time interval.
+    """
+    field :exchange_volume, list_of(:exchange_volume) do
+      arg(:exchange, :string)
```
IvanIvanoff: non_null

lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -338,6 +361,63 @@ defmodule Sanbase.Clickhouse.EthTransfers do
...
+          from IN (?1) AND NOT to IN (?1)
+          AND dt >= toDateTime(?2)
+          AND dt <= toDateTime(?3)
+      GROUP BY dt
+     ) USING dt
+    )
+    ALL INNER JOIN
+    (
+     SELECT 
+      toStartOfDay(dt) as dt, AVG(price_usd) as "price_usd"
```
IvanIvanoff: Can we make this configurable via the `interval` param and not working only with `1d`

tspenov: yeah that will be better but I don't think it will be very easy to implement cause the query is quite complex now. Do you have any suggestion how to implement it ?

lib/sanbase/model/exchange_address.ex
```diff
@@ -26,8 +26,34 @@ defmodule Sanbase.Model.ExchangeAddress do
...
+  @doc ~s"List all exchange addresses"
+  @spec list_all() :: list(String.t())
   def list_all() do
     Repo.all(__MODULE__)
     |> Enum.map(fn %__MODULE__{address: address} -> address end)
   end
+
+  @doc ~s"List all exchange names"
+  @spec list_all_exchanges() :: list(String.t())
+  def list_all_exchanges() do
```
IvanIvanoff: these two names `list_all` and `list_all_exchanges` are a little bit confusing.
`list_all` sounds like it should return a list of `%__MODULE__`.
I suggest to have only `list_all` that returns just `Repo.all(__MODULE__)` and fix the only one place where it is currently used - In the `mark_exchanges.ex` and add the `Enum.map` there.

In your case you'll define here a function `exchange_names` as `list_all() |> Enum.map(fn %{name: name} -> name end) |> Enum.uniq()`

tspenov: yeah `exchange_names` sounds better, also agree that `list_all` should return list of `__MODULE__`. But I think uniqueness is better practice to be in the query with `distinct`

lib/sanbase/clickhouse/eth_transfers.ex
```diff
@@ -135,8 +146,14 @@ defmodule Sanbase.Clickhouse.EthTransfers do
...
   @doc ~s"""
-  Returns the historical balances of given etherium address in all intervals between two datetimes.
+  Returns the historical balances of given ethereum address in all intervals between two datetimes.
   """
+  @spec historical_balance(
+          String.t(),
+          %DateTime{},
+          %DateTime{},
+          non_neg_integer()
+        ) :: {:ok, list(historical_balance)} | {:ok, []}
```
IvanIvanoff: `list()` covers the `[]`. `[]` is not covered by `nonempty_list()`

PR #865, opened 2018-12-07
Add network growth api

tspenov:
```graphql
    {
      networkGrowth(
        slug: "santiment",
        from: "2018-12-01T00:00:00Z",
        to: "2018-12-05T00:00:00Z",
        interval: "1d"
      )
      {
        datetime,
        newAddresses
      }
    }
```
result:
```graphql
{
  "data": {
    "networkGrowth": [
      {
        "datetime": "2018-12-01T00:00:00Z",
        "newAddresses": "3"
      },
      {
        "datetime": "2018-12-02T00:00:00Z",
        "newAddresses": "2"
      },
      {
        "datetime": "2018-12-03T00:00:00Z",
        "newAddresses": "6"
      },
      {
        "datetime": "2018-12-04T00:00:00Z",
        "newAddresses": "2"
      },
      {
        "datetime": "2018-12-05T00:00:00Z",
        "newAddresses": "1"
      }
    ]
  }
}
```

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
@@ -6,7 +6,28 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   alias Sanbase.Clickhouse.Erc20Transfers
   alias Sanbase.Model.Project
 
+  alias Sanbase.Clickhouse.NetworkGrowth
+
   @one_hour_seconds 3600
+
+  def network_growth(_root, args, _resolution) do
+    with {:ok, contract, _} <- args.slug |> Project.by_slug() |> Project.contract_info(),
+         interval <- DateTimeUtils.compound_duration_to_seconds(args.interval),
+         {:ok, network_growth} =
+           NetworkGrowth.network_growth(contract, args.from, args.to, interval) do
+      {:ok, network_growth}
+    else
+      error ->
+        Logger.warn("Can't calculate network growth. Reason: #{inspect(error)}")
+
+        {:ok, []}
+    end
+  rescue
+    e ->
+      Logger.error("Exception raised while calculating network growth. Reason: #{inspect(e)}")
+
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -6,8 +6,29 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   alias Sanbase.Clickhouse.Erc20Transfers
   alias Sanbase.Model.Project
 
+  alias Sanbase.Clickhouse.NetworkGrowth
+
   @one_hour_seconds 3600
 
+  def network_growth(_root, args, _resolution) do
```
IvanIvanoff: I think this function should be in the `NetworkGrowth` module and not the `Erc20Transfers`

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -6,8 +6,29 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+    with {:ok, contract, _} <- args.slug |> Project.by_slug() |> Project.contract_info(),
```
IvanIvanoff: We have `def by_slug(slug) when is_binary(slug) do` but the `slug` arg can be `nil`. I think it's best to have the schema definition as: `arg(:slug, non_null(:string))` and pattern match in the function header (as we do this in each resolver)

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -6,8 +6,29 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+         interval <- DateTimeUtils.compound_duration_to_seconds(args.interval),
```
IvanIvanoff: This will always succeed, you can take it out of the `with`

PR #867, opened 2018-12-07
Add logs for usage of api queries. Log the corresponding token or user_id

IvanIvanoff:
#### Summary
Example log:
```
{"token": "some_token","complexity":3,"level":"info","message":"Apikey usage: token: some_token, query: allErc20Projects, complexity: 3","query":"allErc20Projects","request_id":"2ln1s66fjif1gjvids0016s1","timestamp":"2018-12-07T18:46:54.146"}
```

This json will be parsed by ES and we'll be able to run queries on top of it

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/middlewares/api_usage.ex
+++ b/lib/sanbase_web/graphql/middlewares/api_usage.ex
@@ -56,10 +56,22 @@ defmodule SanbaseWeb.Graphql.Middlewares.ApiUsage do
     resolution
   end
 
-  def call(%{definition: definition} = resolution, _) do
+  def call(%{definition: definition, context: %{remote_ip: remote_ip}} = resolution, _) do
     metadata = Logger.metadata()
-    Logger.metadata(user_id: nil, query: definition.name, complexity: definition.complexity)
-    Logger.info("Apikey usage: query: #{definition.name}, complexity: #{definition.complexity}")
+    remote_ip = to_string(:inet_parse.ntoa(remote_ip))
+
+    Logger.metadata(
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/middlewares/api_usage.ex
```diff
@@ -0,0 +1,62 @@
...
+      }"
+    )
+
+    Logger.reset_metadata([:user_id, :query, :complexity])
+
+    resolution
+  end
+
+  def call(%{definition: definition} = resolution, _) do
+    Logger.metadata(user_id: nil, query: definition.name, complexity: definition.complexity)
```
tspenov: can we take some info from Plug.Conn like the ip address and put it here ?

IvanIvanoff: Yes, I've pushed such a change. I am just wondering if we should always log the IP?
cc @valo @tspenov

lib/sanbase_web/graphql/middlewares/api_usage.ex
```diff
@@ -0,0 +1,78 @@
+defmodule SanbaseWeb.Graphql.Middlewares.ApiUsage do
+  @moduledoc """
+  Track the API usage by logging some info to the standard output.
+  These logs are analyzed with Elasticsearch
+  """
+  @behaviour Absinthe.Middleware
+  # @behaviour Absinthe.Plugin
```
tspenov: leftover

PR #869, opened 2018-12-10
Hist balance fixes

tspenov:
* Downcase address
* handle ethereum slug differently


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -35,6 +35,13 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
       {:ok, []}
   end
 
+  defp calc_historical_balances(
```
IvanIvanoff: Let's remove the non-slug case and add `non_null(:string)` as the slug's type in the schema

PR #883, opened 2018-12-18
Implement average token age consumed in days GQL API

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/blockchain/burn_rate.ex
+++ b/lib/sanbase/blockchain/burn_rate.ex
@@ -45,6 +45,32 @@ defmodule Sanbase.Blockchain.BurnRate do
     end
   end
 
+  def average_token_age_consumed_in_days(contract, from, to, interval, token_decimals \\ 0) do
+    with {:ok, token_age_consumed} <- burn_rate(contract, from, to, interval, token_decimals),
+         {:ok, transaction_volume} <-
+           Sanbase.Blockchain.TransactionVolume.transaction_volume(
+             contract,
+             from,
+             to,
+             interval,
+             token_decimals
+           ) do
+      average_token_age_consumed_in_days =
+        Enum.zip(token_age_consumed, transaction_volume)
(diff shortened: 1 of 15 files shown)


lib/sanbase/blockchain/burn_rate.ex
```diff
@@ -45,6 +45,32 @@ defmodule Sanbase.Blockchain.BurnRate do
...
+             to,
+             interval,
+             token_decimals
+           ) do
+      average_token_age_consumed_in_days =
+        Enum.zip(token_age_consumed, transaction_volume)
+        |> Enum.map(fn {%{burn_rate: br, datetime: dt}, %{transaction_volume: tv}} ->
+          value = %{
+            datetime: dt,
+            token_age_in_days: if(tv >= 0.1, do: br / tv, else: 0) * 15 / 86400
```
tspenov: why `15` here ? Can we extract the formula in private function and explain ?

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -48,8 +47,43 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
   end
 
   @doc ~S"""
-    Return the transaction volume for the given slug and time period.
-    Uses the influxdb cached values instead of issuing a GET request to etherbi
+  Return the average age of the tokens that wer transaction for the given slug and time period.
```
tspenov: `that were transacted`

lib/sanbase/blockchain/burn_rate.ex
IvanIvanoff: Yes, private function + explanation will be better.
We do this approximation because the average ethereum block is created every 15 seconds. The burn rate is amount multiplied by blocks, so we need to do this in order to get the age in days. We won't have this problem though with the new approach as we'll have timestamps

PR #905, opened 2019-01-08
Switch GitHub api. Remove sanbase-workers

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/utils/date_time_utils.ex
```diff
@@ -56,6 +56,7 @@ defmodule Sanbase.DateTimeUtils do
   end
 
   def compound_duration_to_seconds(interval) do
+    IO.inspect(interval, label: "IASNASLDNALSDHJLASJHDLASJLSJALJA")
```
tspenov: :D leftover

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -1,7 +1,4 @@
 defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
-  alias Sanbase.Clickhouse.Github
-  alias Sanbase.Model.Project
```
tspenov: why remove the aliases ?

IvanIvanoff: I tend to not use `alias` when the module is used only once and does not extend the line width

PR #917, opened 2019-01-10
Daa signal realtime

tspenov:
Signal is run every 5 minutes. If notifications is triggered - adds a row - project, notif_type, data=current_daa_value. Next check checks for the last value that is triggered today for the project and substracts it from the current value. 
Caching the averages for the last 30 days for the current day so we don't calculate these.

### DIFF
--- a/lib/sanbase/clickhouse/erc20_daily_active_addresses.ex
+++ b/lib/sanbase/clickhouse/erc20_daily_active_addresses.ex
@@ -10,7 +10,7 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
 
   @type contracts :: String.t() | list(String.t())
 
-  @type contract_daa_tupple :: {
+  @type contract_daa_tuple :: {
           String.t(),
           non_neg_integer()
         }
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/erc20_daily_active_addresses.ex
```diff
@@ -72,6 +72,31 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
     end)
   end
 
+  @doc ~s"""
+  Gets the current value for active addresses for today.
+  Returns a list of tuples {contract, active_addresses}
+  """
+
```
IvanIvanoff: empty line

IvanIvanoff:
The only thing that kind of bothers me is that the signal will never be triggered in the first few hours after 00:00 UTC (probably there will be some extreme cases). But before these changes it was always triggered the next day, so it's ok probably

tspenov:
@IvanIvanoff yeah it will mostly notify towards the end of the day (UTC). But still better than what we have now. We can iterate with new ideas on top of this.

PR #923, opened 2019-01-14
Lowercase the exchange addresses taken from postgres

IvanIvanoff:
All clickhouse addresses are lowercase, but we have addresses in
postgres that could contain upper letters. This leads to wrong
comparisons.

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


tspenov:
Shouldn't we also include a migration that downcases everything in exchange_addresses ?

IvanIvanoff:
@tspenov no need to this now. This will fix the current issue and I've added a card for migrating the data but let's not include it in the current sprint.

IvanIvanoff:
Reason for not needing to migrate the data is because this `MarkExchangeAddresses` service is the only one that communicates with the DB - the API does not.

PR #932, opened 2019-01-15
Add user settings

tspenov:
#### Summary

response `user_settings`
```
{
signalNotifyTelegram
signalNotifyEmail
user {
  email
  ...
}
}
```

```graphql
    mutation {
      settingsToggleChannel(signal_notify_telegram: true) {
        signalNotifyTelegram
      }
    }
```
response - `user_settings`

```graphql
    mutation {
      settingsToggleChannel(signal_notify_email: true) {
        signalNotifyEmail
      }
    }
```
response - `user_settings`

telegram integration - here: https://github.com/santiment/sanbase2/pull/937


IvanIvanoff:
I would prefer to have an argument for setting it to true/false, so the query depends only on the arguments and not the DB state

tspenov:
hmm, maybe that will be better - also combine the first 2 in one query with proper argument

IvanIvanoff:
I am not sure we'll be needing `telegramUrl`. The way the telegram notifications are going to work is in the following way:
1.) For the user with id X we generate a long random string Y. Persist the pair {X,Y}
2.) Generate a link `https://telegram.me/<bot_name>?start={Y}`
3.) Present this link to the user. When he/she clicks on it and then on `Start` in telegram, sanbase will receive a message and we'll match Y with X and save the `chat_id` - this will be destination we'll send messages to.

Bots cannot send messages to users, only to channels

lib/sanbase_web/graphql/schema/account_types.ex
```diff
@@ -25,6 +26,10 @@ defmodule SanbaseWeb.Graphql.AccountTypes do
     field :apikeys, list_of(:string) do
       resolve(&ApikeyResolver.apikeys_list/3)
     end
+
+    field :settings, :user_settings do
```
IvanIvanoff: `user_settings`
Are we going to use this for a wide variety of settings and not only notification settings? If not - maybe `user_notification_settings`

tspenov: my plan is to be used for all user related settings - not only for notification.

PR #937, opened 2019-01-16
Implement Telegram Bot functionality

IvanIvanoff:
#### Summary
API:
```
{ getTelegramDeepLink }
```

When this link is followed and the `Start` button is pressed in telegram, the sanbase and telegram accounts will be linked

--

Also, in #932 the user settings API was added - it includes the mutation:
```
mutation {
      settingsToggleChannel(signal_notify_telegram: true) {
        signalNotifyTelegram
      }
    }
```
Signals should be sent if the user has connected their telegram and sanbase accounts and enabled the telegram signals.

-- 
One additional field will be added to the user type - `hasTelegramConnected` or something similar

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -767,6 +768,17 @@ defmodule SanbaseWeb.Graphql.Schema do
...
     end
+
+    @desc """
+    Get a URL for deep-linking sanbase and telegram accounts. It carries a unique
+    random token that is associated with the user. The link leads to a telegram chat
+    with Santiment's notification bot. When the `Start` button is pressed, telegram
+    and sanbase accounts are linked and the user can receive sanbase signals in telegram.
+    """
+    field :get_telegram_deep_link, :string do
+      # middleware(JWTAuth)
```
valo: Why the JWTAuth token is commented out here?

lib/sanbase_web/router.ex
```diff
@@ -54,6 +54,14 @@ defmodule SanbaseWeb.Router do
   scope "/", SanbaseWeb do
     pipe_through(:browser)
 
+    # Only us and telegram know the token. This makes sure that no malicious party
+    # could counterfeit a request
+    post(
+      "/telegram/start/#{System.get_env("TELEGRAM_NOTIFICATIONS_TOKEN")}",
```
valo: This will be evaluated compile time. Is that ok?

lib/sanbase_web/router.ex
```diff
@@ -5,7 +5,7 @@ defmodule SanbaseWeb.Router do
     plug(:accepts, ["html"])
     plug(:fetch_session)
     plug(:fetch_flash)
-    plug(:protect_from_forgery)
+    # plug(:protect_from_forgery)
```
valo: Why is this commented out?

lib/sanbase_web/graphql/resolvers/user/account_resolver.ex
```diff
@@ -31,6 +31,17 @@ defmodule SanbaseWeb.Graphql.Resolvers.AccountResolver do
...
   def current_user(_root, _args, _context), do: {:ok, nil}
 
+  def telegram_deep_link(_root, _args, %{context: %{auth: %{current_user: user}}}) do
+    link = Sanbase.Telegram.get_or_create_start_link(user.id)
+    {:ok, link}
+  end
+
+  def telegram_deep_link(_root, _args, _resolution) do
+    # {:error, "Only logged in users can call this method"}
+    link = Sanbase.Telegram.get_or_create_start_link(1)
```
valo: This looks like some debug code

lib/sanbase_web/router.ex
IvanIvanoff: @valo it has some testing and local development stuff left - they will be deleted before removing the WIP label

lib/sanbase/telegram/telegram.ex
```diff
@@ -0,0 +1,82 @@
...
+      %{
+        parse_mode: "markdown",
+        chat_id: chat_id,
+        text: text
+      }
+      |> Jason.encode!(),
+      [{"Content-Type", "application/json"}]
+    )
+    |> case do
+      {:ok, %HTTPoison.Response{status_code: 200}} -> :ok
```
tspenov: We must also test the limits  and at some point also `429` has to be handled: https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this

lib/sanbase_web/controllers/telegram_controller.ex
```diff
@@ -0,0 +1,20 @@
...
+    Telegram.store_chat_id(user_token, chat_id)
+
+    conn
+    |> resp(200, "ok")
+    |> send_resp()
+  end
+
+  def start(conn, _params) do
+    conn
+    |> resp(200, "ok")
```
tspenov: Should it return `200` on unknown params - probably 404 ?

IvanIvanoff: The user can enter an arbitrary text in the chat with the bot, we need to just ignore it

lib/sanbase_web/router.ex
IvanIvanoff: nope, I will research how to use REPLACE_OS_VARS as we cannot use the `Config` module here for some reason

lib/sanbase/telegram/telegram.ex
IvanIvanoff: Good point. I've addressed that by changing to Tesla and adding the rate limiting middleware

lib/sanbase_web/graphql/schema.ex
```diff
@@ -769,6 +770,17 @@ defmodule SanbaseWeb.Graphql.Schema do
...
```
tspenov: that is still commented although it is marked as resolved

PR #939, opened 2019-01-18
Add a mutation to add/remove an ethereum address to already signed in user

IvanIvanoff:
#### Summary
```
{
  mutation{
    addUserEthAddress(signature: "<signature>", address: "<address>", messageHash: "<hash>"){
      ethAccounts{ address }
  }
}
```
This mutation accepts the same parameters as the metamask login. It must be executed for a signed in user and instead of login in, it will add the ethereum address to the user. This can be used to add metamask to an account that is logged in with an email. Note that every input that provides 3 correct parameters can be added, metamask only gives an easy way to sign a message.

---

```
{ mutation { removeUserEthAddress(address: "SomeAddress") { ethAccounts{ address } } } }
```

Remove an etehreum address from an account. This will only work if the user has either an email set or another ethereum address added.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/auth/user.ex
```diff
@@ -327,6 +328,23 @@ defmodule Sanbase.Auth.User do
     end
   end
 
+  def add_eth_account(%User{id: user_id}, address) do
+    EthAccount.changeset(%EthAccount{}, %{user_id: user_id, address: address})
```
tspenov: Lets also add unique constraint in the changeset: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/auth/eth_account.ex#L18

lib/sanbase_web/graphql/schema.ex
```diff
@@ -828,6 +828,31 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+      arg(:signature, non_null(:string))
+      arg(:address, non_null(:string))
+      arg(:message_hash, non_null(:string))
+
+      middleware(JWTAuth)
+      resolve(&AccountResolver.add_user_eth_address/3)
+    end
+
+    @desc ~s"""
+    Add the given `address` for the currently logged in user.
```
tspenov: `Remove` instead of `Add`

PR #941, opened 2019-01-21
Retry the same query in case of timeout instead of crashing the process

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap/graph_data.ex
+++ b/lib/sanbase/external_services/coinmarketcap/graph_data.ex
@@ -128,13 +128,13 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
     |> case do
       {:ok, %Tesla.Env{status: 429} = resp} ->
         wait_rate_limit(resp)
-        Process.exit(self(), :normal)
+        fetch_all_time_prices(coinmarketcap_id)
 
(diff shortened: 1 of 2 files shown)


lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -128,13 +128,13 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
     |> case do
       {:ok, %Tesla.Env{status: 429} = resp} ->
         wait_rate_limit(resp)
-        Process.exit(self(), :normal)
+        fetch_all_time_prices(coinmarketcap_id)
```
tspenov: I am thinking that generic `retry` mechanism will be better -  function that that takes func and args and tries for ex. 5 times and then gives up.  Using the `Registry`  as the global state to store `pid_module_func` and count. That way we won't bother for stuff that is always failing. Is it too much for this case ?

IvanIvanoff: In the case of rate limit it makes more sense to implement it with recursion, otherwise again after 5 fails we don't know what result to return. I reworked it because this was returning `:ok` and `Enum.flat_map/2` was failing

PR #949, opened 2019-01-23
Create custom tag insights

tspenov:
Create custom tag when creating insight

### DIFF
--- a/lib/sanbase/voting/post.ex
+++ b/lib/sanbase/voting/post.ex
@@ -148,7 +148,15 @@ defmodule Sanbase.Voting.Post do
 
   # Helper functions
   defp tags_cast(changeset, %{tags: tags}) do
-    tags = Tag |> where([t], t.name in ^tags) |> Repo.all()
+    params =
+      tags
+      |> Enum.map(fn tag -> %{name: tag} end)
+
+    Repo.insert_all(Tag, params, on_conflict: :nothing)
+
+    tags =
+      from(t in Tag, where: t.name in ^tags)
+      |> Repo.all()
 
     changeset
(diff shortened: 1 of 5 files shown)


lib/sanbase/voting/post.ex
```diff
@@ -148,7 +148,15 @@ defmodule Sanbase.Voting.Post do
...
-    tags = Tag |> where([t], t.name in ^tags) |> Repo.all()
+    params =
+      tags
+      |> Enum.map(fn tag -> %{name: tag} end)
+
+    Repo.insert_all(Tag, params, on_conflict: :nothing)
+
+    tags =
+      from(t in Tag, where: t.name in ^tags)
+      |> Repo.all()
```
IvanIvanoff: in case we are inserting the missing tags we no longer need to filter them here

IvanIvanoff: Also, should only insight creation be able to create tags? It will be kind of confusing to write an insight about trending words and needing to create the tag yourself. If this happens then how we'll check on the frontend when an insight is connected to a trending word, for example?

tspenov: 1. I am not checking for insert result so if smth goes wrong I just take these tags that are in the db and add in assoc. And I also need `%Tag{}` structs with id for the `assoc`.
2. @Partysun is adding custom tags on `createPost` query enough ? 
I am not sure what the flow will be exactly. I imagine on the trending words page, clicking `Create Insight` and going to `insight/new/` where you've remebered that you've come from the trending words `Create Insight` button and adding the tag automatically when creating insight.

PR #951, opened 2019-01-24
Make the histogram instrumenter buckets linear

IvanIvanoff:
When they were exponential the buckets were 100ms, 1s, 10s, 100s and
this is not very informative about the query times really

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/prometheus/histogram_instrumenter.ex
```diff
@@ -1,5 +1,13 @@
 defmodule SanbaseWeb.Graphql.Prometheus.HistogramInstrumenter do
   use AbsintheMetrics,
     adapter: AbsintheMetrics.Backend.PrometheusHistogram,
-    arguments: [buckets: {:exponential, 10, 2, 12}]
+    arguments: [
```
tspenov: Can you add a comment what these numbers represent ?

PR #955, opened 2019-01-24
Small optimizations

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/middlewares/api_timeframe_restriction.ex
```diff
@@ -13,6 +13,12 @@ defmodule SanbaseWeb.Graphql.Middlewares.ApiTimeframeRestriction do
   alias Sanbase.Auth.User
 
   @allow_access_without_staking ["santiment"]
+  @required_san_stake_full_access Config.get(:required_san_stake_full_access)
+                                  |> String.to_integer()
+
+  @restrict_to_in_days -1 * (Config.get(:restrict_to_in_days) |> String.to_integer())
+
+  @restrict_from_in_days -1 * (Config.get(:restrict_from_in_days) |> String.to_integer())
```
valo: Isn't it a problem that these will be evaluated during compile time? Isn't it possible to tweak them through ENV vars?

IvanIvanoff: true, the question is do we want to be able to tweak them without recompiling (probably could happen once in a few months) or do this small optimization and do not read it from the config on each query.

tspenov: Does it really save any time cause these calls will be quite fast I guess. Also we might not use it often but its good pattern to be able to control parts of live system without recompiling.

valo: Is there evidence that taking the values during runtime is a performance problem?

IvanIvanoff: I doubt this has a noticeable difference

tspenov:
👍 for the postgres dataloader, but maybe we keep configuration as it is now

IvanIvanoff:
@valo @tspenov removed the config changes

PR #956, opened 2019-01-25
Implement Trigger internal structure

tspenov:
Internal structure of signal triggers.


lib/sanbase/signals/trigger.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Sanbase.Signals.Trigger do
```
IvanIvanoff: Also we probably should have public/private indicator for a signal. It makes sense to have it on the top level along the `:trigger` field

lib/sanbase/signals/trigger.ex
```diff
@@ -0,0 +1,13 @@
...
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  embedded_schema do
+    field(:trigger, :map)
```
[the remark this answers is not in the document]
tspenov: added `validate_required` in the changeset

lib/sanbase/signals/trigger.ex
tspenov: Added `is_public` with default `false` in the emb schema.

lib/sanbase/signals/trigger/daa_trigger.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Signals.Trigger.DaaTrigger do
```
IvanIvanoff: `DailyActiveAddressesTrigger` would be better

lib/sanbase/signals/trigger.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Sanbase.Signals.Trigger do
```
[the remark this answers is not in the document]
tspenov: @IvanIvanoff  Lets try to follow this:

* The model/schema/resource I think should be singular
* db table plural
* for the intermediate folders of context -> plural most of the time (like `signals`)

quote from:
https://hexdocs.pm/phoenix/contexts.html#adding-an-accounts-context

> Naming things is hard. If you’re stuck when trying to come up with a context name when the grouped functionality in your system isn’t yet clear, you can simply use the plural form of the resource you’re creating. For example, a Users context for managing users. As you grow your application and the parts of your system become clear, you can simply rename the context to a more refined name at a later time.

I'll rename `UserTriggers` -> `UserTrigger` and `UserSettings` -> `UserSetting` following this.

tspenov:
@IvanIvanoff added

lib/sanbase/signals/user_trigger.ex
```diff
@@ -0,0 +1,191 @@
...
+    end
+  end
+
+  @spec get_trigger(%User{}, String.t()) :: map()
+  def get_trigger(%User{id: _user_id} = user, trigger_id) do
+    triggers_for(user)
+    |> find_trigger_by_id(trigger_id)
+  end
+
+  @spec create_trigger(%User{}, map()) :: {:ok, list(trigger_struct)} | {:error, String.t()}
```
IvanIvanoff: Any particular reason why this is returning a list of trigger structs?

IvanIvanoff:
@tspenov if I pass `:cooldown` to `create_trigger` function I get an error.

tspenov:
@IvanIvanoff `cooldown` should be passed in the outer map: 
`UserTrigger.create_trigger(user, %{is_public: true, trigger: trigger, cooldown: 3600})`

lib/sanbase/signals/user_trigger.ex
```diff
@@ -0,0 +1,191 @@
...
+    end
+  end
+
+  def create_trigger(_, _), do: {:error, "Trigger structure is invalid"}
+
+  @spec update_trigger(%User{}, map()) :: {:ok, list(trigger_struct)} | {:error, String.t()}
+  def update_trigger(%User{id: user_id} = user, %{id: id} = params) do
+    trigger_data = Map.get(params, :trigger)
+
+    if is_nil(trigger_data) || is_valid?(trigger_data) do
```
IvanIvanoff: What's the idea behind this check? Shouldn't it be `not is_nil(..) and is_valid?(..)`?

tspenov: we have params like this: 
```elixir
%{
trigger: %{ ... }
is_public: ...,
cooldown: ..
}
```
where `trigger` is the `json/map` and some top level fields like `is_public` , etc
if `trigger` map is empty I still want to update  the top level params. If its present I want to check the structure of the `json/map`.
I drop the params that are `nil` using : https://github.com/santiment/sanbase2/pull/956/files#diff-0bfb76c7d4ebf51a58200169f20fa40bR185

lib/sanbase/signals/user_trigger.ex
```diff
@@ -0,0 +1,191 @@
...
+      end
+    else
+      {:error, "Trigger structure is invalid"}
+    end
+  end
+
+  def create_trigger(_, _), do: {:error, "Trigger structure is invalid"}
+
+  @spec update_trigger(%User{}, map()) :: {:ok, list(trigger_struct)} | {:error, String.t()}
+  def update_trigger(%User{id: user_id} = user, %{id: id} = params) do
```
IvanIvanoff: Is it possible to update a trigger only by the trigger id, which is already unique?

lib/sanbase/signals/user_trigger.ex
tspenov: no, it should be maps on `create` , `update`, struct on `triggers_for`

lib/sanbase/signals/user_trigger.ex
```diff
@@ -0,0 +1,191 @@
...
+  alias Sanbase.Signals.Trigger
+  alias Sanbase.Repo
+  alias Sanbase.Signals.Trigger.{DailyActiveAddressesTrigger, PriceTrigger}
+
+  @type trigger_struct :: %Trigger{}
+  @type trigger_data_strct :: %DailyActiveAddressesTrigger{} | %PriceTrigger{}
+
+  schema "user_triggers" do
+    belongs_to(:user, User)
+    embeds_many(:triggers, Trigger)
```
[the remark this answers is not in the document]
tspenov: ok, lets discuss this. With embeds many its hard to implement but we have most of it now.

lib/sanbase/signals/user_trigger.ex
IvanIvanoff: I mean does it need to return `trigger1`, `trigger2`, etc. when creating `triggerN`

tspenov: now returns only the created trigger

lib/sanbase/signals/user_trigger.ex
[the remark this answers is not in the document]
tspenov: changed with `embeds_one` and simplified logic

lib/sanbase/signals/user_trigger.ex
tspenov: the embedded schema has to be updated through the parrent via cast_embed or put_embed, so it has to go through `user_triggers` for which we need the user.

lib/sanbase/signals/trigger.ex
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  embedded_schema do
+    field(:trigger, :map)
```
IvanIvanoff: This is probably better to be called `settings` (or something similar). Currently when I try to fetch all UserTriggers where the type is something specific the SQL looks like this:
` trigger->'trigger'->'type' = ?` 
it would be more understandable if it is `trigger->'settings'->'type' = ?`

tspenov: yeah `trigger.trigger` is not good but didn't have good name for it. `settings` seems good - will change it

PR #958, opened 2019-01-28
Custom signals trigger evaluator

IvanIvanoff:
#### Summary
Basic idea: `Evaluator.run` gets a list of triggers and returns a sublist of it, consisting of the triggers which should send a signal.
Each trigger generates a cache_key which includes only the params that determine the outcome (hence ignoring the channel, if it is repeating, etc.). This cache_key is used to execute the triggers with same params only once. This is achieved by using `ConCache.get_or_store` where if 2+ processes try to access the data for the same cache_key only one of them will run the function that calculates it and the other will wait. 

Further optimizations are possible, though they are not needed at the current scale. Such optimizations are for example:
- embedding triggers: If trigger X needs the mean value for 30 days and trigger Y - for 60 days, we can fetch the average for each day for 60 days and each trigger will take the average that it needs

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -0,0 +1,56 @@
...
+  defp has_cooldown?(%{cooldown: cd, last_triggered: %DateTime{} = lt}) do
+    Timex.compare(
+      Timex.shift(lt, minutes: cd),
+      Timex.now()
+    ) == 1
+  end
+
+  defp filter_triggered({_type, []}), do: []
+
+  defp filter_triggered({_type, triggers}) do
```
IvanIvanoff: This could be reworked to use protocols instead of behaviors and we won't care to know the module name, but use `ProtocolName.triggered?`

lib/sanbase/utils/parallel.ex
```diff
@@ -58,4 +77,17 @@ defmodule Sanbase.Parallel do
...
+
+  def pfilter_concurrent(collection, func, opts \\ []) when is_function(func, 1) do
+    filter_func = fn x -> {func.(x), x} end
+
+    pmap_concurrent(collection, filter_func, opts)
+    |> Enum.filter(fn {bool, _item} -> bool === true end)
+    |> Enum.map(fn {_bool, item} -> item end)
+  end
+
+  def preject_concurrent(collection, func, opts \\ []) when is_function(func, 1) do
```
valo: This function doesn't seem to be used

lib/sanbase/utils/parallel.ex
```diff
@@ -58,4 +77,17 @@ defmodule Sanbase.Parallel do
         end)
     end
   end
+
+  def pfilter_concurrent(collection, func, opts \\ []) when is_function(func, 1) do
```
valo: Do we really need such helper? This operation is done in one place at the moment and I feel like the use case might not be super reusable.

lib/sanbase/signals/evaluator/evaluator.ex
valo: Looks to me that using a protocol here might be the better option. Dynamic module names don't look like a good idea to me. With protocols we should have more compile time checks

lib/sanbase/utils/parallel.ex
valo: Also it could be confusing what is the difference between `pfilter` and `pfilter_concurrent` (parallel filter concurrent?)

lib/sanbase/signals/evaluator/evaluator.ex
tspenov: Also for protocols for `cache_key` and `triggered` and whatever common

lib/sanbase/signals/trigger/daily_active_addresses_trigger.ex
```diff
@@ -5,8 +5,79 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesTriggerSettings do
...
+        [trigger.type, trigger.target, trigger.time_window, trigger.percent_threshold]
+        |> Jason.encode!()
+
+      :crypto.hash(:sha256, data)
+      |> Base.encode16()
+    end
+  end
+
+  defimpl String.Chars, for: DailyActiveAddressesTriggerSettings do
+    def to_string(%{} = trigger) do
```
tspenov: I think only `trigger`  param is not enough to construct the payload ? For example here you might want to show the current DAA value (which is calculated somewhere before)

lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -0,0 +1,37 @@
...
+  alias Sanbase.Signals.{UserTrigger, Trigger}
+
+  @doc ~s"""
+  Takes a list of triggers and returns its subset that evaluate to true at the given moment.
+  """
+  @spec run(list()) :: list()
+  def run(user_triggers) do
+    user_triggers
+    |> remove_triggers_on_cooldown()
+    |> Sanbase.Parallel.pfilter_concurrent(
```
tspenov: I think we need to pass down a `map` with some values calculated by evaluator. In DAA case these are `current_daa`, `avg_daa`, `percent_change` and maybe more. So this can be a `pmap`, then a filter on top of it but passing down the trigger and the map with evaluated values ?

IvanIvanoff: @tspenov one possible idea is to add a new struct that is representing a message/signal - it will have channel(email, telegram, etc) and payload, so the next step can just take it and send it.

tspenov: sounds good 👍

lib/sanbase/signals/trigger/daily_active_addresses_trigger.ex
```diff
@@ -5,8 +5,79 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesTriggerSettings do
...
+
+  alias __MODULE__
+  alias Sanbase.Model.Project
+  alias Sanbase.Clickhouse.{Erc20DailyActiveAddresses, EthDailyActiveAddresses}
+  alias Sanbase.Signals.Evaluator.Cache
+
+  def type(), do: @trigger_type
+
+  def get_data(trigger) do
+    {:ok, contract, _token_decimals} = Project.contract_info_by_slug(trigger.target)
```
tspenov: `trigger.target` can be also a list -> so we should use `List.wrap` and work with it like its a list

lib/sanbase/signals/trigger/daily_active_addresses_trigger.ex
```diff
@@ -5,8 +5,79 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesTriggerSettings do
...
+      Cache.get_or_store("daa_#{contract}_current", fn ->
+        average_daily_active_addresses(
+          contract,
+          Timex.shift(Timex.now(), days: -1),
+          Timex.now()
+        )
+      end)
+
+    average_daa =
+      Cache.get_or_store("daa_#{contract}_prev_#{to_string(trigger.time_window)}", fn ->
```
tspenov: I think `trigger.time_window` should be a string like `'1h'`, `'30d'` etc and use https://github.com/santiment/sanbase2/blob/master/lib/sanbase/utils/datetime/utils.ex#L58 to convert it to seconds.
Also I think in the frontend we should specify limited options for every signal - like for DAA `1d`, `1w`, `30d` - so we have most values in cache

IvanIvanoff: No point in converting `time_window` for the cache key. Though I can remove the `to_string`. You can check the new version I just committed.

lib/sanbase/signals/trigger/daily_active_addresses_trigger.ex
IvanIvanoff: I don't think we should allow lists here, it will complicate the things a lot with no visible benefits. Also it will introduce a lot of ambiguities - is a list `OR` or `AND` . And how to calculate cool downs? I think if you want a list you need to compose signals - and then the user has power to choose `AND`, `OR`, etc.

tspenov: In the notion of the trigger I think it should be `OR`. To achieve what we have now with signals in discord - you should be able to choose for target all projects and send signal for those that fulfill the trigger conditions. I don't think we'll be able to achieve that wit composability

tspenov: of course we can start with this version and upgrade later

lib/sanbase/signals/trigger/daily_active_addresses_trigger.ex
```diff
@@ -1,12 +1,118 @@
...
+
+    def evaluate(%DailyActiveAddressesSettings{} = settings) do
+      {current_daa, average_daa} = DailyActiveAddressesSettings.get_data(settings)
+
+      case percent_change(average_daa, current_daa) >= settings.percent_threshold do
+        true ->
+          %DailyActiveAddressesSettings{
+            settings
+            | triggered?: true,
+              payload: payload(settings, current_daa, average_daa)
```
tspenov: we're updating these temporarily here -> do you think it makes sense to update them permanently at the and with `last_triggered` ?

IvanIvanoff: @tspenov Do you mean to persist the last message that was sent?

tspenov: yes, but it doesn't make much sense

IvanIvanoff: Even if we persist them now we won't use them for anything. We can revisit this idea once the need for it arises.

IvanIvanoff:
@tspenov I'll update the payload and see how to handle the embeds, in the current version it won't work

tspenov:
@IvanIvanoff Though some failing tests

PR #960, opened 2019-01-29
Add min_volume optional param to all_* queries

IvanIvanoff:
#### Summary
```
{
  allProjects(minVolume: 1000000000){
    name
    volumeUsd
  }
}
```

Response:
```
{
  "data": {
    "allProjects": [
      {
        "name": "Bitcoin",
        "volumeUsd": 5921087113.58367
      },
      {
        "name": "Ethereum",
        "volumeUsd": 2934782296.40992
      },
      {
        "name": "Tether",
        "volumeUsd": 4255229740.47791
      }
    ]
  }
}
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


Partysun:
@IvanIvanoff It's better to have the same filter option for watchlists.

lib/sanbase/model/project/project_list.ex
```diff
@@ -1,155 +1,212 @@
...
-  alias Sanbase.Model.{Project, Infrastructure}
+  alias Sanbase.Model.{Project, Infrastructure, LatestCoinmarketcapData}
 
   @preloads [
     :eth_addresses,
     :latest_coinmarketcap_data,
     icos: [ico_currencies: [:currency]]
   ]
 
+  defguard valid_volume(volume) when is_number(volume) and volume >= 0
```
tspenov: the convention for defguard is to start with `is_` so `is_valid_volume`

lib/sanbase/model/project/project_list.ex
```diff
@@ -1,155 +1,212 @@
...
-    projects_query()
+  def projects_count(min_volume) do
+    projects_query(min_volume)
     |> select([p], fragment("count(*)"))
     |> Repo.one()
   end
 
-  defp projects_query() do
-    from(p in Project, where: not is_nil(p.coinmarketcap_id), preload: ^@preloads)
+  defp projects_query(nil) do
```
tspenov: can we just make it `projects_query()` with no arguments ? Passing `nil` seems kinda useless to me.

IvanIvanoff: If there is no param then the dispatching between the query with and without `min_volume` will be really clumsy (using case, etc.). An option is to add make the default parameter `nil` but given that it's a private function it will lead to more code and won't make thing more clear as you need to check the default params when reading it

IvanIvanoff: Note that this default `min_volume` param is made in the public function, so the users of this module won't known about these internals

lib/sanbase/model/project/project_query.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Sanbase.Model.ProjectQuery do
+  defmacro coalesce(left, right) do
+    quote do
+      fragment("COALESCE(?, ?)", unquote(left), unquote(right))
+    end
+  end
+
+  defmacro volume_above(_, nil) do
```
tspenov: Are we using those 2 implementations ?

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -102,14 +102,40 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+
+  defp user_list_filter_min_volume(user_list, nil), do: user_list
+
+  defp user_list_filter_min_volume(user_list, min_volume) do
+    %{list_items: list_items} = user_list
+
+    list_items =
+      list_items
+      |> Enum.each(fn x ->
+        IO.inspect(x)
```
tspenov: Inspect left

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -102,14 +102,40 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+      {:ok, user_list}
+    else
+      error -> error
+    end
+  end
+
+  defp user_list_filter_min_volume(user_list, nil), do: user_list
+
+  defp user_list_filter_min_volume(user_list, min_volume) do
+    %{list_items: list_items} = user_list
```
tspenov: you can pattern match in function header `defp user_list_filter_min_volume(%{list_items: list_items} = user_list, min_volume) do`

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -102,14 +102,40 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+
+    list_items =
+      list_items
+      |> Enum.each(fn x ->
+        IO.inspect(x)
+      end)
+
+    %{user_list | list_items: list_items}
```
tspenov: Is that function doing anything?  it updates with the same `list_items`

tspenov:
@IvanIvanoff I think we need basic tests for public project queries with minVolume and userLists with minVolume. We can add it them in separate PR (if queries without minVolume are covered and working)

lib/sanbase/model/project/project_query.ex
IvanIvanoff: nope

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
IvanIvanoff: I'll remove all min_volume from watchlists for now and will implement them later

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
IvanIvanoff: I removed all code related to watchlist. Now I removed that, too. This will be done in a separate PR

IvanIvanoff:
@tspenov I've added tests

test/sanbase_web/graphql/projects/project_api_min_volume_test.exs
```diff
@@ -0,0 +1,340 @@
...
+      query = """
+      {
+        allProjects{
+          name,
+        }
+      }
+      """
+
+      result =
+        context.conn
```
kalkov: You can extract this to a function.

test/sanbase_web/graphql/projects/project_api_min_volume_test.exs
```diff
@@ -0,0 +1,340 @@
...
+        |> post("/graphql", query_skeleton(query, "allProjects"))
+        |> json_response(200)
+
+      assert result["data"]["allProjects"] |> length == context.projects_with_volume_count
```
kalkov: Can we also assert which projects are included?

test/sanbase_web/graphql/projects/project_api_min_volume_test.exs
```diff
@@ -0,0 +1,340 @@
...
+        }
+      }
+      """
+
+      result =
+        context.conn
+        |> post("/graphql", query_skeleton(query, "allErc20Projects"))
+        |> json_response(200)
+
+      assert result["data"]["allErc20Projects"] |> length ==
```
kalkov: Can we also assert which projects are included?

test/sanbase_web/graphql/projects/project_api_min_volume_test.exs
```diff
@@ -0,0 +1,340 @@
...
+        }
+      }
+      """
+
+      result =
+        context.conn
+        |> post("/graphql", query_skeleton(query, "allCurrencyProjects"))
+        |> json_response(200)
+
+      assert result["data"]["allCurrencyProjects"] |> length ==
```
kalkov: Can we also assert which projects are included?

test/sanbase_web/graphql/projects/project_api_min_volume_test.exs
```diff
@@ -0,0 +1,340 @@
...
+  alias Sanbase.Model.{
+    Project,
+    Infrastructure
+  }
+
+  alias Sanbase.Repo
+  alias Sanbase.Model.LatestCoinmarketcapData
+  import SanbaseWeb.Graphql.TestHelpers
+
+  setup do
```
kalkov: I was going to propose to use `setup_all` because your tests are only requesting data and not creating or editing it but I could not make it work.

```
 12) Sanbase.Graphql.ProjectApiMinVolumeQueriesTest: failure on setup_all callback, all tests have been invalidated
     ** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.915.0>.
     
     When using ownership, you must manage connections in one
     of the four ways:
     
     * By explicitly checking out a connection
     * By explicitly allowing a spawned process
     * By running the pool in shared mode
     * By using :caller option with allowed process
     
     The first two options require every new process to explicitly
     check a connection out or be allowed by calling checkout or
     allow respectively.
     
     The third option requires a {:shared, pid} mode to be set.
     If using shared mode in tests, make sure your tests are not
     async.
     
     The fourth option requires [caller: pid] to be used when
     checking out a connection from the pool. The caller process
     should already be allowed on a connection.
     
     If you are reading this error, it means you have not done one
     of the steps above or that the owner process has crashed.
     
     stacktrace:
       (db_connection) lib/db_connection.ex:934: DBConnection.checkout/2
       (db_connection) lib/db_connection.ex:750: DBConnection.run/3
       (db_connection) lib/db_connection.ex:798: DBConnection.transaction/3
       (ecto) lib/ecto/repo/schema.ex:125: Ecto.Repo.Schema.insert!/4
       test/sanbase_web/graphql/projects/project_api_min_volume_test.exs:17: Sanbase.Graphql.ProjectApiMinVolumeQueriesTest.__ex_unit_setup_all_1/1
       test/sanbase_web/graphql/projects/project_api_min_volume_test.exs:1: Sanbase.Graphql.ProjectApiMinVolumeQueriesTest.__ex_unit__/2
```

IvanIvanoff: @kalkov in the ConnCase definition only `setup` is handled right now. Though if there is no significant performance benefit I prefer to keep the tests as separate as possible

IvanIvanoff: PS: There should not be a great performance benefit as everything is executed in a transaction, so it stays in memory in the database and is never persisted

PR #964, opened 2019-02-01
Use dataloader watchlists projects

tspenov:
Fix n+1 queries by using dataloader for watchlists

### DIFF
--- a/lib/sanbase_web/graphql/schema/user_list_types.ex
+++ b/lib/sanbase_web/graphql/schema/user_list_types.ex
@@ -12,9 +12,14 @@ defmodule SanbaseWeb.Graphql.UserListTypes do
   end
 
   object :list_item do
-    field :project, :project do
-      resolve(&UserListResolver.project_by_list_item/3)
-    end
+    field(:project, :project,
+      resolve:
+        dataloader(SanbaseRepo, :project,
+          args: %{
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/schema/user_list_types.ex
```diff
@@ -12,9 +12,14 @@ defmodule SanbaseWeb.Graphql.UserListTypes do
...
 
   object :list_item do
-    field :project, :project do
-      resolve(&UserListResolver.project_by_list_item/3)
-    end
+    field(:project, :project,
+      resolve:
+        dataloader(SanbaseRepo, :project,
+          args: %{
+            preload: [:latest_coinmarketcap_data, icos: [ico_currencies: [:currency]]]
```
IvanIvanoff: `eth_addresses` also needs to be included. I propose you to add a function to `Project` that quest returns the `@preload` module attribute and use it.

tspenov: do we always need the same preloads for Project ? In that case I'll remove args and the conditional here: https://github.com/santiment/sanbase2/pull/964/files#diff-673fb1190da3e54dc78c0f3d0e6be103R18 and use the predefined preload the way you suggested

PR #965, opened 2019-02-01
Fetch the word context async and cache the result

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/schema/social_data_types.ex
+++ b/lib/sanbase_web/graphql/schema/social_data_types.ex
@@ -2,6 +2,7 @@ defmodule SanbaseWeb.Graphql.SocialDataTypes do
   use Absinthe.Schema.Notation
 
   alias SanbaseWeb.Graphql.Resolvers.SocialDataResolver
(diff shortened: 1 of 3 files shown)


tspenov:
Can you try locally with stage tech-indicators how faster it is and whether there is a problem with the async calls ?

lib/sanbase_web/graphql/schema/social_data_types.ex
```diff
@@ -10,7 +10,7 @@ defmodule SanbaseWeb.Graphql.SocialDataTypes do
 
   object :word_with_context do
     field :context, list_of(:word_context) do
-      resolve(&SocialDataResolver.word_context/3)
```
tspenov: pff, I always test without caching and then forget to bring it back :(

tspenov: ah this is the second level resolver - that's why it didn't have `cache_resolve`

tspenov: But if `trending_words` has top level `cache_resolve` - wouldn't it cache everything ?

PR #975, opened 2019-02-06
Signal trending words

tspenov:
#### Summary
Send trending words signal

Create trending words trigger.
`trigger_time`: ISO UTC string

```graphql
mutation {
      createTrigger(
        cooldown: "24h",
        settings: "{\"channel\":\"telegram\",\"trigger_time\":\"12:00:00\",\"type\":\"trending_words\"}"
      ) {
        id
        settings
      }
    }
```

#### Screenshots

image

### DIFF
--- a/config/notifications_config.exs
+++ b/config/notifications_config.exs
@@ -70,5 +70,9 @@ config :sanbase, Sanbase.Scheduler,
     price_absolute_change_sonar_singal: [
       schedule: "4-59/5 * * * *",
       task: {Sanbase.Signals.Scheduler, :run_price_absolute_change_signals, []}
+    ],
+    trending_words_sonar_singal: [
+      schedule: "@hourly",
+      task: {Sanbase.Signals.Scheduler, :run_trending_words_signals, []}
     ]
   ]
--- a/config/notifications_config.exs
+++ b/config/notifications_config.exs
@@ -72,7 +72,7 @@ config :sanbase, Sanbase.Scheduler,
       task: {Sanbase.Signals.Scheduler, :run_price_absolute_change_signals, []}
     ],
     trending_words_sonar_singal: [
-      schedule: "@hourly",
+      schedule: "*/5 * * * *",
       task: {Sanbase.Signals.Scheduler, :run_trending_words_signals, []}
     ]
   ]
--- a/lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
+++ b/lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
@@ -22,7 +22,6 @@ defmodule Sanbase.Signals.Trigger.TrendingWordsTriggerSettings do
     def triggered?(%TrendingWordsTriggerSettings{triggered?: triggered}), do: triggered
 
     def evaluate(%TrendingWordsTriggerSettings{} = trigger) do
-
       with true <- time_to_signal?(trigger),
            {:ok, payload} <- trigger_payload() do
         %TrendingWordsTriggerSettings{
--- a/lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
+++ b/lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
@@ -1,124 +1,121 @@
 defmodule Sanbase.Signals.Trigger.TrendingWordsTriggerSettings do
   @derive [Jason.Encoder]
   @trigger_type "trending_words"
-  @enforce_keys [:type, :channel, :trigger_time_iso_utc]
+  @trending_words_size 10
+  @trending_words_hours [1, 8, 14]
+  @enforce_keys [:type, :channel, :trigger_time]
 
   defstruct type: @trigger_type,
             channel: nil,
             # ISO8601 string time in UTC
-            trigger_time_iso_utc: nil,
+            trigger_time: nil,
             triggered?: false,
             payload: nil
 
+  use Vex.Struct
   import Sanbase.Utils.Math, only: [to_integer: 1]
+  import Sanbase.Signals.Utils
   alias __MODULE__
+
+  validates(:channel, inclusion: available_channels)
+  validates(:trigger_time, &__MODULE__.valid_trigger_time?/1)
 
   def type(), do: @trigger_type
 
-  defimpl Sanbase.Signals.Settings, for: TrendingWordsTriggerSettings do
-    @trending_words_size 10
-    @trending_words_hours [1, 8, 14]
+  def get_data(%TrendingWordsTriggerSettings{trigger_time: trigger_time}) do
+    trigger_time = Time.from_iso8601!(trigger_time)
 
+    if trigger_time.hour == Timex.now().hour do
+      get_today_top_words()
+    else
+      :error
+    end
+  end
+
+  def valid_trigger_time?(trigger_time) when is_binary(trigger_time) do
+    case Time.from_iso8601(trigger_time) do
+      {:ok, _time} ->
+        :ok
+
+      _ ->
+        {:error, "#{trigger_time} isn't a valid iso time"}
+    end
+  end
+
+  def valid_trigger_time?(_), do: :error
+
+  # private functions
+
+  defp get_today_top_words() do
+    {from, to, hour} = get_trending_word_query_params()
+
+    Sanbase.SocialData.trending_words(
+      :all,
+      @trending_words_size,
+      hour,
+      from,
+      to
+    )
+    |> case do
+      {:ok, [%{top_words: top_words}]} ->
+        {:ok, top_words}
+
+      error ->
+        :error
+    end
+  end
+
+  defp get_trending_word_query_params() do
+    now = Timex.now()
+
+    @trending_words_hours
+    |> Enum.filter(&(&1 < now.hour))
+    |> case do
+      [] ->
+        {
+          Timex.beginning_of_day(Timex.shift(now, days: -1)),
+          Timex.end_of_day(Timex.shift(now, days: -1)),
+          @trending_words_hours |> Enum.max()
+        }
+
+      hours ->
+        {
+          Timex.beginning_of_day(now),
+          Timex.end_of_day(now),
+          hours |> Enum.max()
+        }
+    end
+  end
+
+  defimpl Sanbase.Signals.Settings, for: TrendingWordsTriggerSettings do
     def triggered?(%TrendingWordsTriggerSettings{triggered?: triggered}), do: triggered
 
-    def evaluate(%TrendingWordsTriggerSettings{} = trigger) do
-      with true <- time_to_signal?(trigger),
-           {:ok, payload} <- trigger_payload() do
-        %TrendingWordsTriggerSettings{
-          trigger
-          | triggered?: true,
-            payload: payload
-        }
-      else
-        error ->
-          %TrendingWordsTriggerSettings{trigger | triggered?: false}
-      end
-    end
-
-    def cache_key(%TrendingWordsTriggerSettings{} = trigger) do
-      data =
-        [trigger.trigger_time_iso_utc]
-        |> Jason.encode!()
-
-      :crypto.hash(:sha256, data)
-      |> Base.encode16()
-    end
-
-    defp time_to_signal?(%TrendingWordsTriggerSettings{trigger_time_iso_utc: trigger_time}) do
-      case Time.from_iso8601(trigger_time) do
-        {:ok, time} ->
-          time.hour == Timex.now().hour
-
-        _ ->
-          false
-      end
-    end
-
-    defp get_trending_word_query_params() do
-      now = Timex.now()
-
-      @trending_words_hours
-      |> Enum.filter(&(&1 < now.hour))
-      |> case do
-        [] ->
-          {
-            Timex.beginning_of_day(Timex.shift(now, days: -1)),
-            Timex.end_of_day(Timex.shift(now, days: -1)),
-            @trending_words_hours |> Enum.max()
-          }
-
-        hours ->
-          {
-            Timex.beginning_of_day(now),
-            Timex.end_of_day(now),
-            hours |> Enum.max()
-          }
-      end
-    end
-
-    defp get_today_top_words() do
-      {from, to, hour} = get_trending_word_query_params()
-
-      Sanbase.SocialData.trending_words(
-        :all,
-        @trending_words_size,
-        hour,
-        from,
-        to
-      )
-      |> case do
-        {:ok, [%{top_words: top_words}]} ->
-          {:ok, top_words}
-
-        error ->
-          :error
-      end
-    end
-
-    defp get_max_len(top_words) do
-      top_words
-      |> Enum.map(fn tw -> String.length(tw.word) end)
-      |> Enum.max()
-    end
-
-    defp trigger_payload() do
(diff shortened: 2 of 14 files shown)


config/notifications_config.exs
```diff
@@ -66,5 +66,9 @@ config :sanbase, Sanbase.Scheduler,
     price_percent_change_sonar_singal: [
       schedule: "3-59/5 * * * *",
       task: {Sanbase.Signals.Scheduler, :run_price_percent_change_signals, []}
+    ],
+    trending_words_sonar_singal: [
+      schedule: "@hourly",
```
IvanIvanoff: Let's check with @ivanklimuk when exactly does a new set of trending words become available. If the calculation starts and round hours and takes 10 minutes, it makes more sense to schedule this for every `hh:20:00`

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+  def type(), do: @trigger_type
+
+  defimpl Sanbase.Signals.Settings, for: TrendingWordsTriggerSettings do
+    @trending_words_size 10
+    @trending_words_hours [1, 8, 14]
+
+    def triggered?(%TrendingWordsTriggerSettings{triggered?: triggered}), do: triggered
+
+    def evaluate(%TrendingWordsTriggerSettings{} = trigger) do
+      with true <- time_to_signal?(trigger),
```
IvanIvanoff: This way it will be hard to mock in tests. The approach in the other trigger settings is to have a generic `get_data` function in the signals module (not the implementation of the protocol) that accepts the settings and returns all data needed.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+
+  def type(), do: @trigger_type
+
+  defimpl Sanbase.Signals.Settings, for: TrendingWordsTriggerSettings do
+    @trending_words_size 10
+    @trending_words_hours [1, 8, 14]
+
+    def triggered?(%TrendingWordsTriggerSettings{triggered?: triggered}), do: triggered
+
+    def evaluate(%TrendingWordsTriggerSettings{} = trigger) do
```
IvanIvanoff: `trigger` -> `settings`

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
   @derive [Jason.Encoder]
   @trigger_type "trending_words"
-  @enforce_keys [:type, :channel, :time_window]
+  @enforce_keys [:type, :channel, :trigger_time_iso_utc]
+
   defstruct type: @trigger_type,
+            channel: nil,
             # ISO8601 string time in UTC
-            trigger_time_iso_utc: nil
+            trigger_time_iso_utc: nil,
```
IvanIvanoff: Can we just call this `trigger_time`?

config/notifications_config.exs
ivanklimuk: The calculation of new trending words takes up to 10 minutes, It is being calculated 3 times a day (for now, it is possible to run up to 24 calculations - 1 every hour)

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+      end
+    end
+
+    def cache_key(%TrendingWordsTriggerSettings{} = trigger) do
+      data =
+        [trigger.trigger_time_iso_utc]
+        |> Jason.encode!()
+
+      :crypto.hash(:sha256, data)
+      |> Base.encode16()
```
IvanIvanoff: I think we need to put this functionality in the utils (where the percent_change is). It will accept a list and do the encoding/hashing.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+    end
+
+    defp time_to_signal?(%TrendingWordsTriggerSettings{trigger_time_iso_utc: trigger_time}) do
+      case Time.from_iso8601(trigger_time) do
+        {:ok, time} ->
+          time.hour == Timex.now().hour
+
+        _ ->
+          false
```
IvanIvanoff: I think we should never allow this to happen by validating the params. In case of wrong datetime we'll just never trigger the signal and never find out that something is wrong.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+      end
+    end
+
+    defp get_trending_word_query_params() do
+      now = Timex.now()
+
+      @trending_words_hours
+      |> Enum.filter(&(&1 < now.hour))
+      |> case do
```
IvanIvanoff: `@trending_words_hours |> Enum.filter() |> Enum.max(fn -> @trending_words_hours |> Enum.max end)` is shorter, but I am not sure if it's clearer

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,138 @@
...
+          :error
+      end
+    end
+
+    defp build_payload(top_words, max_len) do
+      top_words_strings =
+        top_words
+        |> Enum.sort_by(fn tw -> tw.score end, &>=/2)
+        |> Enum.map(fn tw ->
+          ~s(#{String.pad_trailing(tw.word, max_len)} | #{to_integer(tw.score)})
```
IvanIvanoff: Isn't `~s/` more correct here, as we have `(` and `)` inside the sigil, but there is no `/`?

config/notifications_config.exs
tspenov: As user has supplied an hour in which he wants to receive the signal I think we have to respect that and send in that hour.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: ok, will change

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: yep - will move once your PR is merged

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: yeah will change - put it more descriptive so the frontend knows what needs to send

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: seems to work - but yeah will change the sigil delimiters

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: yeah validation will help here - its exactly why we need it so we don't do defensive programming all around the execution

config/notifications_config.exs
IvanIvanoff: @tspenov here we discuss the scheduler, not the user defined time. Also we'll have a warning that the trending words are generated at 01:15, 08:15 and 14:15

lib/sanbase/signals/validation.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Sanbase.Signals.Validation do
+  @notification_channels ["telegram", "email"]
+
+  def valid_notification_channels(), do: @notification_channels
+
+  def valid_percent?(percent) when is_float(percent), do: percent > 0
```
IvanIvanoff: hm, -10% is also a valid percentage, showing 10% decrease in price, for example:
image

mix.exs
```diff
@@ -107,7 +107,8 @@ defmodule Sanbase.Mixfile do
       {:libcluster, "~> 3.0"},
       {:number, "~> 0.5.7"},
       {:remote_ip, "~> 0.1"},
-      {:cachex, "~> 3.1"}
+      {:cachex, "~> 3.1"},
+      {:vex, "~> 0.8.0", override: true}
```
IvanIvanoff: What is the conflicting dependency here?

IvanIvanoff: elasticsearch, let's try it and hope it does not break

config/notifications_config.exs
tspenov: @IvanIvanoff @Partysun @valo 
Scheduling is tightly bound with UI options here. So lets decide what is the entire flow here ?
1. From UI what options for receiving the signal we present to user ?
All round hours: 9,10,11 ... OR only `01:15`, `08:15`, `14:15` with a message that trending words are generated then ?


2. When do we schedule sending -
All round hours or at 01:15, 08:15 and 14:15

mix.exs
tspenov: we're not that heavily using it except for the social stats and I didn't see any breaking changes in vex - so should be fine

config/notifications_config.exs
IvanIvanoff: @tspenov we can easily just schedule the signal to run every 5-15 minutes, it does not make any heavy computations so it should be fine.

lib/sanbase/signals/validation.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Sanbase.Signals.Validation do
+  @notification_channels ["telegram", "email"]
+
+  def valid_notification_channels(), do: @notification_channels
+
+  def valid_percent?(percent) when is_float(percent), do: true
+  def valid_percent?(_), do: false
+
+  def valid_time_window?(time_window) when is_binary(time_window) do
+    Regex.match?(~r/^\d+[mhdw]$/, time_window)
```
IvanIvanoff: Should we also allow `s` here?

lib/sanbase/signals/user_trigger.ex
```diff
@@ -132,6 +132,7 @@ defmodule Sanbase.Signals.UserTrigger do
 
   defp is_valid?(trigger) do
     with {:ok, trigger_struct} <- load_in_struct(trigger),
+         true <- Vex.valid?(trigger_struct),
```
IvanIvanoff: This is using the defined `validates` clauses in the different modules?

tspenov: yes. There is a test here: https://github.com/santiment/sanbase2/pull/975/files#diff-1870a349a85f59bea51031e0dbbd8b22R52 -> the channel is in module validates

lib/sanbase/signals/validation.ex
tspenov: I don't think we should  allow seconds for the `time_window` ?

IvanIvanoff: `3600s` is a valid equivalent of `1h` and `60m`. If we disallow seconds we need to implement handling such errors so it's clear why the API call failed with a valid interval. Anything less than `5m` (if the signal runs every 5 minutes) will be treated as 5 minutes

config/notifications_config.exs
tspenov: @IvanIvanoff that the frontend should always send with `1h` or smth < `24h` cooldown so we don't send more than once for that hour and for that day.

IvanIvanoff: Yeah, but that validation will be used for many signals that could trigger more than once a day

tspenov: And also it doesn't matter if its run every 5 minutes cause new data is ready in `01:15, 08:15 and 14:15`

lib/sanbase/signals/validation.ex
tspenov: ok, makes sense for the seconds. I think we should just reject anything smaller than 5 minutes with custom validation so we don't have these checks in the execution logic ?

IvanIvanoff: @tspenov if in the future we change the scheduling time to `1m` (or less) instead of `5m` we'll need to make it clear to all users that that thing changed.

config/notifications_config.exs
valo: I think we can easily run this every 5min and respect the hour the user specified. And have a cool down of 24h. That way the signal will work as expected. I think later we can change the trending words to be computed more frequently, which will improve the freshness of the signal.

tspenov: Ok, scheduled every 5 mins with cooldown 24h. @valo thanks for input

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,169 @@
...
+
+  @spec type() :: String.t()
+  def type(), do: @trigger_type
+
+  @spec get_data(%TrendingWordsTriggerSettings{}) :: {:ok, list(top_word_type)} | :error
+  def get_data(%TrendingWordsTriggerSettings{trigger_time: trigger_time}) do
+    now = Timex.now()
+    trigger_time = Time.from_iso8601!(trigger_time)
+
+    if trigger_time.hour == now.hour do
```
IvanIvanoff: But doesn't this mean that if you create a trigger for 19:59 it will be executed at 19:00?
The more correct way here is to check if the time of the day is greater or equal to the trigger time, i.e. it will be executed once the time is >= 19:59

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
```diff
@@ -1,8 +1,169 @@
...
+      get_today_top_words(now)
+    end
+  end
+
+  def valid_trigger_time?(trigger_time) when is_binary(trigger_time) do
```
IvanIvanoff: shouldn't this be in the module with the other validations named as `valid_iso8601_datetime_string`

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: yes but I don't see any reason why we should allow minutes when specifying the signal time

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
tspenov: Its not used anywhere else - so might stay until we reuse it. Now i don't see where other signal where we will use it.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
IvanIvanoff: @tspenov all limitations such as these are not obvious for the users and we better not spend time and UI text boxes to explain why a valid time is not valid for this signal only. The less assumptions - the best.

lib/sanbase/signals/trigger/trending_words_trigger_settings.ex
IvanIvanoff: I mentioned it so the trending words module does not contain validation logic

tspenov: ok, lets move it there so trigger settings modules are cleaner

PR #976, opened 2019-02-06
Fix historica balance last bucket

tspenov:
#### Summary

image


IvanIvanoff:
Aren't all balances shifted here, not only the last one?

valo:
Let's rewrite this API to use the `eth_balances` and `erc20_balances` tables that should be shipped on production soon.

PR #983, opened 2019-02-12
Support list of slugs and watchlist as target in signals

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/project/project.ex
+++ b/lib/sanbase/model/project/project.ex
@@ -320,6 +320,13 @@ defmodule Sanbase.Model.Project do
     |> Repo.one()
   end
 
+  def by_slugs(slugs) when is_list(slugs) do
+    Project
+    |> where([p], p.coinmarketcap_id in ^slugs)
+    |> preload(^@preloads)
+    |> Repo.all()
+  end
+
   def by_id(id) when is_integer(id) or is_binary(id) do
     Project
     |> where([p], p.id == ^id)
--- a/lib/sanbase/model/project/project_list.ex
+++ b/lib/sanbase/model/project/project_list.ex
@@ -166,8 +166,7 @@ defmodule Sanbase.Model.Project.List do
   def slugs_by_ids(ids) do
     from(
       p in Project,
-      where: p.id in ^ids,
-      where: not is_nil(p.coinmarketcap_id),
+      where: p.id in ^ids and not is_nil(p.coinmarketcap_id),
       select: p.coinmarketcap_id
     )
     |> Repo.all()
--- a/lib/sanbase/model/project/project_list.ex
+++ b/lib/sanbase/model/project/project_list.ex
@@ -162,4 +162,14 @@ defmodule Sanbase.Model.Project.List do
     )
     |> Repo.all()
   end
+
+  def slugs_by_ids(ids) do
+    from(
+      p in Project,
+      where: p.id in ^ids,
+      where: not is_nil(p.coinmarketcap_id),
+      select: p.coinmarketcap_id
+    )
+    |> Repo.all()
+  end
 end
--- a/lib/sanbase/signals/evaluator/evaluator.ex
+++ b/lib/sanbase/signals/evaluator/evaluator.ex
@@ -25,8 +25,7 @@ defmodule Sanbase.Signals.Evaluator do
     Cache.get_or_store(
       Trigger.cache_key(trigger),
       fn ->
-        res = %UserTrigger{user_trigger | trigger: Trigger.evaluate(trigger)}
-        res
+        %UserTrigger{user_trigger | trigger: Trigger.evaluate(trigger)}
       end
     )
   end
--- a/lib/sanbase/signals/evaluator/evaluator.ex
+++ b/lib/sanbase/signals/evaluator/evaluator.ex
@@ -1,6 +1,6 @@
 defmodule Sanbase.Signals.Evaluator do
   @moduledoc ~s"""
-  A module that takes a list of triggers and returns the ones that must be triggered
+  A module that takes a list of triggers and returns the ones that are triggered.
   """
 
   alias Sanbase.Signals.Evaluator.Cache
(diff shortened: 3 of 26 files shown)


lib/sanbase/signals/trigger/price_absolute_change_settings.ex
```diff
@@ -52,30 +54,76 @@ defmodule Sanbase.Signals.Trigger.PriceAbsoluteChangeSettings do
     )
   end
 
+  def get_data(%__MODULE__{target: target}) when is_binary(target) do
```
tspenov: can we somehow reuse `get_data` in DAA and PricePercent, maybe by providing a callback as second argument ?

IvanIvanoff: @tspenov When I start working on a different trigger I'll see what are the common parts that can be abstracted.

lib/sanbase/model/project/project_list.ex
```diff
@@ -162,4 +162,14 @@ defmodule Sanbase.Model.Project.List do
     )
     |> Repo.all()
   end
+
+  def slugs_by_ids(ids) do
+    from(
+      p in Project,
+      where: p.id in ^ids,
+      where: not is_nil(p.coinmarketcap_id),
```
tspenov: Can we just use and here instead of double where twice ?

lib/sanbase/model/project/project.ex
```diff
@@ -320,6 +320,13 @@ defmodule Sanbase.Model.Project do
     |> Repo.one()
   end
 
+  def by_slugs(slugs) when is_list(slugs) do
+    Project
+    |> where([p], p.coinmarketcap_id in ^slugs)
```
tspenov: I think we use dominantly `from (p in Project ...`style. What do you say we stick to one style of querying using ecto in future ?

lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -22,18 +21,12 @@ defmodule Sanbase.Signals.Evaluator do
...
-      Trigger.has_cooldown?(trigger)
-    end)
-  end
-
   defp evaluate(%UserTrigger{trigger: trigger} = user_trigger) do
     Cache.get_or_store(
       Trigger.cache_key(trigger),
       fn ->
-        %UserTrigger{user_trigger | trigger: Trigger.evaluate(trigger)}
+        res = %UserTrigger{user_trigger | trigger: Trigger.evaluate(trigger)}
```
tspenov: why is the assignment needed here ?

lib/sanbase/model/project/project.ex
IvanIvanoff: Piping is a lot more flexible. We can define a set of functions for a lot of the shared behavior (not null slugs, id in a list, volume bigger than x, etc.) and just plug them in the pipe instead of writing the (for example) `where` clause everywhere. I would rather prefer this to be the more used way.

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -47,17 +47,35 @@ defmodule Sanbase.Signals.Scheduler do
     |> Sanbase.Parallel.pmap_concurrent(
       fn %UserTrigger{} = user_trigger ->
         case Signal.send(user_trigger) do
-          :ok ->
-            %{user: user, trigger: %{id: trigger_id}} = user_trigger
-            UserTrigger.update_user_trigger(user, %{id: trigger_id, last_triggered: Timex.now()})
-            :ok
+          list when is_list(list) ->
+            %{user: user, trigger: %{id: trigger_id, last_triggered: last_triggered}} =
```
tspenov: Can we extract this case into function `update_last_triggered` that takes `user_trigger` and the result `list`. I think it will be more readable that way.

lib/sanbase/signals/trigger.ex
```diff
@@ -23,22 +23,32 @@ defmodule Sanbase.Signals.Trigger do
...
-    field(:cooldown, :string)
+    field(:last_triggered, :map, default: %{})
+    field(:cooldown, :string, default: "24h")
   end
 
   @doc false
   def changeset(schema, params) do
     schema
     |> cast(params, [:settings, :is_public, :cooldown, :last_triggered])
+    |> validate_required([:settings, :is_public, :cooldown])
```
tspenov: `cooldown` and `is_public` are not required. I think `settings` is also not required on update also which uses the same changeset.

lib/sanbase/signals/trigger.ex
```diff
@@ -50,13 +60,53 @@ defmodule Sanbase.Signals.Trigger do
...
+        Timex.compare(
+          Timex.shift(target_last_triggered,
+            seconds: Sanbase.DateTimeUtils.compound_duration_to_seconds(cd)
+          ),
+          Timex.now()
+        ) == 1
+    end
+  end
+
+  def set_last_triggered(
```
tspenov: This is not used anywhere

lib/sanbase/signals/types.ex
```diff
@@ -1,6 +1,10 @@
...
-  @type channel :: String.t()
+  alias Sanbase.UserLists.UserList
+
   @type trigger_type :: String.t()
+  @type channel :: String.t()
+  @type target :: String.t()
+  @type complex_target :: target | list(target) | map()
   @type time_window :: String.t()
-  @type payload :: String.t()
+  @type payload :: {target, String.t()}
```
tspenov: isn't payload a map ?

lib/sanbase/signals/evaluator/evaluator.ex
IvanIvanoff: debugging leftover

lib/sanbase/model/project/project.ex
tspenov: hmm I think the same can be achieved with composing queries 
```elixir
 def non_null_slugs(query) do
    from(p in query, where: not is_nil(p.slug))
  end
    
  def by_ids(query, ids) do
    from(p in query, where: p.id in ^ids)
  end

Project
|> Project.non_null_slugs()
|> Project.by_ids(ids)
```

IvanIvanoff: In order not to use subqueries that could become impossible to write or slow, these should be macros instead of functions, so you build a single query in the end. That's exactly what this `where` macro does.

PR #984, opened 2019-02-12
Api historical trigger occurrencies

tspenov:
### daily_active_addresses
```graphql
 {
  historicalTriggerPoints(
    cooldown:"2d"
    settings: "{\"channel\":\"telegram\",\"percent_threshold\":200.0,\"repeating\":false,\"target\":\"naga\",\"time_window\":\"30d\",\"type\":\"daily_active_addresses\"}"
  )
}
```

```graphql
...
      {
        "active_addresses": 17,
        "datetime": "2019-01-10T00:00:00Z",
        "percent_change": -30,
        "price": 0.07062305541440246,
        "triggered?": false
      },
      {
        "active_addresses": 12,
        "datetime": "2019-01-11T00:00:00Z",
        "percent_change": -50,
        "price": 0.0656760396585122,
        "triggered?": false
      },
      {
        "active_addresses": 193,
        "datetime": "2019-01-12T00:00:00Z",
        "percent_change": 718,
        "price": 0.07118857891523572,
        "triggered?": true
      },
```

### price_absolute_change

```graphql
 {
  historicalTriggerPoints(
    cooldown: "1h"
    settings: "{\"channel\":\"telegram\",\"above\":0.400417,\"repeating\":false,\"target\":\"santiment\",\"type\":\"price_absolute_change\"}"
  )
}
```

```
      {
        "datetime": "2019-02-20T07:00:00Z",
        "price": 0.5267121707551666,
        "triggered?": true
      },
      {
        "datetime": "2019-02-20T08:00:00Z",
        "price": 0.5307551205455,
        "triggered?": false
      },
      {
        "datetime": "2019-02-20T09:00:00Z",
        "price": 0.5335671901080714,
        "triggered?": true
      },
      {
        "datetime": "2019-02-20T10:00:00Z",
        "price": 0.5369475239806667,
        "triggered?": false
      }
```

### price_percent_change

```graphql
 {
  historicalTriggerPoints(
    cooldown:"1h"
    settings: "{\"channel\":\"telegram\",\"percent_threshold\":5.0,\"repeating\":false,\"target\":\"santiment\",\"time_window\":\"1h\",\"type\":\"price_percent_change\"}"
  )
}
```

```graphql
{
      {
        "datetime": "2018-11-25T11:00:00Z",
        "percent_change": 6.0,
        "price": 0.2235804541695652,
        "triggered?": true
      },
      {
        "datetime": "2018-11-25T12:00:00Z",
        "percent_change": 1.0,
        "price": 0.22570803848636364,
        "triggered?": false
      },
      {
        "datetime": "2018-11-25T13:00:00Z",
        "percent_change": 0.0,
        "price": 0.22634788076956525,
        "triggered?": false
      },
...
```


test/sanbase/signals/triggers_test.exs
```diff
@@ -10,10 +10,10 @@ defmodule Sanbase.Signals.TriggersTest do
 
     trigger_settings = %{
       type: "daily_active_addresses",
-      target: "santiment",
+      target: "naga",
```
IvanIvanoff: why 😃

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -29,19 +43,16 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
           triggered?: boolean(),
           payload: Type.payload()
         }
+  @type historical_trigger_points_type :: %{
+          datetime: %DateTime{},
+          active_addresses: non_neg_integer(),
+          triggered: boolean()
```
IvanIvanoff: `triggered?`

IvanIvanoff:
@tspenov maybe we need to start separating the functions into modules, i.e. one for current and one for historical calculations

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -92,6 +144,80 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+    filtered_points =
+      points
+      |> Enum.filter(fn %{datetime: dt} ->
+        DateTime.to_unix(dt) >= DateTime.to_unix(from) &&
+          DateTime.to_unix(dt) <= DateTime.to_unix(from)
+      end)
+
+    sum =
+      filtered_points
+      |> Enum.reduce(0, fn %{active_addresses: daa}, acc -> acc + daa end)
```
IvanIvanoff: You cannot sum daily active addresses. The only aggregation that makes sense for it is taking the average/min/max

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -92,6 +144,80 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+         {datetime, active_addresses},
+         daa_result,
+         first_possible_trigger_date,
+         time_window_in_days
+       ) do
+    case DateTime.compare(datetime, first_possible_trigger_date) do
+      :lt ->
+        false
+
+      cmp when cmp in [:gt, :eq] ->
```
IvanIvanoff: This is equivalent to `_ ->`

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -92,6 +144,80 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+            Timex.shift(datetime, days: -time_window_in_days),
+            Timex.shift(datetime, days: -1)
+          )
+
+        percent_change(avg_time_window, active_addresses) >= settings.percent_threshold
+    end
+  end
+
+  defp get_daily_active_addresses(settings) do
+    {:ok, contract, _token_decimals} = Project.contract_info_by_slug(settings.target)
```
IvanIvanoff: Will we have the whole `settings` struct here? The idea is to test multiple times until the user is satisfied with the data and then persist in the DB. If so - we need to organize the flow in such a way that this scenario is possible (api -> struct -> test) without persisting.

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -92,6 +144,80 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+    one_day_in_seconds = 3600 * 24
+
+    if time_window_in_seconds <= @minimal_time_window_in_days * one_day_in_seconds do
+      @minimal_time_window_in_days
+    else
+      Sanbase.Utils.Math.to_integer(time_window_in_seconds / one_day_in_seconds)
+    end
+  end
+
+  defp calc_average_daa_for_interval(points, from, to) do
```
IvanIvanoff: This function does too much imo. We can have the filtering + map to just the `daa` values outside and make a generic `average` function in the `Math` module.
Another idea is to have `average_by` similar to `Enum.sort_by` where you pass the access function as an argument.

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -78,6 +89,47 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+          {:ok, list(historical_trigger_points_type)} | {:error, String.t()}
+  def historical_trigger_points(%__MODULE{target: target} = settings) when is_binary(target) do
+    settings
+    |> get_daily_active_addresses()
+    |> case do
+      {:ok, daa_result} when is_list(daa_result) and daa_result != [] ->
+        time_window_in_days = time_window_in_days(settings.time_window)
+
+        first_possible_trigger_date =
+          first_possible_trigger_date(daa_result, time_window_in_days + 1)
```
IvanIvanoff: This is not used outside `point_triggered?` and all arguments needed to calculate this are passed to the `point_triggered?`. You can reduce the number of arguments and calculate it in the function body.

test/sanbase/signals/triggers_test.exs
tspenov: hahah was testing with `naga` cause there was a signal for it on stage and was copy pasta from the test file in iex

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
tspenov: yeah it won't be persisted - just create struct from the map and use it

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -120,6 +133,120 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
       ])
     end
 
+    # @spec historical_trigger_points(%DailyActiveAddressesSettings{}) ::
```
IvanIvanoff: the spec is in a comment

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
```diff
@@ -120,6 +133,120 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
...
+    end
+
+    defp first_possible_trigger_date(daa_result, days_shift) do
+      daa_result
+      |> List.first()
+      |> Map.get(:datetime)
+      |> Timex.shift(days: days_shift)
+    end
+
+    def time_window_in_days(time_window) do
```
IvanIvanoff: can we have this in the date time utils module? It sounds really like `compound_duration_in_seconds`

lib/sanbase/signals/user_trigger.ex
```diff
@@ -42,6 +42,14 @@ defmodule Sanbase.Signals.UserTrigger do
     |> validate_required([:user_id])
   end
 
+  def historical_trigger_points(%{settings: settings} = params) do
```
IvanIvanoff: This belongs to the `Trigger` module. Also you can do it like:
```
def historical_trigger_points(%Trigger{}), do: ...
def historical_trigger_points(%{settings: settings}), do: settings |> load_in_struct |> historical_trigger_points()
```

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
tspenov: I am passing it because point_triggered? is in a loop - so to not calculate it every time

lib/sanbase/signals/user_trigger.ex
tspenov: Lets leave it here cause all functions that resolver uses are in UserTrigger and in Trigger will have to expose `load_in_struct`. Otherwise made it with pattern matching

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
tspenov: simplified and split into 2 funcs for filtering and avg

lib/sanbase/signals/trigger/daily_active_addresses_settings.ex
tspenov: extracted in `compound_duration_in_days`

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,131 @@
...
+
+    @type historical_trigger_points_type :: %{
+            datetime: %DateTime{},
+            active_addresses: non_neg_integer(),
+            triggered?: boolean()
+          }
+
+    @spec historical_trigger_points(%DailyActiveAddressesSettings{}) ::
+            {:ok, list(historical_trigger_points_type)} | {:error, String.t()}
+    def historical_trigger_points(%DailyActiveAddressesSettings{target: target} = settings)
```
IvanIvanoff: One idea: You can calculate moving average (mapping `{datetime: datetime, active_addresses: active_addresses}` -> `{datetime: datetime, active_addresses: active_addresses, average_daa: average_daa}`). After that the trigger is just another map that just puts `true` if `percent_change(value, average_daa) >= percent_threshold`

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,100 @@
...
+            datetime: %DateTime{},
+            active_addresses: non_neg_integer(),
+            price: float(),
+            triggered?: boolean()
+          }
+
+    @spec historical_trigger_points(%DailyActiveAddressesSettings{}) ::
+            {:ok, list(historical_trigger_points_type)} | {:error, String.t()}
+    def historical_trigger_points(%DailyActiveAddressesSettings{target: target} = settings) do
+      with {:ok, contract, _token_decimals} = Project.contract_info_by_slug(settings.target),
```
IvanIvanoff: you have `=` here

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,100 @@
...
+    defp get_daily_active_addresses(contract, from, to, interval) do
+      if contract == "ETH" do
+        EthDailyActiveAddresses.average_active_addresses(from, to, interval)
+      else
+        Erc20DailyActiveAddresses.average_active_addresses(contract, from, to, interval)
+      end
+    end
+
+    defp time_window_in_days(time_window) do
+      case Sanbase.DateTimeUtils.compound_duration_to_days(time_window) do
```
IvanIvanoff:
```
Sanbase.DateTimeUtils.compound_duration_to_days(time_window) |> max(@minimal_time_window_in_days)
```

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,121 @@
...
+    @spec historical_trigger_points(%PriceAbsoluteChangeSettings{}) ::
+            {:ok, list(PricesHistory.historical_trigger_points_type())} | {:error, String.t()}
+    def historical_trigger_points(%PriceAbsoluteChangeSettings{target: target} = settings) do
+      settings
+      |> PricesHistory.historical_trigger_points()
+      |> case do
+        {:ok, prices} ->
+          prices =
+            prices
+            |> Enum.map(fn point ->
```
IvanIvanoff: you can pattern match on the anon function to remove the `if`
```
fn %{average: average} = point -> Map.put(point, :triggered?, triggered?)
      _ -> Map.put(point, :triggered?, false)
```
Also why you check for `average` when you don't use it?

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,121 @@
...
+              if Map.get(point, :average) do
+                triggered? =
+                  (is_float(settings.above) and point.price >= settings.above) or
+                    (is_float(settings.below) and point.price <= settings.below)
```
IvanIvanoff: Use `is_number` instead of `is_float`

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,121 @@
...
+      prices =
+        Enum.map(prices, fn [dt, price | _rest] ->
+          %{
+            datetime: dt,
+            price: price,
+            triggered?: false
+          }
+        end)
+
+      {:ok, sma} = moving_average_excluding_last(prices, 2, :price)
```
IvanIvanoff: I don't think the historical triggers for prices can be calculated with moving average. In the DAA example it's natural as you are comparing the current value to an average historical.
In the case of percent change you'll need the concrete value at `{time - time_period, time}` and the absolute percent change can be calculated only by the concrete value at a given time.

IvanIvanoff:
One crucial part that I see missing is the `cooldown`. IMO it should be implemented in the evaluator - if a point is triggered, you need to jump over `time_window` interval of points, otherwise it is not understandable when it will trigger

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+            end
+
+          _price, {accumulated_calculations, cooldown_left} ->
+            {[false | accumulated_calculations], cooldown_left - 1}
+        end)
+
+      absolute_price_calculations = absolute_price_calculations |> Enum.reverse()
+
+      points =
+        Enum.zip([datetimes, prices, absolute_price_calculations])
```
IvanIvanoff: Why do you split the result and after that zip the whole thing back?
Can't you just invoice the reduce on top of the result returned from the database?

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+
+    def historical_trigger_points(
+          %PricePercentChangeSettings{target: target} = settings,
+          cooldown
+        )
+        when is_binary(target) do
+      {:ok, prices, datetimes} = PricesHistory.get_prices(settings)
+
+      time_window_in_hours =
+        max(2, Sanbase.DateTimeUtils.compound_duration_to_hours(settings.time_window))
```
IvanIvanoff: It is a little strange that we allow arbitrary cool down, but cannot run the historical one with less than 2 hours. It should be documented or we need to handle it differently

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+      percent_change_calculations =
+        prices
+        |> Enum.chunk_every(time_window_in_hours, 1, :discard)
+        |> Enum.map(fn chunk -> {List.first(chunk), List.last(chunk)} end)
+        |> percent_change_calculations_with_cooldown(
+          settings.percent_threshold,
+          cooldown_in_hours
+        )
+
+      empty_calculations = for _ <- 1..(time_window_in_hours - 1), do: {0.0, false}
```
IvanIvanoff: I would personally prefer `Stream.cycle([{0.0, false}]) |> Enum.take(time_window_in_hours-1)` is more readable

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+            price: price,
+            triggered?: triggered?,
+            percent_change: percent_change
+          }
+        end)
+
+      {:ok, points}
+    end
+
+    def historical_trigger_points(_, _), do: {:error, "Not implemented"}
```
IvanIvanoff: let it crash

lib/sanbase/signals/history/utils.ex
```diff
@@ -0,0 +1,30 @@
...
+        percent_change, {accumulated_calculations, 0} ->
+          if percent_change > percent_threshold do
+            {[{percent_change, true} | accumulated_calculations], cooldown}
+          else
+            {[{percent_change, false} | accumulated_calculations], 0}
+          end
+
+        percent_change, {accumulated_calculations, cooldown_left} ->
+          {[{percent_change, false} | accumulated_calculations], cooldown_left - 1}
+      end)
```
IvanIvanoff: You could avoid the if:
```elixir
|> Enum.reduce({[], 0}, fn
      percent_change, {accumulated_calculations, 0} when percent_change > percent_threshold ->
          {[{percent_change, true} | accumulated_calculations], cooldown}
      percent_change, {accumulated_calculations, 0} ->
          {[{percent_change, false} | accumulated_calculations], 0}
      percent_change, {accumulated_calculations, cooldown_left} ->
        {[{percent_change, false} | accumulated_calculations], cooldown_left - 1}
    end)
```

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+      {:ok, prices, datetimes} = PricesHistory.get_prices(settings)
+      above = Map.get(settings, :above)
+      below = Map.get(settings, :below)
+      cooldown_in_hours = Sanbase.DateTimeUtils.compound_duration_to_hours(cooldown)
+
+      {absolute_price_calculations, _} =
+        prices
+        |> Enum.reduce({[], 0}, fn
+          price, {accumulated_calculations, 0} ->
+            if (is_number(above) and price >= above) or (is_number(below) and price <= below) do
```
IvanIvanoff:
```elixir
defguard outside_interval(value, low, high) when is_number(value) and is_number(low) and is_number(high) and (value <= low or value >= high)
```
...
```elixir

|> Enum.reduce({[], 0}, fn
      price, {accumulated_calculations, 0} when outside_interval(price, below, above) ->
          {[true | accumulated_calculations], cooldown_in_hours}
        price, {accumulated_calculations, 0} ->
          {[false | accumulated_calculations], 0}
      _price, {accumulated_calculations, cooldown_left} ->
        {[false | accumulated_calculations], cooldown_left - 1}
    end)
```

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+
+    def historical_trigger_points(_, _), do: {:error, "Not implemented"}
+  end
+
+  defimpl Sanbase.Signals.History, for: PricePercentChangeSettings do
+    alias Sanbase.Signals.History.PricesHistory
+
+    @spec historical_trigger_points(%PricePercentChangeSettings{}, String.t()) ::
+            {:ok, list(PricesHistory.historical_trigger_points_type())} | {:error, String.t()}
+
```
IvanIvanoff: empty line.

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,136 @@
...
+        |> Enum.chunk_every(time_window_in_hours, 1, :discard)
+        |> Enum.map(fn chunk -> {List.first(chunk), List.last(chunk)} end)
+        |> percent_change_calculations_with_cooldown(
+          settings.percent_threshold,
+          cooldown_in_hours
+        )
+
+      empty_calculations = for _ <- 1..(time_window_in_hours - 1), do: {0.0, false}
+
+      points =
```
IvanIvanoff: points = some_separate_function()

We aim to avoid long functions or more specifically functions that do a lot of things - complex functions are impossible to compose, reason about and test.

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,109 @@
...
+  alias Sanbase.Prices.Store, as: PricesStore
+  require Logger
+
+  defimpl Sanbase.Signals.History, for: DailyActiveAddressesSettings do
+    @historical_days_from 90
+    @historical_days_to 1
+    @historical_days_interval "1d"
+    @minimal_time_window_in_days 2
+
+    @type historical_trigger_points_type :: %{
```
IvanIvanoff: At some other places this type is defined inside the `defimpl` block. Outside the block is the better option I think. I don't think you can access it from a different module if it's here.

lib/sanbase/signals/history/prices_history.ex
tspenov: niiice

lib/sanbase/signals/history/prices_history.ex
tspenov: its time window. Its due to data being bucketed in `1h` intervals. Added a comment.

lib/sanbase/signals/history/prices_history.ex
tspenov: I think its simpler to reduce only on top of price values and the accumulator to return array with true/false and zip it again instead of moving also the other values in all other steps.

test/sanbase/signals/history/utils_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  property "average is always between min and max when list of integers" do
+    check all list <- list_of(positive_integer(), min_length: 1) do
+      average = Utils.average(list)
+      min = Enum.min(list)
+      max = Enum.max(list)
+      assert average >= min and average <= max
+    end
+  end
+
+  property "average is always between min -1 and max + 1 when floats" do
```
IvanIvanoff: 🎉

test/support/data_case.ex
```diff
@@ -16,6 +16,7 @@ defmodule Sanbase.DataCase do
 
   using do
     quote do
+      use ExUnitProperties
```
IvanIvanoff: We could consider using https://github.com/proper-testing/proper after researching more about the license. GPL is more restrictive than MIT, but I am not sure what is the case when distillery excludes the `

tspenov: Also there is a nice book: https://pragprog.com/book/fhproper/property-based-testing-with-proper-erlang-and-elixir
But chose this one based on the comment that they want to include it in the elixir core: https://github.com/whatyouhide/stream_data.

IvanIvanoff: It was scheduled to be included in Elixir 1.7 but it was later decided that it will stay a separate library

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,125 @@
...
+          price: float(),
+          triggered?: boolean(),
+          percent_change: float()
+        }
+
+  defimpl Sanbase.Signals.History, for: DailyActiveAddressesSettings do
+    @historical_days_from 90
+    @historical_days_to 1
+    @historical_days_interval "1d"
+    # Minimal time window is 2 days. That is due to data being bucketed in `1 day` intervals.
```
IvanIvanoff: You can leave only `Because the data is bucketed in daily intervals`

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -0,0 +1,125 @@
...
+           {:ok, prices} when is_list(prices) and prices != [] <-
+             PricesStore.fetch_prices_with_resolution(
+               measurement,
+               Timex.shift(from, days: -1),
+               to,
+               interval
+             ) do
+        prices = Enum.map(prices, fn [dt, price | _rest] -> %{datetime: dt, price: price} end)
+
+        daa_result =
```
IvanIvanoff: Can you add a short comment in the different steps as this function does a lot of different and not obvious transformations

lib/sanbase/signals/history/prices_history.ex
```diff
@@ -0,0 +1,160 @@
...
+  @historical_days_from 90
+  @historical_days_interval "1h"
+
+  @type historical_trigger_points_type :: %{
+          datetime: %DateTime{},
+          price: float(),
+          triggered?: boolean()
+        }
+
+  def get_prices(settings) do
```
IvanIvanoff:
```elixir
def get_prices(%{target: target} = settings) when is_binary(target) do
```

lib/sanbase/signals/history/utils.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Sanbase.Signals.History.Utils do
+  import Sanbase.Signals.Utils
+
+  def percent_change_calculations_with_cooldown(values, percent_threshold, cooldown) do
```
IvanIvanoff: documentation

lib/sanbase/utils/datetime/utils.ex
```diff
@@ -92,6 +92,28 @@ defmodule Sanbase.DateTimeUtils do
...
 
+  def compound_duration_to_days(interval) do
+    interval_in_seconds = compound_duration_to_seconds(interval)
+    one_day_in_seconds = 3600 * 24
+
+    if interval_in_seconds < one_day_in_seconds do
+      0
+    else
+      Sanbase.Utils.Math.to_integer(interval_in_seconds / one_day_in_seconds)
+    end
```
IvanIvanoff: you can just use `div(interval_in_seconds, one_day_in_seconds)` and avoid the if and the `to_integer` function

lib/sanbase/utils/datetime/utils.ex
```diff
@@ -92,6 +92,28 @@ defmodule Sanbase.DateTimeUtils do
...
+
+  def compound_duration_to_hours(interval) do
+    interval_in_seconds = compound_duration_to_seconds(interval)
+    one_hour_in_seconds = 3600
+
+    if interval_in_seconds < one_hour_in_seconds do
+      0
+    else
+      Sanbase.Utils.Math.to_integer(interval_in_seconds / one_hour_in_seconds)
+    end
```
IvanIvanoff: you can just use `div(interval_in_seconds, one_day_in_seconds)` and avoid the if and the `to_integer` function

PR #987, opened 2019-02-15
Add tags to UserTrigger module

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/user_trigger.ex
```diff
@@ -88,9 +126,13 @@ defmodule Sanbase.Signals.UserTrigger do
       preload: [{:user, :user_settings}]
```
tspenov: we should preload tags here

lib/sanbase/signals/user_trigger.ex
```diff
@@ -88,9 +126,13 @@ defmodule Sanbase.Signals.UserTrigger do
...
+    |> Enum.map(&trigger_in_struct/1)
   end
 
+  @doc ~s"""
+  Create a new user trigger that is used to fire signals.
+  To create a new trigger `settings` and `title` parameters must be present
+  """
   @spec create_user_trigger(%User{}, map()) ::
           {:ok, %__MODULE__{}} | {:error, String.t()} | {:error, %Ecto.Changeset{}}
   def create_user_trigger(%User{id: user_id} = _user, %{settings: settings} = params) do
```
tspenov: we should preload tags on `create_user_trigger`

IvanIvanoff: https://github.com/santiment/sanbase2/pull/987/files#diff-0242dcfda9cd696d35ce332c83bd7b54R326
This test works

lib/sanbase/signals/user_trigger.ex
IvanIvanoff: This is not exposed in the public API, only to the evaluator, which does not need the tags (for now)

test/support/Untitled-3
```diff
@@ -0,0 +1,12 @@
+Първата лекция на курса „Функционално програмиране с Elixir“ ще се проведе утре, 19.02.2019 г., от 19:00  ч. в зала 02.
```
tspenov: :D

tspenov: 😄

IvanIvanoff: ooooops

PR #997, opened 2019-02-22
Remove repeating field from triggers settings

tspenov:
We're not using it now, so lets remove it and introduce it when there are mockups for it.


tspenov:
closing this because we decided to implement it

PR #1001, opened 2019-02-22
Fix cooldown trending words

tspenov:
Trending words cooldown was not working with the introductions of cooldown per target

### DIFF
--- a/lib/sanbase/signals/evaluator/scheduler.ex
+++ b/lib/sanbase/signals/evaluator/scheduler.ex
@@ -16,7 +16,7 @@ defmodule Sanbase.Signals.Scheduler do
     TrendingWordsTriggerSettings
   }
 
-  alias Sanbase.Signals.{UserTrigger, Trigger}
+  alias Sanbase.Signals.UserTrigger
   alias Sanbase.Signals.Evaluator
   alias Sanbase.Signal
 
@@ -47,22 +47,9 @@ defmodule Sanbase.Signals.Scheduler do
   defp run(type) do
     type
     |> UserTrigger.get_triggers_by_type()
-    |> remove_triggers_on_cooldown()
     |> Evaluator.run()
     |> send_and_mark_as_sent()
     |> log_sent_messages_stats(type)
-  end
-
-  # remove triggers that has no `target` and are in cooldown
-  defp remove_triggers_on_cooldown(triggers) do
-    triggers
-    |> Enum.reject(fn
-      %UserTrigger{trigger: %Trigger{settings: %{target: _}}} ->
-        false
-
-      %UserTrigger{trigger: trigger} ->
-        Trigger.has_cooldown?(trigger)
-    end)
   end
 
   defp send_and_mark_as_sent(triggers) do
--- a/lib/sanbase/signals/evaluator/scheduler.ex
+++ b/lib/sanbase/signals/evaluator/scheduler.ex
@@ -16,7 +16,7 @@ defmodule Sanbase.Signals.Scheduler do
     TrendingWordsTriggerSettings
   }
 
-  alias Sanbase.Signals.UserTrigger
+  alias Sanbase.Signals.{UserTrigger, Trigger}
   alias Sanbase.Signals.Evaluator
   alias Sanbase.Signal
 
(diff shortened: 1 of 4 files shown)


lib/sanbase/signals/trigger/trigger.ex
```diff
@@ -110,6 +110,14 @@ defmodule Sanbase.Signals.Trigger do
     Sanbase.Signals.Settings.cache_key(trigger_settings)
   end
 
+  def has_cooldown?(%Trigger{last_triggered: nil}), do: false
```
IvanIvanoff: This is impossible scenario. `last_triggered` defaults to `%{}`

lib/sanbase/signals/trigger/trigger.ex
```diff
@@ -110,6 +110,14 @@ defmodule Sanbase.Signals.Trigger do
...
+  def has_cooldown?(%Trigger{cooldown: nil}), do: false
```
IvanIvanoff: This should also be disallowed imo.

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -47,11 +47,24 @@ defmodule Sanbase.Signals.Scheduler do
   defp run(type) do
     type
     |> UserTrigger.get_triggers_by_type()
+    |> remove_triggers_on_cooldown()
```
IvanIvanoff: The rest of the signals (except trending words) will call `remove_triggers_on_cooldown` twice now, right?
In that case can't we change the trending words to use the cooldown as the rest of the signals? If there is no `target` then we'll put the trigger type as a key to indicate that it's global?

tspenov: if I put `target="all"` seems that I have to put have to put also `filtered_target_list` so these without target will have 2 more fields in struct. Otherwise it skips the by pattern matching and checking if target exist

IvanIvanoff: The `filtered_target_list` is an evil I still wonder how we can improve. If we don't have it we need to support N different formats for getting data and that's much worse.
If we have it - cases such this one are not pretty.

lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -50,6 +52,11 @@ defmodule Sanbase.Signals.Trigger.TrendingWordsTriggerSettings do
   def type(), do: @trigger_type
 
   @spec get_data(%TrendingWordsTriggerSettings{}) :: {:ok, list(top_word_type)} | :error
+  def get_data(%TrendingWordsTriggerSettings{filtered_target_list: filtered_target_list})
+      when filtered_target_list == [] do
```
IvanIvanoff: just 
```elixir
def get_data(%TrendingWordsTriggerSettings{filtered_target_list: []}), do: :error
```
will do it

lib/sanbase/signals/trigger/trigger.ex
```diff
@@ -91,17 +91,17 @@ defmodule Sanbase.Signals.Trigger do
     %Trigger{trigger | settings: trigger_settings}
   end
 
+  def evaluate(%Trigger{settings: trigger_settings} = trigger) do
```
IvanIvanoff: If we have `all` for `target` why do we need to handle it this way? Also, will this case ever match if we have `target="all"`. Moreover, if we handle this case with the previous `evaluate` function we'll allow setting up cool downs for trending words. It will be possible to configure it to send it once a week, for example

tspenov: this function existed - just moving it on top so its close to the other signature of evaluate. Cooldowns are handled by the other function. Maybe we don't need this one at all ?

IvanIvanoff: Hm, best to check the test coverage if something passes through this function

tspenov: `evaluate` for triggers without target is not used. Trending words was the last one. Removing it.

PR #1005, opened 2019-02-25
Create table and store executed signals

tspenov:
Create a table where executed triggers will be stored along with their payload.
The table will be used to feed the graphql API endpoint for user signals activity timeline.


lib/sanbase/signals/user_signal.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Sanbase.Signals.UserSignal do
```
IvanIvanoff: This name will be confused a lot with `UserTrigger` and `Trigger`, let's think of a different name. `Signals.HistoricalActivity`?

lib/sanbase/utils/parallel.ex
```diff
@@ -1,10 +1,12 @@
 defmodule Sanbase.Parallel do
   @doc ~s"""
-
+  Module implementing parallel and concurrent versions of enumerable map and filter functions.
```
IvanIvanoff: we definitely need to improve on the names, `pmap` and `pmap_concurrent` do not have much meaning. The idea of the second one is that it's started under a TaskSupervisor and the concurrency can be controlled, while in the plain `pmap` it cannot. Another possibility is to completely remove `pmap` and leave only `pmap_concurrent` with default params.

tspenov: BTW this 2 lines are not needed, right ?
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/utils/parallel.ex#L9

tspenov: Also `pmap_concurrent` uses Stream instead of Enum

tspenov: Are you sure we can replace with `pmap_concurrent` here for example https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex#L82 ?

lib/sanbase/signals/user_signal.ex
tspenov: ok, sounds good

lib/sanbase/utils/parallel.ex
tspenov: Replaced all places with `pmap_concurrent`

IvanIvanoff: @tspenov note that this change will lead to performance degradation. The `max_concurrency` defaults to `System.online_schedulers/0` so everywhere where `pmap` was used the concurrency will be reduced a lot.
This will have a noticeable effect when the mapped function is IO/network bound

IvanIvanoff: Also let's use the name `pmap`. Both implementations are concurrent and, if possible, parallel. So the name `pmap_concurrent` is not that meaningful.

tspenov: @IvanIvanoff will change name.
`System.online_schedulers` defaults to the machine cores so that might be too little.
We use these mostly `pmap` in clickhouse_dataloader, influx_dataloader, timescale_dataloader and for eth_spent which is again clickhouse.
We can put `max_concurrency` for db interactions `10` but all these are used at once in allProjects so it should be even lower. What do you think ?

IvanIvanoff: Old `pmap` is equivalent to old `pmap_concurrent` with max_concurrency infinity. I think we set leave the default to `2 * System.online_schedulers` and to explicitly set the concurrency where needed to something reasonable for that case. i.e. in the case of databases 10-15 should be ok, I think.

tspenov: The problem here is it is used in 4 places and all 4 places are used in allProjects query - so setting it to 10-15 means 10-15 * 4 but in the other case it was infinity so should be ok, I guess

IvanIvanoff: We have a connection pool, so it will be min(10-15 * 4, pool_size)

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -45,16 +45,41 @@ defmodule Sanbase.Signals.Scheduler do
...
+    |> Sanbase.Parallel.pmap(
+      fn %UserTrigger{} = user_trigger ->
+        [
+          %HistoricalActivity{}
+          |> HistoricalActivity.changeset(%{
+            user_id: user_trigger.user_id,
+            user_trigger_id: user_trigger.id,
+            payload: user_trigger.trigger.settings.payload
+          })
+          |> Sanbase.Repo.insert!()
```
IvanIvanoff: should we abort persisting the other signals in case the first one fails?

lib/sanbase/signals/historical_activity.ex
```diff
@@ -0,0 +1,28 @@
...
+  import Ecto.Changeset
+  alias Sanbase.Signals.UserTrigger
+  alias Sanbase.Auth.User
+
+  alias __MODULE__
+
+  schema "signals_historical_activity" do
+    belongs_to(:user, User)
+    belongs_to(:user_trigger, UserTrigger)
+    field(:payload, :map)
```
IvanIvanoff: I think we'll need a `triggered_datetime` as the `created_at` and `updated_at` are carrying metadata for the row itself and not the row's meaning (my opinion).

lib/sanbase/utils/parallel.ex
```diff
@@ -78,16 +46,16 @@ defmodule Sanbase.Parallel do
...
     filter_func = fn x -> {func.(x), x} end
 
-    pmap_concurrent(collection, filter_func, opts)
+    pmap(collection, filter_func, opts)
     |> Enum.filter(fn {bool, _item} -> bool === true end)
     |> Enum.map(fn {_bool, item} -> item end)
   end
 
-  def preject_concurrent(collection, func, opts \\ []) when is_function(func, 1) do
+  def preject(collection, func, opts \\ []) when is_function(func, 1) do
```
IvanIvanoff: now that I see this it looks like mistyped `project`. On a second though, do we need the `p` prefix as each function is in the `Parallel` module and that's pretty self explanatory

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -19,12 +19,15 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
...
+      fn organizations ->
+        {:ok, dev_activity} = Clickhouse.Github.total_dev_activity(organizations, from, to)
 
-      dev_activity
-      |> Enum.map(fn {organization, dev_activity} -> {organization, dev_activity / days} end)
-    end)
+        dev_activity
+        |> Enum.map(fn {organization, dev_activity} -> {organization, dev_activity / days} end)
+      end,
+      map_type: :flat_map
```
IvanIvanoff: we need to add `max_concurrency: 10-15`

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -34,7 +37,7 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     eth_spent =
       eth_addresses(args)
       |> Enum.chunk_every(10)
-      |> Sanbase.Parallel.flat_pmap(&eth_spent(&1, args))
+      |> Sanbase.Parallel.pmap(&eth_spent(&1, args), map_type: :flat_map)
```
IvanIvanoff: max_concurrency: 10-15 also

lib/sanbase_web/graphql/dataloader/timescaledb_dataloader.ex
```diff
@@ -14,15 +14,18 @@ defmodule SanbaseWeb.Graphql.TimescaledbDataloader do
...
-      |> Enum.map(fn {contract_address, addresses} ->
-        {contract_address, addresses}
-      end)
-    end)
+        daily_active_addresses
+        |> Enum.map(fn {contract_address, addresses} ->
+          {contract_address, addresses}
+        end)
+      end,
+      map_type: :flat_map
```
IvanIvanoff: max_concurrency: 10-15 also

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
```diff
@@ -100,7 +102,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectTransactionsResolver do
         _resolution
       ) do
     Project.List.erc20_projects()
-    |> Sanbase.Parallel.pmap(&calculate_eth_spent_over_time_cached(&1, from, to, interval).(),
+    |> Sanbase.Parallel.pmap(
+      &calculate_eth_spent_over_time_cached(&1, from, to, interval).(),
```
IvanIvanoff: hm.. Previously this was spawning hundreds of processes. I think we can put `max_concurrency: 50`, despite the ClickHouse pool size is 30. Also put a comment

priv/repo/migrations/20190225144858_add_signals_historical_activity_table.exs
```diff
@@ -0,0 +1,16 @@
...
+  def change do
+    create table(@table) do
+      add(:user_id, references("users"), null: false)
+      add(:user_trigger_id, references("user_triggers"), null: false)
+      add(:payload, :jsonb)
+
+      timestamps()
+    end
+
+    create(index(@table, [:user_id]))
```
IvanIvanoff: could probably want an index on the user trigger, too?

tspenov: For the timeline we need only to fetch by user. If we want history of all activity for given trigger - then we need this too, yes @Partysun what do you think ?

lib/sanbase/signals/historical_activity.ex
tspenov: will add `triggered_at` and remove the other timestamps()
hm, thats a good point I think we can use max `last_triggered` from `Trigger` so we don't have difference in time but still separate the sending from saving to db.

lib/sanbase/signals/evaluator/scheduler.ex
tspenov: I don't think so. What is your idea ?
Also I am thinking on changing it with pmap, chunk and `insert_all` - I think will be more efficient.

lib/sanbase/utils/parallel.ex
tspenov: sounds good

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -45,34 +48,65 @@ defmodule Sanbase.Signals.Scheduler do
...
+      %UserTrigger{
+        trigger: %{
+          settings: %{triggered?: true, payload: _payload}
+        }
+      } ->
+        true
+
+      _ ->
+        false
+    end)
```
IvanIvanoff: why not just ` %UserTrigger{trigger: %{settings: %{triggered?: triggered?}}} -> triggered?` ?

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -100,6 +134,35 @@ defmodule Sanbase.Signals.Scheduler do
...
+          user_trigger_id: id,
+          user_id: user_id,
+          payload: payload,
+          triggered_at: max_last_triggered(last_triggered)
+        }
+
+      _ ->
+        nil
+    end)
+    |> Enum.reject(fn persist_args -> persist_args == nil end)
```
IvanIvanoff: `Enum.reject(&is_nil/1)`

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -100,6 +134,35 @@ defmodule Sanbase.Signals.Scheduler do
...
+          settings: %{triggered?: true, payload: payload},
+          last_triggered: last_triggered
+        }
+      }
+      when is_non_empty_map(last_triggered) ->
+        %{
+          user_trigger_id: id,
+          user_id: user_id,
+          payload: payload,
+          triggered_at: max_last_triggered(last_triggered)
```
IvanIvanoff: We persist many different payloads here (`payload` is a map with possibly many values) but a single last_triggered. What is the reason behind this?

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -34,7 +40,10 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     eth_spent =
       eth_addresses(args)
       |> Enum.chunk_every(10)
-      |> Sanbase.Parallel.flat_pmap(&eth_spent(&1, args))
+      |> Sanbase.Parallel.map(&eth_spent(&1, args),
+        map_type: :flat_map,
+        max_concurrency: @max_concurrency
```
IvanIvanoff: Note that this is with much less concurrency than before this way - previously all chunks were processed at the same time (the chunks are hundreds) but here only 100 projects can be processed at a time

lib/sanbase_web/graphql/dataloader/influxdb_dataloader.ex
```diff
@@ -16,22 +18,25 @@ defmodule SanbaseWeb.Graphql.InfluxdbDataloader do
 
     measurements
     |> Enum.chunk_every(100)
-    |> Sanbase.Parallel.pmap(fn measurements ->
-      volumes_last_24h = Prices.Store.fetch_mean_volume(measurements, yesterday, now)
+    |> Sanbase.Parallel.map(
+      fn measurements ->
```
IvanIvanoff: You can manually put this `fn` on the same row as the map and the formatter should not change it. This way you'll reduce the nesting by 1 level as now you have a lot of changes because of this.

lib/sanbase/signals/evaluator/scheduler.ex
tspenov: if `triggered?` key does not exist it will raise

IvanIvanoff: There is no signal without `triggered?` field, right?

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -45,34 +48,65 @@ defmodule Sanbase.Signals.Scheduler do
...
+            {:ok, updated_user_trigger} = update_last_triggered(user_trigger, list)
+
+            user_trigger =
+              put_in(
+                user_trigger.trigger.last_triggered,
+                updated_user_trigger.trigger.last_triggered
+              )
 
-            list
+            {user_trigger, list}
```
IvanIvanoff: I am confused why this started returning `updated_user_triggers` while it's function is only to send.

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -100,6 +134,35 @@ defmodule Sanbase.Signals.Scheduler do
...
+          triggered_at: max_last_triggered(last_triggered)
+        }
+
+      _ ->
+        nil
+    end)
+    |> Enum.reject(fn persist_args -> persist_args == nil end)
+    |> Enum.chunk_every(200)
+    |> Enum.each(fn chunk ->
+      Sanbase.Repo.insert_all(HistoricalActivity, chunk, on_conflict: :nothing)
```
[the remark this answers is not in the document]
tspenov: there is no unique constraint on the fields so conflicts shouldn't happen. In this case the default behaviour of raising should be fine and the last opt should be removed.

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -45,34 +48,65 @@ defmodule Sanbase.Signals.Scheduler do
   # Private functions
 
   defp run(type) do
-    type
-    |> UserTrigger.get_triggers_by_type()
-    |> Evaluator.run()
-    |> send_and_mark_as_sent()
+    {updated_user_triggers, sent_list_results} =
```
IvanIvanoff: The function returns `user_trigger` (single) but here it's misleadingly called `updated_user_triggers` (plural).

lib/sanbase/signals/evaluator/scheduler.ex
IvanIvanoff: I think it will be much better to separate both steps - first persist the triggered signals and only after that send them in a different phase in the pipeline. This also has the nice benefit that if something with telegram fails and signal is not sent it will still be persisted.

tspenov: Here it also updates the `last_triggered` field. So I can get `last_triggered` without having to select it.
Why should we persist something that is not sent and why should sending wait for persisting in db ?

lib/sanbase/signals/evaluator/scheduler.ex
tspenov: We have one trigger - so there will be one user trigger activity. Behind that frontend will get a map of the payloads so it can show all things that are send (but that will be the details).

lib/sanbase/signals/evaluator/scheduler.ex
tspenov: no, there isn't. updated

tspenov:
@IvanIvanoff changes to the Prallel module are extracted in separate PR so we can revert easy if we they make things slower: https://github.com/santiment/sanbase2/pull/1026

lib/sanbase/signals/evaluator/scheduler.ex
tspenov: it returns https://github.com/santiment/sanbase2/pull/1005/files/0fdec86193b12d95df0fb30ccd9918857ecf1d5d#diff-272226fdcdfc828cb77104e1ed90b8c8R109 which returns exactly a tuple of 2 lists and the first element `updated_user_triggers `

PR #1007, opened 2019-02-27
Add pagination and cache to all_insights query

IvanIvanoff:
Also all_insights now returns only the published posts but not the
current user's not published ones

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/voting/post.ex
+++ b/lib/sanbase/voting/post.ex
@@ -134,6 +134,17 @@ defmodule Sanbase.Voting.Post do
     |> get_only_published_or_own_posts(user_id)
   end
 
+  def published_posts(page, page_size) do
+    from(
+      p in Post,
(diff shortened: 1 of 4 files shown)


valo:
Should we use a cursor based pagination? What will happen if a new insight is being added? If we put the insights on a time-line which updates in real time, the cursor based approach will be better?

IvanIvanoff:
@valo I suggested to add this as a fast fix and once we lower the rate limit we can work in peace on a better and more complex solution.

valo:
OK

tspenov:
👍 If the frontend is happy with the ordering by datetime.

DmitriVanGuard:
> 👍 If the frontend is happy with the ordering by datetime.

Sure! That's what we did on every time user fetches insights 😄

PR #1008, opened 2019-02-27
Secure that tag name cannot be nil. Remove tags without name

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/migrations/20190227153724_add_tag_name_null_constrain.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.AddTagNameNullConstrain do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tags) do
+      modify(:name, :string, null: true)
```
tspenov: `null: false`

IvanIvanoff: I had one job and I failed.

lib/sanbase/tag/tag.ex
```diff
@@ -22,24 +23,33 @@ defmodule Sanbase.Tag do
...
-    tags =
-      from(t in __MODULE__, where: t.name in ^tags)
-      |> Sanbase.Repo.all()
+      |> Enum.map(&changeset(%Tag{}, %{name: &1}))
+      |> Enum.map(&Repo.insert(&1, on_conflict: :nothing))
+      |> Enum.map(fn
+        {:ok, %Tag{name: name}} -> name
+        _ -> nil
+      end)
+      |> Enum.reject(&is_nil/1)
```
tspenov: I think this implementation is a step back. Check http://blog.plataformatec.com.br/2016/12/many-to-many-and-upserts/

IvanIvanoff: @tspenov reworked it according to the blogpost, but not with `insert_all` as I want it to pass through the changeset

priv/repo/migrations/20190227153610_drop_tags_with_null_name.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Sanbase.Repo.Migrations.DropTagsWithNullName do
+  use Ecto.Migration
+  import Ecto.Query
+
+  def up() do
+    from(t in Sanbase.Tag, where: is_nil(t.name))
+    |> Sanbase.delete_all()
```
tspenov: should be `Sanbase.Repo`

tspenov:
@IvanIvanoff You should run the migrations locally and commit the changed `structure.sql`

lib/sanbase/tag/tag.ex
```diff
@@ -22,21 +23,29 @@ defmodule Sanbase.Tag do
...
+    Ecto.Changeset.add_error(changeset, :tags, "Cannot add more than 10 tags for a record")
+  end
 
+  def put_tags(%Ecto.Changeset{} = changeset, %{tags: tags}) do
     tags =
-      from(t in __MODULE__, where: t.name in ^tags)
-      |> Sanbase.Repo.all()
+      tags
+      |> Enum.map(
+        &Repo.insert!(%__MODULE__{} |> changeset(%{name: &1}),
```
tspenov: I still prefer using `insert_all` instead of inserting one by one and hitting db multiple times.
Now that we have uniq constraint in the db you can just use the changeset for filter the valid tags.
```elixir
valid_tags = tags |> Enum.filter(fn tag -> changeset(%__MODULE__{}, %{name: tag}).valid? end)
```

```elixir
    valid_tags = tags |> Enum.filter(fn tag -> changeset(%__MODULE__{}, %{name: tag}).valid? end)

    params = valid_tags |> Enum.map(fn tag -> %{name: tag} end)
    Repo.insert_all(__MODULE__, params, on_conflict: :nothing, conflict_target: [:name])

    valid_tag_structs = from(t in __MODULE__, where: t.name in ^valid_tags) |> Sanbase.Repo.all()

    changeset
    |> put_assoc(:tags, valid_tag_structs)
```

when we upgrade to ecto 3 there is `:replace_all_except_primary_key` which we can use in `insert_all` so we can skip the last select.

IvanIvanoff: In that case the `insert_all` approach is acceptable as the check can be purely done by the changeset.

priv/repo/migrations/20190227153724_add_tag_name_null_constrain.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.AddTagNameNullConstrain do
+  use Ecto.Migration
+
+  def change do
```
tspenov: this needs to be reversible, so maybe
```elixir
  def up do
    alter table(:tags) do
      modify(:name, :string, null: false)
    end
  end

  def down do
    alter table(:tags) do
      modify(:name, :string, null: true)
    end
  end
```

IvanIvanoff: It is, you need to refresh your page

PR #1009, opened 2019-02-27
cursor API for user signals historical activity

tspenov:
Should be merged after: https://github.com/santiment/sanbase2/pull/1005

```graphql
{
    signalsHistoricalActivity(
      limit:1,
      cursor:{
        type:BEFORE,
        datetime:"2019-03-11T11:56:42.970284"
      }
    ) {
      cursor {
        before,
        after
      }
      activity {
        payload,
        triggeredAt,
        userTrigger {
          trigger {
            title,
            description
          }
        }
      }
    }
}
```

```graphql
{
  "data": {
    "signalsHistoricalActivity": {
      "activity": [
        {
          "payload": {
            "all": "Trending words for: `2019-03-11`\n\n```\nWord    | Score\n------- | -------\ntheta   | 704\nxlm     | 425\nqlc     | 224\nnxs     | 210\nmining  | 207\nnano    | 196\nherpes  | 178\nxrp     | 178\nasic    | 152\ntattoos | 128\n```\nMore info: http://localhost:4000/sonar\n"
          },
          "triggeredAt": "2019-03-11T11:56:41.970284Z",
          "userTrigger": {
            "trigger": {
              "description": null,
              "title": "alabala"
            }
          }
        }
      ],
      "cursor": {
        "after": "2019-03-11T11:56:41.970284Z",
        "before": "2019-03-11T11:56:41.970284Z"
      }
    }
  }
}
```

After cursor: takes latest signal activities.
```
    signalsHistoricalActivity(
      limit:1,
      cursor:{
        type:AFTER,
        datetime:"2019-03-11T11:56:42.970284Z"
      }
    )
```

Before cursor: takes older signal activities.
```
    signalsHistoricalActivity(
      limit:1,
      cursor:{
        type:BEFORE,
        datetime:"2019-03-11T11:56:42.970284Z"
      }
    )
```


IvanIvanoff:
You might want to check:
https://github.com/duffelhq/paginator
https://github.com/drewolson/scrivener

tspenov:
@IvanIvanoff Hmm `scrivener` is more used but I think `paginator` is exactly what we need here. Although not sure we need it cause with composed queries it comes quite nice.

tspenov:
I think the implementation is simple enough (~10LOC)  that it doesn't need to use external lib for now.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -849,6 +852,21 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+    * `before` cursor argument gives those executed before certain datetime 
+    * `after` cursor argument gives those executed after certain datetime
+    * `limit` argument defines the size of the page. Default value is 25
+    """
+    field :signals_historical_activity, :signal_historical_activity_paginated do
+      arg(:before, :naive_datetime)
+      arg(:after, :naive_datetime)
+      arg(:limit, :integer, default_value: 25)
+
+      resolve(&SignalsHistoricalActivityResolver.fetch_historical_activity_for/3)
```
IvanIvanoff: you need to put the JWT auth middleware as otherwise calling by anon users will cause a function clause match error as there won't be `current_user` and the resolver will raise an exception

lib/sanbase_web/graphql/schema/pagination_types.ex
```diff
@@ -0,0 +1,8 @@
+defmodule SanbaseWeb.Graphql.PaginationTypes do
+  use Absinthe.Schema.Notation
+
+  object :cursor do
+    field(:before, :naive_datetime)
+    field(:after, :naive_datetime)
```
IvanIvanoff: another way is to have `datetime` and another field that is enum with two possible values `BEFORE` and `AFTER` and you won't need checks if only one of them is passed

test/sanbase_web/graphql/triggers_api_test.exs
```diff
@@ -411,6 +412,180 @@ defmodule SanbaseWeb.Graphql.TriggersApiTest do
...
+
+    _oldest =
+      insert(:signals_historical_activity,
+        user: user,
+        user_trigger: user_trigger,
+        payload: %{"all" => "oldest"},
+        triggered_at: NaiveDateTime.from_iso8601!("2019-01-20T00:00:00")
+      )
+
+    sga1 =
```
IvanIvanoff: what is `sga` :smiley:

lib/sanbase_web/graphql/schema.ex
tspenov: yeah, left it out cause I moved it from currentUser :(

test/sanbase_web/graphql/triggers_api_test.exs
tspenov: wow, don't know, I probably meant `sha` like signal_historical_activity but something changed along the way :D

lib/sanbase_web/graphql/schema/pagination_types.ex
tspenov: @IvanIvanoff Updated using enum for cursor input arguments. Usage in PR description

lib/sanbase_web/graphql/schema/pagination_types.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  enum(:cursor_type, values: [:before, :after])
+
+  input_object :cursor_input do
+    field(:type, non_null(:cursor_type))
+    field(:datetime, non_null(:naive_datetime))
+  end
+
+  object :cursor do
+    field(:before, :naive_datetime)
```
IvanIvanoff: we use `datetime` in all APIs, not naive

tspenov: The problem is db uses `naive_datetime` so its a lot easier implementing it that way than converting the values to datetime.

tspenov: but will implement it with `datetime` type for the sake of consistency

PR #1013, opened 2019-03-01
Add daily active deposits metric

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/daily_active_deposits.ex
+++ b/lib/sanbase/clickhouse/daily_active_deposits.ex
@@ -1,0 +1,70 @@
+defmodule Sanbase.Clickhouse.DailyActiveDeposits do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate daily active deposits
+  """
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  import Sanbase.Math, only: [to_integer: 1]
+
+  alias Sanbase.DateTimeUtils
+
+  @type active_deposits :: %{
+          datetime: DateTime.t(),
+          active_deposits: non_neg_integer()
+        }
+
+  @spec active_deposits(
+          String.t(),
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(active_deposits)} | {:error, String.t()}
+  def active_deposits(contract, from, to, interval) do
+    {query, args} = active_deposits_query(contract, from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [dt, active_deposits] ->
+      %{
+        datetime: DateTime.from_unix!(dt),
+        active_deposits: active_deposits |> to_integer()
+      }
+    end)
+  end
+
+  defp active_deposits_query(contract, from, to, interval) do
+    contract = String.downcase(contract)
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
+    span = div(to_datetime_unix - from_datetime_unix, interval) |> max(1)
+
+    query = """
+    SELECT
+      toUnixTimestamp(time) AS time,
+      SUM(value) AS active_deposits
+    FROM (
+      SELECT
+        toDateTime(intDiv(toUInt32(?4 + number * ?1), ?1) * ?1) AS time,
+        toUInt32(0) AS value
+      FROM numbers(?2)
+
+      UNION ALL
+
+      SELECT
+        toDateTime(intDiv(toUInt32(dt), ?1) * ?1) AS time,
+        SUM(total_addresses) AS value
+      FROM daily_active_deposits
+      PREWHERE
+        contract = ?3 AND
+        dt >= toDateTime(?4) AND
+        dt <= toDateTime(?5)
+      GROUP BY time
+    )
+    GROUP BY time
+    ORDER BY time
+    """
+
+    args = [interval, span, contract, from_datetime_unix, to_datetime_unix]
+
+    {query, args}
+  end
+end
--- a/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
@@ -3,7 +3,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
 
   alias Sanbase.Model.Project
   alias Sanbase.DateTimeUtils
-  alias Sanbase.Clickhouse.{HistoricalBalance, MVRV, NetworkGrowth}
+  alias SanbaseWeb.Graphql.Helpers.Utils
+  alias Sanbase.Clickhouse.{DailyActiveDeposits, HistoricalBalance, MVRV, NetworkGrowth}
+
+  @one_hour_seconds 3600
 
   def network_growth(_root, args, _resolution) do
     interval = DateTimeUtils.compound_duration_to_seconds(args.interval)
@@ -33,6 +36,35 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
         )
 
         {:error, "Can't calculate MVRV ratio"}
+    end
+  end
+
+  def daily_active_deposits(
+        _root,
+        %{slug: slug, from: from, to: to, interval: interval},
+        _resolution
+      ) do
+    with {:ok, contract, _} <- Project.contract_info_by_slug(slug),
+         {:ok, from, to, interval} <-
+           Utils.calibrate_interval(
+             DailyActiveDeposits,
+             contract,
+             from,
+             to,
+             interval,
+             @one_hour_seconds,
+             50
+           ),
+         {:ok, active_deposits} <-
+           DailyActiveDeposits.active_deposits(contract, from, to, interval) do
+      {:ok, active_deposits}
+    else
+      {:error, error} ->
+        error_msg =
+          "Can't calculate daily active deposits for project with coinmarketcap_id: #{slug}."
+
+        Logger.warn(error_msg <> " Reason: #{inspect(error)}")
+        {:error, error_msg}
     end
   end
 
--- a/lib/sanbase_web/graphql/schema.ex
+++ b/lib/sanbase_web/graphql/schema.ex
@@ -428,6 +428,21 @@ defmodule SanbaseWeb.Graphql.Schema do
       cache_resolve(&EtherbiResolver.daily_active_addresses/3)
     end
 
+    @desc ~s"""
+    Fetch daily active deposits for a project within a given time period.
+    Projects are referred to by a unique identifier (slug).
+    """
+    field :daily_active_deposits, list_of(:active_deposits) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "1d")
+
+      complexity(&Complexity.from_to_interval/3)
+      middleware(ApiTimeframeRestriction)
+      cache_resolve(&ClickhouseResolver.daily_active_deposits/3)
+    end
+
     @desc "Fetch the currently running poll."
     field :current_poll, :poll do
       cache_resolve(&VotingResolver.current_poll/3)
--- a/lib/sanbase_web/graphql/schema/clickhouse_types.ex
+++ b/lib/sanbase_web/graphql/schema/clickhouse_types.ex
@@ -1,18 +1,23 @@
 defmodule SanbaseWeb.Graphql.ClickhouseTypes do
   use Absinthe.Schema.Notation
 
+  object :active_deposits do
+    field(:datetime, non_null(:datetime))
+    field(:active_deposits, non_null(:integer))
+  end
+
   object :historical_balance do
     field(:datetime, non_null(:datetime))
     field(:balance, :float)
   end
 
-  object :network_growth do
-    field(:datetime, non_null(:datetime))
-    field(:new_addresses, :integer)
-  end
-
   object :mvrv_ratio do
     field(:datetime, non_null(:datetime))
     field(:ratio, :float)
+  end
+
+  object :network_growth do
+    field(:datetime, non_null(:datetime))
+    field(:new_addresses, :integer)
   end
 end
--- a/test/sanbase/clickhouse/daily_active_deposits_test.exs
+++ b/test/sanbase/clickhouse/daily_active_deposits_test.exs
@@ -1,0 +1,95 @@
+defmodule SanbaseWeb.Clickhouse.DailyActiveDepositsTest do
+  use Sanbase.DataCase
+  import Mock
+  import Sanbase.Factory
+  import Sanbase.DateTimeUtils, only: [from_iso8601_to_unix!: 1, from_iso8601!: 1]
+  alias Sanbase.Clickhouse.DailyActiveDeposits
+  require Sanbase.ClickhouseRepo
+
+  setup do
+    project =
+      insert(:project, %{
+        coinmarketcap_id: "santiment",
+        ticker: "SAN",
+        main_contract_address: "0x123"
+      })
+
+    [
+      contract: project.main_contract_address,
+      from: from_iso8601!("2019-01-01T00:00:00Z"),
+      to: from_iso8601!("2019-01-03T00:00:00Z"),
+      interval: "1d"
+    ]
+  end
+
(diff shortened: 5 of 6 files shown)


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -30,6 +30,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
       {:error, "Can't calculate network growth"}
   end
 
+  def daily_active_deposits(_root, args, _resolution) do
+    interval = DateTimeUtils.compound_duration_to_seconds(args.interval)
```
IvanIvanoff: @tspenov @kalkov let's decide on a uniform way whether to transform the interval in the resolver or the function using it. My opinion is that the resolvers must be the cleanest function and should just call the modules.
One discrepancy that happens if we transform here is that all functions except these communicating with influxdb will transform the arguments, as the `GROUP BY` in influxb is the same string that is passed here.

IvanIvanoff: Also the `calibrate_interval` that is used a lot also returns a string `"<seconds>s"`

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,60 @@
...
+  @type active_deposits :: %{
+          datetime: %DateTime{},
+          active_deposits: non_neg_integer()
+        }
+
+  @spec active_deposits(
+          String.t(),
+          %DateTime{},
+          %DateTime{},
+          Number.t()
```
IvanIvanoff: You can check in iex that `t Number.t` says `Could not load module Number, got: nofile`
Number type specs are on of:
- `non_neg_integer()`
- `pos_integer()`
- `neg_integer()`
- `integer()`
- `float()`

There's also `number()` which is an alias for `integer() | float()`

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -30,6 +30,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+         {:ok, active_deposits} <-
+           DailyActiveDeposits.active_deposits(contract, args.from, args.to, interval) do
+      {:ok, active_deposits}
+    else
+      error ->
+        Logger.error("Can't calculate daily active deposits. Reason: #{inspect(error)}")
+
+        {:error, "Can't calculate daily active deposits"}
+    end
+  rescue
```
IvanIvanoff: I really dislike this pattern we have at many places. I suggest to leave it like this but create a favro card that handles this on a middleware level (that we could probably plug for all queries?).
These 2 blogposts are on that topic:
1. Introducing the problem and solving it this way: https://tech.decisiv.com/handling-elixir-exceptions-in-absinthe-9656a9108c2d
2. Solving it with a middleware: https://tech.decisiv.com/handling-elixir-exceptions-in-absinthe-part-2-using-middleware-d20e7df7cc62

The proposed solution seems to wrap all middlewares in such safety box, though I think we'll need it only in some places and let other queries just crash.

The main place where we must handle all errors is the `all_*` queries, because if a single field there fails, the whole request fails and the overview page does not load.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -416,6 +416,21 @@ defmodule SanbaseWeb.Graphql.Schema do
...
 
+    @desc ~s"""
+    Fetch daily active deposits for a project within a given time period.
+    Projects are referred to by a unique identifier (slug).
+    """
+    field :daily_active_deposits, list_of(:active_deposits) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "1d")
```
IvanIvanoff: `interval` can also have a default value of `""`. In that case you can use the `calibrate_interval` function that is crafted so it transforms the interval in such a way that in the time period `from - to` there are a specific number of data points, so it looks good on the chart that the frontend draws.

IvanIvanoff: Manually calling and handling `calibrate_interval` in each resolver is also not nice, as it's very verbose. This is again one thing we could probably modify with a middleware.
Saying this I figured out that we have an issue - at some places we handle `""` and set a specific value to it, but in all calculations for complexity the `""` interval is handled as if it is `"1d"`

test/sanbase_web/graphql/clickhouse/active_deposits_test.exs
```diff
@@ -0,0 +1,55 @@
...
+
+  setup do
+    project = insert(:project, %{main_contract_address: "0x123"})
+
+    [
+      project: project
+    ]
+  end
+
+  test "active deposits works", context do
```
IvanIvanoff: you can locally run `MIX_TEST=test mix coveralls.html` and check the `cover/<name>.html` file so you can see what other parts you need to cover in your test (the error cases and empty cases, for example).

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
kalkov: @IvanIvanoff Good arguments, my vote goes to not transform in the resolver.

lib/sanbase/clickhouse/daily_active_deposits.ex
kalkov: Tried to guess it, failed.

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
tspenov: Also agree on keeping short and sweat the resolvers

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -6,7 +6,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   alias Sanbase.Clickhouse.Erc20Transfers
   alias Sanbase.Model.Project
 
-  alias Sanbase.Clickhouse.NetworkGrowth
+  alias Sanbase.Clickhouse.{ NetworkGrowth, DailyActiveDeposits }
```
tspenov: `{NetworkGrowth, DailyActiveDeposits}` extra space on both sides.
We don't have a style guide  but we run `mix format` before commit - it should fix these style things

tspenov: here is a snippet (that I've borrowed form @IvanIvanoff ) to put in `.git/hooks/pre-commit`
```bash
#!/bin/bash

cd `git rev-parse --show-toplevel`
echo "Running pre-commit hook: Check elixir formatting with mix format"
mix format --check-formatted
if [ $? == 1 ]; then
  echo "mix format --check-formatted found formatting issues."
  exit 1
fi
```

test/sanbase_web/graphql/clickhouse/active_deposits_test.exs
valo: I am wondering if we should plug the coverage stats to GitHub

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: On the other hand, I think it's ok to leave the complexity calculations like this, as the current behavior is more restrictive and will not allow some queries, which is ok.

test/sanbase_web/graphql/clickhouse/active_deposits_test.exs
kalkov: @valo You mean to add a coverage badge?

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,61 @@
...
+  alias Sanbase.DateTimeUtils
+
+  @type active_deposits :: %{
+          datetime: %DateTime{},
+          active_deposits: non_neg_integer()
+        }
+
+  @spec active_deposits(
+          String.t(),
+          %DateTime{},
```
IvanIvanoff: for some reasons we have this `%DateTime{}` in the specs, though there is the `DateTime.t()` available

test/sanbase_web/graphql/clickhouse/active_deposits_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule SanbaseWeb.Graphql.ClickhouseActiveDepositsTest do
+  use SanbaseWeb.ConnCase
+  require Sanbase.ClickhouseRepo
```
IvanIvanoff: Here you're only mocking it so there should be no need to `require` it. `require` is needed when you use a macro from that module

test/sanbase_web/graphql/clickhouse/active_deposits_test.exs
valo: Sure, let's try it. This service https://coveralls.io supports open source repos for free, so we can try it too. I think that will be a separate task and not included in this PR.

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,61 @@
...
+          %DateTime{},
+          String.t()
+        ) :: {:ok, list(active_deposits)} | {:error, String.t()}
+  def active_deposits(contract, from, to, interval) do
```
IvanIvanoff: You can add `active_deposits_query` function

test/sanbase_web/graphql/clickhouse/daily_active_deposits_test.exs
```diff
@@ -0,0 +1,138 @@
...
+    with_mock DailyActiveDeposits,
+      active_deposits: fn _, _, _, _ ->
+        {:ok,
+         [
+           %{active_deposits: 100, datetime: from_iso8601!("2019-01-01T00:00:00Z")},
+           %{active_deposits: 200, datetime: from_iso8601!("2019-01-02T00:00:00Z")}
+         ]}
+      end do
+      response = execute_query(context)
+      ratios = parse_response(response)
```
tspenov: this should be called probably `deposits`

tspenov:
👍 LGTM - small comment in test

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,69 @@
...
+  Uses ClickHouse to calculate daily active deposits
+  """
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  import Sanbase.Math, only: [to_integer: 1]
+
+  alias Sanbase.DateTimeUtils
+
+  @type active_deposits :: %{
+          datetime: %DateTime{},
```
IvanIvanoff: Here the datetime type can be changed, too

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,69 @@
...
+        SUM(total_addresses) AS value
+      FROM daily_active_deposits
+      PREWHERE
+        contract = ?3 AND
+        dt >= toDateTime(?4) AND
+        dt <= toDateTime(?5)
+      GROUP BY time
+    )
+    GROUP BY dt
+    ORDER BY dt
```
IvanIvanoff: I am a little bit confused by the redefinitions of `dt` and `time`. Is it possible to do something like `toUnixTimestamp(time) as time`?

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -36,6 +39,34 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+    else
+      {:error, error} ->
+        Logger.warn(
+          "Can't calculate daily active deposits for project with coinmarketcap_id: #{args.slug}. Reason: #{
+            inspect(error)
+          }"
+        )
+
+        {:error,
+         "Can't calculate daily active deposits for project with coinmarketcap_id: #{args.slug}."}
```
IvanIvanoff:
```elixir
error_msg = "Can't calculate daily active deposits for project with coinmarketcap_id: #{args.slug}."
Logger.warn(error_msg <> " Reason: #{inspect(error)}"
{:error, error_msg}
```

lib/sanbase_web/graphql/schema.ex
```diff
@@ -428,6 +428,21 @@ defmodule SanbaseWeb.Graphql.Schema do
...
 
+    @desc ~s"""
+    Fetch daily active deposits for a project within a given time period.
+    Projects are referred to by a unique identifier (slug).
+    """
+    field :daily_active_deposits, list_of(:active_deposits) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, non_null(:string), default_value: "1d")
```
IvanIvanoff: GraphiQL is confused by non_null + default values. Not sure if it's an issue with graphiql though

lib/sanbase_web/graphql/schema.ex
```diff
@@ -428,6 +428,21 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+
+      complexity(&Complexity.from_to_interval/3)
+      middleware(ApiTimeframeRestriction, %{allow_historical_data: true})
```
IvanIvanoff: Is this the desired access or we want to cut off the historical data, too?

lib/sanbase_web/graphql/schema/clickhouse_types.ex
```diff
@@ -1,18 +1,23 @@
 defmodule SanbaseWeb.Graphql.ClickhouseTypes do
   use Absinthe.Schema.Notation
 
-  object :historical_balance do
+  object :active_deposits do
     field(:datetime, non_null(:datetime))
-    field(:balance, :float)
+    field(:active_deposits, :integer)
```
IvanIvanoff: If we're filling empty values with `0` (the correct behavior) then it can be non_null here

test/sanbase/clickhouse/daily_active_deposits_test.exs
```diff
@@ -0,0 +1,95 @@
...
+  import Sanbase.DateTimeUtils, only: [from_iso8601_to_unix!: 1, from_iso8601!: 1]
+  alias Sanbase.Clickhouse.DailyActiveDeposits
+  require Sanbase.ClickhouseRepo
+
+  setup do
+    project =
+      insert(:project, %{
+        coinmarketcap_id: "ethereum",
+        ticker: "ETH",
+        main_contract_address: "0x123"
```
IvanIvanoff: That is confusing because when you call `contract_info` on `"ethereum"` it has a different effect - it does not check the contract field but returns `ETH` directly.

lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -0,0 +1,69 @@
...
+      }
+    end)
+  end
+
+  defp active_deposits_query(contract, from, to, interval) do
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
+    span = div(to_datetime_unix - from_datetime_unix, interval) |> max(1)
+
```
IvanIvanoff: You'll need to downcase the contract as this is the behavior in ClickHouse. We have (mostly) downcased addresses in our DB but for example the `contract_info` for ethereum would give `ETH` while the clickhouse contract for it is `eth`

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -36,6 +39,34 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     end
   end
 
+  def daily_active_deposits(_root, args, _resolution) do
```
IvanIvanoff: In most places we pattern match the args in the function head. It is arguably nicer to read and you'll fail early with a more meaningful error when not correct params are given.

lib/sanbase_web/graphql/schema.ex
kalkov: I think not, fixed

lib/sanbase/clickhouse/daily_active_deposits.ex
kalkov: Addressed, please check the fix

PR #1014, opened 2019-03-01
Improve setup

kalkov:
#### Summary

In development docker was not compiling assets(needed for admin panel). The reason was that `brunch` was not run. Now it should be fixed.

Also removed yarn dependency from `Dockerfile`. We must test that this change will not broke something on staging. 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,7 @@ podTemplate(label: 'sanbase-builder', containers: [
 
         sh "docker build -t sanbase-test:${scmVars.GIT_COMMIT}-${env.BUILD_ID}-${env.CHANGE_ID} -f Dockerfile-test ."
         sh "docker run --rm --name test-postgres-${scmVars.GIT_COMMIT}-${env.BUILD_ID}-${env.CHANGE_ID} -d timescale/timescaledb:0.10.1-pg10"
-        sh "docker run --rm --name test-influxdb-${scmVars.GIT_COMMIT}-${env.BUILD_ID}-${env.CHANGE_ID} -d influxdb:1.4-alpine"
+        sh "docker run --rm --name test-influxdb-${scmVars.GIT_COMMIT}-${env.BUILD_ID}-${env.CHANGE_ID} -d influxdb:1.7-alpine"
         try {
           sh "docker run --rm \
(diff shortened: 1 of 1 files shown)


Dockerfile-test
```diff
@@ -1,7 +1,7 @@
 # Elixir and phoenix assets build image
 FROM elixir:1.8.0
 
-RUN apt-get update && apt-get install -y  git postgresql-client-9.6
+RUN apt-get update && apt-get install -y git postgresql-client
```
IvanIvanoff: It will be best to have this version tagged the same as in the devops repo and update together in the future.
Also we're running influxdb-1.7, so you can fix these things in the Jenkins file, too

kalkov: @IvanIvanoff Can you link the devops file?

IvanIvanoff: https://github.com/santiment/devops/blob/master/prod/rds/main.tf#L39-L40

bin/setup.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
 
 docker-compose build && \
-  docker-compose run sanbase sh -c "mix deps.get && mix ecto.setup_all && mix run priv/repo/seeds.exs && cd assets/; yarn install"
+  docker-compose run sanbase sh -c "mix deps.get &&
+                                    mix ecto.setup_all &&
+                                    mix run priv/repo/seeds.exs &&
+                                    cd assets &&
+                                    npm install &&
+                                    npx brunch build"
```
tspenov: typo - `npm brunch build` ?

kalkov: @tspenov Not a typo : https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b

Dockerfile-test
kalkov: @IvanIvanoff postgresql-client-9.6 is back

kalkov: @IvanIvanoff Updated [influx in Jenkinsfile](https://github.com/santiment/sanbase2/pull/1014/commits/f1eea6b9ccb582f76b760377b6bd5ea143c810d6)

PR #1021, opened 2019-03-05
Update elixir dependencies

IvanIvanoff:
#### Summary
Requesting review from the frontend team, too. This is because here the graphql server behavior will change to return errors with http code 200 instead of 400. For example when the complexity of a query is too high it will now return that info with http code 200.

Also now there won't be any default time parameters (`to` was defaulting to the current time in many cases). Currently I know that it is used in the /assets/all for the total market price query

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


IvanIvanoff:
Note to myself: We cannot probably merge this as Absinthe will compile the default `DateTime.utc_now()` values in `historyPrice` query and it will stop working.
In order to unblock this I need to remove the default param there and the @DmitriVanGuard or @Partysun need to change the `historyPrice` used in the `/assets/all` page to not depend on it.

PR #1022, opened 2019-03-05
Implement historical balances with the balances table

IvanIvanoff:
#### Summary
Example usage:
```elixir
iex(1)>Sanbase.Clickhouse.HistoricalBalance.EthBalance.historical_balance(
  "0x447442a426b7f83A3a647120edD7a98420CeFe75",
   Timex.shift(Timex.now(), days: -110),
   Timex.now(),
   "4w")
{:ok,
  [
    %{balance: 0.0457971968247546, datetime: #DateTime<2018-11-29 00:00:00Z>},
    %{balance: 0.19575028849984766, datetime: #DateTime<2018-12-27 00:00:00Z>},
    %{balance: 0.02576241474915534, datetime: #DateTime<2019-01-24 00:00:00Z>},
    %{balance: 0.17211438421983474, datetime: #DateTime<2019-02-21 00:00:00Z>}
  ]}
```

Example GraphQL:
Request:
```
{
  historicalBalance(
    address:"0x6dd5a9f47cfbc44c04a0a4452f0ba792ebfbcc9a",
    slug:"ethereum",
    from:"2017-07-01T00:00:00Z",
    to:"2019-02-16T00:00:00Z",
  	interval:"4w") {
      datetime
      balance
  }
}
```

Response:
```
{
  "data": {
    "historicalBalance": [
      {
        "balance": 45000.95027004,
        "datetime": "2017-07-13T00:00:00Z"
      },
      {
        "balance": 41000.947750039995,
        "datetime": "2017-08-10T00:00:00Z"
      },
      {
        "balance": 37000.94607004,
        "datetime": "2017-09-07T00:00:00Z"
      },
      {
        "balance": 36000.94312976,
        "datetime": "2017-10-05T00:00:00Z"
      },
      {
        "balance": 33999.28656088823,
        "datetime": "2017-11-02T00:00:00Z"
      },
...
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
@@ -1,0 +1,124 @@
+defmodule Sanbase.Clickhouse.HistoricalBalance.Erc20Balance do
+  @doc ~s"""
+  Returns the historical balances of given address and erc20 contract
+  in all intervals between two datetimes.
+  """
+
+  use Ecto.Schema
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+
+  @type historical_balance :: %{
+          datetime: non_neg_integer(),
+          balance: float
+        }
+
+  @table "erc20_balances"
+  schema @table do
+    field(:datetime, :utc_datetime, source: :dt)
+    field(:contract, :string)
+    field(:address, :string, source: :to)
+    field(:value, :float)
+    field(:sign, :integer)
+  end
+
+  @spec changeset(any(), any()) :: no_return()
+  def changeset(_, _attrs \\ %{}),
+    do: raise("Should not try to change eth daily active addresses")
+
+  @spec historical_balance(
+          String.t(),
+          String.t(),
+          non_neg_integer(),
+          DateTime.t(),
+          DateTime.t(),
+          non_neg_integer()
+        ) :: {:ok, list(historical_balance)}
+  def historical_balance(address, contract, token_decimals, from, to, interval) do
+    token_decimals = Sanbase.Math.ipow(10, token_decimals)
+    address = String.downcase(address)
+    contract = String.downcase(contract)
+
+    {query, args} = historical_balance_query(address, contract, from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [dt, value, sign] ->
+      %{
+        datetime: Sanbase.DateTimeUtils.from_erl!(dt),
+        balance: value / token_decimals,
+        sign: sign
+      }
+    end)
+    |> case do
+      {:ok, result} ->
+        # Clickhouse fills empty buckets with 0 while we need it filled with the last
+        # seen value. As the balance changes happen only when a transfer occurs
+        # then we need to fetch the whole history of changes in order to find the balance
+        result =
+          result
+          |> Enum.reduce({0, []}, fn
+            %{sign: 1, balance: balance, datetime: dt}, {_last_seen, acc} ->
+              {balance, [%{balance: balance, datetime: dt} | acc]}
+
+            %{sign: 0, datetime: dt}, {last_seen, acc} ->
+              {last_seen, [%{balance: last_seen, datetime: dt} | acc]}
+          end)
+          |> elem(1)
+          |> Enum.reverse()
+          |> Enum.drop_while(fn %{datetime: dt} -> DateTime.compare(dt, from) == :lt end)
+
+        {:ok, result}
+
+      error ->
+        error
+    end
+  end
+
+  @first_datetime ~N[2015-10-29 00:00:00]
+                  |> DateTime.from_naive!("Etc/UTC")
+                  |> DateTime.to_unix()
+  defp historical_balance_query(address, contract, _from, to, interval) do
+    interval = Sanbase.DateTimeUtils.compound_duration_to_seconds(interval)
+    to_unix = DateTime.to_unix(to)
+    span = div(to_unix - @first_datetime, interval) |> max(1)
+
+    query = """
+    SELECT time, SUM(value), toInt32(SUM(sign))
+      FROM (
+        SELECT
+          toDateTime(intDiv(toUInt32(?4 + number * ?1), ?1) * ?1) as time,
+          toFloat64(0) AS value,
+          toInt32(0) as sign
+        FROM numbers(?2)
+
+        UNION ALL
+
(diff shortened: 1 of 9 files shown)


lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
```diff
@@ -0,0 +1,124 @@
...
+      }
+    end)
+    |> case do
+      {:ok, result} ->
+        # Clickhouse fills empty buckets with 0 while we need it filled with the last
+        # seen value. As the balance changes happen only when a transfer occurs
+        # then we need to fetch the whole history of changes in order to find the balance
+        result =
+          result
+          |> Enum.reduce({0, []}, fn
```
tspenov: Lets extract this reducer and add unit tests with edge cases on it

lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
```diff
@@ -0,0 +1,124 @@
...
+
+        UNION ALL
+
+    SELECT toDateTime(intDiv(toUInt32(dt), ?1) * ?1) as time, argMax(value, dt), argMax(sign, dt)
+      FROM (
+        SELECT any(value) as value, dt, 1 as sign
+        FROM #{@table}
+        PREWHERE address = ?3
+        AND contract = ?4
+        AND sign = 1
```
tspenov: why only with `sign=1` ?

IvanIvanoff: `sign=-1` is the thing that is "removed from the stack" which means that it holds the old balance, `sign=1` is the new balance after the transfer

tspenov: maybe worth adding a comment to the schema field ?

tspenov:
lets also delete unused things here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/common.ex

lib/sanbase/clickhouse/historical_balance/eth_balance.ex
```diff
@@ -0,0 +1,106 @@
...
+
+        UNION ALL
+
+    SELECT toDateTime(intDiv(toUInt32(dt), ?1) * ?1) as time, argMax(value, dt), argMax(sign, dt)
+      FROM (
+        SELECT any(value) as value, dt, 1 as sign
+        FROM #{@table}
+        PREWHERE address = ?3
+        AND sign = 1
+        AND dt >= toDateTime(?4)
```
tspenov: To calculate the balances shouldn't we have to go through all records for the address from the beginning ?  Because the `value` is the balance change, right - not the actual balance at this moment. So we need all changes from the beginning to calculate the balance in a certain time.

IvanIvanoff: @tspenov `value` with `sign=1` is the current balance. If you want to find the change you need to find the difference between `value`s with the same address and datetime but different signs

IvanIvanoff: Also, it starts going from the beginning but for a different reason - to fill the gaps with the previous value

IvanIvanoff: @tspenov the datetime here is hardcoded as the first datetime that we have data for, so indeed we can remove it as this check is not doing anything

test/sanbase_web/graphql/clickhouse/historical_balances_test.exs
```diff
@@ -24,34 +37,45 @@ defmodule SanbaseWeb.Graphql.Clickhouse.HistoricalBalancesTest do
         {:ok,
          %{
            rows: [
-             [from_iso8601_to_unix!("2017-05-13T00:00:00Z"), 2000],
-             [from_iso8601_to_unix!("2017-05-14T00:00:00Z"), 1800],
-             [from_iso8601_to_unix!("2017-05-15T00:00:00Z"), 1800],
-             [from_iso8601_to_unix!("2017-05-16T00:00:00Z"), 1500],
-             [from_iso8601_to_unix!("2017-05-17T00:00:00Z"), 1900],
-             [from_iso8601_to_unix!("2017-05-18T00:00:00Z"), 1000]
+             [{{2017, 05, 11}, {0, 0, 0}}, :math.pow(10, 18) * 0, 1],
```
kalkov: Do we need so many examples?

IvanIvanoff: I think mocking 10 results from the database is not that much. Real scenarios will return much more, so it's ok to test not only with 2-3

lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
```diff
@@ -0,0 +1,113 @@
...
+    end)
+    |> case do
+      {:ok, result} ->
+        # Clickhouse fills empty buckets with 0 while we need it filled with the last
+        # seen value. As the balance changes happen only when a transfer occurs
+        # then we need to fetch the whole history of changes in order to find the balance
+        result =
+          result
+          |> fill_gaps_last_seen_balance()
+          |> Enum.drop_while(fn %{datetime: dt} -> DateTime.compare(dt, from) == :lt end)
```
tspenov: maybe an optimization will be instead of selecting for 4 years and dropping all before `from` to split in 2 queries. 1st to find the balance for the `from` bucket which will be the last `sign=1` event with date <= `from` and in the second query use `from >= toDateTime(from)` to take only the values between `from` and `to` ?

test/sanbase_web/graphql/clickhouse/historical_balances_test.exs
kalkov: I mean from a reader perspective, not performance one. It's much easier to read 2-3 records and spot the difference than 10 of which 2-3 are different, others similar. Also we need to support these tests.

lib/sanbase/clickhouse/historical_balance/erc20_balance.ex
IvanIvanoff: We can, but let's first see if it is slow on production. It cannot be done in a subquery as this new `from` will affect the `span`. It's questionable if having 2 queries will be faster as the transporting/connecting/transferring data will make a huge difference. And this optimization will have a measurable effect only for wallets that have many transfers - they are not that much and if people use them a lot, they will be served from the cache most of the time

tspenov: cool, lets try it like this and optimize later if needed

PR #1023, opened 2019-03-05
Calibrate the GQL time intervals with a middleware

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/middlewares/calibrate_interval.ex
+++ b/lib/sanbase_web/graphql/middlewares/calibrate_interval.ex
@@ -1,0 +1,96 @@
+defmodule SanbaseWeb.Graphql.Middlewares.CalibrateInterval do
+  @moduledoc """
+  TODO
+  """
+
+  @behaviour Absinthe.Middleware
+
+  alias Absinthe.Resolution
+
+  alias Sanbase.Blockchain.{
+    TokenAgeConsumed,
+    DailyActiveAddresses,
+    TransactionVolume,
+    TokenCirculation,
+    ExchangeFundsFlow
+  }
+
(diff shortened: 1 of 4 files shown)


IvanIvanoff:
@tspenov @kalkov an issue I have with this approach is that the `calibrate_interval` and the calculations are done in different phases so they need both to access the DB to retrieve some info.
Another bad thing is that the cache won't affect the calibration, so it will be executed every time, so even cached queries will hit the database, which is a big no.

Do you think we can think of an elegant way the keeps the calibration separate and the resolvers clean but without any performance issues?

lib/sanbase_web/graphql/middlewares/calibrate_interval.ex
```diff
@@ -0,0 +1,96 @@
...
+      {:ok, from, to, interval} =
+        Utils.calibrate_interval(
+          Sanbase.Clickhouse.Github,
+          github_organization,
+          from,
+          to,
+          "",
+          min_interval_seconds,
+          data_points_count
+        )
```
valo: Why do we need to special case the GitHub module?

IvanIvanoff: Because it's the only module that needs to extract the organization from the slug to check the `first_datetime`

valo: May be we can define some unified behavior where we have a function `first_datetime` and we can call this for a slug?

IvanIvanoff: yeah, we could pass the `slug` and move the slug to organization extraction to the `first_datetime` function

PR #1025, opened 2019-03-06
Add complexity middlewares to more queries. Rearrange middlewares so the complexity is first

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -630,10 +630,9 @@ defmodule SanbaseWeb.Graphql.Schema do
       arg(:to, :datetime)
       arg(:interval, non_null(:string), default_value: "1d")
 
+      complexity(&Complexity.from_to_interval/3)
       middleware(ApiTimeframeRestriction)
-
-      complexity(&TechIndicatorsComplexity.topic_search/3)
-      resolve(&TechIndicatorsResolver.topic_search/3)
+      cache_resolve(&TechIndicatorsResolver.topic_search/3)
```
IvanIvanoff: Let's see how it performs with this query. For such cases we can try to use a different cache backend (Cachex) as it support LRU cache with a limit of N entries, which fits perfectly this use case. In order to use this the `cache_resolve` macro needs to be able to be configurable to use a different cache provider

PR #1026, opened 2019-03-06
Refactor parallel module

tspenov:
Extracted into a separate PR because we might experience performance issues going from parallel execution without `max_concurrency` to one with fixed.


lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -2,6 +2,8 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
   alias Sanbase.Clickhouse
   alias Sanbase.Model.Project
 
+  @max_concurrency 10
```
IvanIvanoff: Let's increase it to match the previous behavior?

lib/sanbase_web/graphql/dataloader/influxdb_dataloader.ex
```diff
@@ -16,22 +18,25 @@ defmodule SanbaseWeb.Graphql.InfluxdbDataloader do
 
     measurements
     |> Enum.chunk_every(100)
-    |> Sanbase.Parallel.pmap(fn measurements ->
-      volumes_last_24h = Prices.Store.fetch_mean_volume(measurements, yesterday, now)
+    |> Sanbase.Parallel.map(
+      fn measurements ->
```
IvanIvanoff: you can put this on the same line as `map` to reduce the nesting

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -34,7 +40,10 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     eth_spent =
       eth_addresses(args)
       |> Enum.chunk_every(10)
-      |> Sanbase.Parallel.flat_pmap(&eth_spent(&1, args))
+      |> Sanbase.Parallel.map(&eth_spent(&1, args),
+        map_type: :flat_map,
+        max_concurrency: @max_concurrency
```
IvanIvanoff: Here only 100 projects will be calculated in parallel instead of the previous behavior of all projects. It's arguable what is better but let's keep the behavior for now

IvanIvanoff:
@tspenov what's the status here?

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
tspenov: increased to 100

tspenov:
@IvanIvanoff resolved conflicts and renamed newly added

PR #1027, opened 2019-03-06
Add MVRV metric

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/mvrv.ex
+++ b/lib/sanbase/clickhouse/mvrv.ex
@@ -1,0 +1,52 @@
+defmodule Sanbase.Clickhouse.MVRV do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate MVRV Ratio(Market-Value-to-Realized-Value)
+  """
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  alias Sanbase.DateTimeUtils
+  alias Sanbase.Model.Project
+
+  @type mvrv_ratio :: %{
+          datetime: %DateTime{},
+          ratio: float
+        }
+
+  @spec mvrv_ratio(
+          String.t(),
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(mvrv_ratio)} | {:error, String.t()}
+  def mvrv_ratio(slug, from, to, interval) do
+    ticker = Project.ticker_by_slug(slug)
+    ticker_slug = "#{ticker}_#{slug}"
+
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
+
+    query = """
+    SELECT
+        toUnixTimestamp((intDiv(toUInt32(toDateTime(dt)), ?1) * ?1)) AS time,
+        avgIf(value, metric = 'marketcap_usd') / avgIf(value, metric = 'stack_realized_value_usd') AS ratio
+    FROM daily_metrics
+    PREWHERE
+      ticker_slug = ?2 AND
+      dt >= toDate(?3) AND
+      dt <= toDate(?4) AND
+      metric IN ('stack_realized_value_usd', 'marketcap_usd')
+    GROUP BY time
+    ORDER BY time ASC
+    """
+
+    args = [interval, ticker_slug, from_datetime_unix, to_datetime_unix]
+
+    ClickhouseRepo.query_transform(query, args, fn [dt, mvrv_ratio] ->
+      %{
+        datetime: DateTime.from_unix!(dt),
+        ratio: mvrv_ratio
+      }
+    end)
+  end
+end
--- a/lib/sanbase/clickhouse/mvrv.ex
+++ b/lib/sanbase/clickhouse/mvrv.ex
@@ -19,6 +19,18 @@ defmodule Sanbase.Clickhouse.MVRV do
           String.t()
         ) :: {:ok, list(mvrv_ratio)} | {:error, String.t()}
   def mvrv_ratio(slug, from, to, interval) do
+    {query, args} = mvrv_query(slug, from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [dt, mvrv_ratio] ->
+      %{
+        datetime: DateTime.from_unix!(dt),
+        ratio: mvrv_ratio
+      }
+    end)
+    |> fill_last_seen()
+  end
+
+  defp mvrv_query(slug, from, to, interval) do
     ticker = Project.ticker_by_slug(slug)
     ticker_slug = "#{ticker}_#{slug}"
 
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/mvrv.ex
```diff
@@ -0,0 +1,58 @@
...
+    ticker = Project.ticker_by_slug(slug)
+    ticker_slug = "#{ticker}_#{slug}"
+
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
+    decimals = :math.pow(10, token_decimals)
+
+    query = """
+    SELECT toUnixTimestamp(time) AS dt,
```
IvanIvanoff: if you do not have filling with zeroes this is not needed

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -30,6 +27,24 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
   end
 
+  def mvrv_ratio(_root, args, _resolution) do
+    with {:ok, contract, token_decimals} <- Project.contract_info_by_slug(args.slug),
+         {:ok, mvrv_ratio} <-
+           Mvrv.mvrv_ratio(contract, args.slug, args.from, args.to, args.interval, token_decimals) do
+      {:ok, mvrv_ratio}
+    else
+      error ->
+        Logger.error("Can't calculate MVRV ratio. Reason: #{inspect(error)}")
```
IvanIvanoff: `for project with coinmarketcap_id: #{args.slug}`

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -30,6 +27,24 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+
+        {:error, "Can't calculate MVRV ration"}
+    end
+  rescue
```
IvanIvanoff: Let's rescue only exceptions for functions that are used in `all_*` queries because otherwise they'll fail the whole query. In that case it's ok to let it crash

lib/sanbase_web/graphql/schema.ex
```diff
@@ -815,6 +815,18 @@ defmodule SanbaseWeb.Graphql.Schema do
       cache_resolve(&ClickhouseResolver.network_growth/3)
     end
 
+    @desc "Returns MVRV(Market-Value-to-Realized-Value)"
+    field :mvrv_ratio, list_of(:mvrv_ratio) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, non_null(:string), default_value: "")
```
IvanIvanoff: You have `non_null` and `default_value` - you need to remove the `non_null`. Moreover, I just tested locally what happens if you have both and the result was that it just hangs.. graphiql complains about a non_null param not provided, but Absinthe didn't straight reject the query. That's very strange behaviour

lib/sanbase/clickhouse/mvrv.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Sanbase.Clickhouse.Mvrv do
```
IvanIvanoff: Personal opinion - `Sanbase.Clickhouse.MVRV` will look better as it's an abbreviation

lib/sanbase/clickhouse/mvrv.ex
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    args = [interval, ticker_slug, from_datetime_unix, to_datetime_unix]
+
+    ClickhouseRepo.query_transform(query, args, fn [dt, mvrv_ratio] ->
+      %{
+        datetime: DateTime.from_unix!(dt),
+        ratio: mvrv_ratio
+      }
+    end)
```
IvanIvanoff: You can pipe this into `case` where in the `{:ok, result}` case you run the result through a reduce that looks like this: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/historical_balance/utils.ex#L19

The difference is in the condition where you fill with last seen - in your case this will be when the `ratio` is `null`. The correct behavior of MVRV. This is not the 100% correct behavior, as the `MV` part of the `MVRV` could change and `RV` is what is the same as before, but this cannot be fixed in the Elixir code. Still, filling with the last known is better than giving `null`s

lib/sanbase/clickhouse/mvrv.ex
```diff
@@ -0,0 +1,52 @@
...
+
+    query = """
+    SELECT
+        toUnixTimestamp((intDiv(toUInt32(toDateTime(dt)), ?1) * ?1)) AS time,
+        avgIf(value, metric = 'marketcap_usd') / avgIf(value, metric = 'stack_realized_value_usd') AS ratio
+    FROM daily_metrics
+    PREWHERE
+      dt >= toDate(?3) AND
+      dt <= toDate(?4) AND
+      ticker_slug = ?2 AND
```
IvanIvanoff: nitpicking: in most places the slug/address/contract check goes before all others

lib/sanbase_web/graphql/schema.ex
```diff
@@ -815,6 +815,18 @@ defmodule SanbaseWeb.Graphql.Schema do
...
 
+    @desc "Returns MVRV(Market-Value-to-Realized-Value)"
+    field :mvrv_ratio, list_of(:mvrv_ratio) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "")
+
+      middleware(ApiTimeframeRestriction)
+
```
IvanIvanoff: Add the complexity middleware before the timeframe restriction

lib/sanbase_web/graphql/schema/clickhouse_types.ex
```diff
@@ -10,4 +10,9 @@ defmodule SanbaseWeb.Graphql.ClickhouseTypes do
     field(:datetime, non_null(:datetime))
     field(:new_addresses, :integer)
   end
+
+  object :mvrv_ratio do
+    field(:datetime, non_null(:datetime))
+    field(:ratio, :float)
```
IvanIvanoff: If you make the above mentioned change and replace the `null`s with the last seen value you can make this `non_null` also

test/sanbase/clickhouse/mvrv_test.exs
```diff
@@ -0,0 +1,112 @@
...
+         }}
+      end do
+      response = execute_query(context, "2d")
+      ratios = parse_response(response)
+
+      assert ratios == [
+               %{"ratio" => 0.0, "datetime" => "2019-01-01T00:00:00Z"},
+               %{"ratio" => 0.22, "datetime" => "2019-01-02T00:00:00Z"},
+               %{"ratio" => 0.33, "datetime" => "2019-01-03T00:00:00Z"},
+               %{"ratio" => 0.44, "datetime" => "2019-01-04T00:00:00Z"}
```
IvanIvanoff: again - after the gap filling you can in the middle here a `null` and assert that it's correctly filled by the previous value

test/sanbase/clickhouse/mvrv_test.exs
```diff
@@ -0,0 +1,127 @@
+defmodule SanbaseWeb.Graphql.Clickhouse.MVRVTest do
```
valo: I think the module name is not correct here. Does not match the file path

test/sanbase/clickhouse/mvrv_test.exs
```diff
@@ -0,0 +1,127 @@
...
+  use SanbaseWeb.ConnCase, async: false
```
valo: This use is probably not needed here, as this is not a web test.

IvanIvanoff: In this case we use `Sanbase.DataCase`

test/sanbase/clickhouse/mvrv_test.exs
```diff
@@ -0,0 +1,125 @@
+defmodule SanbaseWeb.Clickhouse.MVRVTest do
+  use Sanbase.DataCase
+  require Sanbase.ClickhouseRepo
```
IvanIvanoff: nitpick: according to https://github.com/christopheradams/elixir_style_guide#modules we put the `require` last (although we did not always follow this, now we try to)

kalkov: @IvanIvanoff fixed

PR #1028, opened 2019-03-07
POC test using phoenix test server

tspenov:
Instead of mocking Tesla we can use phoenix test server that is running on 4001 as a test server.

### DIFF
--- a/lib/sanbase/telegram/telegram.ex
+++ b/lib/sanbase/telegram/telegram.ex
@@ -15,11 +15,12 @@ defmodule Sanbase.Telegram do
   @type message :: String.t() | iolist()
 
   use Tesla
+  adapter(Tesla.Adapter.Hackney)
   @rate_limiting_server :telegram_bot_rate_limiting_server
   alias Sanbase.ExternalServices.{RateLimiting, ErrorCatcher}
   plug(ErrorCatcher.Middleware)
   plug(RateLimiting.Middleware, name: @rate_limiting_server)
-  plug(Tesla.Middleware.BaseUrl, "https://api.telegram.org/bot#{Config.get(:token)}/")
+  plug(Tesla.Middleware.BaseUrl, base_url())
   plug(Tesla.Middleware.Headers, [{"Content-Type", "application/json"}])
 
   @doc ~s"""
@@ -132,5 +133,9 @@ defmodule Sanbase.Telegram do
 
   defp generate_link(user_token) do
     "https://telegram.me/#{Config.get(:bot_username)}?start=#{user_token}"
+  end
+
+  defp base_url() do
+    "#{Config.get(:api_base_url)}#{Config.get(:token)}/"
   end
 end
--- a/lib/sanbase_web/router.ex
+++ b/lib/sanbase_web/router.ex
@@ -74,6 +74,12 @@ defmodule SanbaseWeb.Router do
 
   get("/env.js", SanbaseWeb.RootController, :react_env)
 
+  if Mix.env() == :test do
+    scope "/test", SanbaseWeb do
+      post("/telegramtoken/sendMessage", TestTelegramController, :send_message)
+    end
+  end
+
   if Mix.env() == :dev do
     pipeline :nextjs do
       plug(:accepts, ["html"])
--- a/test/sanbase/signals/trigger_evaluator_test.exs
+++ b/test/sanbase/signals/trigger_evaluator_test.exs
@@ -158,11 +158,6 @@ defmodule Sanbase.Signals.EvaluatorTest do
   end
 
   test "signal setting cooldown works for trending words" do
-    Tesla.Mock.mock_global(fn
-      %{method: :post} ->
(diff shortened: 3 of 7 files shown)


IvanIvanoff:
I am still not convinced that this is a good idea and what are the benefits over using a library designed for that as seen here: https://github.com/santiment/newsletter_generator/blob/master/test/newsletter_generator/discourse/discourse_publish_test.exs#L6

cc @kalkov @valo for comments

lib/sanbase_web/router.ex
```diff
@@ -74,6 +74,12 @@ defmodule SanbaseWeb.Router do
 
   get("/env.js", SanbaseWeb.RootController, :react_env)
 
+  if Mix.env() == :test do
+    scope "/test", SanbaseWeb do
+      post("/telegramtoken/sendMessage", TestTelegramController, :send_message)
```
IvanIvanoff: This should be done with a variable like this `post("/:path", TestController, :dispatch_test)` which will dispatch to the correct test. Otherwise our router will quickly become 90% tests and 10% actual usage

lib/sanbase/telegram/telegram.ex
```diff
@@ -15,11 +15,12 @@ defmodule Sanbase.Telegram do
   @type message :: String.t() | iolist()
 
   use Tesla
+  adapter(Tesla.Adapter.Hackney)
```
IvanIvanoff: Isn't this by default so?

test/sanbase/signals/trigger_evaluator_test.exs
```diff
@@ -158,11 +158,6 @@ defmodule Sanbase.Signals.EvaluatorTest do
   end
 
   test "signal setting cooldown works for trending words" do
-    Tesla.Mock.mock_global(fn
```
IvanIvanoff: What exactly is the issue with the global mock that's inside a test that's not in an async suit? We have it also here: https://github.com/santiment/sanbase2/blob/master/test/sanbase/signals/trigger_evaluator_price_test.exs#L75

tspenov: nothing, just not needed

lib/sanbase/telegram/telegram.ex
tspenov: yes, but overwritten by test config which uses Tesla.Mock

tspenov:
Benefits
* phoenix server is already running on 4001 so we don't need to add a dependency.
* It will work for all http libs Tesla, HTTPoison, Mint, etc . 
* Its much closer to the real flow.
* Also when you want to test something that makes a couple of requests but you are testing other functionality - it hits the happy path and you don't need to add anything irrelevant to the test.

lib/sanbase_web/router.ex
tspenov: we can have a `TestRouter` module and `match "/test/*",  to: Test.Router` or smth like this

valo:
Is it possible to define what response the test server should return for a given request? Here is an example where we want to receive a specific request and return a specific test response https://github.com/santiment/newsletter_generator/blob/master/test/newsletter_generator/discourse/discourse_publish_test.exs#L11 Is it possible to handle such scenarios with the test server?

tspenov:
@valo we can send different responses based on the passed params: https://github.com/santiment/sanbase2/pull/1028/files#diff-53e7765d2a70fcf78172e4086ebe86b5R4 and take advantage of pattern matching

valo:
I looked at this `bypass` package and I like how the test setup is in the test case, as opposed into a test router module, which lives somewhere else. It seems to me that using `bypass` will accomplish the same goals as this PR, with the added benefit of having a convenient DSL for defining the expected requests: https://github.com/PSPDFKit-labs/bypass#example

What are the downsides of using `bypass`, apart from an additional dependency?

IvanIvanoff:
@valo to use bypass the functions should be rewritten to accept `url` as the parameter because the port is assigned runtime (if I remember correctly). This could cause issues if in the test we have for example something that spawn processes that chain functions and somewhere there there is an HTTP call - how to pass the `url` from the test to that function?

Probably there is a better way though, but for the purposes of the newsletter generator I did not need to try to solve this issue.

On the other hand coding with this in mind keeps you away from writing code that's hard to test and that's a good thing.

tspenov:
@valo One example from the codebase is: testing `cooldown` for the trending words. We want to test whether if we send one one message, have cooldown, try to send again - the cooldown will work as expected. But to do that we have to mock in the test 2 network calls: one to `tech_indicators`, second to `telegram` and we're testing something different. With the test server it will work like global mock (based on params) so we don't have to do anything if we have the green path implemented.

valo:
May be we can have a global bypass instance, which is being run on a fixed port and can be used for global mocking? I like the DSL of bypass and how it allows you to define the mocks in the tests

tspenov:
OK, let's close this for now.

PR #1031, opened 2019-03-07
Resolve some warnings

kalkov:
#### Summary
There were a lot of warnings in the beginning of tests execution. Resolved most of them but few are left, I don't feel confident enough yet to touch them, so help is welcomed.

```
warning: variable "notification" is unused
  lib/ex_admin/show.ex:5

warning: def run_query/4 has multiple clauses and also declares default values. In such cases, the default values should be defined in a header. Instead of:

    def foo(:first_clause, b \\ :default) do ... end
    def foo(:second_clause, b) do ... end

one should write:

    def foo(a, b \\ :default)
    def foo(:first_clause, b) do ... end
    def foo(:second_clause, b) do ... end

  lib/sanbase_web/admin/model/ico.ex:10

warning: SanbaseWeb.Graphql.Resolvers.PriceResolver.history_price/3 is deprecated. Use history price by slug
  lib/sanbase_web/graphql/schema.ex:188
```
[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/signals/history/prices_history.ex
+++ b/lib/sanbase/signals/history/prices_history.ex
@@ -1,6 +1,6 @@
 defmodule Sanbase.Signals.History.PricesHistory do
   @moduledoc """
-  Implementations of historical trigger points for price_percent_change and 
+  Implementations of historical trigger points for price_percent_change and
   price_absolute_change triggers. Historical prices are bucketed at `1 hour` intervals and goes
   `90 days` back.
   """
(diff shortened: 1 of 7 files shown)


lib/sanbase/signals/history/prices_history.ex
```diff
@@ -26,7 +26,7 @@ defmodule Sanbase.Signals.History.PricesHistory do
           triggered?: boolean()
         }
 
-  def get_prices(%{target: target} = settings) when is_binary(target) do
+  def get_prices(%{target: target} = _settings) when is_binary(target) do
```
IvanIvanoff: in that case you can omit the while matching as the `settings` var is not used

kalkov: @IvanIvanoff But target is used.

tspenov: yes you van leave only`%{target: target}`

PR #1033, opened 2019-03-08
Fix the price-volume diff when result size argument is not 0

IvanIvanoff:
The shift is in concrete number of intervals but the drop must not be.
To support in a single operation both `size = 0` (return all datapoint from tech_indicators) and `size > 0` (return only the last size results) we can use `drop_while` instead of dropping concrete number of points.

If we drop a concrete number of datapoint the behavior is broken for example in the case of `size=1`. This is because we'll try to remove ~21 number of points from a result with only 1 point.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/eth_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/eth_balance.ex
@@ -66,7 +66,7 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.EthBalance do
     interval = Sanbase.DateTimeUtils.compound_duration_to_seconds(interval)
     from_unix = DateTime.to_unix(from)
     to_unix = DateTime.to_unix(to)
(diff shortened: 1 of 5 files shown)


lib/sanbase/notifications/price_volume_diff.ex
```diff
@@ -87,7 +87,7 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
             volume_change: nil_to_zero(volume_change)
           }
 
-        _ ->
+        data ->
```
tspenov: why named and not used later ?

tspenov:
some related tests failing, though

lib/sanbase/notifications/price_volume_diff.ex
IvanIvanoff: debugging purposes, will make it again `_`

IvanIvanoff:
@tspenov yes, they are fixed now

PR #1035, opened 2019-03-11
Fix handling of empty results in the DAA signals

IvanIvanoff:
To simplify things the following is done:
1. Rewrite the cache to support :error tuples that are not persisted
2. Add a Clickhouse DAA dispatch module for easier access

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/daily_active_addresses.ex
```diff
@@ -0,0 +1,35 @@
...
+end
+
+def average_active_addresses(contract, from, to) do
+  Erc20.average_active_addresses(contract, from, to)
+end
+
+def average_active_addresses(eth, from, to, interval) when eth in @ethereum do
+  Eth.average_active_addresses(from, to, interval)
+
+  def average_active_addresses(contract, from, to, interval) do
```
tspenov: that should be outside of the other function :D

PR #1038, opened 2019-03-12
Featured Insights, Watchlists and Triggers

IvanIvanoff:
#### Summary
3 APIs are available now: `featuredInsights`, `featuredWatchlists` and `featuredUserTriggers`. Each one of them returns a list of the corresponding GQL types.

Example usage:
```graphql
{
  featuredUserTriggers{
    trigger{
      title
      settings 
    }
  }
}
```
```graphql
{
  "data": {
    "featuredUserTriggers": [
      {
        "trigger": {
          "settings": {
            "channel": "telegram",
            "filtered_target_list": [],
            "payload": null,
            "percent_threshold": 300.0,
            "repeating": false,
            "target": "santiment",
            "time_window": "1d",
            "triggered?": false,
            "type": "daily_active_addresses"
          },
          "title": "Generic title"
        }
      }
    ]
  }
}
```

## Admin page:
Seeing the list of featured items:
image

Also visible from the page of the insights/watchlists/user triggers:
image

Can be set when editing (it's one of the few fields that can be edited):
image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/migrations/20190312142628_add_featured_items_table.exs
```diff
@@ -0,0 +1,20 @@
...
+      timestamps()
+    end
+
+    fk_check = """
+    (CASE WHEN post_id IS NULL THEN 0 ELSE 1 END) +
+    (CASE WHEN user_list_id IS NULL THEN 0 ELSE 1 END) +
+    (CASE WHEN user_trigger_id IS NULL THEN 0 ELSE 1 END) = 1
+    """
+
+    create(constraint(:featured_items, :only_one_fk, check: fk_check))
```
tspenov: Also we need UNIQ constraint on each separately, so we don't insert twice something as featured.

lib/sanbase/featured_items/featured_item.ex
```diff
@@ -0,0 +1,100 @@
...
+    featured_items
+    |> cast(attrs, [:post_id, :user_list_id, :user_trigger_id])
+  end
+
+  def insights() do
+    insights_query()
+    |> join(:inner, [fi], fi in assoc(fi, :post))
+    |> select([_fi, post], post)
+    |> Repo.all()
+    |> Repo.preload([:user, :poll])
```
tspenov: Here we're preloading more asociations.
https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/post_resolver.ex#L16

Generally I think most of the time preloaded associations will be the same so I think a good idea is to extract them
in module attribute and public function if needed (like in this case) in its own module.
like here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/model/project/project.ex#L21

tspenov: That way we will be sure that we didn't miss anything.

tspenov:
```elixir
Repo.preload(Post.preloads())
```

IvanIvanoff: 👍 Let's do the rework in a separate PR as it will probably require us to rework many other modules, too

IvanIvanoff: @tspenov some of the fields are actually resolved with a custom resolver, so we can skip preloading them when they are not needed.

lib/sanbase/featured_items/featured_item.ex
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  @doc ~s"""
+  Changeset for the FeaturedItem module.
+  There is a database check that exactly one of `post_id`, `user_list_id` and
+  `user_trigger_id` fields is set
+  """
+  def changeset(%__MODULE__{} = featured_items, attrs \\ %{}) do
+    featured_items
+    |> cast(attrs, [:post_id, :user_list_id, :user_trigger_id])
```
tspenov: Also a `|> unique_constraint()` on these 3 can be added in the changeset.

lib/sanbase/featured_items/featured_item.ex
tspenov: ok, but please make sure that all needed fields are returned cause seems to me we're not returning foe ex: `tags` and `images` for featured insights.

IvanIvanoff: tags and images are resolved via the dataloader when needed

PR #1043, opened 2019-03-13
Add realized value metric

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/realized_value.ex
+++ b/lib/sanbase/clickhouse/realized_value.ex
@@ -1,0 +1,121 @@
+defmodule Sanbase.Clickhouse.RealizedValue do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate Realized Value
+  """
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  alias Sanbase.DateTimeUtils
+  alias Sanbase.Model.Project
+
+  @type realized_value :: %{
+          datetime: DateTime.t(),
+          realized_value: integer,
+          non_exchange_realized_value: integer
+        }
+
+  @spec realized_value(
+          String.t(),
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(realized_value)} | {:error, String.t()}
+  def realized_value(slug, from, to, interval) do
+    {query, args} = realized_value_query(slug, from, to, interval)
+
+    ClickhouseRepo.query_transform(
+      query,
+      args,
+      fn [dt, realized_value, non_exchange_realized_value] ->
+        %{
+          datetime: DateTime.from_unix!(dt),
+          realized_value: realized_value,
+          non_exchange_realized_value: non_exchange_realized_value
+        }
+      end
+    )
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/realized_value.ex
```diff
@@ -0,0 +1,65 @@
...
+  @spec realized_value(
+          String.t(),
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(realized_value)} | {:error, String.t()}
+  def realized_value(slug, from, to, interval) do
+    {query, args} = realized_value_query(slug, from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [
```
IvanIvanoff: the formatter is acting weird here. You can make all arguments on a separate line and it look like this:
```elixir
ClickhouseRepo.query_transform(
      query,
      args,
      fn [dt, realized_value, non_exchange_realized_value] ->
        %{
          datetime: DateTime.from_unix!(dt),
          realized_value: realized_value,
          non_exchange_realized_value: non_exchange_realized_value
        }
      end
    )
```

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -36,6 +36,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     end
   end
 
+  def realized_value(
+        _root,
+        %{slug: slug, from: from, to: to, interval: interval},
+        _resolution
+      ) do
+    with {:ok, realized_value} <- RealizedValue.realized_value(slug, from, to, interval) do
```
IvanIvanoff: you can keep this but also such cases are better suited for `case` - you have only 2 values to match on but no chaining of success matches (nested `case` otherwise`

lib/sanbase/clickhouse/realized_value.ex
```diff
@@ -0,0 +1,121 @@
+defmodule Sanbase.Clickhouse.RealizedValue do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate Realized Value
+  """
```
IvanIvanoff: Some more description for such non-trivial metrics (daa, volume, etc.) will be helpful

lib/sanbase/clickhouse/realized_value.ex
```diff
@@ -0,0 +1,121 @@
...
+      |> elem(2)
+      |> Enum.reverse()
+
+    {:ok, filled_result}
+  end
+
+  defp fill_last_seen(data), do: data
+
+  defp fill_element(
+         %{realized_value: nil, non_exchange_realized_value: nil, datetime: dt},
```
IvanIvanoff: is it possible that only one of them is nil?

kalkov: @IvanIvanoff : Not 100% sure but based on quick check in grafana can conclude it is not possible.

PR #1049, opened 2019-03-15
Refactor and catch errors daa signal

tspenov:
#### Summary

sentry stage issue:
image

### DIFF
--- a/lib/sanbase/notifications/discord/daa_signal.ex
+++ b/lib/sanbase/notifications/discord/daa_signal.ex
@@ -27,20 +27,11 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
          {:ok, today_daa_for_projects} <- all_projects_daa_for_today(projects) do
       notification_type = Type.get_or_create("daa_signal")
 
-      projects_to_signal =
-        projects_to_signal(
-          projects,
-          avg_daa_for_projects,
-          today_daa_for_projects,
-          notification_type
-        )
+      projects
+      |> projects_to_signal(avg_daa_for_projects, today_daa_for_projects, notification_type)
+      |> send_and_persist(notification_type)
 
-      if Enum.count(projects_to_signal) > 0 do
-        send_and_persist(projects_to_signal, notification_type)
-      else
-        Logger.info("Daily Active Addresses Signal finished with nothing to publish")
-        :ok
-      end
+      :ok
     else
       {:error, error} ->
         Logger.error(
@@ -59,7 +50,9 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
   defp all_projects() do
     from(
       p in Project,
-      where: not is_nil(p.coinmarketcap_id) and not is_nil(p.main_contract_address)
+      where:
+        not is_nil(p.coinmarketcap_id) and
+          (not is_nil(p.main_contract_address) or p.coinmarketcap_id == "ethereum")
     )
     |> Repo.all()
   end
@@ -82,6 +75,9 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
     |> Enum.reject(&is_nil/1)
     |> Enum.sort_by(fn {_, _, _, _, change, _} -> change end, &>=/2)
   end
+
+  defp send_and_persist([], _),
+    do: Logger.info("Daily Active Addresses Signal finished with nothing to publish")
 
   defp send_and_persist(projects_to_signal, notification_type) do
     projects_to_signal
@@ -108,10 +104,10 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
       Logger.info(
         "DAA signal check: #{project.coinmarketcap_id}, #{avg_daa}, #{current_daa}, #{
           last_triggered_daa
-        }, #{percent_change}%, #{hours}, #{percent_change > threshold_change() * 100}"
+        }, #{percent_change}%, #{hours}, #{percent_change > trigger_percent(project) * 100}"
       )
 
-      if percent_change > threshold_change() * 100 do
+      if percent_change > trigger_percent(project) * 100 do
         {project, avg_daa, current_daa, last_triggered_daa, percent_change, hours}
       else
         nil
@@ -205,7 +201,7 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
         timeframe_to()
       )
     end)
-    |> handle_errors()
+    |> handle_result()
   end
 
   defp get_daa_contract(contract, all_projects_daa) do
@@ -222,10 +218,10 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
     |> Enum.map(fn contracts ->
       DailyActiveAddresses.realtime_active_addresses(contracts)
     end)
-    |> handle_errors()
+    |> handle_result()
   end
 
-  defp handle_errors(daa_for_projects) do
+  defp handle_result(daa_for_projects) do
     daa_for_projects
     |> Enum.find(&match?({:error, _}, &1))
     |> case do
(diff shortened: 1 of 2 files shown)


lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -149,21 +172,39 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
-        {:ok, daa_result} =
-          Erc20DailyActiveAddresses.average_active_addresses(
-            contracts,
-            timeframe_from(),
-            timeframe_to()
-          )
-
-        daa_result
-      end)
+    cache_id = :signals_cache
```
IvanIvanoff: better be module attribute

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -149,21 +172,39 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
+    cache_key = "daa_signal_#{today_str()}_averages"
+
+    ConCache.get(cache_id, cache_key)
```
IvanIvanoff: Let's think how to reuse modules like https://github.com/santiment/sanbase2/blob/master/lib/sanbase/signals/evaluator/cache.ex
as they are handling more edge cases.

The difference will be that the 2 caches will have different TTL/name. Otherwise, we can expose a way to start the cache itself under the supervision tree instead of just starting the con cache and using it in the cache implementation

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -149,21 +172,39 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
+        end
+
+      avg_daa ->
+        {:ok, avg_daa}
+    end
+  end
+
+  defp get_avg_daa(projects) do
+    projects
+    |> Enum.map(&Project.contract_address/1)
```
IvanIvanoff: you can plug `reject is_nil` here also

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -176,10 +217,26 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
-    |> Enum.flat_map(fn contracts ->
-      {:ok, today_daa} = Erc20DailyActiveAddresses.realtime_active_addresses(contracts)
-      today_daa
+    |> Enum.map(fn contracts ->
+      Erc20DailyActiveAddresses.realtime_active_addresses(contracts)
     end)
+    |> catch_errors()
+  end
+
+  defp catch_errors(daa_for_projects) do
```
IvanIvanoff: I though it was something to do with `throw/catch`, `handle_errors` would be more appropriate

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -149,21 +173,39 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
+    end
+  end
+
+  defp get_avg_daa(projects) do
+    projects
+    |> Enum.map(&Project.contract_address/1)
+    |> Enum.reject(&is_nil/1)
+    |> Enum.chunk_every(100)
+    |> Enum.map(fn contracts ->
+      Erc20DailyActiveAddresses.average_active_addresses(
```
IvanIvanoff: Use the new dispatch module `Sanbase.Clickhouse.DailyActiveAddresses`

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -175,11 +217,31 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
-      {:ok, today_daa} = Erc20DailyActiveAddresses.realtime_active_addresses(contracts)
-      today_daa
+    |> Enum.map(fn contracts ->
+      Erc20DailyActiveAddresses.realtime_active_addresses(contracts)
     end)
+    |> handle_errors()
+  end
+
+  defp get_projects_contracts(projects) do
+  end
```
IvanIvanoff: empty

lib/sanbase/notifications/discord/daa_signal.ex
tspenov: Good suggestion. Unfortunately we'll have to pass the cache_name/cache_id which I don't think is so bad. But lets sit on for a while.

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -149,21 +173,39 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
+    end
+  end
+
+  defp get_avg_daa(projects) do
+    projects
+    |> Enum.map(&Project.contract_address/1)
+    |> Enum.reject(&is_nil/1)
+    |> Enum.chunk_every(100)
+    |> Enum.map(fn contracts ->
+      DailyActiveAddresses.average_active_addresses(
```
IvanIvanoff: This will now support `ethereum` or `ETH` being passed as a contract. You can change the `all_projects()` function to: `where: not is_nil(p.coinmarketcap_id) and (not is_nil(p.main_contract_address) or p.coinmarketcap_id = "ethereum")`

IvanIvanoff: that's the case if it is passed as a single param and not part of list, though

IvanIvanoff: Though ethereum won't ever trigger the x3 increase.. You can check here how ethereum is handled in the exchange inflow signal with a different threshold: https://github.com/santiment/sanbase2/commit/84b48f14441215c66b3a0c8a0467b1fc215fd7c3

IvanIvanoff: but this can be done in a separate PR

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -13,46 +13,44 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
-      projects_to_signal
-      |> Enum.map(&create_notification_content/1)
-      |> Enum.each(fn {project, payload, embeds, current_daa} ->
-        payload
-        |> Discord.encode!(publish_user(), embeds)
-        |> publish("discord")
-
-        Notification.insert_triggered(project, notification_type, "#{current_daa}")
-      end)
+      if Enum.count(projects_to_signal) > 0 do
```
IvanIvanoff: instead of `if` + count add (note that it returns `:ok` so no need to add it):
```elixir
defp send_and_persist([], _), do: Logger.info("Daily Active Addresses Signal finished with nothing to publish")
```

lib/sanbase/notifications/discord/daa_signal.ex
```diff
@@ -175,11 +217,28 @@ defmodule Sanbase.Notifications.Discord.DaaSignal do
...
+    |> Enum.find(&match?({:error, _}, &1))
+    |> case do
+      {:error, error} ->
+        {:error, error}
+
+      nil ->
+        {
+          :ok,
+          daa_for_projects
+          |> Enum.flat_map(fn {:ok, daa} -> daa end)
```
IvanIvanoff: it's strange to me that the function `handle_errors` mainly handles the :ok case. It should be `handle_result`

tspenov: Its main function is to check if there are any errors but it does handle also the ok case. `handle_result` sounds ok to me

lib/sanbase/notifications/discord/daa_signal.ex
tspenov: yeah, have to merge somehow result for list erc2 contracts and eth. Lets make it in separate PR

PR #1050, opened 2019-03-15
Add api gainers losers

tspenov:
```graphql
    {
      topSocialGainersLosers(
        status: ALL, 
        from: "2018-01-09T00:00:00Z",
        to:"2018-01-10T00:00:00Z",
        range: "15d",
        size: 1
      ) {
        datetime,
        projects {
          project,
          change,
          status
        }
      }
    }
```

```graphql
    {
      socialGainersLosersStatus(
        slug: "qtum",
        from: "2018-01-09T00:00:00Z",
        to:"2018-01-10T00:00:00Z",
        range: "15d",
      ) {
        datetime,
        change,
        status
      }
    }
```


lib/sanbase_web/graphql/schema.ex
```diff
@@ -849,6 +849,30 @@ defmodule SanbaseWeb.Graphql.Schema do
 
       resolve(&SignalsHistoricalActivityResolver.fetch_historical_activity_for/3)
     end
+
+    @desc """
+    Top social gainers/losers for given range.
```
IvanIvanoff: @ivanklimuk can you write a short description of the social gainers/losers that goes into a little bit more details?

lib/sanbase_web/graphql/schema.ex
```diff
@@ -849,6 +849,30 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+    * `status` can be one of: `ALL`, `GAINER`, `LOSER`, `NEWCOMER`
+    """
+    field :top_social_gainers_losers, list_of(:top_social_gainers_losers) do
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:status, non_null(:social_gainers_losers_status_enum))
+      arg(:size, :integer, default_value: 10)
+      arg(:range, non_null(:string))
```
IvanIvanoff: Is this the `interval` we use everywhere else? If so, let's name it `interval` even though `tech_indicators` uses range. We need to be as consistent as possible in the API.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -849,6 +849,30 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+
+      cache_resolve(&SocialDataResolver.top_social_gainers_losers/3)
+    end
+
+    @desc "Social gainers/losers for given slug"
+    field :social_gainers_losers_status, list_of(:social_gainers_losers_status) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:range, non_null(:string))
```
IvanIvanoff: The same comment for `range` as above.

lib/sanbase_web/graphql/schema.ex
ivanklimuk: @IvanIvanoff no it's not, it's a specific parameter for this API only, has nothing to do with the aggregation interval of the data being fetched

lib/sanbase_web/graphql/schema.ex
ivanklimuk: @IvanIvanoff shall I do it here or on discourse or somewhere else? there was a description in the corresponding PR

ivanklimuk: and in the README in the repo: https://github.com/santiment/social_gainers_losers

lib/sanbase_web/graphql/schema.ex
tspenov: @IvanIvanoff Yeah,  I named it `interval`in the first commit but its different than the bucket intervals we use in the other API endpoints. Here we don't have control over how long time one bucket will be. 
ex: `from=yesterday`, `to=now`, range="15d" and you will have buckets/intervals with datetimes between `from` and `to` (probably computation time) but the `change` is function of the `range`.
@ivanklimuk Am I right ?

ivanklimuk: @tspenov correct. The interval between `from` and `to` in this API is always 1 hour

tspenov: @ivanklimuk BTW what is the min and max value for `range` ?

lib/sanbase_web/graphql/schema.ex
tspenov: 👍  cool, I'll use these

lib/sanbase_web/graphql/schema.ex
ivanklimuk: @tspenov min `range` is probably `2d`, max is theoretically not limited, although I don't think we will ever use a value higher than let's say `30d`

tspenov: @ivanklimuk 👍

test/sanbase/social_data/social_data_test.exs
```diff
@@ -183,4 +183,132 @@ defmodule Sanbase.SocialDataTest do
...
+    )
+
+    args = %{
+      slug: "qtum",
+      from: DateTime.from_naive!(~N[2019-03-15 12:57:28], "Etc/UTC"),
+      to: DateTime.from_naive!(~N[2019-03-15 13:57:28], "Etc/UTC"),
+      range: "15d"
+    }
+
+    # As the HTTP call is mocked these arguemnts do no have much effect, though you should try to put the real ones that are used
```
kalkov: Ditto

test/sanbase_web/graphql/social_data_api_test.exs
```diff
@@ -307,6 +307,186 @@ defmodule Sanbase.SocialDataApiTest do
     end
 
     assert capture_log(result_fn) =~
-             "Error status 500 fetching word_trend_score for word merry: Internal Server Error"
+             "Error status 500 fetching word trend score for word merry: Internal Server Error"
+  end
+
+  test "successfully fetch top social gainers losers", %{conn: conn} do
+    mock(
```
kalkov: I think it's better idea to mock `Sanbase.SocialData` methods here instead of the HTTP calls. @valo,  @IvanIvanoff, @tspenov what is your opinion?

test/sanbase_web/graphql/social_data_api_test.exs
```diff
@@ -307,6 +307,186 @@ defmodule Sanbase.SocialDataApiTest do
...
+      :get,
+      {:ok,
+       %HTTPoison.Response{
+         body:
+           "[{\"timestamp\": 1552654800, \"range\": \"15d\", \"projects\": [{\"project\": \"qtum\", \"change\": 137.13186813186815, \"status\": \"gainer\"}, {\"project\": \"abbc-coin\", \"change\": -1.0, \"status\": \"loser\"}]}]",
+         status_code: 200
+       }}
+    )
+
+    query = """
```
kalkov: Can be extracted into a private method and reused.

test/sanbase_web/graphql/social_data_api_test.exs
```diff
@@ -307,6 +307,186 @@ defmodule Sanbase.SocialDataApiTest do
...
+        projects {
+          project,
+          change,
+          status
+        }
+      }
+    }
+    """
+
+    result =
```
kalkov: Can be extracted into a private method and reused.

test/sanbase_web/graphql/social_data_api_test.exs
tspenov: Good suggestion because we already have tests for `SocialData` module. I'll have a look and refactor these.

tspenov:
@kalkov updated comments

test/sanbase_web/graphql/social_data_api_test.exs
```diff
@@ -1,115 +1,322 @@
...
+  alias Sanbase.DateTimeUtils
+
+  @error_response "Error executing query. See logs for details."
+
+  test "successfully fetch trending words", %{conn: conn} do
+    success_response = [
+      %{
+        datetime: DateTimeUtils.from_iso8601!("2018-11-10T00:00:00Z"),
+        top_words: [
+          %{score: 167.74716011726295, word: "pele"},
```
kalkov: :soccer:

PR #1053, opened 2019-03-19
Add query to metadata

IvanIvanoff:
#### Summary
Now we'll have the following logs:
```
[16:43:47.754][info][request_id=2m6m6j5t2oj1fr564o0010m6 query=allPublicTriggers ] Sent 200 in 856µs
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/middlewares/api_usage.ex
+++ b/lib/sanbase_web/graphql/middlewares/api_usage.ex
@@ -40,7 +40,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.ApiUsage do
       }, Complexity: #{definition.complexity}"
     )
 
-    Logger.reset_metadata(metadata)
+    Logger.reset_metadata([{:query, definition.name} | metadata])
 
     resolution
   end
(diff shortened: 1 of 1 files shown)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -723,17 +723,17 @@ defmodule SanbaseWeb.Graphql.Schema do
 
     @desc "Fetch all favourites lists for current_user."
     field :fetch_user_lists, list_of(:user_list) do
-      resolve(&UserListResolver.fetch_user_lists/3)
+      cache_resolve(&UserListResolver.fetch_user_lists/3)
```
tspenov: IMO we shouldn't cache things the user can mutate. Otherwise there will be situations where user fetches , the creates or updates, fetches again - but don't see updates or new stuff.

tspenov:
Otherwise 👍 for logging

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: These things are slow though. I'll remove them for now but we'll need to implement cache invalidation soon.

PR #1057, opened 2019-03-20
Add gas used metric

kalkov:
#### Summary
It has generic name because in the future we can support other projects as well - e.g ETC. 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/gas_used.ex
+++ b/lib/sanbase/clickhouse/gas_used.ex
@@ -1,0 +1,56 @@
+defmodule Sanbase.Clickhouse.GasUsed do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate used gas.
+  """
+
+  import Sanbase.Math, only: [to_integer: 1]
+  alias Sanbase.DateTimeUtils
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+
+  @type gas_used :: %{
+          datetime: DateTime.t(),
+          eth_gas_used: non_neg_integer()
+        }
+
+  @spec gas_used(
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(gas_used)} | {:error, String.t()}
+  def gas_used(from, to, interval) do
+    {query, args} = gas_used_query(from, to, interval)
+
+    ClickhouseRepo.query_transform(
+      query,
+      args,
+      fn [dt, gas_used] ->
+        %{
+          datetime: DateTime.from_unix!(dt),
+          eth_gas_used: gas_used |> to_integer()
+        }
+      end
+    )
+  end
+
+  defp gas_used_query(from, to, interval) do
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
+
+    query = """
+    SELECT
+      toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS time,
+      sum(gasUsed) AS gas_used
+    FROM eth_blocks
+    PREWHERE
+      dt >= toDateTime(?2) AND
+      dt <= toDateTime(?3)
+    GROUP BY time
+    ORDER BY time ASC
+    """
+
+    args = [interval, from_datetime_unix, to_datetime_unix]
+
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/gas_used.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Sanbase.Clickhouse.GasUsed do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate used gas.
+  """
+
+  alias Sanbase.DateTimeUtils
+  import Sanbase.Math, only: [to_integer: 1]
```
tspenov: the convention we try to follow `use`, `import`, `alias`, `require` sequence

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -15,6 +16,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
 
   @one_hour_seconds 3600
 
+  def gas_used(
+        _root,
+        %{from: from, to: to, interval: interval},
+        _resolution
+      ) do
+    with {:ok, gas_used} <- GasUsed.gas_used(from, to, interval) do
```
tspenov: I think `case` is preferable over `with` for only one clause

lib/sanbase_web/graphql/schema.ex
```diff
@@ -830,6 +830,19 @@ defmodule SanbaseWeb.Graphql.Schema do
       cache_resolve(&ClickhouseResolver.nvt_ratio/3)
     end
 
+    @desc """
+    Returns used GAS.
```
tspenov: May better description here

kalkov: @tspenov Don't know what to write here, can you suggest something?

lib/sanbase/clickhouse/gas_used.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Sanbase.Clickhouse.GasUsed do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate used gas.
+  """
```
IvanIvanoff: We usually have the schema of the table here even though it's not used.
It serves as a description of what the table is and also could be used when we start spawning ClickHouse server in the tests instead of mocking everything.

lib/sanbase_web/graphql/schema.ex
```diff
@@ -830,6 +830,19 @@ defmodule SanbaseWeb.Graphql.Schema do
...
+    """
+    field :gas_used, list_of(:gas_used) do
```
IvanIvanoff: this is better to be `eth_gas_used` or support a single slug `ethereum` for now

lib/sanbase/clickhouse/gas_used.ex
kalkov: I will add schema info for all GQL Clickhouse metrics in a separate PR aimed to improve consistency.

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -15,6 +16,22 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+  def gas_used(
+        _root,
+        %{from: from, to: to, interval: interval},
+        _resolution
+      ) do
+    case GasUsed.gas_used(from, to, interval) do
+      {:ok, gas_used} ->
+        {:ok, gas_used}
+
+      {:error, error} ->
```
IvanIvanoff: you can add one test that passes through the `{:error, String.t()}` path

lib/sanbase_web/graphql/schema.ex
IvanIvanoff: https://kb.myetherwallet.com/gas/what-is-gas-ethereum.html

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
kalkov: There is [one](https://github.com/santiment/sanbase2/blob/03ff83e9fe2e3e2f9fb647dd2b07f2357d3bf39f/test/sanbase_web/graphql/clickhouse/gas_used_test.exs#L62)

PR #1060, opened 2019-03-20
Rework the filtered_target_list

IvanIvanoff:
#### Summary
Now instead of just strings and lists it supports specifiying the type.
To keep the old behaviour strings and lists are treated as slugs and
lists of slugs. The new behaviour will require you to pass %{slug: slug}
maps as targets.

This is implemented so a single signal can support different types of targets.
An example is the signal for tracking ETH wallets. The signal target can be a
project's slug, so all of the project's wallets are tracked. The target can also be
a specific ETH address or a list of ETH addresses that the user finds interesting.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/map_utils.ex
+++ b/lib/map_utils.ex
@@ -1,0 +1,5 @@
+defmodule Sanbase.MapUtils do
+  def drop_diff_keys(left, right) do
+    Map.drop(left, Map.keys(left) -- Map.keys(right))
+  end
+end
--- a/lib/map_utils.ex
+++ b/lib/map_utils.ex
@@ -1,4 +1,24 @@
 defmodule Sanbase.MapUtils do
+  @doc ~s"""
+  Return a subset of `left` map that has only the keys that are also present in `right`.
+
+  #### Examples:
+
+  iex> Sanbase.MapUtils.drop_diff_keys(%{}, %{})
+  %{}
+
+  iex> Sanbase.MapUtils.drop_diff_keys(%{a: 1}, %{a: 1})
+  %{a: 1}
+
+  iex> Sanbase.MapUtils.drop_diff_keys(%{a: 1, b: 2}, %{a: 1})
+  %{a: 1}
+
+  iex> Sanbase.MapUtils.drop_diff_keys(%{a: 1}, %{a: "ASDASDASDA"})
+  %{a: 1}
+
+  iex> Sanbase.MapUtils.drop_diff_keys(%{a: 1, d: 555, e: "string"}, %{b: 2, c: 3, f: 19})
+  %{}
+  """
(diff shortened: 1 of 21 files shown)


lib/sanbase/signals/trigger/trigger.ex
```diff
@@ -145,11 +146,28 @@ defmodule Sanbase.Signals.Trigger do
...
-    target_list
-    |> Enum.reject(&Sanbase.Signals.Trigger.has_cooldown?(trigger, &1))
+  defp remove_targets_on_cooldown(%{slug: slug}, trigger)
+       when is_binary(slug) or is_list(slug) do
+    slug
+    |> List.wrap()
+    |> remove_targets_on_cooldown(trigger, :slug)
+  end
+
+  defp remove_targets_on_cooldown(%{eth_address: address}, trigger)
```
tspenov: Here for `target` we have  `%{eth_address: eth_address | list_of_eth_addresses}` but do we want to connect these to the `slug` also (for example in the signal payload) because this will change the structure yet again.

IvanIvanoff: We cannot connect `eth_address` to `slug`. Almost all addresses do not belong to a project. If you want to follow the project's eth addresses you'll use its slug and the logic will know how to handle it and make the best payload possible.

lib/map_utils.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Sanbase.MapUtils do
```
tspenov: may add a simple `iex` on top

lib/sanbase/signals/trigger/validation.ex
```diff
@@ -54,6 +54,22 @@ defmodule Sanbase.Signals.Validation do
   def valid_target?(target),
     do: {:error, "#{inspect(target)} is not a valid target"}
 
+  def valid_eth_wallet_target?(%{address: address}) when is_binary(address) or is_list(address) do
```
tspenov: here it is `eth_address`: https://github.com/santiment/sanbase2/pull/1060/files#diff-c4cf6d5e5082cdc55ae98ac9b9e99169R159

lib/map_utils.ex
IvanIvanoff: what do you mean by that?

tspenov: like this: https://github.com/santiment/sanbase2/blob/b414141a3e22ef283aa17c61c3af76e8006f8134/lib/sanbase/utils/math.ex#L11

IvanIvanoff: :+1:

PR #1061, opened 2019-03-21
Hide draft insights

tspenov:


lib/sanbase/insight/post.ex
```diff
@@ -135,6 +135,14 @@ defmodule Sanbase.Insight.Post do
     |> get_only_published_or_own_posts(user_id)
   end
 
+  def all_insights_for_user(user_id) do
```
IvanIvanoff: `user_published_insights` :? At least let's have `published` in the name

PR #1063, opened 2019-03-21
Improve NVT ratio description

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/schema.ex
+++ b/lib/sanbase_web/graphql/schema.ex
@@ -826,7 +826,8 @@ defmodule SanbaseWeb.Graphql.Schema do
     @desc """
     Returns NVT (Network-Value-to-Transactions-Ratio
     Daily Market Cap / Daily Transaction Volume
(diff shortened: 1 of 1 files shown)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -826,7 +826,8 @@ defmodule SanbaseWeb.Graphql.Schema do
     @desc """
     Returns NVT (Network-Value-to-Transactions-Ratio
     Daily Market Cap / Daily Transaction Volume
-    Since Daily Transaction Volume gets rather noisy and often includes duplicate transactions,
+    Since Daily Transaction Volume gets rather noisy and easy to manipulate
+    by transferring the same tokens through couple of addresses over and over again,
```
IvanIvanoff: Hm, this description is really similar to token circulation which aims to solve the same issue. I am not sure that NVT is used for the thing it's written here

IvanIvanoff: NVM, I missed the part that follows these lines

PR #1067, opened 2019-03-21
Remove String.to_atom on external data

tspenov:


IvanIvanoff:
it's not external data but only 2 strings that are being passed by the caller. Nevertheless, it's not good to have it in our code.

tspenov:
yes you are right - but still an improvement

PR #1073, opened 2019-03-25
Improve triggers api docs

tspenov:


docs/user-triggers-api.md
```diff
@@ -16,7 +16,7 @@ These are the fields describing a trigger.
 ### Settings fields
 
 - **type** Defines the type of the trigger. Can be one of: `["daily_active_addresses", "price_absolute_change", "price_percent_change", "trending_words", "price_volume_difference"]`
-- **target**:  slug or list of slugs or watchlist - `"naga" | ["ethereum", "santiment"] | {"user_list": user_list_id}`
+- **target**:  slug or list of slugs or watchlist - `{"slug": "naga"} | {"slug": ["ethereum", "santiment"]} | {"user_list": user_list_id}`
```
IvanIvanoff: also the `eth_address` type is supported, though it should be noted that it's not for every signal

PR #1074, opened 2019-03-26
Timeline events API

tspenov:
```graphql
{
      timelineEvents {
        cursor {
          after
          before
        }
        events {
          eventType,
          createdAt,
          user {
            id
          },
          userList {
            name
            isPublic
          }
          post {
            id
          }
          userTrigger {
            trigger {
              title,
              description
            }
          }
        }
      }
    }
```

```elixir
[
  %{
    "cursor" => %{
      "after" => "2019-03-27T15:43:07.109578Z",
      "before" => "2019-03-27T15:43:07.051777Z"
    },
    "events" => [
      %{
        "insertedAt" => "2019-03-27T15:43:07.109578Z",
        "eventType" => "create_public_trigger",
        "post" => nil,
        "user" => %{"id" => "4420"},
        "userList" => nil,
        "userTrigger" => %{
          "trigger" => %{
            "description" => "DAA going up 300%",
            "title" => "my trigger"
          }
        }
      },
      %{
        "insertedAt" => "2019-03-27T15:43:07.101046Z",
        "eventType" => "create_watchlist",
        "post" => nil,
        "user" => %{"id" => "4420"},
        "userList" => %{"isPublic" => true, "name" => "My Test List"},
        "userTrigger" => nil
      },
      %{
        "insertedAt" => "2019-03-27T15:43:07.051777Z",
        "eventType" => "publish_insight",
        "post" => %{"id" => "1889"},
        "user" => %{"id" => "4420"},
        "userList" => nil,
        "userTrigger" => nil
      }
    ]
  }
]
```


lib/sanbase/insight/post.ex
```diff
@@ -81,6 +85,28 @@ defmodule Sanbase.Insight.Post do
...
+    draft = draft()
+
+    with {:own_post?, %Post{user_id: ^user_id}} <- {:own_post?, post},
+         {:draft?, %Post{ready_state: ^draft}} <- {:draft?, post},
+         {:ok, %{post: post}} <- publish_post(post, user_id) do
+      {:ok, post}
+    else
+      {:error, _, error, _} ->
+        error_message = "Can't publish post with id #{post_id}"
+        Logger.error("#{error_message}, #{inspect(error)}")
```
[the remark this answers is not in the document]
tspenov: Good question - here I catch transaction error which could be db down or some kind of network error connecting to external services like discourse and discord, so IMO should be `Logger.error`. I think `warn` should be that something odd happened but it does not affect the response and `error` for something that happened and response is changed in this situation.

IvanIvanoff:
Do you think we can benefit from using something like:
https://elixirforum.com/t/polymorphic-field-in-absinthe/17740/2
https://hexdocs.pm/absinthe/Absinthe.Type.Union.html
https://hexdocs.pm/absinthe/Absinthe.Type.Interface.html

lib/sanbase/discourse/insight.ex
```diff
@@ -0,0 +1,40 @@
...
+  alias Sanbase.Discourse
+
+  def create_discourse_topic(%Post{id: id, title: title, inserted_at: inserted_at}) do
+    link = posts_url(id)
+
+    text = ~s"""
+      This topic hosts the discussion about [#{link}](#{link})
+    """
+
+    title = "##{id} | #{title} | #{DateTime.to_naive(inserted_at) |> to_string}"
```
IvanIvanoff:
```suggestion
    title = "##{id} | #{title} | #{inserted_at |> DateTime.truncate(:second) |> DateTime.to_naive()}"
```

`to_string` is not needed as DateTime implements the String.Chars protocol. `truncate` is for better title as we don't care for microseconds here

lib/sanbase/discourse/insight.ex
```diff
@@ -0,0 +1,40 @@
...
+
+    Discourse.Api.publish(title, text)
+    |> case do
+      {:ok,
+       %{
+         "topic_id" => topic_id,
+         "topic_slug" => topic_slug
+       }} ->
```
IvanIvanoff: You can change the formatting here to `  {:ok, %{"topic_id" => topic_id, "topic_slug" => topic_slug}} ->` and the formatter will keep it

lib/sanbase/discourse/insight.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  defp discourse_topic_url(topic_id, topic_slug) do
+    discourse_url()
+    |> URI.parse()
+    |> URI.merge("/t/#{topic_slug}/#{topic_id}")
+    |> URI.to_string()
+  end
+
+  defp posts_url(id), do: "#{sanbase_url()}/insights/read/#{id}"
+  defp sanbase_url(), do: Config.module_get(SanbaseWeb.Endpoint, :frontend_url)
```
IvanIvanoff: You can directly use `SanbaseWeb.Endpoint.frontend_url/0` and don't care what's the config name here

lib/sanbase/following/user_follower.ex
```diff
@@ -39,4 +39,11 @@ defmodule Sanbase.Following.UserFollower do
   end
 
   def unfollow(_, _), do: {:error, "User can't unfollow oneself"}
+
+  def following(user_id) do
```
IvanIvanoff: It's not clear what `following` means. Is this the users that `user_id` follows? I would call this `followed_by(user_id)` or something like this

lib/sanbase/insight/post.ex
```diff
@@ -86,6 +91,33 @@ defmodule Sanbase.Insight.Post do
...
+      {:nil?, nil} ->
+        {:error, "Can't publish post with id #{post_id}"}
+
+      {:draft?, _} ->
+        {:error, "Can't publish already published post with id: #{post_id}"}
+
+      {:own_post?, _} ->
+        {:error, "Can't publish not own post with id: #{post_id}"}
+
+      {:error, _, error, _} ->
```
IvanIvanoff: why 4 params here?

lib/sanbase/notifications/insight.ex
```diff
@@ -11,12 +11,14 @@ defmodule Sanbase.Notifications.Insight do
     |> publish()
     |> case do
       {:ok, %HTTPoison.Response{status_code: 204}} ->
-        :ok
+        {:ok, "Successfully published insight"}
 
       {:ok, %HTTPoison.Response{status_code: status_code}} ->
-        Logger.error("Cannot publish insight [#{post.id}] in discord: code[#{status_code}]")
+        {:error, "Can't publish insight"}
+        Logger.error("Can't publish insight [#{post.id}] in discord: code[#{status_code}]")
```
IvanIvanoff: You need to switch these 2 lines

lib/sanbase/signals/user_trigger.ex
```diff
@@ -146,9 +149,13 @@ defmodule Sanbase.Signals.UserTrigger do
   def create_user_trigger(%User{id: user_id} = _user, %{settings: settings} = params) do
     with {:nil?, false} <- {:nil?, is_nil(settings)},
          :ok <- valid?(settings) do
-      %UserTrigger{}
-      |> create_changeset(%{user_id: user_id, trigger: params})
-      |> Repo.insert()
+      changeset = %UserTrigger{} |> create_changeset(%{user_id: user_id, trigger: params})
+
+      if Map.get(params, :is_public) do
+        create_trigger_and_log_event(changeset)
```
IvanIvanoff: Do we really need the trigger creation to fail if there is some issue with the timeline event?
If not, I would make it
```elixir
Repo.insert(changeset)
|> maybe_log_event()
```
where `maybe_log_event/1` has one clause with `is_public=false` that does nothing and a second clause that logs the event

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,125 @@
...
+    field(:event_type, :string)
+    belongs_to(:user, User)
+    belongs_to(:post, Post)
+    belongs_to(:user_list, UserList)
+    belongs_to(:user_trigger, UserTrigger)
+
+    timestamps()
+  end
+
+  def publish_insight(), do: @publish_insight
```
IvanIvanoff: I would expect `publish_insight` function to be a function that publishes an insight as we have functions named exactly like this with this behavior. This will be better named `publish_insight_type`

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,125 @@
...
+    |> cast(attrs, [:event_type, :user_id, :post_id, :user_list_id, :user_trigger_id, :created_at])
+  end
+
+  def events(
+        %User{id: user_id},
+        %{limit: limit, cursor: %{type: cursor_type, datetime: cursor_datetime}}
+      ) do
+    TimelineEvent
+    |> events_by_followed_users(user_id, limit)
+    |> by_cursor(cursor_type, cursor_datetime)
```
IvanIvanoff: I would put the limit inside the `by_cursor` - it's more clear to limit it there

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,125 @@
...
+    from(
+      event in query,
+      where: event.created_at > ^datetime
+    )
+  end
+
+  defp events_with_cursor([]), do: {:ok, %{events: [], cursor: %{}}}
+
+  defp events_with_cursor(events) do
+    before_datetime = events |> List.last() |> Map.get(:created_at)
```
IvanIvanoff: Should be put an upper bound on `limit` as this can potentially iterate over huge lists

lib/sanbase/user_lists/user_list.ex
```diff
@@ -50,11 +54,33 @@ defmodule Sanbase.UserList do
...
 
-    UserList.by_id(id)
-    |> Repo.preload(:list_items)
-    |> update_changeset(params)
-    |> Repo.update()
+    if watchlist.is_public and Map.get(params, :list_items) do
+      update_watchlist_and_log_event(changeset)
+    else
+      Repo.update(changeset)
+    end
```
IvanIvanoff: I think in the future we need to have something like an event hub and send a notification there when a watchlist/insight/trigger/etc. is updated/created/deleted. `update_user_list` should only update the user lists and not performing any other side effects than that

lib/sanbase/insight/post.ex
```diff
@@ -188,6 +220,30 @@ defmodule Sanbase.Insight.Post do
...
+      publish_changeset(post, %{
+        discourse_topic_url: discourse_topic_url,
+        ready_state: Post.published()
+      })
+      |> Repo.update()
+    end)
+    |> Multi.run(:publish_in_discord, fn %{post: post} ->
+      Sanbase.Notifications.Insight.publish_in_discord(post)
+    end)
+    |> Multi.run(:create_timeline_event, fn %{post: post} ->
```
IvanIvanoff: This introduces too much coupling between the post and the timeline modules.
https://medium.com/elixirlabs/decoupled-modules-with-elixir-eventbus-a709b1479411

IvanIvanoff:
@tspenov @Partysun I see this is staying for a month, what is the priority here?

lib/sanbase/insight/post.ex
tspenov: Ecto.Multi returns `{:error, failed_operation, failed_value, changes_so_far}`

lib/sanbase/timeline/timeline_event.ex
tspenov: makes sense, but you can also query without cursor and still use the limit. You query with cursor when you want before or after certain datetime but you may want to fetch the items ordered by datetime with limit.

lib/sanbase/signals/user_trigger.ex
tspenov: reworked to be not coupled

test/sanbase_web/graphql/timeline_event_api_test.exs
```diff
@@ -0,0 +1,125 @@
...
+      )
+
+    insert(:timeline_event,
+      user_trigger: user_trigger,
+      user: user_to_follow,
+      event_type: TimelineEvent.create_public_trigger()
+    )
+
+    result = timeline_events_query(conn, "limit: 3")
+    assert result |> hd() |> Map.get("events") |> length() == 3
```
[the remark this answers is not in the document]
tspenov: Added one for cursor datetimes.

lib/sanbase/user_lists/user_list.ex
tspenov: Retry is good to have, though these are simple db calls and are not crucial if they fail (user won't be notified for event of followed user). I think current implementation is very simple but enough for the case. I mainly boils to calling these function when event happens.
```elixir
  def create_event_async(resource, type) do
    Task.Supervisor.async_nolink(Sanbase.TaskSupervisor, fn ->
      create_event(resource, %{
        event_type: type,
        user_id: resource.user_id
      })
    end)
  end
```

lib/sanbase/insight/post.ex
tspenov: Answered above.

lib/sanbase/timeline/timeline_event.ex
tspenov: Added a max_limit 100

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,135 @@
...
+
+  alias __MODULE__
+
+  @publish_insight_type "publish_insight"
+  @update_watchlist_type "update_watchlist"
+  @create_public_trigger_type "create_public_trigger"
+
+  @max_limit 100
+
+  @timestamps_opts [inserted_at: :created_at, updated_at: false, type: :utc_datetime]
```
IvanIvanoff: Let's not rename - we're not doing this in the other place

IvanIvanoff: The only thing we do is `updated_at: false` but this is only on tables we do not manage (clickhouse, timescale)

lib/sanbase/user_lists/user_list.ex
```diff
@@ -76,10 +79,32 @@ defmodule Sanbase.UserList do
...
+      by_id(id)
+      |> Repo.preload(:list_items)
+
+    changeset =
+      watchlist
+      |> update_changeset(params)
+
+    if watchlist.is_public and Map.get(params, :list_items) do
+      Repo.update(changeset)
+      |> log_timeline_event()
```
IvanIvanoff: I strongly disagree to have such logic that decides whether or not to add it to the timeline inside the user_list module. In the long run this will be really hard to maintain.

Instead - just do `maybe_log_timeline_event(watchlisit, changeset)` and let the function decide based on `is_public` and `%{changes: %{list_items: ...}} = changeset} whether to log.

lib/sanbase/user_lists/user_list.ex
```diff
@@ -76,10 +79,32 @@ defmodule Sanbase.UserList do
...
-    |> Repo.update()
+    watchlist =
+      by_id(id)
+      |> Repo.preload(:list_items)
+
+    changeset =
+      watchlist
+      |> update_changeset(params)
+
+    if watchlist.is_public and Map.get(params, :list_items) do
```
IvanIvanoff: If part of the changeset is to make the list private this will still create a timeline event for the other changes that are part of the update.

lib/sanbase/timeline/timeline_event.ex
tspenov: In this case it sounds better to me that event is created instead of inserted cause it is created out of other action.

lib/sanbase/user_lists/user_list.ex
tspenov: this is a good suggestion - I'll rework it so the logic is in `TimelineEvent`

lib/sanbase/discourse/insight.ex
```diff
@@ -0,0 +1,35 @@
...
+  require Sanbase.Utils.Config, as: Config
+
+  alias Sanbase.Insight.Post
+  alias Sanbase.Discourse
+
+  def create_discourse_topic(%Post{id: id, title: title, inserted_at: inserted_at}) do
+    link = posts_url(id)
+
+    text = ~s"""
+      This topic hosts the discussion about [#{link}](#{link})
```
IvanIvanoff: this will add leading whitespaces - it can safely be made with just `"`

lib/sanbase/notifications/insight.ex
```diff
@@ -11,13 +11,15 @@ defmodule Sanbase.Notifications.Insight do
     |> publish()
     |> case do
       {:ok, %HTTPoison.Response{status_code: 204}} ->
-        :ok
+        {:ok, "Successfully published insight"}
```
IvanIvanoff: Not needed as no one will log/return/read this

lib/sanbase/signals/user_trigger.ex
```diff
@@ -146,9 +148,10 @@ defmodule Sanbase.Signals.UserTrigger do
   def create_user_trigger(%User{id: user_id} = _user, %{settings: settings} = params) do
     with {:nil?, false} <- {:nil?, is_nil(settings)},
          :ok <- valid?(settings) do
-      %UserTrigger{}
-      |> create_changeset(%{user_id: user_id, trigger: params})
-      |> Repo.insert()
+      changeset = %UserTrigger{} |> create_changeset(%{user_id: user_id, trigger: params})
+
+      Repo.insert(changeset)
+      |> log_timeline_event(Map.get(params, :is_public))
```
IvanIvanoff: Updating a trigger is missing timeline log

tspenov: The create event is more important here. Also we are calling update trigger internally which makes things harder to differentiate. So skipping and will add if really necessary

lib/sanbase/following/user_follower.ex
```diff
@@ -39,4 +39,11 @@ defmodule Sanbase.Following.UserFollower do
   end
 
   def unfollow(_, _), do: {:error, "User can't unfollow oneself"}
+
+  def followed_by(user_id) do
+    Repo.get(User, user_id)
+    |> Repo.preload(:following)
+    |> Map.get(:following)
+    |> Enum.map(&Map.get(&1, :user_id))
```
IvanIvanoff: This will load into memory the whole user structs (~20+ fields).
The whole thing is equivalent to selecting all fields from the `UserFollower` table where `follower == user_id`

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,135 @@
...
+  alias Sanbase.Signals.UserTrigger
+  alias Sanbase.Following.UserFollower
+
+  alias __MODULE__
+
+  @publish_insight_type "publish_insight"
+  @update_watchlist_type "update_watchlist"
+  @create_public_trigger_type "create_public_trigger"
+
+  @max_limit 100
```
IvanIvanoff: Please add documentation and a more meaningful name - `@max_events_returned`.

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def create_event(%UserTrigger{id: id}, params) do
+    create_event(:user_trigger_id, id, params)
+  end
+
+  # private functions
+
+  defp create_event(type, id, params) do
+    %__MODULE__{} |> create_changeset(Map.put(params, type, id)) |> Repo.insert()
```
IvanIvanoff: Some documentation is needed here about what is happening

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,135 @@
...
+  defp by_cursor(query, :after, datetime) do
+    from(
+      event in query,
+      where: event.created_at > ^datetime
+    )
+  end
+
+  defp events_with_cursor([]), do: {:ok, %{events: [], cursor: %{}}}
+
+  defp events_with_cursor(events) do
```
IvanIvanoff: I am confused by the naming of functions here. We have `by_cursor` and `with_cursor` and semantically they sound the same to me. Documentation is needed

tspenov: `events_with_cursor` wraps the events with the cursor, `by_cursor` returns an ecto query filtering by given cursor. Any suggestion is welcome how to make them more self explaining.

lib/sanbase/following/user_follower.ex
```diff
@@ -39,4 +42,30 @@ defmodule Sanbase.Following.UserFollower do
...
+  Returns all user ids of users that are followed by certain user
+  """
+  def followed_by(user_id) do
+    from(
+      uf in __MODULE__,
+      where: uf.follower_id == ^user_id,
+      select: [:user_id]
+    )
+    |> Repo.all()
+    |> Enum.map(&Map.get(&1, :user_id))
```
IvanIvanoff: `from(uf in __MODULE__, where: uf.follower_id ==^user_id, select: uf.id) |> Repo.all` will return only the ids. I am not sure it has the expected behavior with `select: [:user_id]` as this returns everything and the select is kind of ignored

lib/sanbase/following/user_follower.ex
```diff
@@ -39,4 +42,30 @@ defmodule Sanbase.Following.UserFollower do
...
+  Returns all user ids of users that follow certain user
+  """
+  def following(user_id) do
+    from(
+      uf in __MODULE__,
+      where: uf.user_id == ^user_id,
+      select: [:follower_id]
+    )
+    |> Repo.all()
+    |> Enum.map(&Map.get(&1, :follower_id))
```
IvanIvanoff: Same as above

lib/sanbase/insight/post.ex
```diff
@@ -88,6 +92,33 @@ defmodule Sanbase.Insight.Post do
...
   def draft(), do: @draft
 
+  def publish(post_id, user_id) do
+    post_id = String.to_integer(post_id)
+    post = Repo.get(Post, post_id)
+    draft = draft()
+
+    with {:nil?, %Post{id: ^post_id}} <- {:nil?, post},
+         {:own_post?, %Post{user_id: ^user_id}} <- {:own_post?, post},
+         {:draft?, %Post{ready_state: ^draft}} <- {:draft?, post},
```
IvanIvanoff:
```suggestion
         {:draft?, %Post{ready_state: ^@draft}} <- {:draft?, post},
```

lib/sanbase/insight/post.ex
```diff
@@ -88,6 +92,33 @@ defmodule Sanbase.Insight.Post do
   def published(), do: @published
   def draft(), do: @draft
 
+  def publish(post_id, user_id) do
+    post_id = String.to_integer(post_id)
+    post = Repo.get(Post, post_id)
+    draft = draft()
```
IvanIvanoff:
```suggestion
```
You can remove this one batched with the change below

lib/sanbase/insight/post.ex
```diff
@@ -167,6 +198,28 @@ defmodule Sanbase.Insight.Post do
 
   # Helper functions
 
+  defp publish_post(post) do
+    publish_changeset = publish_changeset(post, %{ready_state: Post.published()})
+
+    with {:ok, post} <-
+           publish_changeset |> Repo.update(),
```
IvanIvanoff: can go on one line

lib/sanbase/insight/post.ex
```diff
@@ -167,6 +198,28 @@ defmodule Sanbase.Insight.Post do
...
+         {:ok, discourse_topic_url} <- Sanbase.Discourse.Insight.create_discourse_topic(post),
+         {:ok, post} <-
+           publish_changeset(post, %{discourse_topic_url: discourse_topic_url}) |> Repo.update() do
+      Task.Supervisor.async_nolink(Sanbase.TaskSupervisor, fn ->
+        Sanbase.Notifications.Insight.publish_in_discord(post)
+      end)
+
+      TimelineEvent.maybe_create_event_async(
```
IvanIvanoff: This can go inside the task above with a conditional, so the event will be created only if the topic is published :?

lib/sanbase/signals/user_trigger.ex
```diff
@@ -267,6 +270,13 @@ defmodule Sanbase.Signals.UserTrigger do
     end
   end
 
+  defp maybe_create_event({:ok, user_trigger}, changeset, event_type) do
+    TimelineEvent.maybe_create_event_async(event_type, user_trigger, changeset)
+    {:ok, user_trigger}
+  end
+
+  defp maybe_create_event(error_result, _, _), do: error_result
```
IvanIvanoff: cool, this way the `maybe_create_event` is pipe-able 😄

lib/sanbase/insight/post.ex
tspenov: You mean to try to create event only if it is published in discord ? I think it doesn't make sense cause these are not depending on each other.

lib/sanbase/user_lists/user_list.ex
```diff
@@ -76,10 +79,11 @@ defmodule Sanbase.UserList do
   def update_user_list(%{id: id} = params) do
     params = update_list_items_params(params, id)
 
-    by_id(id)
-    |> Repo.preload(:list_items)
-    |> update_changeset(params)
-    |> Repo.update()
+    watchlist = by_id(id) |> Repo.preload(:list_items)
+    changeset = watchlist |> update_changeset(params)
```
IvanIvanoff: Why is this broken into steps and cannot simply be:
```elixir
changeset = by_id |> preload |> update_chageset
```

tspenov: good catch - leftover form previous implementation

lib/sanbase/timeline/timeline_event.ex
IvanIvanoff: Having a consistent naming convention is preferable imo. If we use `timstamps()` and rename one of the fields and remove the other it has nothing in common with `timetsamps()` and could be confusing. We can make a custom `created_at` field in this case.

lib/sanbase/insight/post.ex
```diff
@@ -167,6 +197,27 @@ defmodule Sanbase.Insight.Post do
...
 
+  defp publish_post(post) do
+    publish_changeset = publish_changeset(post, %{ready_state: Post.published()})
+
+    with {:ok, post} <- publish_changeset |> Repo.update(),
+         {:ok, discourse_topic_url} <- Sanbase.Discourse.Insight.create_discourse_topic(post),
+         {:ok, post} <-
+           publish_changeset(post, %{discourse_topic_url: discourse_topic_url}) |> Repo.update() do
+      Task.Supervisor.async_nolink(Sanbase.TaskSupervisor, fn ->
+        Sanbase.Notifications.Insight.publish_in_discord(post)
```
IvanIvanoff: Isn't the `create_discourse_topic` already publishing in discord (it calls `Discourse.Api.publish(title, text)`)

These 2 functions are in 2 different modules but they do the same thing it seems.

lib/sanbase/timeline/timeline_event.ex
tspenov: It is not exactly like this because we have to make trigger to update `inserted_at`, `updated_at` at create/update event, which is better to be handled by the built in functionality. Though I renamed it for consistency which is really important cause it is exposed in the API.

lib/sanbase/insight/post.ex
IvanIvanoff: On a second though - no. Though it is kind of confusing why only Discord needs to happen async

lib/sanbase/notifications/insight.ex
```diff
@@ -14,10 +14,12 @@ defmodule Sanbase.Notifications.Insight do
         :ok
 
       {:ok, %HTTPoison.Response{status_code: status_code}} ->
-        Logger.error("Cannot publish insight [#{post.id}] in discord: code[#{status_code}]")
+        Logger.error("Can't publish insight [#{post.id}] in discord: code[#{status_code}]")
+        {:error, "Can't publish insight"}
```
IvanIvanoff:
```suggestion
        {:error, "Can't publish insight creation notification to Discord"}
```

lib/sanbase/notifications/insight.ex
```diff
@@ -14,10 +14,12 @@ defmodule Sanbase.Notifications.Insight do
...
 
       {:error, error} ->
         Logger.error("Cannot publish insight [#{post.id}] in discord: " <> inspect(error))
+        {:error, "Can't publish insight"}
```
IvanIvanoff:
```suggestion
        {:error, "Can't publish insight creation notification to Discord"}
```

lib/sanbase/insight/post.ex
tspenov: Thats because we are creating a discourse topic and updating the insight with the link to the discussion.

lib/sanbase/signals/user_trigger.ex
```diff
@@ -38,6 +39,7 @@ defmodule Sanbase.Signals.UserTrigger do
 
     has_one(:featured_item, Sanbase.FeaturedItem, on_delete: :delete_all)
     has_many(:signals_historical_activity, HistoricalActivity, on_delete: :delete_all)
+    has_many(:timeline_events, TimelineEvent, on_delete: :delete_all)
```
IvanIvanoff: Is deleting of notifications desirable here?

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -0,0 +1,212 @@
...
+    TimelineEvent
+    |> events_by_followed_users(user_id, min(limit, @max_events_returned))
+    |> Repo.all()
+    |> events_with_cursor()
+  end
+
+  def events(_, _), do: {:error, "Bad arguments"}
+
+  @doc """
+    Asynchronously create a timeline event only if all criterias are met.
```
IvanIvanoff: Indenting inside `@doc` is not really correct - I was doing it before but after looking at the Elixir source code seems like no one else is doing it.

lib/sanbase_web/graphql/resolvers/timeline/timeline_event_resolver.ex
```diff
@@ -0,0 +1,17 @@
+defmodule SanbaseWeb.Graphql.Resolvers.TimelineEventResolver do
+  require Logger
+
+  alias Sanbase.Timeline.TimelineEvent
+
+  def timeline_events(_root, args, %{
+        context: %{auth: %{current_user: current_user}}
+      }) do
+    case TimelineEvent.events(current_user, args) do
```
IvanIvanoff: just `TimelineEvent.events` will do as in both cases you return what you get

lib/sanbase/signals/user_trigger.ex
tspenov: yeah, probably we should keep the events

tspenov: Though these are foreign keys so if we don't delete them they won't be very useful cause the attached resource will be missing. Also DB will throw a constraint error on trying to delete the resource cause it is a foreign key in another table.

PR #1083, opened 2019-03-29
Consistency to graphql clickhouse

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/daily_active_deposits.ex
+++ b/lib/sanbase/clickhouse/daily_active_deposits.ex
@@ -1,12 +1,23 @@
 defmodule Sanbase.Clickhouse.DailyActiveDeposits do
   @moduledoc ~s"""
-  Uses ClickHouse to calculate daily active deposits
+  Uses ClickHouse to calculate daily active deposits.
+  The number of unique deposit addresses that have been active
   """
+
+  use Ecto.Schema
 
   require Sanbase.ClickhouseRepo, as: ClickhouseRepo
   import Sanbase.Math, only: [to_integer: 1]
 
   alias Sanbase.DateTimeUtils
+
+  @table "daily_active_deposits"
+  schema @table do
+    field(:dt, :utc_datetime)
+    field(:contract, :string)
+    field(:exchange, :string)
+    field(:total_addresses, :integer)
+  end
 
   @type active_deposits :: %{
           datetime: DateTime.t(),
@@ -52,7 +63,7 @@ defmodule Sanbase.Clickhouse.DailyActiveDeposits do
       SELECT
         toDateTime(intDiv(toUInt32(dt), ?1) * ?1) AS time,
         SUM(total_addresses) AS value
-      FROM daily_active_deposits
+      FROM #{@table}
       PREWHERE
         contract = ?3 AND
         dt >= toDateTime(?4) AND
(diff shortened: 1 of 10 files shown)


lib/sanbase/clickhouse/daily_active_deposits.ex
```diff
@@ -1,13 +1,24 @@
 defmodule Sanbase.Clickhouse.DailyActiveDeposits do
   @moduledoc ~s"""
-  Uses ClickHouse to calculate daily active deposits
+  Uses ClickHouse to calculate daily active deposits.
+  The number of unique deposit addresses that have been active
```
IvanIvanoff: This second sentence is strange - better combine them into `Uses ClickHouse to calculate the number of unique daily active deposits` or something similar

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -146,17 +147,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
 
-        {:error, "Can't calculate historical balances for project with coinmarketcap_id #{slug}"}
+      {:error, error} ->
+        error_msg = generate_error_message_for_project("historical balances", slug)
+        log_error(error_msg, error)
+        {:error, error_msg}
     end
   end
+
+  defp log_error(error_msg, error) do
```
IvanIvanoff: we can have these functions in `error_handling.ex` and use them in other places also. Also it will be nice to have some UUID returned in the api so we can easily trace the errors (again as in the ErrorHandling)

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -146,17 +147,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+    Logger.warn(error_msg <> " Reason: #{inspect(error)}")
+  end
+
+  defp generate_error_message(metric_name) do
+    "Can't calculate #{metric_name}."
+  end
+
+  defp generate_error_message_for_project(metric_name, slug) do
+    "Can't calculate #{metric_name} for project with coinmarketcap_id: #{slug}."
```
IvanIvanoff: I would prefer using `slug` in the future instead of coimarketcap_id. the cmc is an implementation detail

test/sanbase/clickhouse/daily_active_deposits_test.exs
```diff
@@ -7,12 +7,7 @@ defmodule SanbaseWeb.Clickhouse.DailyActiveDepositsTest do
   require Sanbase.ClickhouseRepo
 
   setup do
-    project =
-      insert(:project, %{
-        coinmarketcap_id: "santiment",
-        ticker: "SAN",
-        main_contract_address: "0x123"
-      })
+    project = insert(:project, %{main_contract_address: "0x123"})
```
IvanIvanoff: Even if Santiment is default let's keep it so the tests are more clear

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -28,24 +28,25 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
       {:ok, network_growth}
     else
-      error ->
-        Logger.error("Can't calculate network growth. Reason: #{inspect(error)}")
-
-        {:error, "Can't calculate network growth"}
+      {:error, error} ->
+        error_msg = generate_error_message_for_project("network growth", slug)
+        log_error(error_msg, error)
+        {:error, error_msg}
```
IvanIvanoff: I am not sure if we need to keep this behavior. In case there is no contract address this is not really an error case I guess? It's more like a case where `null` should be returned

PR #1087, opened 2019-04-01
Fix tests with datetimes outside free tier

tspenov:
For some reason these are not failing though they have only 3 months historical access. But still this should be better.


IvanIvanoff:
These tests are for santiment, so historical data can be accessed

IvanIvanoff:
We could have a lot of problems with tests with dynamic dates - having fixed datetimes is better so when you run the tests 2 times they'll use the same data.
What we can do is instead of a simple connection, to use a jwt connection and a user with san tokens

tspenov:
@IvanIvanoff Yes, missed that it is actually `santiment` project. Reverted old fixes with dynamic data and added authenticated and staked user.

PR #1089, opened 2019-04-01
Better seeds

kalkov:
#### Summary
Few manual steps were needed to have a properly working project - adding contract and adjusting token decimals. Should not be the case anymore.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -5,10 +5,8 @@
 # Inside the script, you can read and write to any of your
 # repositories directly:
 #
-#     Sanbase.Repo.insert!(%Sanbase.SomeSchema{})
+#     Sanbase.Repo.insert(%Sanbase.SomeSchema{})
 #
-# We recommend using the bang functions (`insert!`, `update!`
-# and so on) as they will fail if something goes wrong.
 
 alias Sanbase.Model.Project
 alias Sanbase.Model.ProjectEthAddress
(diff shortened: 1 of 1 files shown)


priv/repo/seeds.exs
```diff
@@ -164,39 +167,39 @@ end
...
+|> Enum.each(insert_on_conflict_nothing)
 
-user =
-  %User{
-    email: "john.d@santiment.net",
-    username: "John Dow",
-    salt: "LgcwR3e98PR/gvgV7Ph1+ZXnw4yhTz25k08QLi/39qdCt/V0XOGlJRiL938NtJk0"
-  }
-  |> Repo.insert!()
+defmodule InsertUser do
```
tspenov: I think we can take it a step further and extract these module into own files and in seeds file we have only the raw data and calls to`InsertUser.run, etc...` functions. Also for the `Projects`, `InsertExchangeAddresses`, `Infrastructure` etc... What do you think ?

tspenov: Also all of these `.run` functions to be made to handle a list of tuples of data.

tspenov: But of course these changes are still an improvement so feel free to skip if you think it is too out of scope.

IvanIvanoff: https://github.com/santiment/sanbase2/blob/master/priv/timescale_repo/seeds.exs

kalkov: @tspenov Good suggestion, will step further but in separate PR. The idea behind this PR was to quickly fix the seeds so I can continue working.

PR #1093, opened 2019-04-03
Percent of DAD in DAA

kalkov:
#### Summary
Apart from adding the share of deposits to the Daily Active Addresses I have also refactored the codebase to solely use Clickhouse. 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -110,6 +111,43 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+           DailyActiveAddresses.average_active_addresses(
+             contract,
+             from,
+             to,
+             interval
+           ) do
+      {:ok, daily_active_addresses}
+    else
+      {:error, {:missing_contract, error_msg}} ->
+        {:error, error_msg}
```
IvanIvanoff: Let's start returning `{:ok, []}` 🤔

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
```diff
@@ -112,47 +114,88 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
...
+      if(SUM(addresses) != 0, SUM(deposits)/SUM(addresses), 0) * 100 AS share_of_deposits
     FROM (
       SELECT
-        toDateTime(intDiv(toUInt32(?4 + (number + 1) * ?1), ?1) * ?1) as time,
-        toUInt32(0) AS value
-      FROM numbers(?2)
+        toDateTime(intDiv(toUInt32(?4 + (number + 1) * ?1), ?1) * ?1) AS time,
+        toUInt32(0) AS addresses,
+        toUInt32(0) AS deposits
+      FROM numbers(10)
```
IvanIvanoff: probably `numbers(span)`

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
```diff
@@ -112,47 +114,88 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
...
         FROM erc20_daily_active_addresses
-        PREWHERE contract = ?3 AND
-        dt < toDateTime(today()) AND
-        dt >= toDateTime(?4) AND
-        dt <= toDateTime(?5)
+        PREWHERE
+          contract = ?3 AND
+          dt < toDateTime(today()) AND
+          dt >= toDateTime(?4) AND
+          dt <= toDateTime(?5)
```
IvanIvanoff: I guess the part
```
 dt < toDateTime(today()) AND
dt <= toDateTime(?5)
```
serves to not allow querying data for today as it's not full. On the other hand,  if we exclude the realtime data what's the point of using ClickHouse in this case? The CH DAA's advantage is that it's realtime (opposed to Timescale where it's updated once a day)

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
```diff
@@ -112,47 +114,88 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
...
         GROUP BY contract, dt
 
         UNION ALL
 
-        SELECT toStartOfDay(dt) as dt, uniq(address) as total_addresses
+        SELECT
+          toStartOfDay(dt) AS dt,
+          uniq(address) AS total_addresses
```
IvanIvanoff: Is this needed because of possible duplicates? I think running `uniq` on large intervals could be a performance problem. Have you tested it?

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
```diff
@@ -112,47 +114,88 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
...
+        SELECT
+          sum(total_addresses) AS deposits,
+          dt
+        FROM(
+          SELECT DISTINCT
+            exchange,
+            dt,
+            contract,
+            total_addresses
+          FROM daily_active_deposits
```
IvanIvanoff: I think we need to have 2 different queries for when the client requests DAA and when deposit addresses are requested.
In case of only requesting active addresses the query will still look through all N million deposit addresses.

kalkov: Do we have such an example in out codebase?

IvanIvanoff: The frontend is using only daily active addresses and will continue to do so after this is deployed. Also the API allows such behavior so it will happen in the future, too.

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
kalkov:
```sql
SELECT count(address)
FROM erc20_daily_active_addresses_list 
WHERE contract = '0x0d8775f648430679a709e98d2b0cb6250d2887ef'

┌─count(address)─┐
│         531168 │
└────────────────┘

1 rows in set. Elapsed: 0.220 sec. Processed 1.45 million rows, 147.69 MB (6.58 million rows/s., 670.98 MB/s.) 
```
```sql
SELECT uniq(address)
FROM erc20_daily_active_addresses_list 
WHERE contract = '0x0d8775f648430679a709e98d2b0cb6250d2887ef'

┌─uniq(address)─┐
│        277444 │
└───────────────┘

1 rows in set. Elapsed: 0.269 sec. Processed 1.45 million rows, 147.67 MB (5.38 million rows/s., 548.56 MB/s.)
```

```sql
SELECT count(address)
FROM eth_daily_active_addresses_list 

┌─count(address)─┐
│      188936548 │
└────────────────┘

1 rows in set. Elapsed: 0.905 sec. Processed 188.94 million rows, 9.64 GB (208.88 million rows/s., 10.65 GB/s.) 
```
```sql
SELECT uniq(address)
FROM eth_daily_active_addresses_list 

┌─uniq(address)─┐
│      47496486 │
└───────────────┘

1 rows in set. Elapsed: 1.780 sec. Processed 188.94 million rows, 9.64 GB (106.17 million rows/s., 5.41 GB/s.) 
```

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
kalkov: I understand it this way: 
Get `total_addresses`(precalculated number of unique addresses per day) from `erc20_daily_active_addresses` table for the days before today and union with the number of unique addresses for today(calculated at the moment) from `erc20_daily_active_addresses_list` table.

lib/sanbase/clickhouse/daily_active_addresses.ex
```diff
@@ -23,11 +23,20 @@ defmodule Sanbase.Clickhouse.DailyActiveAddresses do
...
 
-  def average_active_addresses(eth, from, to, interval) when eth in @ethereum do
-    Eth.average_active_addresses(from, to, interval)
+  def average_active_addresses(eth, from, to, interval, requested_fields) when eth in @ethereum do
+    Eth.average_active_addresses(from, to, interval, requested_fields)
   end
 
-  def average_active_addresses(contract, from, to, interval) do
-    Erc20.average_active_addresses(contract, from, to, interval)
+  def average_active_addresses(contract, from, to, interval, requested_fields) do
```
IvanIvanoff: This breaks the DAA signals - we need to have a default for the `requested_fields` param and the default should keep the old behavior.

I would prefer to have different queries for both:
- `average_active_addresses`
- `average_active_addresses_with_deposits` or something like this.

lib/sanbase/clickhouse/daily_active_addresses.ex
```diff
@@ -23,11 +23,20 @@ defmodule Sanbase.Clickhouse.DailyActiveAddresses do
...
+    if include_deposits_share?(requested_fields) do
+      Erc20.average_active_addresses_with_deposits_share(contract, from, to, interval)
+    else
+      Erc20.average_active_addresses(contract, from, to, interval)
+    end
+  end
+
+  defp include_deposits_share?(requested_fields) do
+    Enum.member?(requested_fields, "activeDeposits") ||
+      Enum.member?(requested_fields, "shareOfDeposits")
```
IvanIvanoff: Probably won't catch all situations as the API allows you to use both camel case and snake case

lib/sanbase/clickhouse/daily_active_addresses.ex
```diff
@@ -28,6 +28,14 @@ defmodule Sanbase.Clickhouse.DailyActiveAddresses do
   end
 
   def average_active_addresses(contract, from, to, interval) do
-    Erc20.average_active_addresses(contract, from, to, interval)
+    Erc20.average_active_addresses_with_deposits(contract, from, to, interval)
```
valo: I suspect this needs to be 

```suggestion
    Erc20.average_active_addresses(contract, from, to, interval)
```

kalkov:
Found an issue.
When you request: 
```graphql
{ 
  projectBySlug(slug: "ethereum") {averageDailyActiveAddresses(from: "2018-01-16T00:00:00.000Z", to: "2019-04-04T00:00:00.000Z")}
}
```

it will actually execute `Sanbase.Clickhouse.Erc20DailyActiveAddresses.average_active_addresses(contract, from, to)` instead of `Sanbase.Clickhouse.EthDailyActiveAddresses.average_active_addresses(from, to)`

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
valo: Have you tested this query if it updates the DAA as the day progresses? I suspect this query won't include the active address from today as `toDateTime(today())` returns the beginning of today.

valo: Nevermind. I misunderstood the query. Seems to be right.

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
valo: This query will be run only on the data from today, so it will be super fast (because of the `dt >= toDateTime(today())` clause)

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
```diff
@@ -106,67 +141,159 @@ defmodule Sanbase.Clickhouse.Erc20DailyActiveAddresses do
...
+          PREWHERE
+            contract = ?3 AND
+            dt < toDateTime(today()) AND
```
valo: What about if we request the current daily active deposits? Isn't this cause going to filter this out?

lib/sanbase/clickhouse/daily_active_addresses/eth_daily_active_addresses.ex
```diff
@@ -91,60 +127,144 @@ defmodule Sanbase.Clickhouse.EthDailyActiveAddresses do
...
+          dt
+        FROM(
+          SELECT DISTINCT
+            exchange,
+            dt,
+            contract,
+            total_addresses
+          FROM daily_active_deposits
+          PREWHERE
+            dt < toDateTime(today()) AND
```
valo: I suspect this should be `dt <= toDateTime(today())`

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
valo: I suspect this should be dt <= toDateTime(today())

lib/sanbase/clickhouse/daily_active_addresses/eth_daily_active_addresses.ex
```diff
@@ -91,60 +127,144 @@ defmodule Sanbase.Clickhouse.EthDailyActiveAddresses do
...
+        SELECT
+          toStartOfDay(dt) AS dt,
+          anyLast(total_addresses) AS total_addresses
         FROM eth_daily_active_addresses
-        WHERE
-        dt < toDateTime(today()) AND
-        dt >= toDateTime(?3) AND
-        dt <= toDateTime(?4)
+        PREWHERE
+          dt < toDateTime(today()) AND
```
valo: I suspect this should be dt <= toDateTime(today())

lib/sanbase_web/graphql/schema/clickhouse_types.ex
```diff
@@ -1,6 +1,13 @@
 defmodule SanbaseWeb.Graphql.ClickhouseTypes do
   use Absinthe.Schema.Notation
 
+  object :active_addresses do
+    field(:datetime, non_null(:datetime))
+    field(:active_addresses, non_null(:integer))
+    field(:active_deposits, non_null(:integer))
+    field(:share_of_deposits, :float)
```
valo: I think we should have implemented this as a separate time series... It would have saved quite a bit of "hacks" like https://github.com/santiment/sanbase2/pull/1093/files#diff-1cf42fc6714488fe1249edf4db53048bR154

lib/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses.ex
valo: I think we are making things unnecessary complex by trying to put the 3 metrics into a single query. Let's discuss on Monday

kalkov:
We tried to do too much in this PR, closing. 
The new metric is introduced in #1107 and a PR for refactoring DAA to use Clickhouse instead of TimescaleDB will follow.

PR #1098, opened 2019-04-04
Added social dominance query and schema for it

valo:
Broke down the social dominance query into a separate file, as the main
schema is getting out of hand.


lib/sanbase_web/graphql/schema.ex
```diff
@@ -585,6 +586,8 @@ defmodule SanbaseWeb.Graphql.Schema do
       resolve(&TechIndicatorsResolver.social_volume/3)
     end
 
+    import_fields :social_data_queries
```
IvanIvanoff: :tada:

lib/sanbase_web/graphql/schema/social_data_queries.ex
```diff
@@ -0,0 +1,37 @@
+defmodule SanbaseWeb.Graphql.Schema.SocialDataQueries do
+  use Absinthe.Schema.Notation
+
+  alias SanbaseWeb.Graphql.Resolvers.TechIndicators
+  alias SanbaseWeb.Graphql.Complexity
+  alias SanbaseWeb.Graphql.Middlewares.ApiTimeframeRestriction
+
+  import_types(SanbaseWeb.Graphql.SocialDataTypes)
```
IvanIvanoff: Will this work if you have the `import_types` in `schema.ex`? I am not sure how well this approach will work when we need the same types in multiple `Queries` modules

valo: It won't work. That's why this change: https://github.com/santiment/sanbase2/pull/1098/files#diff-e77f0e25b129cf9d83250e9c1ef10d6cL58

We will need to take care of circular dependencies somehow. If certain types are shared we might not be able to break the queries into multiple files.

lib/sanbase/social_data/social_dominance.ex
```diff
@@ -0,0 +1,103 @@
...
+    end
+  end
+
+  defp social_dominance_request(
+         datetime_from,
+         datetime_to,
+         interval,
+         source
+       ) do
+    cache_key = Cache.cache_key(:api_request, %{
```
IvanIvanoff: `:social_dominance_api_request` to include the name of the query

IvanIvanoff:
@also ru the formatter

test/sanbase/social_data/social_dominance_test.exs
```diff
@@ -0,0 +1,148 @@
...
+
+    [
+      project: project
+    ]
+  end
+
+  describe "social_dominance/5" do
+    test "response: success" do
+      from = 1_523_876_400
+      to = 1_523_880_000
```
IvanIvanoff:
```
from = DateTime.from_naive!(~N[2018-04-16 10:00:00], "Etc/UTC")
to = DateTime.from_naive!(~N[2018-04-16 22:00:00], "Etc/UTC")
```
and you can drop the `from_unix!`

valo:
Ran the format

PR #1099, opened 2019-04-04
The DocumentProvider recognizes interpolated datetimes

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/cache/cache.ex
+++ b/lib/sanbase_web/graphql/cache/cache.ex
@@ -210,13 +210,12 @@ defmodule SanbaseWeb.Graphql.Cache do
 
     ttl = base_ttl + ({name, slug} |> :erlang.phash2(max_ttl_offset))
 
-    args_hash =
+    args =
       args
       |> convert_values(ttl)
-      |> Jason.encode_to_iodata!()
 
     cache_key =
-      [name, args_hash]
+      [name, args]
       |> :erlang.phash2()
 
     {cache_key, ttl}
@@ -224,17 +223,24 @@ defmodule SanbaseWeb.Graphql.Cache do
 
   # Convert the values for using in the cache. A special treatement is done for
   # `%DateTime{}` so all datetimes in a @ttl sized window are treated the same
-  defp convert_values(args, ttl) do
+  defp convert_values(%DateTime{} = v, ttl) do
+    div(DateTime.to_unix(v, :second), ttl)
+  end
+
+  defp convert_values(%_{} = v, _), do: Map.from_struct(v)
+
+  defp convert_values(args, ttl) when is_list(args) or is_map(args) do
     args
     |> Enum.map(fn
-      {k, %DateTime{} = v} ->
-        {k, div(DateTime.to_unix(v, :second), ttl)}
+      {k, v} ->
+        [k, convert_values(v, ttl)]
 
-      pair ->
-        pair
+      data ->
+        convert_values(data, ttl)
     end)
-    |> Map.new()
   end
+
+  defp convert_values(v, _), do: v
 
   defp captured_mfa_name(captured_mfa) do
     captured_mfa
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/document/document_provider.ex
```diff
@@ -74,11 +67,43 @@ defmodule SanbseWeb.Graphql.Phase.Document.Execution.Resolution do
...
 
-    {:ok, %{bp_root | result: result}}
+    SanbaseWeb.Graphql.Cache.cache_key("bp_root", bp_root |> remove())
+    |> IO.inspect(label: "CACHE_KEY")
+
+    Absinthe.Phase.Document.Execution.Resolution.run(bp_root, options)
+  end
+
+  @cache_fields [:name, :argument_data, :selection_set, :selections, :fragments, :operations]
+  defp remove(%DateTime{} = dt), do: dt
```
valo: I think we can have a better name for this, like `sanitize_blueprint` or something like that

lib/sanbase_web/graphql/cache/cache.ex
```diff
@@ -210,32 +210,38 @@ defmodule SanbaseWeb.Graphql.Cache do
...
   # Convert the values for using in the cache. A special treatement is done for
   # `%DateTime{}` so all datetimes in a @ttl sized window are treated the same
-  defp convert_values(args, ttl) do
+  defp convert_values(%DateTime{} = v, ttl) do
+    div(DateTime.to_unix(v, :second), ttl)
+  end
+
+  defp convert_values(%_{} = v, _), do: Map.from_struct(v)
+
+  defp convert_values(args, ttl) when is_list(args) or is_map(args) do
```
valo: Is this function really working for lists that doesn't contain tuples?

valo: I just managed to parse the nesting. It will work indeed

lib/sanbase_web/graphql/document/document_provider.ex
```diff
@@ -2,53 +2,39 @@ defmodule SanbaseWeb.Graphql.DocumentProvider do
...
-          SanbseWeb.Graphql.Phase.Document.Execution.Resolution
-        )
-        |> Absinthe.Pipeline.without(Absinthe.Phase.Document.Result)
-    end
+  def pipeline(%{pipeline: pipeline}) do
+    pipeline
+    |> Absinthe.Pipeline.replace(
+      Absinthe.Phase.Document.Execution.Resolution,
+      SanbseWeb.Graphql.Phase.Document.Execution.Resolution
+    )
```
valo: I think that we can plug less "violently" here, by putting our custom resolution before the absinthe one. In that case we can reduce the amount of code in our custom resolution. Potentially we can also name it with a better name, like `SanbseWeb.Graphql.Phase.Document.Execution.QueryCache`

valo:
If we manage to put the `before_send` in the document provider we will have a fully self contained component we can open source. It will be a powerful and smart query level cache for GraphQL :)

lib/sanbase_web/graphql/document/document_provider.ex
IvanIvanoff: Sounds good, will try it

IvanIvanoff:
@valo all of these changes are dependent on the `Cache` module also. If someone else decides to use our caching solution all they need to change is the `before_send` hook as it contains the list of queries that are going to be cached.
I can think about packing the whole Cache + con_cache + this doc provider in a separate library

valo:
> I can think about packing the whole Cache + con_cache + this doc provider in a separate library

Let's think about it and we can eventually do something. For now I think we are good to go.

valo:
Let's also test this in production a bit. We might be missing some issues around our approach.

PR #1100, opened 2019-04-04
Correctly transform the datetimes TTL for the cache

IvanIvanoff:
#### Summary
Some time ago the ttl was changed from `:timer.minutes(5)` to just `300` but the datetime was not changed from `:millisecond` to `:second`. The output of `:timer.minutes(5)` is `300000` which is 300 seconds in milliseconds.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


valo:
Wow! So the cache was not really caching for 5 min, but for 300ms?

PR #1101, opened 2019-04-04
Add long desc projects

tspenov:


lib/sanbase_web/admin/model/project.ex
```diff
@@ -120,6 +120,7 @@ defmodule Sanbase.ExAdmin.Model.Project do
         input(project, :token_decimals)
         input(project, :token_supply)
         input(project, :description)
+        input(project, :long_description, type: :text)
```
IvanIvanoff: isn't it ok to be `:string` here?

priv/repo/migrations/20190404144631_import_project_long_descriptions_from_csv.exs
```diff
@@ -0,0 +1,26 @@
...
+
+    Path.expand("project_long_desc.csv", __DIR__)
+    |> File.stream!()
+    |> CSV.decode()
+    |> Stream.drop(1)
+    |> Enum.map(fn {:ok, line} ->
+      [_, ticker, name, long_desc] = line |> Enum.map(&String.trim/1)
+
+      Project
+      |> where([p], p.ticker == ^ticker)
```
IvanIvanoff: duplicates

lib/sanbase_web/admin/model/project.ex
tspenov: to display `textarea` `type: :text` is needed otherwise displays input (probably gets it from the schema its still :string)

priv/repo/migrations/20190404144631_import_project_long_descriptions_from_csv.exs
tspenov: using slug now

PR #1102, opened 2019-04-04
Add percent of token supply on exchanges metric

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


test/sanbase_web/graphql/clickhouse/percent_of_tokens_supply_on_exchanges_test.exs
```diff
@@ -0,0 +1,150 @@
+defmodule SanbaseWeb.Graphql.Clickhouse.PercentOfTokenSupplyOnExchangesTest do
```
kalkov: @IvanIvanoff @tspenov Should I log in the user in this test?

valo: I think with the current implementation the test will stop working within a few days, as the anon user will lose access to the historical data.

One option is to login a user with full access, the other option will be to use a `from/to` interval which falls within the free timeframe.

kalkov: @valo I'm using santiment as token, it should have infinite access?

lib/sanbase/clickhouse/percent_of_token_supply_on_exchanges.ex
```diff
@@ -0,0 +1,92 @@
...
+    query = """
+    SELECT
+      time,
+      1 - non_exchange_token_supply / total_supply AS tokens_on_exchanges_ratio
+    FROM (
+      SELECT
+        toUnixTimestamp((intDiv(toUInt32(toDateTime(dt)), ?1) * ?1)) AS time,
+        avg(value / pow(10, token_decimals)) AS total_supply
+      FROM erc20_balances
+      GLOBAL ANY INNER JOIN (
```
tspenov: we can use: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/model/project/project.ex#L342 and take contract and decimals by slug from postgres instead of using join

kalkov: @tspenov : Fixed

lib/sanbase/clickhouse/percent_of_token_supply_on_exchanges.ex
```diff
@@ -0,0 +1,85 @@
...
+      SELECT
+        toUnixTimestamp((intDiv(toUInt32(toDateTime(dt)), ?1) * ?1)) AS time,
+        avg(value / pow(10, token_decimals)) AS total_supply
+      FROM erc20_balances
+      PREWHERE
+        dt >= toDateTime(?3) AND
+        dt <= toDateTime(?4) AND
+        address = 'TOTAL' AND
+        sign = 1 AND
+        contract = (
```
tspenov: This `contract` query also can be substituted by fetching first contract by slug: 
```elixir
{:ok, contract, token_decimals} = Project.contract_info_by_slug(slug)
```

kalkov:
Will address the token decimals into a separate PR.

test/sanbase_web/graphql/clickhouse/percent_of_tokens_supply_on_exchanges_test.exs
valo: Yes. That will work

valo:
OK. Let's fix that later

lib/sanbase/clickhouse/percent_of_token_supply_on_exchanges.ex
```diff
@@ -0,0 +1,92 @@
...
+  def percent_on_exchanges(slug, from, to, interval) do
+    {query, args} = percent_on_exchanges_query(slug, from, to, interval)
+
+    ClickhouseRepo.query_transform(
+      query,
+      args,
+      fn [dt, tokens_on_exchanges_ratio] ->
+        %{
+          datetime: DateTime.from_unix!(dt),
+          percent_on_exchanges: tokens_on_exchanges_ratio * 100
```
IvanIvanoff: deposits, nvt and mvrv use ratio instead of %

valo: These ratios are different from percentages IMO. For example MVRV show be transformed by subtracting 1.0 from it, so that it really becomes a useful percent. We don't do this as it depends how you present it. Here I think rendering as percent will be exactly as it will be shown on the client side.

PR #1104, opened 2019-04-05
Reduce the DocumentProvider cache ttl settings

IvanIvanoff:
The main reason is that the resolver-level caches have random TTL
between 5 and 7 minutes. If we keep the same values for the
DocumentProvider cache we can reach 10+ minutes of cache length

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/document/document_provider.ex
+++ b/lib/sanbase_web/graphql/document/document_provider.ex
@@ -72,7 +72,9 @@ defmodule SanbseWeb.Graphql.Phase.Document.Execution.CacheDocument do
     cache_key =
       SanbaseWeb.Graphql.Cache.cache_key(
         {"bp_root", permissions},
-        santize_blueprint(bp_root)
+        santize_blueprint(bp_root),
+        ttl: 120,
+        max_ttl_offset: 90
       )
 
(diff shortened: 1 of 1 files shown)


lib/sanbase_web/graphql/document/document_provider.ex
```diff
@@ -72,7 +72,9 @@ defmodule SanbseWeb.Graphql.Phase.Document.Execution.CacheDocument do
     cache_key =
       SanbaseWeb.Graphql.Cache.cache_key(
         {"bp_root", permissions},
-        santize_blueprint(bp_root)
+        santize_blueprint(bp_root),
+        ttl: 120,
+        max_ttl_offset: 90
```
valo: Why don't we set the TTL to be so that the max cache interval to be 5min? So that `ttl + max_ttl_offset = 300`

IvanIvanoff: Because we're ignoring that some part of the whole query could have failed and was not cached at resolver level. Reducing this DocumentProvider TTL will allow the backend to heal such issues earlier.

PR #1107, opened 2019-04-06
Share of deposits

kalkov:
#### Summary

Share of deposits from Daily Active Addresses

**Note**
`activeAddresses` & `activeDeposits` are temporary disabled due to
unconsistent data with Daily Active Addresses from TimescaleDB.

The fields will be bringed back after DAA is refactored to use
Clickhouse.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/share_of_deposits/eth_share_of_deposits.ex
+++ b/lib/sanbase/clickhouse/share_of_deposits/eth_share_of_deposits.ex
@@ -1,0 +1,117 @@
+defmodule Sanbase.Clickhouse.EthShareOfDeposits do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate share of deposits from daily active addresses for ETH
+  """
+
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  import Sanbase.Math, only: [to_integer: 1]
+
+  alias Sanbase.DateTimeUtils
+
+  @type share_of_deposits :: %{
+          datetime: %DateTime{},
+          active_addresses: non_neg_integer(),
+          active_deposits: non_neg_integer(),
+          share_of_deposits: number()
+        }
+
+  @spec share_of_deposits(
+          %DateTime{},
+          %DateTime{},
+          String.t()
+        ) :: {:ok, list(share_of_deposits)} | {:error, String.t()}
+  def share_of_deposits(from, to, interval) do
+    {query, args} = share_of_deposits_query(from, to, interval)
+
+    ClickhouseRepo.query_transform(
+      query,
+      args,
+      fn [dt, active_addresses, active_deposits, share_of_deposits] ->
+        %{
+          datetime: DateTime.from_unix!(dt),
+          active_addresses: active_addresses |> to_integer(),
+          active_deposits: active_deposits |> to_integer(),
+          share_of_deposits: share_of_deposits
+        }
+      end
+    )
+  end
+
+  defp share_of_deposits_query(from, to, interval) do
+    interval = DateTimeUtils.compound_duration_to_seconds(interval)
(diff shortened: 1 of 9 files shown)


lib/sanbase/clickhouse/share_of_deposits/eth_share_of_deposits.ex
```diff
@@ -0,0 +1,118 @@
...
+
+  @spec share_of_deposits(
+          %DateTime{},
+          %DateTime{},
+          String.t()
+        ) :: {:ok, list(share_of_deposits)} | {:error, String.t()}
+  def share_of_deposits(from, to, interval) do
+    {query, args} = share_of_deposits_query(from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [
```
IvanIvanoff: can you put each arg on a new line manually and run the formatter after that. The weird function format will be fixed then

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -197,4 +198,36 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+             interval,
+             @one_hour_seconds,
+             50
+           ),
+         {:ok, share_of_deposits} <-
+           ShareOfDeposits.share_of_deposits(contract, from, to, interval) do
+      {:ok, share_of_deposits}
+    else
+      {:error, {:missing_contract, error_msg}} ->
+        {:error, error_msg}
```
IvanIvanoff: Should this return error or just `{:ok, []}` :?

test/sanbase/clickhouse/share_of_deposits/erc20_share_of_deposits_test.exs
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  test "returns share of deposits from daily active addresses",
+       context do
+    with_mock Sanbase.ClickhouseRepo,
+      query: fn _, _ ->
+        {:ok,
+         %{
+           rows: [
+             [from_iso8601_to_unix!("2019-01-01T00:00:00Z"), 200_000, 20_000, 10.0],
```
IvanIvanoff: `DateTime.to_unix!(context.from)`

test/sanbase_web/graphql/clickhouse/share_of_deposits_test.exs
```diff
@@ -0,0 +1,196 @@
...
+  import ExUnit.CaptureLog
+  import Sanbase.Factory
+
+  alias Sanbase.Clickhouse.ShareOfDeposits
+
+  setup do
+    user = insert(:staked_user)
+    conn = setup_jwt_auth(build_conn(), user)
+
+    token = insert(:project, %{main_contract_address: "0x123"})
```
IvanIvanoff: `project`

kalkov: I want to explicitly differentiate between token and ethereum.

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
kalkov: Well, as a user I would expect an error and we have same behavior in the other places in the codebase?

IvanIvanoff: My issue with this `:error` is that it won't be cached and all calls to this API will make DB calls. Also the the metrics will report this as error response while nothing wrong happened - we just don't have data for that project.

kalkov: Let's address in another PR

PR #1113, opened 2019-04-10
Do not throw a 500 error when fetching BTC top holders

valo:
The BTC top holders is still not implemented, but we should not be
throwing a 500 error when calling this API. A properly formatted error
should be returned.

The problem is that we don't fetch the contract info about BTC in the
canonical way.

### DIFF
--- a/test/sanbase/clickhouse/top_holders_test.exs
+++ b/test/sanbase/clickhouse/top_holders_test.exs
@@ -11,6 +11,8 @@ defmodule Sanbase.Clickhouse.TopHoldersTest do
 
     [
       slug: project.coinmarketcap_id,
+      contract: "ETH",
+      token_decimals: 18,
       number_of_holders: 10,
       from: from_iso8601!("2019-01-01T00:00:00Z"),
       to: from_iso8601!("2019-01-03T00:00:00Z")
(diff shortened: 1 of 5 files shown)


test/sanbase/clickhouse/top_holders_test.exs
```diff
@@ -11,6 +11,8 @@ defmodule Sanbase.Clickhouse.TopHoldersTest do
 
     [
       slug: project.coinmarketcap_id,
+      contract: "ETH",
```
kalkov: Let's not hardcode contract and decimals here, we can take them from the factory?
Even better would be to introduce new `project_ethrereum` factory like this: 
```elixir
  def project_ethereum_factory() do
    %Project{
      name: "Ethereum",
      ticker: "ETH",
      coinmarketcap_id: "ethereum",
      token_decimals: 18,
      total_supply: 105_597_654,
      github_link: "https://github.com/ethereum
    }
  end
```

kalkov: It can also be done in another PR, for example the one for adding consistency that will come soon.

valo: I think we will need to do a separate PR which introduces project factories. Seems like we currently use the pattern above in all the places.

PR #1116, opened 2019-04-10
Add interval calibration to the bitcoin metrics

IvanIvanoff:
#### Summary
Clean the rest of the calibrations
Make the calibrations more consistent in the number of data points returned

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/bitcoin/bitcoin.ex
+++ b/lib/sanbase/clickhouse/bitcoin/bitcoin.ex
@@ -50,31 +50,27 @@ defmodule Sanbase.Clickhouse.Bitcoin do
   def token_velocity(from, to, interval) do
     {query, args} = token_velocity_query(from, to, interval)
 
-    ClickhouseRepo.query_transform(
-      query,
-      args,
-      fn [dt, value] ->
-        %{
-          token_velocity: value,
-          datetime: DateTime.from_unix!(dt)
-        }
-      end
-    )
+    ClickhouseRepo.query_transform(query, args, fn [datetime, value] ->
+      %{
+        token_velocity: value,
+        datetime: DateTime.from_unix!(datetime)
+      }
+    end)
   end
 
   def mvrv_ratio(from, to, interval) do
     {query, args} = mvrv_ratio_query(from, to, interval)
 
-    ClickhouseRepo.query_transform(
-      query,
-      args,
-      fn [dt, value] ->
-        %{
-          ratio: value,
-          datetime: DateTime.from_unix!(dt)
-        }
-      end
-    )
+    ClickhouseRepo.query_transform(query, args, fn [datetime, value] ->
+      %{
+        ratio: value,
+        datetime: DateTime.from_unix!(datetime)
+      }
+    end)
+  end
+
+  def first_datetime(_) do
+    ~N[2009-01-09 00:00:00] |> DateTime.from_naive("Etc/UTC")
   end
 
   # Generic way to fetch a metric that is used by making an aggregation over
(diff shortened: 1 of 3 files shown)


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -93,7 +93,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   end
 
   def mvrv_ratio(_root, %{slug: "bitcoin", from: from, to: to, interval: interval}, _resolution) do
-    Bitcoin.mvrv_ratio(from, to, interval)
+    with {:ok, from, to, interval} <-
+           calibrate_interval(Bitcoin, "bitcoin", from, to, interval, 86400, 50) do
```
kalkov: Can you extract to constants `86_400` & `50`?

lib/sanbase/clickhouse/bitcoin/bitcoin.ex
```diff
@@ -158,4 +150,12 @@ defmodule Sanbase.Clickhouse.Bitcoin do
 
     {query, args}
   end
+
+  # Round the interval to whole days only
+  defp round_interval(interval) when is_binary(interval) do
+    Sanbase.DateTimeUtils.compound_duration_to_seconds(interval)
+    |> div(86_400)
```
kalkov: This can go into constant

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -20,38 +28,29 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
         %{slug: "bitcoin", from: from, to: to, interval: interval},
         _resolution
       ) do
-    Bitcoin.token_age_consumed(from, to, interval)
+    with {:ok, from, to, interval} <-
+           calibrate_interval(Bitcoin, "bitcoin", from, to, interval, 86400, 50) do
```
kalkov: Can be taken from constants

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
```diff
@@ -68,26 +67,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.EtherbiResolver do
...
-             Blockchain.TokenAgeConsumed,
-             contract_address,
+           calibrate_interval(
+             TokenAgeConsumed,
+             contract,
              from,
              to,
              interval,
-             60 * 60,
+             3600,
```
kalkov: We have `@one_hour_seconds 3600` in clickhouse_resolver, you can extract it here as well.

lib/sanbase/clickhouse/bitcoin/bitcoin.ex
IvanIvanoff: having a constant for as single value like here will reduce the readability imo

lib/sanbase_web/graphql/resolvers/etherbi_resolver.ex
IvanIvanoff: I don't see much logic in using it. I prefer to use constants when they can change like the number of datapoints returned. The seconds in an hour won't change

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
IvanIvanoff: They are not constants and can vary between queries

kalkov: It's not obvious what 86400 is. Also 50. One way to make it more explicit is by extracting into constant, the other - using named params. The first one is much simpler in the current case.

lib/sanbase/clickhouse/bitcoin/bitcoin.ex
kalkov:
```elixir
Sanbase.DateTimeUtils.compound_duration_to_seconds(interval)
    |> div(@one_day_in_seconds)
```
seems more readable to me

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
IvanIvanoff: If you think this 86400 is so not understandable let's have a card for replacing it as we have in 12-15 places and changing just one and ignoring the others won't do any good

PR #1121, opened 2019-04-11
Add an `all` source to the social dominance metric

valo:
The new source will fetch the social dominance across all sources and
average them.

### DIFF
--- a/lib/sanbase/social_data/social_dominance.ex
+++ b/lib/sanbase/social_data/social_dominance.ex
@@ -11,6 +11,42 @@ defmodule Sanbase.SocialData.SocialDominance do
 
   @recv_timeout 15_000
 
+  @sources Absinthe.Schema.lookup_type(SanbaseWeb.Graphql.Schema, :social_dominance_sources).values
+           |> Map.keys()
+           |> List.delete(:all)
+
+  def social_dominance(
+        slug,
+        datetime_from,
+        datetime_to,
+        interval,
+        :all
+      ) do
+    result =
+      @sources
+      |> Sanbase.Parallel.pmap(fn source ->
+        {:ok, result} = social_dominance(slug, datetime_from, datetime_to, interval, source)
+
+        result
+      end)
+      |> List.zip()
+      |> Enum.map(&Tuple.to_list/1)
+      |> Enum.map(fn all_sources_point ->
+        average_dominance =
+          all_sources_point
+          |> Enum.map(fn point -> point.dominance end)
+          |> Enum.sum()
+          |> Kernel./(length(@sources))
+
+        %{
+          datetime: all_sources_point |> List.first() |> Map.get(:datetime),
+          dominance: average_dominance
+        }
+      end)
+
+    {:ok, result}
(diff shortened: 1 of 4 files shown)


lib/sanbase/social_data/social_dominance.ex
```diff
@@ -11,6 +11,42 @@ defmodule Sanbase.SocialData.SocialDominance do
...
+        result
+      end)
+      |> List.zip()
+      |> Enum.map(&Tuple.to_list/1)
+      |> Enum.map(fn all_sources_point ->
+        average_dominance =
+          all_sources_point
+          |> Enum.map(fn point -> point.dominance end)
+          |> Enum.sum()
+          |> Kernel./(length(@sources))
```
IvanIvanoff: Shouldn't there be some weight depending on the volume?

lib/sanbase_web/graphql/schema/social_data_queries.ex
```diff
@@ -21,13 +21,14 @@ defmodule SanbaseWeb.Graphql.Schema.SocialDataQueries do
         2. "TELEGRAM" - shows the relative social dominance of this project in the telegram crypto channels
         3. "DISCORD" - shows the relative social dominance of this project on discord crypto communities
         4. "REDDIT" - shows the relative social dominance of this project on crypto subreddits
+        5. "ALL" - shows the average value of the social dominance across all sources
```
IvanIvanoff: nitpick: here in the docs if we quote the params they look like string but in reality they are used without quotes

lib/sanbase/social_data/social_dominance.ex
valo: Let's roll it out like this and do fixes as we find issues. As of right now we don't have an idea if volume needs to be taken into account or not.

test/sanbase/social_data/social_dominance_test.exs
```diff
@@ -58,6 +59,44 @@ defmodule Sanbase.SocialDominanceTest do
...
+    test "when computing for all sources" do
+      from = DateTime.from_naive!(~N[2018-04-16 10:00:00], "Etc/UTC")
+      to = DateTime.from_naive!(~N[2018-04-16 22:00:00], "Etc/UTC")
+
+      with_mock(HTTPoison, [],
+        get: fn _, _, _ ->
+          {:ok,
+           %HTTPoison.Response{
+             body:
+               "[{\"BTC_bitcoin\": 5, \"EOS_eos\": 15, \"ETH_ethereum\": 5, \"datetime\": 1523872800}, {\"BTC_bitcoin\": 15, \"EOS_eos\": 5, \"ETH_ethereum\": 10, \"datetime\": 1523916000}]",
```
kalkov: Will be more readable on multiple lines.

lib/sanbase_web/graphql/schema/social_data_queries.ex
valo: True that. I'll remove the quotes

PR #1124, opened 2019-04-12
Add published_at field to the insights

IvanIvanoff:
Populate the published_at for published insights in the following way:
- If the insight is featured, use the `created_at` datetime
- If the insight use the updated_at datetime

This is needed because when an insight is made featured it will have its
updated_at field updated and the frontend will use it to render the dot
on the price chart. This is not the correct behavior.

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/structure.sql
```diff
@@ -2424,4 +2424,5 @@ ALTER TABLE ONLY public.votes
...
...0211202224), (20180215105804), (20180216182032), (20180219102602), (20180219133328), (20180222135838), (20180223114151), (20180227090003), (20180319041803), (20180322143849), (20180323111505), (20180330045410), (20180411112814), (20180411112855), (20180411113727), (20180411120339), (20180412083038), (20180418141807), (20180423115739), (20180423130032), (20180424122421), (20180424135326), (20180425145127), (20180430093358), (20180503110930), (20180504071348), (20180526114244), (20180601085613), (20180620114029), (20180625122114), (20180628092208), (20180704075131), (20180704075135), (20180708110131), (20180708114337), (20180829153735), (20180830080945), (20180831074008), (20180831094245), (20180903115442), (20180912124703), (20180914114619), (20181002095110), (20181029104029), (20181102114904), (20181107134850), (20181129132524), (20181218142658), (20190110101520), (20190114144216), (20190116105831), (20190124134046), (20190128085100), (20190215131827), (20190225144858), (20190227153610), (20190227153724), (20190312142628), (20190327104558), (20190327105353), (20190328141739), (20190329101433), (20190404143453), (20190404144631), (20190405124751), (20190408081738), (20190412100349);
```
valo: I think there should be a migration which adds this field to the table.

IvanIvanoff: oops,

lib/sanbase_web/graphql/schema/insight_types.ex
```diff
@@ -43,6 +43,7 @@ defmodule SanbaseWeb.Graphql.InsightTypes do
     end
 
     field(:updated_at, non_null(:datetime))
+    field(:published_at, non_null(:datetime))
```
kalkov: Why non_null? What it never gets published?

IvanIvanoff: fixed

priv/repo/structure.sql
IvanIvanoff: Added all needed files

lib/sanbase_web/graphql/resolvers/post_resolver.ex
```diff
@@ -184,7 +184,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.PostResolver do
     case Repo.get(Post, post_id) do
       %Post{user_id: ^user_id} = post ->
         post
-        |> Post.publish_changeset(%{ready_state: Post.published()})
+        |> Post.publish_changeset(%{
+          ready_state: Post.published(),
+          published_at: NaiveDatetime.utc_now()
```
tspenov: typo: should be `NaiveDateTime`. One test is failing because of it.

PR #1129, opened 2019-04-12
Rework eth spent calculations to use eth_balances instead of eth_transfers

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance.ex
@@ -1,24 +1,23 @@
 defmodule Sanbase.Clickhouse.HistoricalBalance do
   alias Sanbase.Model.Project
+  alias Sanbase.Clickhouse.HistoricalBalance.{EthBalance, Erc20Balance}
+
+  def eth_balance_change(addresses, from, to) do
+    EthBalance.balance_change(addresses, from, to)
+  end
+
+  def eth_balance_change(addresses, from, to, interval) do
+    EthBalance.balance_change(addresses, from, to, interval)
+  end
 
   def balance_change(address, slug, from, to) do
     with {:ok, contract, token_decimals} <- Project.contract_info_by_slug(slug) do
       case contract do
         "ETH" ->
-          __MODULE__.EthBalance.balance_change(
-            address,
-            from,
-            to
-          )
+          EthBalance.balance_change(address, from, to)
 
         _ ->
-          __MODULE__.Erc20Balance.balance_change(
-            address,
-            contract,
-            token_decimals,
-            from,
-            to
-          )
+          Erc20Balance.balance_change(address, contract, token_decimals, from, to)
       end
     else
       {:error, error} -> {:error, inspect(error)}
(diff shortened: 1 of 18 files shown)


IvanIvanoff:
@valo @kalkov @tspenov 
One issue arises from this implementation:

With the old implementation we were looking only on the transfers and not the balance changes. So if a project has 3 wallets and it moves funds like this `A -> B -> C` and we're missing wallet `B` we were marking ETH as spent while it was actually not.

With the new implementation we're looking only on the balance changes (sum all balance changes for a set of wallets and if it's negative, then ETH is being spent). So in the same scenario the balance of `A` will be reduced by X and the balance of `C` will be increased by `C` so the ETH Spent is 0 even though we're missing wallet B.

This leads to the situation where the numbers reported by the old and the new implementation are visibly different. My theory is that the calculations with the balances are more believable but I want to hear your opinions, too.

lib/sanbase/clickhouse/historical_balance/eth_balance.ex
```diff
@@ -59,32 +79,115 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.EthBalance do
...
+        SELECT
+          toUnixTimestamp(intDiv(toUInt32(?4 + number * ?1), ?1) * ?1) AS time,
+          toFloat64(0) AS change
+        FROM numbers(?2)
+
+      UNION ALL
+
+      SELECT
+        toUnixTimestamp(intDiv(toUInt32(dt), ?1) * ?1) AS time,
+        sign*value as change
```
kalkov: AS

lib/sanbase/clickhouse/historical_balance.ex
```diff
@@ -1,53 +1,144 @@
...
+  @doc ~s"""
+  For a given address or list of addresses returns the ethereum balance change for the
+  from-to period. The returned lists indicates the address, before balance, after balance
+  and the balance change.
+
+  This is special case of balance_change/4 but as ethereum is used a lot for calculating
+  ethereum spent this case avoids a call to the database to obrain the contract
+  """
+  @spec eth_balance_change(address | list(address), DateTime.t(), DateTime.t()) ::
+          {:ok, list({address, {balance_before, balance_after, balance_change}})}
```
tspenov: I would prefer returning a map cause I think its easier to work in the future.

lib/sanbase/clickhouse/historical_balance.ex
```diff
@@ -1,53 +1,144 @@
...
+  def eth_balance_change(addresses, from, to) do
+    EthBalance.balance_change(addresses, from, to)
+  end
+
+  @doc ~s"""
+  For a given address or list of addresses returns the ethereum  balance change for each bucket
+  of size `interval` in the from-to time period
+  """
+  @spec eth_balance_change(address | list(address), DateTime.t(), DateTime.t(), interval) ::
+          {:ok, list({address, %{datetime: DateTime.t(), balance_change: number()}})}
```
tspenov: Also list of map here

lib/sanbase/clickhouse/historical_balance.ex
IvanIvanoff: I am using list of tuples mainly for the improved data locality and performance. I would do this only in such cases with 2-3 element tuples that would never change - in case we need to change it will be reworked to a map

lib/sanbase/clickhouse/historical_balance.ex
```diff
@@ -1,53 +1,144 @@
...
+  """
+  @type interval :: String.t()
+
+  @doc ~s"""
+  For a given address or list of addresses returns the ethereum balance change for the
+  from-to period. The returned lists indicates the address, before balance, after balance
+  and the balance change.
+
+  This is special case of balance_change/4 but as ethereum is used a lot for calculating
+  ethereum spent this case avoids a call to the database to obrain the contract
```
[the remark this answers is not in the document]
IvanIvanoff: You can use the `suggestion` block for typos

lib/sanbase/clickhouse/historical_balance.ex
tspenov: I am concerned that things like this` Enum.map(fn {_, {_, _, change}} -> change end)` won't be immediately obvious while this `Enum.map(fn %{balance_change: change} -> change end)` would be.

PR #1132, opened 2019-04-15
Cache the heavier queries for longer. 

IvanIvanoff:
#### Summary
Make the max_ttl_offset bigger so less queries expire at the same time
The bigger TTLs are for queries that return data that does not change that often:
- Github data (updated once an hour)
- Ethereum spent
- Social data stats

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/schema.ex
+++ b/lib/sanbase_web/graphql/schema.ex
@@ -29,7 +29,7 @@ defmodule SanbaseWeb.Graphql.Schema do
     UserFollowerResolver
   }
 
-  import SanbaseWeb.Graphql.Cache, only: [cache_resolve: 1]
+  import SanbaseWeb.Graphql.Cache, only: [cache_resolve: 1, cache_resolve: 2]
 
   alias SanbaseWeb.Graphql.Complexity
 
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/graphql/schema.ex
```diff
@@ -632,7 +632,7 @@ defmodule SanbaseWeb.Graphql.Schema do
 
       complexity(&Complexity.from_to_interval/3)
       middleware(ApiTimeframeRestriction)
-      cache_resolve(&TechIndicatorsResolver.topic_search/3)
+      cache_resolve(&TechIndicatorsResolver.topic_search/3, ttl: 500, max_ttl_offset: 240)
```
valo: Why 500? This is like 8.3 min which is not a round number.

IvanIvanoff: No real reason, it looks round in seconds :smiley:
I can change it to `600` which is 10 minutes 🤔

valo: I think 10 min makes sense. Let's do that.

IvanIvanoff: Done

PR #1133, opened 2019-04-15
Refactor DAA to get data from Clickhouse

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -59,6 +59,26 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     |> Map.new()
   end
 
+  def query(:average_daily_active_addresses, args) do
```
IvanIvanoff: This will never match as on line 11 there's a clause matching these params

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -129,7 +227,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
              from,
              to,
              interval,
-             3600,
+             @one_hour_in_seconds,
```
IvanIvanoff: I don't think this reads better, but that's my opinion

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
kalkov: Ooops introduces this from the rebase.

kalkov: Fixed

test/sanbase_web/graphql/clickhouse/daily_active_addresses_test.exs
```diff
@@ -0,0 +1,306 @@
...
+      bitcoin_slug: bitcoin.coinmarketcap_id,
+      from: from_iso8601!("2019-01-01T00:00:00Z"),
+      to: from_iso8601!("2019-01-03T00:00:00Z"),
+      interval: "1d"
+    ]
+  end
+
+  describe "for dailyActiveAddresses query" do
+    test "returns active addresses for erc20 tokens", context do
+      with_mock DailyActiveAddresses,
```
IvanIvanoff: You can mock the ClickhouseRepo as in the other tests

test/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses_test.exs
```diff
@@ -0,0 +1,88 @@
...
+      contract1: project1.main_contract_address,
+      contract2: project2.main_contract_address,
+      from: from_iso8601!("2019-01-01T00:00:00Z"),
+      to: from_iso8601!("2019-01-03T00:00:00Z")
+    ]
+  end
+
+  test "returns a single current value for active addresses per contract", context do
+    with_mock Sanbase.ClickhouseRepo,
+      query: fn _, _ ->
```
IvanIvanoff: How does this work?  If you don't pass as second argument to `with_mock` `[:passthrough]` you'll get 
```
** (UndefinedFunctionError) function Sanbase.ClickhouseRepo.query_transform/3 is undefined or private
```

IvanIvanoff: NVM, `query_transform` is a macro

test/sanbase_web/graphql/clickhouse/daily_active_addresses_test.exs
kalkov: There are tests that mock the database and assert the correct behavior of the modules [here](https://github.com/santiment/sanbase2/tree/refactor-daa-to-use-clickhouse/test/sanbase/clickhouse/daily_active_addresses).

A couple of benefits of mocking `DailyActiveAddresses` instead of the database in graphQL tests:

* slimmer SuT = simpler tests. We don't care about `DailyActiveAddresses` internals, it can be refactored and the test will still pass. We are testing here that the API works with different crypto projects, has default interval, actual calculation is not important.
* less test suit maintenance
* speed

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -5,7 +5,12 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   alias Sanbase.DateTimeUtils
   import SanbaseWeb.Graphql.Helpers.Utils, only: [calibrate_interval: 7]
 
+  alias SanbaseWeb.Graphql.SanbaseDataloader
+
+  import Absinthe.Resolution.Helpers
```
IvanIvanoff: You can specify `only: [on_load: 2]` because we have our own helpers module that imports the `async` function and could cause confusion in future if someone imports that module

test/sanbase_web/graphql/clickhouse/daily_active_addresses_test.exs
IvanIvanoff: Agree on all 3 points, mostly on the first one :smiley:

test/sanbase/clickhouse/daily_active_addresses/erc20_daily_active_addresses_test.exs
IvanIvanoff: As discussed in Discord, this behavior is kind of misleading because it only works because `query_transform/{2,3}` is a macro. More explicit behavior will be to pass `[:passthrough]` as a second argument to `with_mock` so it will l continue to work after `query_transform/{2,3}` is refactored to be a function instead of macro.

test/sanbase/clickhouse/daily_active_addresses/eth_daily_active_addresses_test.exs
```diff
@@ -0,0 +1,71 @@
...
+        query: fn _, _ -> {:ok, %{rows: [[100_000]]}} end do
+        result = Eth.average_active_addresses(context.from, context.to)
+
+        assert result == {:ok, 100_000}
+      end
+    end
+
+    test "returns 0 when the database returns nil", context do
+      with_mock Sanbase.ClickhouseRepo,
+        query: fn _, _ -> {:ok, %{rows: [[nil]]}} end do
```
IvanIvanoff: Is this behavior possible?

kalkov: Probably: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/daily_active_addresses/eth_daily_active_addresses.ex#L77

lib/sanbase_web/graphql/schema.ex
```diff
@@ -429,7 +429,7 @@ defmodule SanbaseWeb.Graphql.Schema do
 
       complexity(&Complexity.from_to_interval/3)
       middleware(TimeframeRestriction, %{allow_historical_data: true})
-      cache_resolve(&EtherbiResolver.daily_active_addresses/3)
+      cache_resolve(&ClickhouseResolver.daily_active_addresses/3)
```
IvanIvanoff: For a future change: IMO we should have resolver names that are not linked to the implementation but to the data itself. In this case all blockchain data should be resolved by `BlockchainResolver` which will then know if the data is in Clickhouse, timescaledb or in any other place.

lib/mix/seeds/daily_active_addresses_seed.ex
```diff
@@ -1,38 +0,0 @@
...
-    alias Sanbase.Blockchain.DailyActiveAddresses
-
-    %DailyActiveAddresses{}
-    |> DailyActiveAddresses.changeset(%{
-      contract_address: contract,
-      timestamp: timestamp,
-      active_addresses: active_addresses
-    })
-  end
-end
```
IvanIvanoff: Let's remember that we have these seeds which could be easily reused once we start spinning CH Server for our tests

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -116,6 +122,97 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+    with {:ok, contract_address, _token_decimals} <- Project.contract_info(project) do
+      average_daily_active_addresses =
+        loader
+        |> Dataloader.get(
+          SanbaseDataloader,
+          :average_daily_active_addresses,
+          contract_address
+        )
+
+      {:ok, average_daily_active_addresses || 0}
```
IvanIvanoff: nitpick: Better move `|| 0` in the first binding of `average_daily_active_addresses`. It will be easy to introduce a bug if some code needs to be added in the future between the binding and here

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -116,6 +122,97 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+    else
+      {:error, {:missing_contract, _}} ->
+        {:ok, 0}
+
+      {:error, error} ->
+        error_msg = "Can't fetch average daily active addresses for #{Project.describe(project)}"
+        Logger.warn(error_msg <> "Reason: #{inspect(error)}")
```
IvanIvanoff: An idea on how we can improve the logging/errors here: We could really use some ideas from the `ErrorHandling` module here. Mainly, to include an UUID in the logs and send this UUID as `details` field in the graphql response. This way it can be really easy to find the particular error in the logs

test/sanbase_web/graphql/clickhouse/daily_active_addresses_test.exs
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  setup do
+    token = insert(:project, %{main_contract_address: "0x123"})
+
+    ethereum =
+      insert(:project, %{
+        coinmarketcap_id: "ethereum",
+        ticker: "ETH",
+        main_contract_address: "0x456"
```
IvanIvanoff: Adding contract to ETH is misleading. `contract_info/1` and `contract_info_by_slug/1` are specially handling ethereum  and bitcoin. The most correct behavior currently is to either not provide this field at all, or pass `nil`

IvanIvanoff:
@kalkov @MilanRusev One more thing - this currently uses `eth_daily_active_addresses`, `erc20_daily_active_addresses` and `btc_daily_metrics`. Are these the latest tables for this data?

MilanRusev:
@kalkov @IvanIvanoff about `eth_daily_active_addresses`, `erc20_daily_active_addresses` - they are populated with the latest data at the moment. The `daily_metrics` table should have the latest data as well.

However we will be moving towards the new daily metrics implementation that Tzanko is making at the moment but it is still a work in progress.

IvanIvanoff:
@kalkov let's test on stage

PR #1134, opened 2019-04-15
Dispatch to the correct daily active addresses function in the signals

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/daily_active_addresses.ex
+++ b/lib/sanbase/clickhouse/daily_active_addresses.ex
@@ -21,10 +21,6 @@ defmodule Sanbase.Clickhouse.DailyActiveAddresses do
     Erc20.realtime_active_addresses(contract)
   end
 
-  def average_active_addresses(eth, from, to) when is_binary(eth) and eth in @ethereum do
-    Eth.average_active_addresses(from, to)
-  end
-
   def average_active_addresses(eth, from, to, interval)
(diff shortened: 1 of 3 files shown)


lib/sanbase/signals/trigger/settings/daily_active_addresses_settings.ex
```diff
@@ -79,7 +79,7 @@ defmodule Sanbase.Signals.Trigger.DailyActiveAddressesSettings do
 
   defp average_active_addresses(contract, from, to) do
     Cache.get_or_store("daa_#{contract}_current", fn ->
-      case DailyActiveAddresses.average_active_addresses(contract, from, to) do
+      case DailyActiveAddresses.average_active_addresses([contract], from, to) do
```
kalkov: It was broken? Can we have a test case?

IvanIvanoff: We are mocking the DB in the case Clickhouse - we don't have real tests here :(
We can rewrite all tests to mock the DB calls and not the module itself, but that's a pretty big task in itself and should be done separately

kalkov: Could be a simple test that asserts when you call settings method `DailyActiveAddresses.average_active_addresses` will be executed with the right arguments.

PR #1136, opened 2019-04-16
Order insights by published at

tspenov:

### DIFF
--- a/test/sanbase_web/graphql/post_test.exs
+++ b/test/sanbase_web/graphql/post_test.exs
@@ -274,14 +274,14 @@ defmodule SanbaseWeb.Graphql.PostTest do
     assert result |> hd() |> Map.get("text") == post.text
   end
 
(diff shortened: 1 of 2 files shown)


test/sanbase_web/graphql/post_test.exs
```diff
@@ -274,14 +274,14 @@ defmodule SanbaseWeb.Graphql.PostTest do
     assert result |> hd() |> Map.get("text") == post.text
   end
 
-  test "getting all posts ranked", %{user: user, user2: user2, conn: conn, poll: poll} do
+  test "getting all insights ordered", %{user: user, user2: user2, conn: conn, poll: poll} do
```
kalkov: A separate test just for asserting ordering will guarantee us that will always work as expected. 
Example: Create 3 posts with custom `created_at` values, set `published_at` and assert the ordering. 
Could be in another PR if this one is urgent.

tspenov: sounds good - will add one

PR #1138, opened 2019-04-16
Add checks for valid from-to time range in the timeframe restriction middleware

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


kalkov:
Can you add a test?

IvanIvanoff:
Before this change a request that had `from` and `to` both outside the allowed timeframe was allowed (even though after the middleware passes `to` was a datetime that is before `from`) was executed.

With this change the request is rejected right away. @Partysun @DmitriVanGuard can this change break something?

PR #1139, opened 2019-04-16
Hide not approved insights

tspenov:


lib/sanbase/insight/post.ex
```diff
@@ -88,89 +88,51 @@ defmodule Sanbase.Insight.Post do
...
 
   @doc """
-    Returns published or current user's posts ranked by the ranking algorithm
+  All published and approved insights paginated
   """
-  @spec ranked_published_or_own_posts(integer) :: [%Post{}]
-  def ranked_published_or_own_posts(user_id) do
-    posts_by_score()
-    |> get_only_published_or_own_posts(user_id)
+  def published_and_approved_insights(page, page_size) do
```
IvanIvanoff: I think we can call these `public` 🤔 
published and approved exposes too much implementation details for a public function

tspenov: yes, good suggestion and its lot shorter :)

lib/sanbase_web/graphql/resolvers/post_resolver.ex
```diff
@@ -33,7 +33,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.PostResolver do
 
   def all_insights(_root, %{page: page, page_size: page_size}, _resolution) do
     posts =
-      Post.published_posts(page, page_size)
+      Post.public_insights(page, page_size)
       |> Repo.preload(@preloaded_assoc)
```
IvanIvanoff: can we move this preloads to the `Post` module? Ideally the resolvers won't ever call `Repo`

PR #1143, opened 2019-04-18
 Use TOTAL_ERC20 in historyPrice returning combined mcap and volume for all ERC20 projects

tspenov:
**TODO** - add tests 

```graphql
{
  historyPrice(
    from: "2019-04-10T06:26:56Z", 
    to: "2019-04-17T06:26:56Z", 
    slug: "TOTAL_ERC20",
    interval:"1d"
) {
    datetime
   	marketcap,
    volume
  }
}
```
```
{
  "data": {
    "historyPrice": [
      {
        "datetime": "2019-04-10T00:00:00Z",
        "marketcap": 8836311527.056316,
        "volume": 973921469.2144265
      },
      {
        "datetime": "2019-04-11T00:00:00Z",
        "marketcap": 8164143429.799532,
        "volume": 1060768530.8284835
      },
      {
        "datetime": "2019-04-12T00:00:00Z",
        "marketcap": 8107297450.22629,
        "volume": 898096518.7432971
      },
      {
        "datetime": "2019-04-13T00:00:00Z",
        "marketcap": 8283760089.58144,
        "volume": 800561516.5699275
      },
      {
        "datetime": "2019-04-14T00:00:00Z",
        "marketcap": 8414500342.378368,
        "volume": 804779334.8218791
      },
      {
        "datetime": "2019-04-15T00:00:00Z",
        "marketcap": 8492956066.920276,
        "volume": 876235014.353876
      },
      {
        "datetime": "2019-04-16T00:00:00Z",
        "marketcap": 8399515706.078108,
        "volume": 799923682.2872059
      },
      {
        "datetime": "2019-04-17T00:00:00Z",
        "marketcap": 8541575631.927128,
        "volume": 763322517.3871572
      }
    ]
  }
}
```

### DIFF
--- a/lib/sanbase/prices/store.ex
+++ b/lib/sanbase/prices/store.ex
@@ -51,6 +51,38 @@ defmodule Sanbase.Prices.Store do
     end
   end
 
+  def fetch_prices_with_resolution("TOTAL_ERC20", from, to, resolution) do
+    measurements =
+      Sanbase.Model.Project.List.erc20_projects()
+      |> Enum.map(&Sanbase.Influxdb.Measurement.name_from/1)
+
+    total_erc20 =
+      measurements
+      |> Enum.chunk_every(100)
+      |> Sanbase.Parallel.pmap(fn measurements ->
+        {:ok, result} =
+          fetch_combined_mcap_volume(
+            measurements,
+            from,
+            to,
+            resolution
+          )
+
+        result
+      end)
+      |> Enum.zip()
+      |> Enum.map(&Tuple.to_list/1)
+      |> Enum.map(fn [%{datetime: dt} | _rest] = l ->
+        %{
+          datetime: dt,
+          volume: Enum.reduce(l, 0, fn %{volume: v}, acc -> acc + v end),
+          marketcap: Enum.reduce(l, 0, fn %{marketcap: m}, acc -> acc + m end)
+        }
+      end)
+
+    {:ok, total_erc20}
+  end
+
   def fetch_prices_with_resolution(measurement, from, to, resolution) do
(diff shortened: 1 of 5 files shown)


lib/sanbase/prices/store.ex
```diff
@@ -51,6 +51,38 @@ defmodule Sanbase.Prices.Store do
...
   end
 
+  def fetch_prices_with_resolution("TOTAL_ERC20", from, to, resolution) do
+    measurements =
+      Sanbase.Model.Project.List.erc20_projects()
+      |> Enum.map(&Sanbase.Influxdb.Measurement.name_from/1)
+
+    total_erc20 =
+      measurements
+      |> Enum.chunk_every(20)
```
IvanIvanoff: in the influxdb dataloader we chunk every 100 and it works good

kalkov:
Seems hacky to me to pass fake slug and would expect from this query to return price records per ERC20 project. Also the returned fields are different from the original `historyPrice` behavior. 
Isn't it better to have a separate query like : `totalErc20Data`, we can find better word than "data"?

IvanIvanoff:
@kalkov we already have `TOTAL_MARKET` slug used. like this. This is not the best behavior, but is consistent with what we have

lib/sanbase/prices/store.ex
tspenov: yeah should be 100. Used smaller chunks cause I had some problems locally, which ended up not related with the chunk size.

tspenov:
@kalkov What you're suggesting makes sense but also the frontend wants more unified interface. I think we may use this fake slug on more places like watchlists. Also as @IvanIvanoff pointed its more or less consistent with what we have.

PR #1148, opened 2019-04-19
Dynamic watchlist

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


tspenov:
@IvanIvanoff can we use the existing watchlist and just add support for `field(:function, WatchlistFunction, default: %{})` which is optional? Now if we want to support something for the watchlist, also have to support it for dynamic watchlist - example featured items, but also signals, events ...

IvanIvanoff:
@tspenov We could reuse the database, though I still prefer to introduce a new API as the old one as it has some shortcomings:
- We want to migrate the naming convention from `user_list` to `watchlist`. `user_list` sounds like a list of user and it's actually a list of projects.
- The update API won't work good for big lists - if you want to remove 1 items from a a list of 100 projects, you need to send the remaining 99 projects and they'll overwrite the existing ones.
- some other things also

tspenov:
@IvanIvanoff These 2 won't be fixed here either. We could reuse not only the db but also the schema/context module, even the API cause there is only one difference between the 2 - one has `list of projects`, the other has a map field which is converted to a list of projects.

IvanIvanoff:
@tspenov I am doing a major rework - let's discuss after it's pushed

assets/webpack.config.js
```diff
@@ -1,10 +1,10 @@
-const path = require('path');
-const glob = require('glob');
-const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
-const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
-const webpack = require('webpack');
+const path = require("path");
```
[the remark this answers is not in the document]
IvanIvanoff: I've just rebased to master

lib/sanbase/blockchain/timescaledb.ex
```diff
@@ -127,10 +127,10 @@ defmodule Sanbase.Timescaledb do
...
    ## Examples
 
-      iex> Sanbase.Timescaledb.timestamp_to_datetime({{2018, 8, 7}, {12, 55, 5, 00}})
-      #DateTime<2018-08-07 12:55:05Z>
-      iex> Sanbase.Timescaledb.timestamp_to_datetime({{2015, 1, 17}, {12, 55, 37, 00005}})
-      #DateTime<2015-01-17 12:55:37Z>
+      iex> Sanbase.Timescaledb.timestamp_to_datetime({{2018, 8, 7}, {12, 55, 5, 00}}) |> DateTime.to_iso8601()
+      "2018-08-07T12:55:05Z"
+      iex> Sanbase.Timescaledb.timestamp_to_datetime({{2015, 1, 17}, {12, 55, 37, 00005}}) |> DateTime.to_iso8601()
+      "2015-01-17T12:55:37Z"
```
IvanIvanoff: This is because in Elixir 1.9 there's a new `~U` sigil and the date times will no longer be displayed as `#DateTime...` but `~U...`. Transforming the date time to string representation will work in both versions without issues.

lib/sanbase/user_lists/watchlist_function.ex
```diff
@@ -0,0 +1,81 @@
...
+  use Vex.Struct
+
+  defstruct name: "empty", args: []
+
+  alias Sanbase.Model.Project
+
+  @behaviour Ecto.Type
+  def type, do: :map
+
+  def evaluate(%__MODULE__{name: "market_segment", args: args}) do
```
tspenov: As a future enhancement we can add a category foreign key and group certain projects based on that. That way we can achieve a dynamic watchlist for `US based projects` for ex. and many more that can't be achieved with `market_segment`. But we can do that later based on @santiment/frontend team feedback and needs.

lib/sanbase/user_lists/user_list.ex
```diff
@@ -25,107 +27,95 @@ defmodule Sanbase.UserList do
...
 
   def fetch_all_public_lists() do
     query =
       from(
-        ul in UserList,
+        ul in __MODULE__,
         where: ul.is_public == true
       )
 
-    {:ok, Repo.all(query) |> Repo.preload(:list_items)}
```
tspenov: Removed preloads are handled by this one `field(:list_items, list_of(:list_item), resolve: &UserListResolver.list_items/3)` ?

test/sanbase_web/graphql/user/user_list_test.exs
```diff
@@ -116,15 +117,15 @@ defmodule SanbaseWeb.Graphql.UserListTest do
   end
 
   test "update user list - remove list items", %{user: user, conn: conn} do
-    {:ok, created_user_list} = UserList.create_user_list(user, %{name: "My Test List"})
+    {:ok, created_UserList} = UserList.create_user_list(user, %{name: "My Test List"})
```
tspenov: Some search&replace leftover. Also more down.

lib/sanbase/user_lists/user_list.ex
IvanIvanoff: Yes. Just preloading the list items is not enough because the function must be evaluated, so now it's done by a custom resolver that calls `get_projects/1` which combines both

PR #1162, opened 2019-05-10
Handle invalid authorization headers

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/context_plug.ex
+++ b/lib/sanbase_web/graphql/context_plug.ex
@@ -42,6 +42,11 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
       :skip ->
         build_context(conn, rest)
 
+      {:error, error} ->
+        conn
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -60,32 +66,37 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
         }
       }
     else
-      _ -> :skip
+      {:has_header?, _} -> :skip
```
tspenov: maybe `:halt` instead of skip ?

IvanIvanoff: When there's no header we need to skip and try the next. Halting is when there is a header but it's wrong

IvanIvanoff: We call it `:try_next`

PR #1168, opened 2019-05-14
Return json for API key errors

kalkov:
#### Summary
Return authentication errors in JSON: 
```json
{"errors": {"detail": "Error message"}} 
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/context_plug.ex
+++ b/lib/sanbase_web/graphql/context_plug.ex
@@ -37,6 +37,10 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
     |> put_private(:absinthe, %{context: context})
   end
 
+  defp build_error_msg(msg) do
+    %{errors: %{details: msg}} |> Jason.encode!()
+  end
+
   defp build_context(conn, [auth_method | rest]) do
     auth_method.(conn)
     |> case do
@@ -44,12 +48,12 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
         build_context(conn, rest)
 
       {:error, error} ->
-        warn_msg = "Bad authorization header: #{error}"
-        Logger.warn(warn_msg)
+        msg = "Bad authorization header: #{error}"
+        Logger.warn(msg)
 
         conn =
           conn
-          |> send_resp(400, warn_msg)
+          |> send_resp(400, build_error_msg(msg))
           |> halt()
 
         {conn, %{}}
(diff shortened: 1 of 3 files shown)


lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -37,19 +37,25 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
     |> put_private(:absinthe, %{context: context})
   end
 
+  defp build_error_msg(msg) do
+    %{errors: %{detail: msg}} |> Jason.encode!()
```
IvanIvanoff:
```suggestion
    %{errors: %{details: msg}} |> Jason.encode!()
```

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -37,19 +37,25 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
...
-        Logger.warn(warn_msg)
+        msg = "Bad authorization header: #{error}"
+        Logger.warn(msg)
+
+        response_msg = build_error_msg(msg)
 
         conn =
           conn
-          |> send_resp(400, warn_msg)
+          |> send_resp(400, response_msg)
```
IvanIvanoff: you can just put `|> send_resp(400, build_error_msg(msg))` ehere

lib/sanbase_web/graphql/context_plug.ex
IvanIvanoff: Also you need to change the tests for this. All other places use plural for details

kalkov: It's "detail" for 500: https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/views/error_view.ex

kalkov:
```json
{"errors":{"detail":"Internal server error"}}
```

IvanIvanoff: I don't think the code you linked is used anywhere

IvanIvanoff: The "details" we see in errors are of the format described here: https://github.com/absinthe-graphql/absinthe/blob/master/guides/errors.md#basic-errors

tspenov: BTW isn't it better to return the same structure as absinthe but without locations:

```
{
  "errors": [ {"message": "message 1"}, ...]
}
```
@kalkov @IvanIvanoff

IvanIvanoff: @kalkov You were right - when the Plug parser breaks (invalid json) then the ErrorView handles the response, so this `detail` you linked is the one you see

PR #1180, opened 2019-05-20
Better error messages

kalkov:
#### Summary
More consistent error messages for Clickhouse and Etherbi resolvers

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/tech_indicators/price_volume_difference.ex
+++ b/lib/sanbase/tech_indicators/price_volume_difference.ex
@@ -89,13 +89,13 @@ defmodule Sanbase.TechIndicators.PriceVolumeDifference do
 
   defp handle_result({:ok, %HTTPoison.Response{status_code: status, body: body}}, project) do
     warn_result(
-      "Error status #{status} fetching price-volume diff for #{Project.describe(project)}: #{body}"
+      "Error status #{status} fetching price-volume diff for #{Project.describe(project)} - #{body}"
     )
   end
 
   defp handle_result({:error, %HTTPoison.Error{} = error}, project) do
     error_result(
-      "Cannot fetch price-volume diff data for #{Project.describe(project)}: #{
+      "Cannot fetch price-volume diff data for #{Project.describe(project)} - #{
         HTTPoison.Error.message(error)
       }"
     )
--- a/lib/sanbase/utils/error_handling.ex
+++ b/lib/sanbase/utils/error_handling.ex
@@ -12,4 +12,16 @@ defmodule Sanbase.Utils.ErrorHandling do
     Logger.warn("[#{log_id}] #{message}")
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
+
+  def graphql_error_msg(metric_name) do
+    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name}"
+  end
+
+  def graphql_error_msg(metric_name, slug) do
+    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name} for project with slug: #{slug}"
+  end
+
+  def log_graphql_error(message, error) do
+    Logger.warn("#{message}" <> ", Reason: #{inspect(error)}")
+  end
 end
--- a/test/sanbase_web/graphql/clickhouse/mining_pools_distribution_test.exs
+++ b/test/sanbase_web/graphql/clickhouse/mining_pools_distribution_test.exs
@@ -23,14 +23,16 @@ defmodule SanbaseWeb.Graphql.Clickhouse.MiningPoolsDistributionTest do
   end
 
   test "works only for ethereum", context do
+    error = "Currently only ethereum is supported!"
+
     with_mock MiningPoolsDistribution,
-      distribution: fn _, _, _, _ -> {:error, "Currently only ethereum is supported!"} end do
+      distribution: fn _, _, _, _ -> {:error, error} end do
       assert capture_log(fn ->
                response = execute_query("unsupported", context)
                result = parse_response(response)
                assert result == nil
              end) =~
-               ~s/[warn] Can't calculate mining pools distribution. Reason: "Currently only ethereum is supported!"/
+              graphql_error_msg("Mining Pools Distribution", error)
     end
   end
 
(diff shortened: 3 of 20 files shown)


lib/sanbase/tech_indicators/price_volume_difference.ex
```diff
@@ -89,13 +89,13 @@ defmodule Sanbase.TechIndicators.PriceVolumeDifference do
 
   defp handle_result({:ok, %HTTPoison.Response{status_code: status, body: body}}, project) do
     warn_result(
-      "Error status #{status} fetching price-volume diff for #{Project.describe(project)}: #{body}"
+      "Error status #{status} fetching price-volume diff for #{Project.describe(project)} #{body}"
```
IvanIvanoff: could probably use a dash here?
`for project with id 5 - <the body goes here>`

lib/sanbase/utils/error_handling.ex
```diff
@@ -12,4 +12,17 @@ defmodule Sanbase.Utils.ErrorHandling do
     Logger.warn("[#{log_id}] #{message}")
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
+
+  def generate_graphql_error_msg(metric_name) do
```
IvanIvanoff: I would omit the `generate_` part as it's clear enough

lib/sanbase/utils/error_handling.ex
```diff
@@ -12,4 +12,17 @@ defmodule Sanbase.Utils.ErrorHandling do
...
+    "Can't calculate #{metric_name}."
+  end
+
+  def generate_graphql_error_msg(metric_name, slug) do
+    "Can't calculate #{metric_name} for project with slug: #{slug}."
+  end
+
+  def log_graphql_error(message, error) do
+    log_id = Ecto.UUID.generate()
+    Logger.warn("[#{log_id}] #{message}" <> " Reason: #{inspect(error)}")
```
IvanIvanoff: If we only generate this UUID while logging and don't send it to the frontend which can include it in the sentry exception there's no point in it.

lib/sanbase/utils/error_handling.ex
```diff
@@ -12,4 +12,17 @@ defmodule Sanbase.Utils.ErrorHandling do
     Logger.warn("[#{log_id}] #{message}")
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
+
+  def generate_graphql_error_msg(metric_name) do
+    "Can't calculate #{metric_name}."
+  end
+
+  def generate_graphql_error_msg(metric_name, slug) do
+    "Can't calculate #{metric_name} for project with slug: #{slug}."
```
IvanIvanoff: I think we better use `fetch`/`get` here. sanbase is not the app that calculates the metrics

lib/sanbase/tech_indicators/price_volume_difference.ex
kalkov: I prefer : "Error status 500 fetching price-volume diff for project with slug: santiment. Internal Server Error"
over 
"Error status 500 fetching price-volume diff for project with slug: santiment. - Internal Server Error"

IvanIvanoff: `describe` does not add `.` at the end. I agree with adding `.` as currently it will be displayed like this: `Error status 500 fetching price-volume diff for project with slug: santiment Internal Server Error` (on a single line)

kalkov: I have changed the `describe` to include `.`

IvanIvanoff: But enforcing the period in `describe` will break messages like:
`"Cannot calculate ETH spent for #{Project.describe(project)}.` (already has period. this is default in most places)
and
`"Error status #{status} fetching price-volume diff for #{Project.describe(project)}: #{body}"` (already has another punctuation)
and places where describe appears in the middle `project #{Project.describe(slug)} has ... `

lib/sanbase/utils/error_handling.ex
kalkov: Chose `fetch`

lib/sanbase/tech_indicators/price_volume_difference.ex
kalkov: Fixed the way you proposed.

lib/sanbase/tech_indicators/price_volume_difference.ex
```diff
@@ -89,13 +89,13 @@ defmodule Sanbase.TechIndicators.PriceVolumeDifference do
...
     warn_result(
-      "Error status #{status} fetching price-volume diff for #{Project.describe(project)}: #{body}"
+      "Error status #{status} fetching price-volume diff for #{Project.describe(project)} - #{body}"
     )
   end
 
   defp handle_result({:error, %HTTPoison.Error{} = error}, project) do
     error_result(
-      "Cannot fetch price-volume diff data for #{Project.describe(project)}: #{
+      "Cannot fetch price-volume diff data for #{Project.describe(project)} #{
```
IvanIvanoff:
```suggestion
      "Cannot fetch price-volume diff data for #{Project.describe(project)} - #{
```

lib/sanbase/utils/error_handling.ex
```diff
@@ -12,4 +12,16 @@ defmodule Sanbase.Utils.ErrorHandling do
     Logger.warn("[#{log_id}] #{message}")
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
+
+  def graphql_error_msg(metric_name) do
+    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name}."
```
IvanIvanoff: You ca remove the `.` from these functions, too

test/sanbase_web/graphql/clickhouse/mining_pools_distribution_test.exs
```diff
@@ -101,14 +103,16 @@ defmodule SanbaseWeb.Graphql.Clickhouse.MiningPoolsDistributionTest do
   end
 
   test "logs warning when calculation errors", context do
+    error = "Some error description here"
+
     with_mock MiningPoolsDistribution,
-      distribution: fn _, _, _, _ -> {:error, "Some error description here"} end do
+      distribution: fn _, _, _, _s -> {:error, error} end do
```
IvanIvanoff: 🤔

kalkov: Typo, will fix

PR #1181, opened 2019-05-20
Add slug to Mining Pools Distribution metric

kalkov:
#### Summary
It was a mistake to omit the slug in this query, so introducing it here. 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/mining_pools_distribution.ex
+++ b/lib/sanbase/clickhouse/mining_pools_distribution.ex
@@ -15,11 +15,12 @@ defmodule Sanbase.Clickhouse.MiningPoolsDistribution do
         }
 
   @spec distribution(
+          String.t(),
           DateTime.t(),
           DateTime.t(),
           String.t()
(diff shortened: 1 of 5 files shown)


lib/sanbase/clickhouse/mining_pools_distribution.ex
```diff
@@ -15,11 +15,22 @@ defmodule Sanbase.Clickhouse.MiningPoolsDistribution do
...
-  def distribution(from, to, interval) do
+  def distribution(slug, from, to, interval) do
+    case slug do
+      "ethereum" ->
+        distribution_for_ethereum(from, to, interval)
+
+      _ ->
+        {:error, "Currently only ethereum is supported!"}
+    end
+  end
```
IvanIvanoff: I would do it like:
```
def distribution("ethereum", from, to, interval), do: distribution_for_ethereum(from, to, interval)
def distribution(_,_,_,_), do: {:error, "Currently only ethereum is supported!"}
```

PR #1191, opened 2019-05-22
Export APII Call data to Kafka

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

#### Additional deploy notes
Adding an init container for creating a Kafka topic https://github.com/santiment/devops/pull/863

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -63,5 +80,76 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
...
+
+  defp extract_caller_data(%{
+         auth: %{auth_method: :basic, san_balance: san_balance}
+       }) do
+    {nil, san_balance, :basic, nil}
+  end
+
+  defp extract_caller_data(_), do: {nil, nil, nil, nil}
+
+  defp user_agent(%{req_headers: headers}) do
```
tspenov: There is a helper `Plug.Conn.get_req_header(conn, "user-agent")` that will simplify this one.

lib/sanbase_web/endpoint_error_handler.ex
```diff
@@ -0,0 +1,69 @@
...
+            :erlang.raise(kind, reason, System.stacktrace())
+        end
+      end
+
+      defp export_api_call_data(conn, kind, reason) do
+        remote_ip = conn.remote_ip |> :inet_parse.ntoa() |> to_string
+        status_code = Map.get(reason, :status_code) || Map.get(reason, :plug_status, 500)
+
+        user_agent =
+          Enum.find(conn.req_headers, &match?({"user-agent", _}, &1))
```
tspenov: You can use - `Plug.Conn.get_req_header(conn, "user-agent")`

config/dev.exs
```diff
@@ -11,7 +11,7 @@ config :sanbase, Sanbase, url: {:system, "SANBASE_URL", "https://sanbase-low-sta
 config :sanbase, SanbaseWeb.Endpoint,
   http: [port: 4000],
   url: [host: "0.0.0.0"],
-  debug_errors: true,
+  debug_errors: false,
```
tspenov: why is that ? It is false by default, though I don't see a reason to change it in dev.exs

IvanIvanoff: Will turn back to `true`. It was `false` so I can test the behavior on prod

lib/sanbase/api_call_data/in_memory_kafka_producer.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Sanbase.InMemoryKafka.Producer do
+  @moduledoc ~s"""
+  In-memory implementation of the SanExporterEx.ProduxerBehaviour used in dev
```
tspenov:
```suggestion
  In-memory implementation of the SanExporterEx.ProducerBehaviour used in dev
```

lib/sanbase/api_call_data/in_memory_kafka_producer.ex
```diff
@@ -0,0 +1,58 @@
...
+    {:ok, _} = Task.start(fn -> send_data(producer, topic, data) end)
+    :ok
+  end
+
+  def get_state(producer \\ @kafka_producer) do
+    Agent.get(producer, & &1)
+  end
+end
+
+defmodule Sanbase.InMemoryKafka.Supervisor do
```
tspenov: IMO, this should go in a new file

lib/sanbase/application.ex
```diff
@@ -99,6 +99,14 @@ defmodule Sanbase.Application do
   @spec common_children() :: [:supervisor.child_spec() | {module(), term()} | module()]
   def common_children() do
     [
+      # Start the Kafka Exporter
+      {SanExporterEx, [kafka_producer_module: kafka_producer()]},
```
tspenov: shouldn't these be started only in the `web` pod ?

lib/sanbase/web.ex
```diff
@@ -19,7 +19,7 @@ defmodule Sanbase.Application.Web do
       Sanbase.TimescaleRepo,
 
       # Start the Clickhouse Repo
-      start_in({Sanbase.ClickhouseRepo, []}, [:dev, :prod]),
```
tspenov: Don't we need `ClickhouseRepo` in :dev too ?

lib/sanbase_web/endpoint_error_handler.ex
```diff
@@ -0,0 +1,69 @@
...
+  This will specifically handle:
+  1. Anything that has either :status_code or :plug_status field such as:
+    - Plug Praser Error - HTTP Status Code 400
+    - Phoenix Routing Error - HTTP Status Code 404
+    - Any other Plug or Phoenix specific error
+   2. Anything that does not have these fields will return HTTP Status Code 500
+
+  If the error is catched here the GraphQL Layer has not been reached, so the
+  `has_graphql_errors` field is set to `nil`. The GraphQL layer is responsible
+  for parsign the query, authorization and extraction of SAN Balance, so these
```
tspenov:
```suggestion
  for parsing the query, authorization and extraction of SAN Balance, so these
```

lib/sanbase/api_call_data/in_memory_kafka_producer.ex
tspenov: Though now that I think it can stay as it is cause it is only used in dev and test.

lib/sanbase_web/endpoint_error_handler.ex
```diff
@@ -0,0 +1,63 @@
+defmodule SanbaseWeb.Endpoint.ErrorHandler do
+  @moduledoc ~s"""
+  Handle errors that happened in the Endpoint.
+
+  This will specifically handle:
+  1. Anything that has either :status_code or :plug_status field such as:
+    - Plug Praser Error - HTTP Status Code 400
```
tspenov:
```suggestion
    - Plug Parser Error - HTTP Status Code 400
```

lib/sanbase/api_call_data/in_memory_kafka_producer.ex
IvanIvanoff: I thought that, too, but in the end I moved it to a separate directory and made it into 2 files

lib/sanbase/application.ex
IvanIvanoff: My idea was to also monitor the prometheus metrics calls, though now I realize this path is not handled. We can safely start this only in web I guess. 
The ugly this is that I want to start this before the `Endpoint` so it will be some `if` here and prepend it to the list only if the pod is `web` or `all`

IvanIvanoff: Handled it in a slightly more elegant way

lib/sanbase/api_call_data/api_call_data_exporter.ex
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  @doc ~s"""
+  Send all available data in the buffers before shutting down.
+
+  The ApiCallRecorder should be started before the Endpoint in the supervison tree.
+  This means that when shutting down it will be stopped after the Endpoint so
+  all API call data will be stored in Kafka and no more API calls are exepcted
+  """
+  def terminate(_reason, state) do
```
tspenov: Can you add some logging here, so we can trace when this happens cause this is some edge case. Maybe info logging inspecting the first element of the state and the size of the state.

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -63,5 +79,69 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
...
+  defp extract_caller_data(%{
+         auth: %{auth_method: :user_token, current_user: user, san_balance: san_balance}
+       }) do
+    {user.id, san_balance, :jwt, nil}
+  end
+
+  defp extract_caller_data(%{
+         auth: %{auth_method: :apikey, current_user: user, token: token, san_balance: san_balance}
+       }) do
+    {user.id, san_balance, :apikei, token}
```
tspenov:
```suggestion
    {user.id, san_balance, :apikey, token}
```

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -29,30 +45,30 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
...
-    # -`:nocache` returend from a resolver
-    # - result is taken from the cache and should not be stored again. Storing
-    # it again `touch`es it and the TTL timer is restarted. This can lead
-    # to infinite storing the same value if there are enough requests
-    should_cache? = !Process.get(:do_not_cache_query)
-    cache_result(should_cache?, blueprint)
+    queries = queries_in_request(blueprint)
+    export_api_call_data(queries, conn, blueprint)
+
+    case blueprint do
```
tspenov: maybe reuse the existing private function here ?
```elixir
if has_graphql_errors?(blueprint) do
  :ok
else
   should_cache? = !Process.get(:do_not_cache_query)
   cache_result(should_cache?, queries, blueprint)
end
```

lib/sanbase/api_call_data/api_call_data_exporter.ex
IvanIvanoff: This terminate will always be called when stopping (deploying new code) so it's not an edge case. I can add logging though

tspenov:
@IvanIvanoff build is failing

lib/sanbase/api_call_data/api_call_data_exporter.ex
```diff
@@ -0,0 +1,140 @@
+defmodule Sanbase.ApiCallDataExporter do
+  @moduledoc ~s"""
+  Module for persisting API calls data to Kafka.
+
+  The module exposes one function that should be used - `persist/1`.
+  This functions adds the data to an internal buffer that is flushed
+  every `kafka_flush_timeout` seconds or when the buffer is big enough.
+
+  The exporter cannot send data more than once every 1 second so the this
```
tspenov:
```suggestion
  The exporter cannot send data more than once every 1 second so the
```

lib/sanbase/api_call_data/api_call_data_exporter.ex
```diff
@@ -0,0 +1,144 @@
...
+    :ok
+  end
+
+  @spec handle_cast({:persist, api_call_data | [api_call_data]}, state) :: {:noreply, state}
+        when state: map()
+  def handle_cast(
+        {:persist, api_call_data},
+        state
+      ) do
+    data = api_call_data |> List.wrap() |> Enum.map(&{"", Jason.encode!(&1)})
```
tspenov: `Enum.map(&{"", Jason.encode!(&1)})` why is a tuple with first element empty string created ?

IvanIvanoff: Because the kafka library works only with {key, value} tuples

IvanIvanoff: I tried reworking the library but in a lot of places the code needs to changes as to introduce such change it should be optional and the default should be the old behavior

lib/sanbase/api_call_data/api_call_data_exporter.ex
```diff
@@ -0,0 +1,144 @@
...
+    data_length = length(data)
```
tspenov: maybe `new_messages_size` ?

lib/sanbase/api_call_data/api_call_data_exporter.ex
```diff
@@ -0,0 +1,144 @@
...
+       size: 0,
+       kafka_flush_timeout: kafka_flush_timeout,
+       buffering_max_messages: buffering_max_messages,
+       can_send_after_interval: can_send_after_interval,
+       can_send_after: DateTime.utc_now() |> DateTime.add(can_send_after_interval, :millisecond)
+     }}
+  end
+
+  @doc ~s"""
+  Assynchronously add the API call data to the buffer.
```
tspenov:
```suggestion
  Asynchronously add the API call data to the buffer.
```

lib/sanbase/api_call_data/api_call_data_exporter.ex
tspenov: But can that be handled by `san_exporter_ex` ?

IvanIvanoff: It can, but in this case the presence of `key` is hidden for the user and leads to unexpected behavior

PR #1193, opened 2019-05-23
Initial stripe integration

tspenov:
### List curentUser subscriptions with plans and products

``` graphql
    {
      currentUser {
        subscriptions {
          plan {
            id
            name
            access
            product {
              name
            }
          }
        }
      }
    }
```

### List all products and corresponding plans
```graphql
    {
      productsWithPlans {
        name
        plans {
          name
          access
        }
      }
    }
```
### Subscribe logged in user for plan with card token
```graphql

    mutation {
      subscribe(card_token: "card_token", plan_id: 1) {
        plan {
          id
          name
          access
          product {
            name
          }
        }
      }
    }
    ```

### CODE REVIEW

File: lib/sanbase/pricing/plan/access_seed.ex [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+
+  @pro %{
+    api_calls_munute: 120,
+    api_calls_month: 150_000,
+    historical_data_in_days: 18 * 30,
+    metrics: @advanced_metrics
+  }
+
+  @premium %{
+    api_calls_munute: 180,
```
IvanIvanoff: typo

lib/sanbase/pricing/product.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Sanbase.Pricing.Product do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+  import Ecto.Query
+
+  alias Sanbase.Pricing.Plan
+  alias __MODULE__
```
IvanIvanoff: not used

lib/sanbase/pricing/product.ex
```diff
@@ -0,0 +1,50 @@
...
+  schema "products" do
+    field(:name, :string)
+    field(:stripe_id, :string)
+
+    has_many(:plans, Plan)
+  end
+
+  def changeset(%__MODULE__{} = product, attrs \\ %{}) do
+    product
+    |> cast(attrs, [:stripe_id])
```
IvanIvanoff: `:name` should be allowed here

lib/sanbase/pricing/plan.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  import Ecto.Changeset
+  import Ecto.Query
+  import Sanbase.Math, only: [to_integer: 1]
+
+  alias Sanbase.Repo
+  alias Sanbase.Pricing.{Product, Subscription}
+  alias __MODULE__
+
+  @api_plans %{
```
IvanIvanoff: We can use an enum similar to `ColorEnum`

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -0,0 +1,50 @@
...
+      resolution
+    else
+      resolution
+      |> Resolution.put_result({
+        :error,
+        """
+        Requested metric #{query} is not provided by the current subscription plan #{
+          Subscription.plan_name(current_subscription)
+        }.
+        Please upgrade to #{Plan.upgrade_plan(current_subscription.plan)} to get access to #{
```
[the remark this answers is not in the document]
tspenov: Yeah, you are correct. Maybe should just return list of plans that contain that query.

lib/sanbase/pricing/product.ex
tspenov: used below

lib/sanbase/pricing/plan.ex
```diff
@@ -0,0 +1,85 @@
...
+    |> cast(attrs, [:stripe_id, :access])
+  end
+
+  def plans_with_metric(query) do
+    from(
+      p in Plan,
+      where: fragment(~s(access @> ?), ^%{metrics: [query]})
+    )
+    |> Repo.all()
+    |> Enum.map(&Map.get(&1, :name))
```
IvanIvanoff: You can move this into a `select: p.name` in the SQL?

lib/sanbase/pricing/plan.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def by_id(plan_id) do
+    Repo.get(__MODULE__, plan_id)
+    |> Repo.preload(:product)
+    |> update_stripe_id_if_not_present()
+  end
+
+  defp update_stripe_id_if_not_present(%__MODULE__{stripe_id: stripe_id} = plan)
```
IvanIvanoff: `create_or_update_stripe_id` :?

lib/sanbase/pricing/plan.ex
```diff
@@ -0,0 +1,85 @@
...
+        {:error, reason}
+    end
+  end
+
+  defp update_stripe_id_if_not_present(%__MODULE__{stripe_id: stripe_id} = plan)
+       when is_binary(stripe_id) do
+    {:ok, plan}
+  end
+
+  defp create_stripe_plan(
```
IvanIvanoff: Can you add some documentation here? Is this pinging the stripe API and creating something on their side?

lib/sanbase/pricing/plan/access_seed.ex
```diff
@@ -0,0 +1,64 @@
+defmodule Sanbase.Pricing.Plan.AccessSeed do
+  @standart_metrics [
```
IvanIvanoff: I would automate this (in another PR after this one) by introducing:
```
field :dev_activity, list_of(:dev_activity) do
   meta price_plan: :standard
   ....
end
```
And then introspecting the schema to fetch the queries

The idea is to touch the smallest subset of modules when you add new metrics

lib/sanbase/pricing/product.ex
```diff
@@ -0,0 +1,50 @@
...
+    product
+    |> cast(attrs, [:name, :stripe_id])
+  end
+
+  def by_id(product_id) do
+    Repo.get(__MODULE__, product_id)
+    |> update_stripe_id_if_not_present()
+  end
+
+  defp update_stripe_id_if_not_present(%__MODULE__{stripe_id: stripe_id} = product)
```
IvanIvanoff: `create_or_update_stripe_id`?
I see `Stripe.Product.create` in the body but the function does not indicate creation

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+    belongs_to(:user, User)
+    belongs_to(:plan, Plan)
+  end
+
+  def changeset(%Subscription{} = subscription, attrs \\ %{}) do
+    subscription
+    |> cast(attrs, [:plan_id, :user_id, :stripe_id])
+  end
+
+  def list_product_with_plans do
```
IvanIvanoff: `products_with_plans`

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+  def user_subscriptions(user) do
+    from(s in Subscription, where: s.user_id == ^user.id, order_by: [desc: s.id])
+    |> Repo.all()
+    |> Repo.preload(plan: [:product])
+  end
+
+  def current_subscription(user) do
+    user
+    |> user_subscriptions()
+    |> List.first()
```
IvanIvanoff: Why is the order (the first one in the list) important?

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  def historical_data_in_days(subscription) do
+    subscription.plan.access["historical_data_in_days"]
+  end
+
+  def is_restricted?(query) do
+    free = AccessSeed.free()[:metrics]
+    all = AccessSeed.all_restricted_metrics()
+
+    query in (all -- free)
```
IvanIvanoff: `query in all and not query in free` -  you can go without constructing new list

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  def plan_name(subscription) do
+    subscription.plan.name
+  end
+
+  def subscribe(user_id, card_token, plan_id) do
+    with %User{} = user <- Repo.get(User, user_id),
+         %Plan{} = plan <- Repo.get(Plan, plan_id),
+         {:ok, _} <- create_or_update_stripe_customer(user, card_token),
+         {:ok, stripe_subscription} <- create_stripe_subscription(user, plan) do
```
IvanIvanoff: What happens if `create_or_update_stripe_customer` succeeds and `create_stripe_subscription` fails?

lib/sanbase/pricing/plan.ex
tspenov: what this function does is: 
checks if `stripe_id` is present in the local table plans
- yes: return the plan
- no: create a new plan in Stripe and update local table `stripe_id`
so `update_stripe_id_if_not_present` describes it better. What do you think about `maybe_create_plan_in_stripe` ?

lib/sanbase/pricing/product.ex
tspenov: how about `maybe_create_resource_in_stripe` ? It will be used for products and plans.

lib/sanbase/pricing/subscription.ex
tspenov: You can have one active subscription for product. The other ones should be inactive. I take the last on as the current one. 
Maybe it should be `current_subscription_for_product` and when you subscribe for some product should deactivate other subscriptions.

lib/sanbase/pricing/plan.ex
IvanIvanoff: But how can you update something if it's not present?

lib/sanbase/pricing/subscription.ex
tspenov: harder to read IMO

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  defp update_subscription_with_coupon(nil, subscription), do: subscription
+
+  defp update_subscription_with_coupon(percent_off, subscription) do
+    {:ok, coupon} = Stripe.Coupon.create(%{percent_off: percent_off, duration: "forever"})
+    Map.put(subscription, :coupon, coupon)
+  end
+
+  defp percent_off(balance) when balance >= 1000, do: 20
```
IvanIvanoff: `percent_discount`

lib/sanbase/pricing/subscription.ex
```diff
@@ -0,0 +1,153 @@
...
+  defp percent_off(balance) when balance >= 200, do: 4
+  defp percent_off(_), nil
+
+  defp subscription_access?(nil, _query), do: true
```
IvanIvanoff: why `nil` gives access?

lib/sanbase_web.ex
```diff
@@ -36,7 +36,7 @@ defmodule SanbaseWeb do
       import Phoenix.Controller, only: [get_flash: 2, view_module: 1]
       import Phoenix.HTML
 
-      import SanbaseWeb.Router.Helpers
+      alias SanbaseWeb.Router.Helpers, as: Routes
```
IvanIvanoff: Is this something that should have been changed in the Phoenix 1.3 -> 1.4 migration?

lib/sanbase/pricing/subscription.ex
tspenov: I can wrap that in `Ecto.Multi`.
There is a lot of going on here: 
* get user
* get plan
* stripe customer exists ?
NO: create stripe customer if not existing and update `User` with stripe_customer_id
YES: update stripe customer with new card token
* create discount if needed, create stripe subscription
* create subscription in db

lib/sanbase/pricing/plan.ex
tspenov: plans/products are present cause they are inserted in db with a migration. If we create the corresponding plan/product in stripe and update with stripe_id in migration this will simplify this code. (this function won't be needed)  Currently this code handles 2 cases - when there is plan/product in stripe - do nothing, otherwise create in stripe and update local table with stripe_id.
Is that more clear ?

lib/sanbase_web.ex
tspenov: yes, i was running Stripe.js locally to test and added some routes, so this was needed.

kalkov:
There a couple of things that can be improved:
1. There are no tests.
2. Mixing "Model" code with calls to external services - Stripe. Should be decoupled. That will allow us to write fast unit tests w/o making external requests. 
3. Splitting of GraphQL schema/queries/types can be in separate PR and merged prior to this one.

IvanIvanoff:
@kalkov @tspenov I agree about the decoupling - let's have a separate module that will handle all external Stripe communication as a good start

tspenov:
@kalkov @IvanIvanoff 
- I've already started 3.) as a separate PR.
- I'll add more tests
- about 2.) - @kalkov How do you imagine decoupling ? @IvanIvanoff How adding a proxy module will help ? External communication is handled by Stripe.* modules.

kalkov:
@tspenov I imagine to have the following modules: 
* StripePlan
* StripeProduct
* StripeCustomer
* StripeSubscription

They need to have create/update/.. functions which will use the `stripity_stripe` lib to make the external calls to Stripe. No calls to Stripe should be made outside these modules. Most of them will need to receive a local record as an argument. Example: 

```elixir
StripePlan.create(local_plan)
``` 

There a couple of benefits with this approach:
* Easy de-featuring. Should be relatively easy to remove Stripe if we decide to. 
* Should be easy to add another payment service in the future.
* Tests for models(Plan/Subscription) will not be dependent on stripe/stripe-mock in any way. Just assert that StripePlan/StripeSubscription has been called with the right arguments w/o actually executing it. Stripe modules will have separate tests probably using stripe-mock.

lib/sanbase/pricing/subscription.ex
tspenov: lets skip using Ecto.Multi for now -> the only thing that can happen is to insert the subscription in Stripe and not insert in postgres so user paid but hasn't access. But that is a rare case.

lib/sanbase_web/graphql/schema/queries/pricing_queries.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  alias SanbaseWeb.Graphql.Middlewares.{
+    JWTAuth
+  }
+
+  import_types(SanbaseWeb.Graphql.Schema.PricingTypes)
+
+  object :pricing_queries do
+    @desc ~s"""
+    List availbale products with corresponding subscription plans.
```
kalkov: available

priv/repo/structure.sql
```diff
@@ -723,6 +758,36 @@ CREATE SEQUENCE public.processed_github_archives_id_seq
 ALTER SEQUENCE public.processed_github_archives_id_seq OWNED BY public.processed_github_archives.id;
 
 
+--
+-- Name: products; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.products (
+    id bigint NOT NULL,
+    name character varying(255),
```
kalkov: Shouldn't be not null?

priv/repo/structure.sql
```diff
@@ -525,6 +525,41 @@ CREATE SEQUENCE public.notifications_id_seq
 ALTER SEQUENCE public.notifications_id_seq OWNED BY public.notifications.id;
 
 
+--
+-- Name: plans; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.plans (
```
kalkov: Maybe we should be more restrictive here and add more not nulls?

test/support/graphql_test_helpers.ex
```diff
@@ -41,6 +44,38 @@ defmodule SanbaseWeb.Graphql.TestHelpers do
     |> ContextPlug.call(%{})
   end
 
+  def execute_query(conn, query, query_name) do
```
kalkov: :clap:

PR #1195, opened 2019-05-27
Marketcap fixes

tspenov:
image

### DIFF
--- a/lib/sanbase/prices/store.ex
+++ b/lib/sanbase/prices/store.ex
@@ -338,7 +338,7 @@ defmodule Sanbase.Prices.Store do
        FROM #{measurements_str}
        WHERE time >= #{DateTime.to_unix(from, :nanosecond)}
        AND time <= #{DateTime.to_unix(to, :nanosecond)}
-       GROUP BY time(#{resolution}) fill(0)/
+       GROUP BY time(#{resolution}) fill(none)/
   end
 
   defp volume_mcap_multiple_measurements_reducer(results, measurement_slug_map) do


IvanIvanoff:
@tspenov I am against this solution - all hell will break loose with it.
The behavior now will nicely fill zeroes when there is missing data, so you have the correct number of intervals.

If we have `fill(none)` then we have no guarantees that people can join results (pandas data frames for example) as you're not guaranteed that row X is for the same date time as row X from the other data frame.

The better solution will be to fill with the last observed value

IvanIvanoff:
We can use `fill(linear)` to fill missing data with the linear interpolation - it is arguable if 0 or interpolation is the better solution here.

haritonasty:
"fill with the last observed value" - looks nice

tspenov:
@IvanIvanoff `historyPrice` only for `TOTAL_ERC20` uses `fill(0)`, otherwise `historyPrice` for all other measurements uses `fill(none)`: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/prices/store.ex#L284
The idea is to unify these. IMO `fill(0)` will be best, though @santiment/frontend doesn't seem to like it for some reason.

haritonasty:
@tspenov
[`projectsListHistoryStats`](https://api.santiment.net/graphiql?query=%7B%0A%20%20projectsListHistoryStats(%0A%20%20%20%20from%3A%20%222019-05-25T21%3A00%3A00.000Z%22%2C%20%0A%20%20%20%20slugs%3A%20%5B%22tether%22%2C%20%22usd-coin%22%2C%20%22trueusd%22%2C%20%22paxos-standard-token%22%2C%20%22dai%22%2C%20%22stasis-eurs%22%2C%20%22gemini-dollar%22%5D%2C%20%0A%20%20%20%20to%3A%20%222019-05-30T21%3A00%3A00.000Z%22%2C%20%0A%20%20%20%20interval%3A%20%221d%22)%20%7B%0A%20%20%20%20datetime%0A%20%20%20%20marketcap%0A%20%20%20%20volume%0A%20%20%7D%0A%7D%0A&variables=%7B%7D)  also return zeros. If you return zeros for this case - we need to update logic for all cases, where we use `historyPrice` and `projectsListHistoryStats`
we can do it, if removing zeros can affect smth

IvanIvanoff:
last observed value functionality is not present in influxdb

tspenov:
fixed on frontend side. We should revisit API responses,  though.

PR #1200, opened 2019-05-30
Add miners balance metric

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/miner_balances_over_time.ex
```diff
@@ -0,0 +1,66 @@
...
+          datetime: DateTime.t(),
+          balance: float
+        }
+
+  @spec miner_balances(
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(miner_balances)} | {:error, String.t()}
+  def miner_balances(from, to, interval) do
```
IvanIvanoff: I think it would be good to have the same name as the functions in `HistoricalBalance`. So this module would be named `MinerHistoricalBalance` and will have functions like `historical_balance`, etc.

lib/sanbase/clickhouse/miner_balances_over_time.ex
```diff
@@ -0,0 +1,66 @@
+defmodule Sanbase.Clickhouse.MinerBalancesOverTime do
+  @moduledoc ~s"""
+  Uses ClickHouse to calculate miner balances over time.
+  """
+
+  alias Sanbase.DateTimeUtils
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+
+  @type miner_balances :: %{
```
IvanIvanoff:
```suggestion
  @type miner_balance :: %{
```

lib/sanbase/clickhouse/miner_balances_over_time.ex
```diff
@@ -0,0 +1,66 @@
...
+          datetime: DateTime.t(),
+          balance: float
+        }
+
+  @spec miner_balances(
+          DateTime.t(),
+          DateTime.t(),
+          String.t()
+        ) :: {:ok, list(miner_balances)} | {:error, String.t()}
```
IvanIvanoff:
```suggestion
        ) :: {:ok, list(miner_balance)} | {:error, String.t()}
```

plural is a list(singular) fits better

lib/sanbase_web/graphql/schema.ex
```diff
@@ -918,6 +918,20 @@ defmodule SanbaseWeb.Graphql.Schema do
       cache_resolve(&ClickhouseResolver.mining_pools_distribution/3)
     end
 
+    @desc """
+    Returns miner balances over time.
+    Currently only ETH is supported.
```
IvanIvanoff: If only ETH is supported let's make it a slug with default value

IvanIvanoff:
@kalkov The metrics is an aggregation for all miners, so I guess you can replace `miner_balance` with `miners_balance` everywhere :?

lib/sanbase/clickhouse/miners_balance.ex
```diff
@@ -0,0 +1,81 @@
...
+      SELECT days, arrayCumSum(values) AS values
+      FROM (
+        SELECT groupArray(time) AS days, groupArray(value) AS values
+        FROM (
+          SELECT
+            toUnixTimestamp(intDiv(toUInt32(toDateTime(date)), ?1) * ?1) AS time,
+            argMax(value, calculation_date) AS value
+          FROM eth_miners_metrics
+          WHERE
+            id = 102 AND
```
tspenov: I think, this should be extracted as module attribute with proper name.

lib/sanbase/clickhouse/miners_balance.ex
```diff
@@ -28,6 +30,22 @@ defmodule Sanbase.Clickhouse.MinersBalance do
 
   def balances(_, _, _, _), do: {:error, "Currently only ethereum is supported!"}
 
+  def first_datetime(_) do
+    query = """
+    SELECT
+      toUnixTimestamp(min(dt))
+    FROM #{@table}
```
IvanIvanoff: you can hardcode this one as such metrics (similar to the metrics for ethereum and bitcoin) because they exist since the creation of the blockchain -in this case this will always be 2015-07-30

lib/sanbase/clickhouse/miners_balance.ex
```diff
@@ -0,0 +1,99 @@
...
+        {:error, "The interval must consist of whole days!"}
+    end
+  end
+
+  def balances(_, _, _, _), do: {:error, "Currently only ethereum is supported!"}
+
+  def first_datetime(_) do
+    query = """
+    SELECT
+      toUnixTimestamp(min(dt))
```
IvanIvanoff:
```suggestion
      toUnixTimestamp(min(date))
```

`dt` is not a valid column here

lib/sanbase/clickhouse/miners_balance.ex
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  defp calculate_balances(from, to, interval_in_seconds) do
+    {query, args} = balances_query(from, to, interval_in_seconds)
+
+    {:ok, balances} =
+      ClickhouseRepo.query_transform(
+        query,
+        args,
+        fn [dt, balance] ->
```
IvanIvanoff: `dt` is misleiding name here - this is actually `dt_unix` or `timestamp` (not really accurate but we use it in other places)

IvanIvanoff:
This falls into the category of historical balances so it could be in that dir (Sanbase.Clickhouse.HistoricalBalance.MinerBalance) and have a function `historical_balance`

lib/sanbase/clickhouse/historical_balance/miners_balance.ex
```diff
@@ -0,0 +1,101 @@
...
+          datetime: DateTime.from_unix!(timestamp),
+          balance: balance
+        }
+      end
+    )
+  end
+
+  defp balances_query(from, to, interval) do
+    from_datetime_unix = DateTime.to_unix(from)
+    to_datetime_unix = DateTime.to_unix(to)
```
IvanIvanoff: iirk, you can pass DateTime structs as params (test though). In many places we convert to unix so we can calculate the `span` which is `(to-from)/interval`

PR #1202, opened 2019-05-30
Fix Elixir version in Dockerfile. Break builds if code is not formatted

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update \
 RUN mix local.hex --force
 RUN mix local.rebar --force
 
+
 WORKDIR /app
 
 COPY mix.lock /app/mix.lock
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -17,22 +17,21 @@ COPY mix.exs /app/mix.exs
 RUN mix deps.get
 RUN mix deps.compile
 
-COPY ./assets/package.json /app/assets/package.json
-COPY ./assets/package-lock.json /app/assets/package-lock.json
-
+COPY ./assets /app/assets
```
tspenov: do we need the whole assets directory ?

IvanIvanoff: We already have `COPY . /app` which copies everything*. Here I explicitly copy the assets folder before so I can run `npm build:prod` before managing the elixir files, which change one every build. This way the building of the js part can be fetched from cache almost every time
 
* - Here by everything I mean everything that's not in .dockerignore (/assets/node_modules is)

tspenov: nice

PR #1203, opened 2019-05-31
Use alpine images to reduce base container size by 400MB

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,9 @@
 # Elixir and phoenix assets build image
-FROM elixir:1.8.2-otp-22 as code_builder
+FROM elixir:1.8.2-otp-22-alpine as code_builder
 
 ENV MIX_ENV prod
 
-RUN apt-get update \
-  && curl -sL https://deb.nodesource.com/setup_8.x | bash \
-  && apt-get install -y nodejs git
+RUN apk add --update nodejs git make g++
 
 RUN mix local.hex --force
 RUN mix local.rebar --force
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,11 +1,9 @@
 # Elixir and phoenix assets build image
-FROM elixir:1.8.2-otp-22 as code_builder
+FROM elixir:1.8.2-otp-22-alpine as code_builder
 
 ENV MIX_ENV prod
 
-RUN apt-get update \
-  && curl -sL https://deb.nodesource.com/setup_8.x | bash \
-  && apt-get install -y nodejs git
+RUN apk add --update nodejs git make g++
```
valo: I think using `--no-cache` is a better option here: https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

```suggestion
RUN apk add --no-cache nodejs git make g++
```

Dockerfile-test
```diff
@@ -1,7 +1,6 @@
-# Elixir and phoenix assets build image
-FROM elixir:1.8.2-otp-22
+FROM elixir:1.8.2-otp-22-alpine
 
-RUN apt-get update && apt-get install -y git postgresql-client-9.6
+RUN apk add --update git postgresql-client make g++
```
IvanIvanoff:
```suggestion
RUN apk add --no-cache git postgresql-client make g++
```

Dockerfile
IvanIvanoff: Should we do this in all places?

Dockerfile
```diff
@@ -31,9 +29,9 @@ RUN mix phx.digest
 RUN mix release
 
 # Release image
-FROM elixir:1.8.2-otp-22
+FROM elixir:1.8.2-otp-22-alpine
 
-RUN apt-get install -y --only-upgrade bash
+RUN apk add --update bash
```
IvanIvanoff:
```suggestion
RUN apk add --no-cache bash
```

PR #1216, opened 2019-06-05
Cookie login

tspenov:
POC for creating a cookie after sign-in.

Flow:
Create session:
- After successful `emailLoginVerify` mutation - {:ok, token: token, user: user} is returned. 
- Add middleware after the resolver and on success put the JWT token in the session cookie.

Fetch session:
- :fetch_session plug in API pipeline
- In context plug get token from session and get the current_user from the token 

### Logout
```graphql
mutation {
  logout {
    success
  }
}

{
  "data": {
    "logout": {
      "success": true
    }
  }
}
```


lib/sanbase_web/router.ex
```diff
@@ -15,7 +15,18 @@ defmodule SanbaseWeb.Router do
 
   pipeline :api do
     plug(:accepts, ["json"])
+
+    plug(Plug.Session,
+      store: :cookie,
+      # 30 days
```
IvanIvanoff: I think we need to reduce it to ~7 days max.
cc @valo @Partysun @tspenov what do you think is appropriate lifespan for this?

lib/sanbase_web/graphql/schema.ex
```diff
@@ -952,6 +952,14 @@ defmodule SanbaseWeb.Graphql.Schema do
       arg(:token, non_null(:string))
 
       resolve(&AccountResolver.email_login_verify/2)
+
+      middleware(fn resolution, _ ->
+        with %{value: %{token: token, user: user}} <- resolution do
+          Map.update!(resolution, :context, fn ctx ->
+            Map.put(ctx, :auth_token, token)
```
valo: Is this the place where we put the session token in the cookie?

lib/sanbase_web/router.ex
valo: Not sure... 30 days might be ok. What could be the problems with a longer period?

lib/sanbase_web/graphql/schema.ex
tspenov: @valo It is here: https://github.com/santiment/sanbase2/pull/1216/files#diff-a8315ddc4492ae95bdeae22950e9cebcR45
Here only put it in context and so it can be fetched by the `before_send` part of the Absinthe pipeline.

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -37,10 +37,15 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     # - result is taken from the cache and should not be stored again. Storing
     # it again `touch`es it and the TTL timer is restarted. This can lead
     # to infinite storing the same value if there are enough requests
+
     should_cache? = !Process.get(:do_not_cache_query)
     cache_result(should_cache?, blueprint)
 
-    conn
+    if auth_token = blueprint.execution.context[:auth_token] do
+      Plug.Conn.put_session(conn, :auth_token, auth_token)
```
valo: Actually this is the place where the token is set in the session

tspenov: yes

lib/sanbase_web/router.ex
Partysun: 30 days used many co. Why do you think it is a critical @IvanIvanoff ?

IvanIvanoff: I do not think it is critical @Partysun.
30 days just seemed time long to me - I've seen JWT's living for 5-7 days mostly

config/dev.exs
```diff
@@ -99,6 +99,9 @@ config :arc,
   storage: Arc.Storage.Local,
   storage_dir: "/tmp/sanbase/filestore/"
 
+config :sanbase, SanbaseWeb.Plug.SessionPlug,
+  domain: {:system, "SANTIMENT_ROOT_DOMAIN", "localhost"}
```
IvanIvanoff: In tests it will still be `santiment.net`

lib/sanbase_web/endpoint.ex
```diff
@@ -33,16 +33,6 @@ defmodule SanbaseWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
 
-  # The session will be stored in the cookie and signed,
-  # this means its contents can be read but not tampered with.
-  # Set :encryption_salt if you would also like to encrypt it.
-  plug(
-    Plug.Session,
```
IvanIvanoff: You are moving the `Plug.Session` to a much later place. Will this cause any issues?

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -37,10 +37,12 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     # - result is taken from the cache and should not be stored again. Storing
     # it again `touch`es it and the TTL timer is restarted. This can lead
     # to infinite storing the same value if there are enough requests
+
     should_cache? = !Process.get(:do_not_cache_query)
     cache_result(should_cache?, blueprint)
 
     conn
+    |> maybe_create_or_drop_session(blueprint.execution.context)
```
IvanIvanoff: Is this the only place where you need to put this as it only covers the success path of a GraphQL?
You can check https://github.com/santiment/sanbase2/pull/1191 here where `persist` is called - in both graphql ok and error cases and when there is some exception

lib/sanbase_web/graphql/schema.ex
```diff
@@ -37,7 +37,9 @@ defmodule SanbaseWeb.Graphql.Schema do
     JWTAuth,
     ApikeyAuth,
     TimeframeRestriction,
-    ApiUsage
+    ApiUsage,
+    CreateSession,
+    DeleteSession
```
IvanIvanoff: Do you need 2 middlewares? Isn't it better to have a single middleware for handling the session and just pattern match there

lib/sanbase_web/graphql/schema.ex
```diff
@@ -952,6 +955,12 @@ defmodule SanbaseWeb.Graphql.Schema do
       arg(:token, non_null(:string))
 
       resolve(&AccountResolver.email_login_verify/2)
+      middleware(CreateSession)
+    end
+
+    field :logout, :logout do
+      resolve(fn _, _ -> {:ok, %{logout_success: true}} end)
+      middleware(DeleteSession)
```
IvanIvanoff: You need to guard this with `middleware(JWTAuth)`

lib/sanbase_web/plug/session_plug.ex
```diff
@@ -0,0 +1,22 @@
...
+  def call(conn, opts) do
+    runtime_opts =
+      opts
+      |> Keyword.put(:domain, domain())
+      |> Plug.Session.init()
+
+    Plug.Session.call(conn, runtime_opts)
+  end
+
+  defp domain(), do: Config.get(:domain)
```
IvanIvanoff: I think we can put such configs in a module attribute - I do not expect our not domain to change so we won't really configure it with ENV vars or change at runtime.

lib/sanbase_web/router.ex
```diff
@@ -15,7 +15,17 @@ defmodule SanbaseWeb.Router do
 
   pipeline :api do
     plug(:accepts, ["json"])
+
+    plug(SanbaseWeb.Plug.SessionPlug,
+      store: :cookie,
+      # 30 days
+      max_age: 24 * 60 * 60 * 30,
+      key: "_sanbase_sid",
+      signing_salt: "bfH+5EQ0"
```
IvanIvanoff: Better generate new signing_salt than the previously used one. Also put a comment that the signing key is generated by the secret base and the signing salt and as [according to José](https://github.com/phoenixframework/phoenix/issues/2146#issuecomment-283891524) it's not needed to have this in a secret also

lib/sanbase_web/endpoint.ex
tspenov: I don't think it was used before for anything. Now it is only for the API pipeline, though maybe it makes sense to be in the endpoint covering also browser pipeline. I am inclined to move again in Endpoint. What do you think ?

lib/sanbase_web/graphql/absinthe_before_send.ex
tspenov: On error this won't do anything cause its pattern matching for specific things set in the middleware after resolver which sets them only on successful resolving.

lib/sanbase_web/graphql/schema.ex
tspenov: Also though about this one. Lets go with one.

lib/sanbase_web/plug/session_plug.ex
tspenov: How will it work with module attribute? Domain has to be different for :dev and :prod env

PR #1222, opened 2019-06-10
Disable caching of all_insights query. Extract queries and mutations in own file.

tspenov:

### DIFF
--- a/lib/sanbase_web/graphql/schema/insight_queries.ex
+++ b/lib/sanbase_web/graphql/schema/insight_queries.ex
@@ -1,0 +1,158 @@
+defmodule SanbaseWeb.Graphql.Schema.InsightQueries do
+  @moduledoc ~s"""
+  Queries and mutations for working with Insights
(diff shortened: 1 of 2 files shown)


lib/sanbase_web/graphql/schema/insight_queries.ex
```diff
@@ -0,0 +1,158 @@
...
+    Fetch a list of all posts/insights.
+    Optionally a list of tags can be passed so it fetches all insights with these tags.
+    It supports paginations with `page` and `page_size` args.
+    """
+    field :all_insights, list_of(:post) do
+      arg(:page, :integer, default_value: 1)
+      arg(:page_size, :integer, default_value: 20)
+      arg(:tags, list_of(:string))
+
+      resolve(&PostResolver.all_insights/3)
```
tspenov: That is the change needed so voting is showed properly. Other changes are moving into a separate file.

PR #1229, opened 2019-06-11
Add statistics for users and watchlists

IvanIvanoff:
#### Summary
image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -16,7 +16,7 @@ defmodule Sanbase.Auth.Statistics do
     %{
       "average_tokens_staked" => Math.average(san_balances),
       "median_tokens_staked" => Math.median(san_balances),
-      "tokens_staked" => Enum.sum(san_balances) |> Float.round(2)
+      "tokens_staked" => Enum.sum(san_balances) |> Kernel.*(1.0) |> Float.round(2)
     }
   end
 
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -13,10 +13,16 @@ defmodule Sanbase.Auth.Statistics do
       |> Repo.all()
       |> Enum.map(&Decimal.to_float/1)
 
+    {min_stake, max_stake} = Sanbase.Math.min_max(san_balances)
+
     %{
       "average_tokens_staked" => Math.average(san_balances),
       "median_tokens_staked" => Math.median(san_balances),
-      "tokens_staked" => Enum.sum(san_balances) |> Kernel.*(1.0) |> Float.round(2)
+      "tokens_staked" => Enum.sum(san_balances) |> Kernel.*(1.0) |> Float.round(2),
+      "biggest_stake" => max_stake |> Kernel.*(1.0) |> Float.round(2),
+      "smallest_stake" => min_stake |> Kernel.*(1.0) |> Float.round(2),
+      "users_with_over_1000_san" => Enum.count(san_balances, fn balance -> balance >= 1000 end),
+      "users_with_over_200_san" => Enum.count(san_balances, fn balance -> balance >= 200 end)
     }
   end
 
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -1,0 +1,103 @@
+defmodule Sanbase.Auth.Statistics do
+  import Ecto.Query
+
+  alias Sanbase.Repo
+  alias Sanbase.Math
+  alias Sanbase.Auth.{User, Settings, UserSettings}
+
+  def tokens_staked() do
+    san_balances =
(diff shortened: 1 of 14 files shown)


lib/sanbase_web/admin/statistics/statistics.ex
```diff
@@ -0,0 +1,35 @@
...
+      Sanbase.Statistics.get_all()
+      |> Enum.map(fn {name, map} ->
+        table do
+          h3("#{format_underscored(name)}")
+
+          Enum.map(map, fn {key, value} ->
+            tr do
+              td("#{format_underscored(key)}")
+              td(" | ")
+              td("#{value}")
```
tspenov:
```suggestion
              td align: "right" do
                "#{value}"
              end
```

lib/sanbase_web/admin/statistics/statistics.ex
```diff
@@ -0,0 +1,35 @@
...
+
+      Sanbase.Statistics.get_all()
+      |> Enum.map(fn {name, map} ->
+        table do
+          h3("#{format_underscored(name)}")
+
+          Enum.map(map, fn {key, value} ->
+            tr do
+              td("#{format_underscored(key)}")
+              td(" | ")
```
tspenov:
```suggestion
              td do
                pre("|", style: "all: initial; margin: 0 10px 0 10px;")
              end
```

tspenov: thats the only way I can think of for inserting some space. `&nbsp;` won'r work here

lib/sanbase_web/admin/statistics/statistics.ex
IvanIvanoff: This does not increase the readability imo

lib/sanbase_web/admin/statistics/statistics.ex
IvanIvanoff: that works :tada:

lib/sanbase/auth/statistics.ex
```diff
@@ -0,0 +1,103 @@
...
+    from_naive = DateTime.to_naive(from)
+    to_naive = DateTime.to_naive(to)
+
+    from(
+      u in User,
+      where: u.inserted_at >= ^from_naive and u.inserted_at <= ^to_naive
+    )
+  end
+
+  defp positive_san_balance(query) do
```
tspenov:
```suggestion
  defp positive_san_balance_query(query) do
```
Lets have a convention for subqueries and to end with `_query` so we know we can plug them. I think you already do it for other subqueries.

lib/sanbase/auth/statistics.ex
```diff
@@ -0,0 +1,103 @@
...
+  import Ecto.Query
+
+  alias Sanbase.Repo
+  alias Sanbase.Math
+  alias Sanbase.Auth.{User, Settings, UserSettings}
+
+  def tokens_staked() do
+    san_balances =
+      from(u in User)
+      |> positive_san_balance()
```
tspenov:
```suggestion
      |> positive_san_balance_query()
```

lib/sanbase/auth/statistics.ex
```diff
@@ -0,0 +1,103 @@
...
+    }
+  end
+
+  @doc ~s"""
+  Return the number of all users that are staking tokens
+  """
+  @spec staking_users() :: non_neg_integer()
+  def staking_users() do
+    from(u in User)
+    |> positive_san_balance()
```
tspenov:
```suggestion
    |> positive_san_balance_query()
```

lib/sanbase/auth/statistics.ex
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  @doc ~s"""
+  Return the number of newly registered users in the given time interval that have
+  positive SAN balance
+  """
+  @spec registered_staking_users(DateTime.t(), DateTime.t()) :: non_neg_integer()
+  def registered_staking_users(%DateTime{} = from, %DateTime{} = to) do
+    registered_users_query(from, to)
+    |> positive_san_balance()
```
tspenov:
```suggestion
    |> positive_san_balance_query()
```

lib/sanbase/auth/statistics.ex
IvanIvanoff: Now that I think about it if we append `_query` to all such functions it will become weird:
```
 registered_users_query(from, to)
    |> positive_san_balance_query()
    |> count_query()
    |> Repo.one()
```

Another option is to append `query` to the function that initializes the query - the one that does not get another query as an arg

tspenov: My idea is to mark these that return query - so these we know that we can plug them most of the time and they will do what they are supposed to do (filter, aggregate ...)

PR #1231, opened 2019-06-12
Add newsletter subscription datetime to the user settings

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/auth/settings.ex
+++ b/lib/sanbase/auth/settings.ex
@@ -10,6 +10,7 @@ defmodule Sanbase.Auth.Settings do
     field(:telegram_chat_id, :integer)
     field(:has_telegram_connected, :boolean, virtual: true)
     field(:newsletter_subscription, :string, default: "OFF")
+    field(:newsletter_subscription_updated_at_unix, :integer, default: nil)
   end
 
   def changeset(schema, params) do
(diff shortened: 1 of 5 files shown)


lib/sanbase/auth/user_settings.ex
```diff
@@ -43,7 +43,10 @@ defmodule Sanbase.Auth.UserSettings do
   end
 
   def change_newsletter_subscription(%User{id: user_id}, params) do
-    settings_update(user_id, params)
+    settings_update(
+      user_id,
+      params |> Map.put(:newsletter_subscribed_at, NaiveDateTime.utc_now())
```
tspenov: maybe `newsletter_subscription_updated_at` cause it is updated on every newsletter option change ?

IvanIvanoff: @tspenov done

lib/sanbase/auth/settings.ex
```diff
@@ -32,10 +33,11 @@ defmodule Sanbase.Auth.Settings do
...
   defp normalize_newsletter_subscription(changeset, field, value) do
-    put_change(
-      changeset,
-      field,
-      value |> Atom.to_string() |> String.upcase()
+    changeset
+    |> put_change(field, value |> Atom.to_string() |> String.upcase())
+    |> put_change(
+      :newsletter_subscription_updated_at_unix,
+      DateTime.utc_now() |> DateTime.to_unix()
```
IvanIvanoff: This is because we cannot compare out of the box DateTimes inside `jsonb` but we can compare integers

PR #1235, opened 2019-06-13
Evaluate signals for users with telegram. Fix caching

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/signals/evaluator/evaluator.ex
+++ b/lib/sanbase/signals/evaluator/evaluator.ex
@@ -6,11 +6,19 @@ defmodule Sanbase.Signals.Evaluator do
   alias Sanbase.Signals.Evaluator.Cache
   alias Sanbase.Signals.{UserTrigger, Trigger}
 
+  require Logger
+
   @doc ~s"""
   Takes a list of triggers and returns its subset that evaluate to true at the given moment.
   """
   @spec run(list()) :: list()
-  def run(user_triggers) do
+  def run([]), do: []
+
+  def run([ut | _] = user_triggers) do
+    Logger.info(
+      "Start evaluating #{length(user_triggers)} signals of type #{ut.trigger.settings.__struct__}"
+    )
+
     user_triggers
     |> Sanbase.Parallel.map(
       &evaluate/1,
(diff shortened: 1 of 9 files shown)


lib/sanbase/signals/user_trigger.ex
```diff
@@ -137,8 +137,24 @@ defmodule Sanbase.Signals.UserTrigger do
     )
     |> Repo.all()
     |> Enum.map(&trigger_in_struct/1)
+    |> Enum.reject(&has_invalid_notification_channel?/1)
   end
 
+  defp has_invalid_notification_channel?(%{
```
tspenov: IMO it is better `has_valid_notification_channel?` with `channel=telegram && has_telegram_connected=true` or
`channel=email && not_nil(email)` and false otherwise

tspenov:
@IvanIvanoff build is failing

lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -21,14 +29,24 @@ defmodule Sanbase.Signals.Evaluator do
...
       )
 
-    %UserTrigger{user_trigger | trigger: trigger}
+    %UserTrigger{
+      user_trigger
+      | trigger: %{
+          original_trigger
+          | settings: %{
+              original_trigger.settings
+              | payload: trigger.settings.payload,
```
tspenov: maybe add a comment that payload is got from the cache if existing ?

lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -21,14 +29,24 @@ defmodule Sanbase.Signals.Evaluator do
...
+                triggered?: trigger.settings.triggered?
```
tspenov: Does `triggered?` has to be got from cache or `original_trigger` ?

tspenov:
I think we should add a test that covers this. Scenario can be: create 2 triggers with same cache key but different settings, evaluate and check which settings changed.

IvanIvanoff:
@tspenov the cache key change is also very important here. If we do not include `last_triggered` to the key it really depends on which trigger will be picked up and calculated first. If you created a trigger a few hours ago and now I create it with the same params and mine executes first you'll get another set of signals even if your cool down is big

PR #1250, opened 2019-06-24
Add historical stats to the watchlist GQL type

IvanIvanoff:
#### Summary
Instead of fetching the whole watchlist and then passing the list of slugs back to the backend to obtain the historical stats, now it can be done with a single query:
```
{
  watchlist(id: 344) {
    historicalStats(from: "2019-06-01T00:00:00Z", to: "2019-06-10T00:00:00Z", interval: "1d") {
      datetime
      marketcap
      volume
    }
    listItems {
      project {
        slug
      }
    }
  }
}
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/schema/types/user_list_types.ex
+++ b/lib/sanbase_web/graphql/schema/types/user_list_types.ex
@@ -2,6 +2,7 @@ defmodule SanbaseWeb.Graphql.UserListTypes do
   use Absinthe.Schema.Notation
 
   import Absinthe.Resolution.Helpers
+  import SanbaseWeb.Graphql.Cache, only: [cache_resolve: 1]
 
   alias SanbaseWeb.Graphql.SanbaseRepo
   alias SanbaseWeb.Graphql.Resolvers.UserListResolver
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/schema/types/user_list_types.ex
```diff
@@ -26,5 +27,13 @@ defmodule SanbaseWeb.Graphql.UserListTypes do
     field(:list_items, list_of(:list_item), resolve: &UserListResolver.list_items/3)
     field(:inserted_at, non_null(:naive_datetime))
     field(:updated_at, non_null(:naive_datetime))
+
+    field(:historical_stats, list_of(:combined_projects_stats)) do
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, non_null(:string), default_value: "1d")
```
kalkov: Default value and non null does not play well, better remove non_null.

kalkov:
Tests?

lib/sanbase_web/graphql/schema/types/user_list_types.ex
IvanIvanoff: It's more of an issue with graphiql - `non_null` and `default_value` are solving different problems.
`interval: null` will not trigger the default value - only missing value will trigger the default.

Nevertheless, I should remove it as the graphql behavior is confusing for most people

PR #1251, opened 2019-06-24
Update subscription

tspenov:
Executed from logged in users.

### upgrade/downgrade subscription without new card

```graphql
    mutation {
      updateSubscription(subscriptionId: #{subscription_id}, planId: #{plan_id}) {
        plan {
          id
          name
          access
          interval
          amount
          product {
            name
          }
        }
      }
    }
```
### Schedule a subscription for cancellation. Cancellation will be done at `scheduledForCancellationAt`.
```
    mutation {
      cancelSubscription(subscriptionId: #{subscription_id}) {
        isScheduledForCancellation
        scheduledForCancellationAt
      }
    }
```
new fields in subscription:
```
    field(:id, :id)
    field(:user, :user)
    field(:plan, :plan)
    field(:current_period_end, :datetime)
    field(:cancel_at_period_end, :boolean)
```


kalkov:
Tests?

lib/sanbase/pricing/product.ex
```diff
@@ -6,13 +6,17 @@ defmodule Sanbase.Pricing.Product do
   alias Sanbase.Pricing.Plan
   alias Sanbase.Repo
 
+  @sanbase_api 1
```
IvanIvanoff: Documentation?
Also all `Sanbase.Pricing.*` modules should have moduledoc's + docs for the functions

tspenov:
@IvanIvanoff @kalkov working on adding both. Forgot to make the PR draft, sorry for confusion.

lib/sanbase_web/graphql/schema/types/pricing_types.ex
```diff
@@ -15,7 +15,16 @@ defmodule SanbaseWeb.Graphql.Schema.PricingTypes do
...
+  object :subscription_plan do
+    field(:id, :id)
+    field(:user, :user)
     field(:plan, :plan)
+    field(:current_period_end, :datetime)
+    field(:cancel_at_period_end, :boolean)
+  end
+
+  object :subscription_cancellation do
+    field(:scheduled_for_cancellation, :boolean)
```
IvanIvanoff: let's call it `is_scheduled_for_cancelation`

lib/sanbase/pricing/plan.ex
```diff
@@ -1,4 +1,9 @@
 defmodule Sanbase.Pricing.Plan do
+  @moduledoc """
+  Module for managing billing plans that define the amount and billing cycle
+  for subscriptions.
+  We have plans with the same name but different interval (`month`, `year`) and amount.
```
IvanIvanoff: I'm not sure I understand this

lib/sanbase/pricing/plan.ex
```diff
@@ -9,8 +14,10 @@ defmodule Sanbase.Pricing.Plan do
 
   schema "plans" do
     field(:name, :string)
+    # amount is in cents
     field(:amount, :integer)
```
IvanIvanoff: `price_in_cents`?

lib/sanbase/pricing/plan.ex
```diff
@@ -9,8 +14,10 @@ defmodule Sanbase.Pricing.Plan do
...
     field(:currency, :string)
+    # interval is one of `month` or `year`
     field(:interval, :string)
```
IvanIvanoff: If interval is `month` or `year` let's name this param differently as we use this name for something else already

lib/sanbase/pricing/product.ex
```diff
@@ -22,6 +31,10 @@ defmodule Sanbase.Pricing.Product do
     Repo.get(__MODULE__, product_id)
   end
 
+  @doc """
+  If product does not have `stripe_id`. Create a product in Stripe and update with
```
IvanIvanoff: The first sentence is not a sentence.
I would reword it 
```
Does nothing if the product has `stripe_id`.
Create a product in Stripe and update it with the received `stripe_id` otherwise.

File: lib/sanbase/pricing/subscription.ex [hunk 7]
```diff
@@ -20,31 +25,131 @@ defmodule Sanbase.Pricing.Subscription do
 
   schema "subscriptions" do
     field(:stripe_id, :string)
+    field(:active, :boolean, null: false, default: false)
```
IvanIvanoff: `is_active`

lib/sanbase/pricing/subscription.ex
```diff
@@ -20,31 +25,131 @@ defmodule Sanbase.Pricing.Subscription do
...
+    field(:current_period_end, :utc_datetime)
+    field(:cancel_at_period_end, :boolean, null: false, default: false)
```
IvanIvanoff: `should_ cancel_at_period_end`

lib/sanbase/pricing/subscription.ex
```diff
@@ -20,31 +25,131 @@ defmodule Sanbase.Pricing.Subscription do
...
   end
 
+  @doc """
+  Upgrade or Downgrade plan:
+
+  - Updates subcription in Stripe with new plan.
+  - Updates local subscription
+  Stripe docs:   https://stripe.com/docs/billing/subscriptions/upgrading-downgrading#switching
+  """
+  def update_subscription(user_id, subscription_id, plan_id) do
```
IvanIvanoff: This function is not understandable - it just does too many things. All the `get` stuff should be handled in the caller function and as this is `update_*` it should only handle the update.

lib/sanbase/pricing/subscription.ex
```diff
@@ -20,31 +25,131 @@ defmodule Sanbase.Pricing.Subscription do
...
+           StripeApi.update_subscription(subscription.stripe_id, %{
+             items: [
+               %{
+                 id: item_id,
+                 plan: subscription.plan.stripe_id
+               }
+             ]
+           }),
+         {:ok, updated_subscription} <-
+           update_subscription_db(subscription, %{plan_id: new_plan.id}) do
```
IvanIvanoff: What happens if this fails? The previous function succeeded `StripeApi.update_subscription` but no one is rolling it back. I think we should have only transactional logic when dealing with payments

lib/sanbase/stripe_api.ex
```diff
@@ -51,6 +51,36 @@ defmodule Sanbase.StripeApi do
...
+  def update_subscription(stripe_id, params) do
+    Stripe.Subscription.update(stripe_id, params)
+  end
+
+  def cancel_subscription(stripe_id) do
+    stripe_id
+    |> update_subscription(%{cancel_at_period_end: true})
+  end
+
+  def get_subscription_first_item_id(stripe_id) do
```
IvanIvanoff: Why is the first item important?

test/sanbase_web/graphql/pricing/subscribe_api_test.exs
```diff
@@ -85,8 +98,71 @@ defmodule SanbaseWeb.Graphql.Pricing.SubscribeApiTest do
...
+        )
+
+      query = cancel_subscription_mutation(subscription.id)
+      response = execute_mutation(context.conn, query, "cancelSubscription")
+
+      assert response["scheduledForCancellation"]
+      assert response["scheduledForCancellationAt"] == DateTime.to_iso8601(tomorrow)
+    end
+
+    test "error when cancelling not own subscription", context do
```
IvanIvanoff: please add tests for each possible step that could fail

lib/sanbase/pricing/plan.ex
tspenov: We have 2 plans `Pro` for product `SanbaseAPI` which are differentiated by the `interval` field (whether id `month` or `year`) and also the amount is different. That is because Stripe regards `month` and `year` plans as different plans - so its easier to keep them in postgres like this.

lib/sanbase/pricing/plan.ex
tspenov: I think its best to stick with the same names that Stripe is using to avoid confusion.

lib/sanbase/pricing/plan.ex
tspenov: Again that is what Stripe uses so I prefer to stick to this.

IvanIvanoff: IMO it's better to keep our API consistent - users of our API don't care about stripe naming

lib/sanbase/pricing/product.ex
tspenov:
```suggestion
  If product does not have `stripe_id` - create a product in Stripe and update with
```

lib/sanbase/pricing/subscription.ex
tspenov: again using Stripe field names - otherwise it will become confusing

lib/sanbase/pricing/subscription.ex
tspenov: ok, lets extract fetching user, plan and subscription in the resolver and see if that will help with understanding

lib/sanbase/pricing/subscription.ex
tspenov: If that fails currently we have subscription that user has paid for but no access. 
We have few options: 
1. reverse the action in Stripe -> in this case update back with initial params
2. retry - this still can fail and the response will be blocked for some time
3.  just get a subscription list periodically and sync with Stripe once 5 minutes. That way we'll keep local and Stripe subscriptions in sync and autocorrect these things.
I am voting for option 3 but that will be implemented in subsequent PR.

lib/sanbase/stripe_api.ex
tspenov: We expect to have one item in Subscription - that is the plan the user is subscribed to. 
We fetch it here so we can update it with the new plan.

lib/sanbase/pricing/plan.ex
tspenov: `interval` is used for timeseries queries. Here it is internal mutation used by frontend so I don't think we're mixing anything.  Also frontend is already using it.

lib/sanbase/pricing/subscription.ex
tspenov: field removed - `cancel_at_period_end` and `current_period_end` to determine whether subscription is active.

lib/sanbase/pricing/subscription.ex
tspenov: Extracted fetching and error handling logic in caller

test/sanbase_web/graphql/pricing/subscribe_api_test.exs
tspenov: Added tests for all scenarios

lib/sanbase/stripe_api.ex
IvanIvanoff: let's just call it `get_subscription_item_id` so it's clear that we expect to have one

lib/sanbase/pricing/subscription.ex
IvanIvanoff: Can we reverse it? Check if user has balance -> give access on our side -> make the user pay for it -> acknowledge. In this case we cannot have the situation where the user pays but has no access. We can also make use of some library like https://github.com/appcues/gen_retry to retry until success

tspenov: I don't think there is method for checking user card balance in Stripe. And why give access before payment - user can decide to not pay, so he can abuse using the system for some time.
As I said it is better in case of db failure to just sync on some intervals, that can lead to some delay but the system will be correct.

PR #1262, opened 2019-06-28
Webhook for payment succeeded event

tspenov:


lib/sanbase/pricing/stripe_event.ex
```diff
@@ -0,0 +1,97 @@
+defmodule Sanbase.Pricing.StripeEvent do
+  @moduledoc """
+  Module for persisting and handling Stripe webhook events.
+
+  Events are persisted for 3 reasons:
+  1. help making event processing idempotent by not processing already processed events
```
kalkov: uppercase first letter of help

priv/repo/migrations/20190709125125_add_stripe_events_table.exs
```diff
@@ -0,0 +1,14 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:stripe_events, primary_key: false) do
+      add(:event_id, :string, primary_key: true)
+      add(:type, :string, null: false)
+      add(:payload, :jsonb, null: false)
+      add(:is_processed, :boolean, default: false)
+
+      timestamps
```
kalkov: Here we should probably have only created_at timestamp?

tspenov: We insert the record when receive the event and update `is_processed` when we handle the event async so `updated_at` should stay.

lib/sanbase/pricing/stripe_event.ex
tspenov:
```suggestion
  1. Help making event processing idempotent by not processing already processed events
```

lib/sanbase_web/endpoint.ex
```diff
@@ -25,6 +25,10 @@ defmodule SanbaseWeb.Endpoint do
   plug(Plug.RequestId)
   plug(Plug.Logger)
 
+  # This plug plug should be before Plug.Parsers because its reading the
```
tspenov:
```suggestion
  # This plug should be placed before Plug.Parsers because its reading the
```

lib/sanbase_web/endpoint.ex
```diff
@@ -25,6 +25,10 @@ defmodule SanbaseWeb.Endpoint do
   plug(Plug.RequestId)
   plug(Plug.Logger)
 
+  # This plug should be placed before Plug.Parsers because its reading the
```
tspenov:
```suggestion
  # This plug should be placed before Plug.Parsers because it is reading the
```

PR #1264, opened 2019-07-02
Aggregate pricing data with LAST instead of MEAN

IvanIvanoff:
LAST is more stable and does not lead to radical differences in case
larger interval are aggregated

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/notifications/price_volume_diff.ex
+++ b/lib/sanbase/notifications/price_volume_diff.ex
@@ -45,7 +45,7 @@ defmodule Sanbase.Notifications.PriceVolumeDiff do
     measurement = Sanbase.Influxdb.Measurement.name_from(project)
 
     with {:ok, [[_dt, volume]]} <-
-           Sanbase.Prices.Store.fetch_mean_volume(measurement, from_datetime, to_datetime) do
+           Sanbase.Prices.Store.fetch_average_volume(measurement, from_datetime, to_datetime) do
       volume >= notification_volume_threshold()
     else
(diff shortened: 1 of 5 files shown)


IvanIvanoff:
@Partysun @valo @haritonasty 
Does it make  sense to change MEAN to LAST in other places than the one that fetches the marketcap (as per request)?
Such change will change historyPrice response significantly for larger periods. But not changing it means that we'll be aggregating the data differently in the 2 charts (watchlist widget and PDP)

kalkov:
@IvanIvanoff It makes sense to me to change it in the other places.

Partysun:
> @Partysun @valo @haritonasty @kalkov @tspenov
> Does it make sense to change MEAN to LAST in other places than the one that fetches the marketcap (as per request)?
> Such change will change historyPrice response significantly for larger periods. But not changing it means that we'll be aggregating the data differently in the 2 charts (watchlist widget and PDP)

Please, can you add more info/example Don't understand the context.

IvanIvanoff:
@Partysun if you request the price/volume/mcap for 7 days intervals currently you'll get the average of them for the interval. This PR changes not to return the average but the last value instead.

This is done to avoid issues like: 
image
image

As the different time intervals use different aggregation intervals, the values of the last one (displayed as the number) differ significantly.

Partysun:
@IvanIvanoff ok. Yeah, we need to show last. Cause it's the true value.

PR #1288, opened 2019-07-10
Change test db config

kalkov:
#### Summary
Add some default credentials to test env and few timeouts so `IEx.pry` can work properly

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -24,14 +24,26 @@ config :sanbase, Sanbase.ApiCallDataExporter,
 # Configure postgres database
 config :sanbase, Sanbase.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
+  username: "postgres",
+  password: "postgres",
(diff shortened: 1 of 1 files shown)


config/test.exs
```diff
@@ -24,14 +24,26 @@ config :sanbase, Sanbase.ApiCallDataExporter,
...
   pool: Ecto.Adapters.SQL.Sandbox,
+  username: "postgres",
+  password: "postgres",
+  hostname: "localhost",
   database: "sanbase_test",
-  pool_size: 5
+  pool_size: 5,
+  pool_timeout: 600_000,
+  timeout: 600_000,
+  ownership_timeout: 600_000
```
IvanIvanoff: Let's have these in `test.secret.exs` as the timeouts show bugs in our tests. Currently we have a few places where DB connections are not properly cleaned and if we get too many of such cases we can reach the pool size without noticing

PR #1289, opened 2019-07-10
Watchlist settings

IvanIvanoff:
#### Summary
1. Restructure the Signal.Validation into two modules, so one of them is more general and is used in the watchlists.
2. Add watchlist settings per user. The behavior is the following:
- If the user or the watchlist creator does not have settings - return default settings
- If the user does not have settings and the creator does have - return the creator's settings
- If the user has settings - return his or her settings

Examples:
```graphql
mutation {
        updateWatchlistSettings(
          id: 10,
          settings: {
            pageSize: 50
            timeWindow: "180d"
            tableColumns: <json map with arbitrary keys>
          }) {
            pageSize
            tableColumns
            timeWindow
          }
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -129,6 +139,23 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+  def update_watchlist_settings(_root, %{id: watchlist_id, settings: settings}, %{
+        context: %{auth: %{current_user: current_user}}
+      }) do
+    UserList.Settings.update_or_create_settings(watchlist_id, current_user.id, settings)
+    |> case do
+      {:ok, %{settings: settings}} ->
+        {:ok, settings}
+
+      {:error, changeset} ->
+        # {:error,
```
kalkov: These are leftovers probably?

priv/repo/migrations/20190710140802_add_watchlist_settings_table.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Sanbase.Repo.Migrations.AddWatchlistSettingsTable do
```
kalkov: Two migrations with same content but diff names?

kalkov:
Can we also have tests for validations?

priv/repo/migrations/20190710140802_add_watchlist_settings_table.exs
IvanIvanoff: Hmm, I couldn't find the migration and created a new one. Obviously it was where it should be :smiley:

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
IvanIvanoff: Yeah, I need to do some cleaning

lib/sanbase/user_lists/settings.ex
```diff
@@ -0,0 +1,144 @@
...
+
+  The settings are stored per user. One watchlist can have many settings for many
+  users. When a user requires the settings for a watchlist the resolution process
+  is the first one that matches:
+  1. If the requesting user has defined settings - return them.
+  2. If the watchlist craetor has defined settings - return them.
+  3. Return the default settings
+  """
+
+  defmodule WatchlistSettings do
```
tspenov: why nested modules ?

IvanIvanoff: no particular reason. I'm just trying how it looks. And this module won't be used by anyone else

lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -129,6 +139,23 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+    |> case do
+      {:ok, %{settings: settings}} ->
+        {:ok, settings}
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        {:error,
+         message: "Cannot update watchlist settings",
+         details: Sanbase.Utils.ErrorHandling.changeset_errors_to_str(changeset)}
+
+        # {:error, inspect(changeset)}
```
tspenov: leftover

lib/sanbase/user_lists/settings.ex
```diff
@@ -0,0 +1,144 @@
...
+  use Ecto.Schema
+
+  import Ecto.Query
+  import Ecto.Changeset
+
+  alias Sanbase.Repo
+  alias Sanbase.Auth.User
+  alias Sanbase.UserList
+  alias __MODULE__.WatchlistSettings
+
```
tspenov: shouldn't we set `@primary_key false` as per: https://hexdocs.pm/ecto/Ecto.Schema.html#module-primary-keys ?

priv/repo/migrations/20190710140802_add_watchlist_settings_table.exs
```diff
@@ -0,0 +1,15 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:watchlist_settings) do
```
tspenov: I think we need `primary_key: false` here and `primary_key: true` on each of the fields making the composite primary_key

lib/sanbase/user_lists/settings.ex
IvanIvanoff: The table still has primary key `id`. I probably should remote the `primary_key: true`

lib/sanbase_web/graphql/schema/queries/watchlist_queries.ex
```diff
@@ -122,6 +122,13 @@ defmodule SanbaseWeb.Graphql.Schema.WatchlistQueries do
       resolve(&UserListResolver.update_user_list/3)
     end
 
+    field :update_watchlist_settings, :watchlist_settings do
+      arg(:id, non_null(:integer))
+      arg(:settings, :watchlist_settings_input_object)
```
tspenov: non_null ?

lib/sanbase/user_lists/settings.ex
```diff
@@ -0,0 +1,144 @@
...
+    settings =
+      get_settings(ul.id, user.id) ||
+        get_settings(ul.id, ul.user_id) ||
+        WatchlistSettings.default_settings()
+
+    {:ok, settings}
+  end
+
+  def settings_for(%UserList{} = ul, _) do
+    get_settings(ul.id, ul.user_id)
```
tspenov: That id getting access to settings of watchlist from not-logged in user: `UserList.Settings.settings_for(watchlist, nil)`. Should we check at least that watchlist id public ?

lib/sanbase/user_lists/settings.ex
IvanIvanoff: Hmm. Or better to rework the migration and indeed have this pair as primary key

lib/sanbase/user_lists/settings.ex
```diff
@@ -0,0 +1,144 @@
...
+  exists:
+  1. User's settings for that watchlist
+  2. The watchlist's creator's settings for that watchlist
+  3. Default settings
+  """
+  @spec settings_for(%UserList{}, %User{} | nil) :: {:ok, WatchlistSettings.t()}
+  def settings_for(%UserList{} = ul, %User{} = user) do
+    settings =
+      get_settings(ul.id, user.id) ||
+        get_settings(ul.id, ul.user_id) ||
```
tspenov: This is another user's watchlist. Shouldn't this check if watchlist is public ?

lib/sanbase/user_lists/settings.ex
tspenov: yes, but looks confusing at first sight. Maybe just place it in the same file but not nested like we did for triggers :?

lib/sanbase_web/graphql/schema/queries/watchlist_queries.ex
IvanIvanoff: :+1:

lib/sanbase/user_lists/settings.ex
IvanIvanoff: `settings` a field in the user_list type and is accessed only if the user is granted access to the watchlist.

lib/sanbase/user_lists/settings.ex
IvanIvanoff: You cannot reach here as you'll be stopped by graphql - settings is being resolved only if you're given access to it

lib/sanbase/user_lists/settings.ex
IvanIvanoff: If not nested the module name will be less readable (UserList.Settings.WatchlistSettings). I can rename it to `Settings.EmbededSchema` for example and separate it?

tspenov: I can live with how it is now.

tspenov:
@IvanIvanoff build is failing.

PR #1291, opened 2019-07-11
Renew cancelled subscription

tspenov:
```graphql
    mutation {
      renewCancelledSubscription(subscriptionId: #{subscription_id}) {
        id 
        currentPeriodEnd
        cancelAtPeriodEnd
        plan {
            name
        }
      }
    }
```

### DIFF
--- a/lib/sanbase/pricing/subscription.ex
+++ b/lib/sanbase/pricing/subscription.ex
@@ -118,7 +118,7 @@ defmodule Sanbase.Pricing.Subscription do
 
   https://stripe.com/docs/billing/subscriptions/canceling-pausing#reactivating-canceled-subscriptions
   """
-  def renew_subscription(subscription) do
+  def renew_cancelled_subscription(subscription) do
     with {:end_period_reached?, :lt} <-
            {:end_period_reached?, DateTime.compare(Timex.now(), subscription.current_period_end)},
          {:ok, stripe_subscription} <-
@@ -130,7 +130,7 @@ defmodule Sanbase.Pricing.Subscription do
            }) do
       {:ok, updated_subscription |> Repo.preload([plan: [:product]], force: true)}
     else
-      {:end_period_reached?, :gt} ->
+      {:end_period_reached?, _} ->
         {:end_period_reached_error,
          "Cancelled subscription has already reached the end period at #{
            subscription.current_period_end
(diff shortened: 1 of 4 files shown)


lib/sanbase/pricing/subscription.ex
```diff
@@ -113,6 +113,34 @@ defmodule Sanbase.Pricing.Subscription do
...
+         {:ok, stripe_subscription} <-
+           StripeApi.update_subscription(subscription.stripe_id, %{cancel_at_period_end: false}),
+         {:ok, updated_subscription} <-
+           update_subscription_db(subscription, %{
+             cancel_at_period_end: false,
+             current_period_end: DateTime.from_unix!(stripe_subscription.current_period_end)
+           }) do
+      {:ok, updated_subscription |> Repo.preload([plan: [:product]], force: true)}
+    else
+      {:end_period_reached?, :gt} ->
```
IvanIvanoff: You're not handling `:eq`. It's extremely unlikely to happen, but you can just leave `_` here instead of `:gt`

lib/sanbase/pricing/subscription.ex
```diff
@@ -113,6 +113,34 @@ defmodule Sanbase.Pricing.Subscription do
     end
   end
 
+  @doc """
+  Renew cancelled subscription if `current_period_end` is not reached.
+
+  https://stripe.com/docs/billing/subscriptions/canceling-pausing#reactivating-canceled-subscriptions
+  """
+  def renew_subscription(subscription) do
```
IvanIvanoff: `renew_cancelled_subscription` :?

lib/sanbase/pricing/subscription.ex
```diff
@@ -113,6 +113,34 @@ defmodule Sanbase.Pricing.Subscription do
     end
   end
 
+  @doc """
+  Renew cancelled subscription if `current_period_end` is not reached.
```
IvanIvanoff: Are we checking somewhere that the subscription that is renewed is indeed cancelled?

tspenov: Yes, also there is a test for that below.

lib/sanbase/pricing/subscription.ex
tspenov: I am 50/50

tspenov: but renamed - sounds more descriptive

PR #1293, opened 2019-07-11
Jenkinsfile improvements

azhelev:
#### Summary
- Generate the pod template from jenkins shared library. The shared lib is [here](https://github.com/santiment/jenkins-libs/blob/master/src/net/santiment/utils/podTemplates.groovy)
- Use the node's docker daemon to do the builds instead of the shared docker-host
- Set up discarding of builds older than 30 days

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


azhelev:
It will discard the old builds as well. The setting takes effect when at leas one build is done with the setting enabled.

PR #1294, opened 2019-07-12
Extend Daily Active Addresses signal to use operation field

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


tspenov:
@IvanIvanoff `percent_threshold` becomes `percent_up`. We have tests only for `percent_up`. I think we should add some very basic tests for other cases too :?

IvanIvanoff:
@tspenov Yes, I've only reworked the tests to fit the new format, now I'll be working on adding more tests (there are currently no tests for absolute values and some of the percentages)

PR #1309, opened 2019-07-19
Add status field. Sync subscriptions on change and on events.

tspenov:
image


lib/sanbase/ecto_enum.ex
```diff
@@ -0,0 +1,15 @@
+import EctoEnum
+
+defenum(ColorEnum, :color, ["none", "blue", "red", "green", "yellow", "grey", "black"])
```
IvanIvanoff: 🤔

tspenov: yes - I think its better to have these just on one place

PR #1311, opened 2019-07-19
Timeframe restriction update

IvanIvanoff:
#### Summary
This PR includes:
- Complete rewrite of TimeframeRestriction middleware. Now it is simplified (no nested with-case-if expressions) and there are more functions, each of them doing less things
- Complete rewrite of how the subscription plan of a given query is determined. We introduce the usage of the `meta` information in every query. As the GraphQL Schema is build compile-time, the `AcecssSeed` module automatically determines all queries and all access restrictions. It builds different sets and provides functions to check the data.
- Rewrite of how the subscription is accessed - it is now determined in the `ContextPlug` because previously it could make 2 DB queries during the execution.
- Rewrite the logic in such a way that communication with the DB is not needed to determine query access. The database data for the metrics in each plan is now not used and could be deprecated in the future.

Additional smaller changes part of this PR:
- Add missing Complexity middleware in some places
- Allow realtime access to GitHub data

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/pricing/subscription.ex
+++ b/lib/sanbase/pricing/subscription.ex
@@ -144,12 +144,12 @@ defmodule Sanbase.Pricing.Subscription do
 
   def update_subscription_db(subscription, params) do
     subscription
-    |> Subscription.changeset(params)
+    |> changeset(params)
     |> Repo.update()
   end
 
   def sync_all() do
-    Subscription
+    __MODULE__
     |> Repo.all()
     |> Enum.each(&sync_with_stripe_subscription/1)
   end
@@ -163,7 +163,7 @@ defmodule Sanbase.Pricing.Subscription do
         },
         db_subscription
       ) do
-    Subscription.update_subscription_db(db_subscription, %{
+    update_subscription_db(db_subscription, %{
       current_period_end: DateTime.from_unix!(current_period_end),
       cancel_at_period_end: cancel_at_period_end,
       status: status,
@@ -171,7 +171,7 @@ defmodule Sanbase.Pricing.Subscription do
     })
   end
 
-  def sync_with_stripe_subscription(%Subscription{stripe_id: stripe_id} = subscription) do
+  def sync_with_stripe_subscription(%__MODULE__{stripe_id: stripe_id} = subscription) do
     StripeApi.retrieve_subscription(stripe_id)
     |> case do
       {:ok,
@@ -181,7 +181,7 @@ defmodule Sanbase.Pricing.Subscription do
          status: status,
          plan: %Stripe.Plan{id: stripe_plan_id}
        }} ->
-        Subscription.update_subscription_db(subscription, %{
+        update_subscription_db(subscription, %{
           current_period_end: DateTime.from_unix!(current_period_end),
           cancel_at_period_end: cancel_at_period_end,
           status: status,
@@ -315,8 +315,8 @@ defmodule Sanbase.Pricing.Subscription do
          user,
          plan
        ) do
-    %Subscription{}
-    |> Subscription.changeset(%{
+    %__MODULE__{}
+    |> changeset(%{
       stripe_id: stripe_id,
       user_id: user.id,
       plan_id: plan.id,
--- a/lib/sanbase/pricing/subscription.ex
+++ b/lib/sanbase/pricing/subscription.ex
@@ -9,7 +9,7 @@ defmodule Sanbase.Pricing.Subscription do
   import Ecto.Changeset
   import Ecto.Query
 
-  alias Sanbase.Pricing.{Plan, Subscription}
+  alias Sanbase.Pricing.Plan
   alias Sanbase.Pricing.Plan.AccessChecker
   alias Sanbase.Auth.User
   alias Sanbase.Repo
@@ -17,11 +17,10 @@ defmodule Sanbase.Pricing.Subscription do
 
   require Logger
 
-  @generic_error_message """
-  Current subscription attempt failed. Please, contact administrator of the site for more information.
-  """
   @percent_discount_1000_san 20
   @percent_discount_200_san 4
+  @generic_error_message "Current subscription attempt failed. Please, contact administrator of the site for more information."
+
   # After `current_period_end` timestamp passes there is some time until `invoice.payment_succeeded` event is generated
   # to update the field with new timestamp value. So we add 1 day gratis in which we should receive payment before
   # we decide that this subscription is not active.
@@ -33,13 +32,14 @@ defmodule Sanbase.Pricing.Subscription do
     field(:current_period_end, :utc_datetime)
     field(:cancel_at_period_end, :boolean, null: false, default: false)
     field(:status, SubscriptionStatusEnum)
+
     belongs_to(:user, User)
     belongs_to(:plan, Plan)
   end
 
   def generic_error_message, do: @generic_error_message
 
-  def changeset(%Subscription{} = subscription, attrs \\ %{}) do
+  def changeset(%__MODULE__{} = subscription, attrs \\ %{}) do
     subscription
     |> cast(attrs, [
       :plan_id,
@@ -51,6 +51,7 @@ defmodule Sanbase.Pricing.Subscription do
     ])
   end
 
+  @spec free_subscription() :: %__MODULE__{}
   def free_subscription() do
     %__MODULE__{plan: Plan.free_plan()}
   end
@@ -230,12 +231,16 @@ defmodule Sanbase.Pricing.Subscription do
   @doc """
   How much historical days a subscription plan can access.
   """
-  def historical_data_in_days(subscription) do
-    subscription.plan.access["historical_data_in_days"]
+  def historical_data_in_days(%__MODULE__{plan: plan}) do
+    plan
+    |> Plan.plan_atom_name()
+    |> AccessChecker.historical_data_in_days()
   end
 
-  def realtime_data_cut_off_in_days(subscription) do
-    subscription.plan.access["realtime_data_cut_off_in_days"]
+  def realtime_data_cut_off_in_days(%__MODULE__{plan: plan}) do
+    plan
+    |> Plan.plan_atom_name()
+    |> AccessChecker.realtime_data_cut_off_in_days()
   end
 
   @doc ~s"""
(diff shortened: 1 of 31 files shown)


lib/sanbase/pricing/subscription.ex
```diff
@@ -181,22 +181,22 @@ defmodule Sanbase.Pricing.Subscription do
     subscription.plan.access["historical_data_in_days"]
   end
 
-  @doc """
-  Checks whether a query is in any plan.
-  """
-  def is_restricted?(query) do
-    query in AccessSeed.all_restricted_metrics()
+  def realtime_data_cut_off_in_days(subscription) do
+    subscription.plan.access["realtime_data_cut_off_in_days"]
```
tspenov: Won't this return `nil` for the free plan  instead of 1 day ?

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -106,7 +109,8 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
           auth: %{
             auth_method: :user_token,
             current_user: current_user,
-            san_balance: san_balance(current_user)
+            san_balance: san_balance(current_user),
+            subscription: Subscription.current_subscription(current_user, @api_product_id)
```
tspenov: should we call the key `subscription_api` because we'll have also for grafana and sheets later ? 
Also we should fetch these with one query but we can update that later. Just something that we should be aware of .

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -7,57 +7,60 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
   """
   @behaviour Absinthe.Middleware
 
+  require SanbaseWeb.Graphql.Schema
+  @mutation_type Absinthe.Schema.lookup_type(SanbaseWeb.Graphql.Schema, :mutation)
+  @mutations_mapset MapSet.new(@mutation_type.fields |> Map.keys())
```
tspenov: Should we extract these to some helper module - cause we'll probably use these on other places ?

lib/sanbase_web/graphql/context_plug.ex
IvanIvanoff: Let's rename once we introduce new subscription

lib/sanbase_web/graphql/middlewares/access_control.ex
IvanIvanoff: I've extracted this.
Another reason to extract is so it can be used in the Schema, so it is applied to all queries. It should be moved because otherwise we'll have circular dependencies

lib/sanbase/pricing/subscription.ex
IvanIvanoff: I guess so.. I can rework it so it does not depend on the DB

tspenov: The other choice is to create a migration to add this field in the free plan map

lib/sanbase_web/graphql/schema/queries/blockchain_queries.ex
```diff
@@ -334,6 +352,8 @@ defmodule SanbaseWeb.Graphql.Schema.BlockchainQueries do
 
     @desc "Fetch a list of all exchange wallets. This query requires basic authentication."
     field :exchange_wallets, list_of(:wallet) do
+      meta(subscription: :restricted)
```
tspenov: Is that continuing to work for basic auth ?

lib/sanbase_web/graphql/schema/queries/tech_indicators_queries.ex
```diff
@@ -45,6 +47,8 @@ defmodule SanbaseWeb.Graphql.Schema.TechIndicatorsQueries do
     * interval - an integer followed by one of: `m`, `h`, `d`, `w`
     """
     field :metric_anomaly, list_of(:anomaly_value) do
+      meta(subscription: :basic)
```
tspenov: shouldn't this be also `free`  - it will stop working on the frontend otherwise

IvanIvanoff: :+1:

lib/sanbase_web/graphql/schema/queries/blockchain_queries.ex
IvanIvanoff: hm.. Should be, because the `AccessControl` will stop based on the check `in premium and not in basic` and this will return `false` (do not restrict)

IvanIvanoff: but I need to add `restricted` to the `all_projects_project_transparency` query as it's also behind BasicAuth

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
+      to = Timex.shift(Timex.now(), days: -2)
+      query = network_growth_query(from, to)
+      result = execute_query(context.conn, query, "networkGrowth")
+
+      assert_called(Sanbase.Clickhouse.NetworkGrowth.network_growth(:_, from, to, :_))
+      assert result != nil
+    end
+
+    test "cannot access PRO metrics for over 3 months", context do
+      from = Timex.shift(Timex.now(), days: -100)
```
kalkov: You can change it to 90 or 91, the one that will forbid access.

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
+      result = execute_query(context.conn, query, "networkGrowth")
 
       refute called(Sanbase.Clickhouse.NetworkGrowth.network_growth(:_, from, to, :_))
       assert result != nil
     end
 
-    test "can access ADVANCED metrics for 3 months", context do
-      from = Timex.shift(Timex.now(), days: -(100 + 1))
+    test "can access BASIC metrics within 90 days and 1 day interval", context do
+      from = Timex.shift(Timex.now(), days: -85)
```
kalkov: I would use 89

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
+      to = Timex.now()
+      query = history_price_query(context.project, from, to)
+      result = execute_query(context.conn, query, "historyPrice")
+
+      assert_called(Sanbase.Prices.Store.fetch_prices_with_resolution(:_, from, to, "1d"))
+      assert result != nil
+    end
+
+    test "cannot access BASIC metrics for over 3 months", context do
+      from = Timex.shift(Timex.now(), days: -100)
```
kalkov: You can change it to 90 or 91, the one that will forbid access.

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
       query = mvrv_query(from, to)
-      result = execute_query(context.conn_apikey_free, query, "mvrvRatio")
+      result = execute_query(context.conn, query, "mvrvRatio")
 
       refute called(Sanbase.Clickhouse.MVRV.mvrv_ratio(:_, from, to, :_))
       assert result != nil
     end
+
+    test "can access ADVANCED withing 90 days and 1 day interval", context do
+      from = Timex.shift(Timex.now(), days: -85)
```
kalkov: Ditto

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
       to = Timex.now()
       query = mvrv_query(from, to)
-      result = execute_query(context.conn_apikey_free, query, "mvrvRatio")
+      result = execute_query(context.conn, query, "mvrvRatio")
 
       refute called(Sanbase.Clickhouse.MVRV.mvrv_ratio(:_, from, to, :_))
       assert result != nil
     end
+
+    test "can access ADVANCED withing 90 days and 1 day interval", context do
```
kalkov: ADVANCED -> PRO

test/sanbase/pricing/access_test.exs
```diff
@@ -10,129 +10,303 @@ defmodule Sanbase.Pricing.AccessTest do
...
+  describe "user with PREMIUM plan" do
+    test "can access FREE metrics for all time", context do
+      insert(:subscription_premium, user: context.user)
+
+      from = Timex.shift(Timex.now(), days: -900)
+      to = Timex.now()
+      query = history_price_query(context.project, from, to)
+      result = execute_query(context.conn, query, "historyPrice")
+
+      assert_called(Sanbase.Prices.Store.fetch_prices_with_resolution(:_, from, to, "1d"))
```
kalkov: Can replace "1d" with "_" for the sake of consistency

lib/sanbase/pricing/subscription.ex
```diff
@@ -9,19 +9,18 @@ defmodule Sanbase.Pricing.Subscription do
...
   alias Sanbase.StripeApi
 
   require Logger
 
-  @generic_error_message """
-  Current subscription attempt failed. Please, contact administrator of the site for more information.
-  """
   @percent_discount_1000_san 20
   @percent_discount_200_san 4
+  @generic_error_message "Current subscription attempt failed. Please, contact administrator of the site for more information."
```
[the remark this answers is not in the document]
IvanIvanoff: Why?

lib/sanbase_web/graphql/middlewares/timeframe_restriction.ex
```diff
@@ -174,19 +201,40 @@ defmodule SanbaseWeb.Graphql.Middlewares.TimeframeRestriction do
...
+        resolution
+
+      _ ->
+        # If we reach here the first time we checked to < from was not true
+        # This means that the middleware rewrote the params in a way that this is
+        # now true. If that happens - both from and to are outside the allowed interval
+        resolution
+        |> Resolution.put_result(
+          {:error,
+           "Both `from` and `to` parameters are outside the allowed intervals you are allowed to query with your current subscription plan."}
```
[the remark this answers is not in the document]
IvanIvanoff: Cannot properly describe the issue in less words

test/sanbase/pricing/access_test.exs
IvanIvanoff: :+1:

PR #1313, opened 2019-07-19
Docker will use diff DBs for postgres and timescale

kalkov:
#### Summary
Running migrations was causing timescale tables to go into
postgres' structure.sql

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -36,7 +36,7 @@ services:
 
     environment:
       DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment
-      TIMESCALE_DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -36,7 +36,7 @@ services:
 
     environment:
       DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment
-      TIMESCALE_DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment
+      TIMESCALE_DATABASE_URL: postgres://postgres:postgres@postgres:5432/santiment_timescale
```
IvanIvanoff: Is this enough? If yes - can we remove the `santiment` db from the `environment`?

kalkov: Yes, it's enough. Removed the environment from `bin/test.sh`

PR #1322, opened 2019-07-23
Handle subscription created event

tspenov:
That is needed for subscriptions created at Stripe dashboard directly.


IvanIvanoff:
I've just merged some more changes that rename files from pricing -> billing (2 PRs in total for this renaming). You might want to rebase to get these changes

lib/sanbase/pricing/stripe_event.ex
```diff
@@ -94,9 +95,43 @@ defmodule Sanbase.Billing.StripeEvent do
...
+       }) do
+    handle_subscription_created(id, "customer.subscription.created", subscription_id)
+  end
+
+  defp handle_subscription_created(id, type, subscription_id) do
+    with {:ok, stripe_subscription} <- StripeApi.retrieve_subscription(subscription_id),
+         {:user_nil?, %User{} = user} <-
+           {:user_nil?, Repo.get_by(User, stripe_customer_id: stripe_subscription.customer)},
+         {:plan_nil?, %Plan{} = plan} <-
+           {:plan_nil?, Plan.by_stripe_id(stripe_subscription.plan.id)},
```
IvanIvanoff: Just a note: Such `plan_nil?` check must be added in other places. We have places where we call `Plan.by_stripe_id(id).id` which fails as it executes `nil.id` in some cases

lib/sanbase/pricing/subscription.ex
```diff
@@ -145,6 +145,28 @@ defmodule Sanbase.Billing.Subscription do
...
+    %__MODULE__{}
+    |> changeset(%{
+      stripe_id: stripe_id,
+      user_id: user.id,
+      plan_id: plan.id,
+      current_period_end: DateTime.from_unix!(current_period_end),
+      cancel_at_period_end: cancel_at_period_end,
+      status: status
+    })
+    |> Repo.insert(on_conflict: :nothing)
```
IvanIvanoff: why `:nothing` and not `:replace_all`?

IvanIvanoff:
@tspenov do we need some tests to cover this feature?

lib/sanbase/pricing/subscription.ex
tspenov: I don't think an update is needed here. These will happen cause subscription is inserted synchronously but I am also handling `subscription.created` event which will come with same data. So doing `:nothing` seems more appropriate here.

lib/sanbase/pricing/stripe_event.ex
tspenov: I saw this. I am almost sure missing plans come from mixed events from stage and localhost testing so these won't happen to prod. Though it is good these nonetheless.

PR #1327, opened 2019-07-24
Implement custom acccess restrictions for some queries

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/billing/plan/access_checker.ex
+++ b/lib/sanbase/billing/plan/access_checker.ex
@@ -79,7 +79,7 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
   if @custom_acess_queries |> Enum.sort() != custom_access_meta |> Enum.sort() do
     IO.warn("""
     The list of GraphQL queries with special access defined in the CustomAccess
-    module and withsubscription meta field `:custom_access` is not the same.
+    module and with subscription meta field `:custom_access` is not the same.
     """)
   end
 
--- a/lib/sanbase/billing/plan/access_checker.ex
+++ b/lib/sanbase/billing/plan/access_checker.ex
@@ -22,6 +22,8 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
   query without a subscription plan defined
   """
 
+  alias Sanbase.Billing.Plan.CustomAccess
+
   defmodule Helper do
     @moduledoc ~s"""
     Contains a single function `get_metrics_with_subscription_plan/1` that examines
@@ -66,6 +68,22 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
       Queries without subscription plan: #{inspect(queries)}
       """)
   end
+
+  @custom_acess_queries_stats CustomAccess.get()
+  @custom_acess_queries @custom_acess_queries_stats |> Map.keys()
+
+  # Emit a compile time warning if there are queries with custom access logic
+  # but the meta field `subscription` is not correct
+  custom_access_meta = Helper.get_metrics_with_subscription_plan(:custom_access)
+
+  if @custom_acess_queries |> Enum.sort() != custom_access_meta |> Enum.sort() do
+    IO.warn("""
+    The list of GraphQL queries with special access defined in the CustomAccess
+    module and withsubscription meta field `:custom_access` is not the same.
+    """)
+  end
+
+  def mutations_mapset(), do: Helper.mutations_mapset()
 
   # Below are defined lists and sets (for fast member check) of all metrics
   # available in a given plan. Each next plan contains all the metrics from theh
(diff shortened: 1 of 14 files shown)


IvanIvanoff:
@tspenov @kalkov My idea is to be able to add one record to the module `CustomAccess` and to change the `subscription` meta key to `:custom_access`. Do you think having to support both of them is a good idea? To me something is better if you need to touch only 1 place to change it.

On the other hand, I've added compile time warnings that will tell you that the meta field is not right if there is data in the CustomAccess module

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -67,6 +69,22 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
   end
 
+  @custom_acess_queries_stats CustomAccess.get()
+  @custom_acess_queries @custom_acess_queries_stats |> Map.keys()
+
+  # Emit a compile time warning if there are queries with custom access logic
+  # but the meta field `subscription` is not correct
+  custom_access_meta = Helper.get_metrics_with_subscription_plan(:custom_access)
+
+  if @custom_acess_queries |> Enum.sort() != custom_access_meta |> Enum.sort() do
```
IvanIvanoff: image
We should be aware of this warning after this gets merged

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -135,13 +156,20 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
+    |> Map.get(:realtime_data_cut_off_in_days)
+  end
+
+  def realtime_data_cut_off_in_days(plan, _query) do
     case plan do
-      :free -> @free_plan_stats[:realtime_data_cut_off_in_days] || 0
-      :basic -> @basic_plan_stats[:realtime_data_cut_off_in_days] || 0
-      :pro -> @pro_plan_stats[:realtime_data_cut_off_in_days] || 0
-      :premium -> @premium_plan_stats[:realtime_data_cut_off_in_days] || 0
-      :custom -> @premium_plan_stats[:realtime_data_cut_off_in_days] || 0
```
IvanIvanoff: We do not need to make it 0 in case it's missing because `nil` is handled as `0`.
If we have kept this we should have added this to a lot more places + `historical_data_in_days` does not have a meaningful integer default value

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -150,21 +178,30 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
   A query can be restricted but still accessible by not-paid users or users with
   lower plans. In this case historical and/or realtime data access can be cut off
   """
-  def is_restricted?(query) do
-    not (query in @free_metrics_mapset)
-  end
+  def is_restricted?(query) when query in @custom_acess_queries, do: true
```
IvanIvanoff: If we do not have custom historical/realtime access changes then this query won't be made custom. Because of that all such queries must go through the restriction logic

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -150,21 +178,30 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
+  def is_restricted?(query), do: not (query in @free_metrics_mapset)
 
   @doc ~s"""
   Check if a query access is given only to users with an advanced plan
   (pro or higher). No access is given to users with lower plans
   """
+  def needs_advanced_plan?(query) when query in @custom_acess_queries do
+    %{accessible_by_plan: [lowest_plan | _]} = Map.get(@custom_acess_queries_stats, query)
```
IvanIvanoff: The list of plans is sorted in ascending order, so the lowest is guaranteed to be first.

lib/sanbase/billing/plan/access_checker.ex
tspenov: we can add the diff between both so we can see which one is missing :?

lib/sanbase/billing/plan/custom_access.ex
```diff
@@ -0,0 +1,75 @@
...
+    the `historical_data_in_days` and/or `realtime_data_cut_off_in_days` keys.
+    If a plan is missing it means that it has no restrictions. If a field in a plan
+    is missing it means that it is not restricted
+  """
+
+  import Sanbase.Billing.Plan, only: [sort_plans: 1]
+  Module.register_attribute(__MODULE__, :metric, accumulate: true)
+
+  @metric %{
+    metric_name: [:mvrv_ratio, :realized_value],
```
tspenov: we can skip this key if we instead of `meta(subscription: :custom_access)` just name the 2 cases differently somehow `meta(subscription: :standart_full_except_last_30d)`,  `meta(subscription: :advanced_full_except_last_30d)`

lib/sanbase/billing/plan/access_checker.ex
IvanIvanoff: image

Here's an example of the update - if you have added history_price in one place and ohlc in the other place

lib/sanbase/billing/plan/custom_access.ex
IvanIvanoff: Currently I've done it so:
- You must add `custom_access`
- You must add a record for it in the `CustomAccess`
- If something is not done correctly - compile time error.

The broad idea is in the future to be able to change these settings via JSON config so it can be edited by the FE team. If we want this, the whole config must be in this map (that can be generated from a json file) and not in the name of the field in GraphQL

lib/sanbase/billing/plan/access_checker.ex
tspenov: niice

lib/sanbase_web/graphql/schema/queries/social_data_queries.ex
```diff
@@ -229,8 +229,7 @@ defmodule SanbaseWeb.Graphql.Schema.SocialDataQueries do
     end
 
     @desc ~s"""
-    Returns the % of the social dominance a given project has over time in a given social channel.
-
+    c
```
tspenov: accidentally deleted ?

PR #1329, opened 2019-07-25
Optimize influx tests

kalkov:
#### Summary
There was a flaky test - `test/sanbase/external_services/coinmarketcap/ticker_fetcher_test.exs` which was failing due to timeouts. 

Dropping a single measurement is a [slow operation](https://github.com/influxdata/influxdb/issues/10306), so first I tried dropping all series but that resulted in inconsistent results - random timeouts or very slow execution of the drop.

So it seems the most performant way  is to re-create the DB on every test that uses influxdb.

Test suite execution has been reduced from 100-150s to 75s on my machine and is much more consistent.

Before
Screenshot from 2019-07-25 11-47-32

After
Screenshot from 2019-07-25 11-53-21

### DIFF
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -21,4 +21,4 @@ RUN mix deps.compile
 COPY . /app
 RUN mix format --check-formatted
 
-CMD mix test_all --trace
+CMD mix test_all --trace --slowest 20
--- a/test/sanbase/external_services/coinmarketcap/graph_data_test.exs
+++ b/test/sanbase/external_services/coinmarketcap/graph_data_test.exs
@@ -5,6 +5,8 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphDataTest do
   alias Sanbase.ExternalServices.Coinmarketcap.PricePoint2, as: PricePoint
   alias Sanbase.Prices.Store
 
+  import Sanbase.InfluxdbHelpers
+
   @total_market_measurement "TOTAL_MARKET_total-market"
 
   test "fetching the first price datetime of a token" do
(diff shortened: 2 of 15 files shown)


Dockerfile-test
```diff
@@ -21,4 +21,4 @@ RUN mix deps.compile
 COPY . /app
 RUN mix format --check-formatted
 
-CMD mix test_all --trace
```
IvanIvanoff: leftover

test/sanbase/external_services/coinmarketcap/graph_data_test.exs
```diff
@@ -88,8 +88,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphDataTest do
     end)
 
     Store.create_db()
-    measurement_name = "TOTAL_MARKET_total-market"
-    Store.drop_measurement(measurement_name)
+    Store.drop_series()
```
IvanIvanoff: This will force every test that touches influxdb to be synchronous.

kalkov: The code will not stay as it is now, the proper place to be executed is in after test callback.
But why `drop_measurement` doesn't force to be sync and `drop_series` does?
BTW all the tests that I have touched are sync except `ticker_fetcher_test.exs`

Dockerfile-test
kalkov: It is on purpose, will display 20 slowest tests in Jenkins.

test/sanbase/external_services/coinmarketcap/graph_data_test.exs
IvanIvanoff: with `drop_measurement` you're dropping only the measurement you're working with. With `drop_series` you're dropping everything - if 2 tests run in parallel and one drops the series, it will affect the other test

PR #1332, opened 2019-07-25
Daa signal historical backtest

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


IvanIvanoff:
@GarageInc adding you so you can follow when this gets merged

lib/sanbase/signals/history/daily_active_addresses_history.ex
```diff
@@ -24,105 +17,120 @@ defmodule Sanbase.Signal.History.DailyActiveAddressesHistory do
...
+              false ->
+                {[Map.put(elem, :triggered?, false) | acc], 0}
+            end
 
-      {from, to, interval}
+          elem, {acc, cooldown_left} ->
+            {[Map.put(elem, :triggered?, false) | acc], cooldown_left - 1}
+        end)
+
+      {:ok, result |> Enum.reverse()}
```
tspenov: what keys are inside each map in result (`daily_active_addresses, percent_change, triggered?`) ? do we need to put the `price` also or frontend will get it with another query ?

IvanIvanoff: We're not putting the price in the other signals, so I've deprecated it here. Putting the price was only increasing the complexity with no real benefits.

Moreover, we recently changed from average to closing price in our charts - if we provide price here it's one more place where we need to take care of

PR #1335, opened 2019-07-26
Fix eth_wallet signal backtest api

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/history/eth_wallet_trigger_history.ex
```diff
@@ -13,25 +13,37 @@ defmodule Sanbase.Signal.History.EthWalletTriggerHistory do
   require Logger
 
   @historical_days_from 365
-  @historical_interval "1d"
+  @historical_interval "1h"
```
IvanIvanoff: This was not caught by the test because the result was mocked

PR #1336, opened 2019-07-26
Add sansheets plans

tspenov:


priv/repo/migrations/20190726123917_sync_new_plans_in_stripe.exs
```diff
@@ -0,0 +1,35 @@
+defmodule Sanbase.Repo.Migrations.SyncNewPlansInStripe do
```
kalkov: Isn't it better to be a one-timer task, we won't need that migration in the future.

tspenov: The migration is also a one-time task (because of the versioning). Probably you talk about something like rake task (mix tasks are the equivalent) where you log in to prod server and run or schedule with a cron job to run once but I don't really see the benefits than using the existing migrations (except that we can run whenever we want - not on startup and reuse if we want to run later again without copying the code in new migration)
This approach has the upside that the code will be executed before the system starts - so we won't forget it and the user won't use the system with plans that doesn't exist in Stripe and it uses the existing infrastructure.

kalkov: Yes, meant mix task. 

The biggest downside of executing code from our codebase in migrations is that when this code is refactored chances are high that it will be left in broken state in the migrations until somebody runs them while setting up the project. 

Also this code should be executed in staging/production but it can be skipped in development, right?

tspenov: For setting up the project it is not needed to run the migrations - just `ecto.load` the schema from `structure.sql` which will load the actual schema of the db. (and the migrations can fail for various reasons so it is not good idea to run the whole set in order to setup the db)
It skips running the migration if you don't have Stripe api key setup so if not configured in .env or dev.secret.exs it will skip it.

kalkov: Here is the article that changed my mind about migrations: https://thoughtbot.com/blog/data-migrations-in-rails

IvanIvanoff: We already have migrations that cannot pass if you try to rerun them, but that's ok as rerunning migrations is never done

tspenov: I certainly agree on the third point for long running migrations (also we have some healthchecks that require the system to start in few minutes so these might break the deployment) but it is not the case here. About the first 2 points as @IvanIvanoff pointed we already have migrations that will fail if rerun and we use the approach with loading the schema form `structure.sql`.

kalkov: My point was not about rerunning migrations but running them. I had an issue last week and it was not fun. Doesn't make sense to me to keep dead code in migrations.

PR #1337, opened 2019-07-26
NOT FOR MERGING: Add inspect to see what is slow in CI test runs

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


IvanIvanoff:
The issue was found:
In tests we're using the same rate limiting rules even though there are no actual calls but only mocks. A particular rate limiter was allowing 5 requests per minute and started to get triggered quite often. Issue is fixed in #1340

PR #1345, opened 2019-07-29
Add access control sanbase plans

tspenov:


lib/sanbase/billing/plan/api_access_checker.ex
```diff
@@ -143,6 +143,8 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
     metrics: @premium_plan_stats
   }
 
+  def all_metrics(), do: @premium_metrics
```
IvanIvanoff: these are not all metrics

IvanIvanoff: the `custom_access` ones are missing. The correct would be to implement this function in the helper module and do not do any filtering

lib/sanbase/billing/plan.ex
```diff
@@ -78,7 +78,7 @@ defmodule Sanbase.Billing.Plan do
   Fetching only with `interval`=`month` because we have yearly metrics with the same metrics
   and names.
   """
-  defdelegate lowest_plan_with_metric(query), to: AccessChecker
+  defdelegate lowest_plan_with_metric(query), to: ApiAccessChecker
```
IvanIvanoff: I think we should have `AccessChecker` module that dispatches to the correct one (Api*, Sanbase*, etc.). Changing to `ApiAccessChecker` here and in the timeframe restriction is not really correct because sanbase and sansheets will go through the same path

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Sanbase.Billing.Plan.SanbaseAccessChecker do
+  alias Sanbase.Billing.Plan.ApiAccessChecker
+
+  @all_metrics ApiAccessChecker.all_metrics()
+  @bundle_metrics [
```
IvanIvanoff: We should not manage this list by hand - if you need this list you should use`CustomAccess.get |> Enum.map(& &1.metric_name)`

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
...
+    :mvrv_ratio,
+    :realized_value,
+    :nvt_ratio,
+    :token_age_consumed,
+    :burn_rate
+  ]
+
+  @free_plan_stats %{
+    historical_data_in_days: 2 * 365,
+    realtime_data_cut_off_in_days: 30,
```
IvanIvanoff: everything is cut off ?

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
...
+    metrics: @all_metrics,
+    signals_limit: 10,
```
IvanIvanoff: Let's make this a map so it's easily extendable.
```elixir
signals: %{
  limit: 10
  ...other keys
}
```

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
...
+    external_data_providers: false,
+    sangraphs_access: false
+  }
+
+  @basic_plan_stats %{
+    historical_data_in_days: 2 * 365,
+    realtime_data_cut_off_in_days: 7,
+    metrics: @all_metrics,
```
IvanIvanoff: This will introduce a discrepancy in how we define the plans. Some of them are defined in the subscription meta field and some are described here.

IvanIvanoff: Moreover, now we'll have `:pro` subscription in the meta field while we have it here in `basic` and `free`

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
...
+    signals_limit: 10,
+    external_data_providers: false,
+    sangraphs_access: true
+  }
+
+  @pro_plan_stats %{
+    historical_data_in_days: 3 * 365,
+    realtime_data_cut_off_in_days: 0,
+    metrics: @all_metrics,
+    external_data_providers: true,
```
IvanIvanoff: document this flag

lib/sanbase/billing/subscription.ex
```diff
@@ -259,27 +259,27 @@ defmodule Sanbase.Billing.Subscription do
...
     |> Plan.plan_atom_name()
-    |> AccessChecker.historical_data_in_days(query)
+    |> ApiAccessChecker.historical_data_in_days(query)
   end
 
   def realtime_data_cut_off_in_days(%__MODULE__{plan: plan}, query) do
     plan
     |> Plan.plan_atom_name()
-    |> AccessChecker.realtime_data_cut_off_in_days(query)
+    |> ApiAccessChecker.realtime_data_cut_off_in_days(query)
```
IvanIvanoff: We should still use `AccessChecker` but pass an additional param `product_id` so `AccessChecker` can properly dispatch to `ApiAccessChecker`. 
Also not providing direct access to `ApiAccessChecker` will let us modify it much easier

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,45 @@
...
+    sangraphs_access: true
+  }
+
+  @enterprise_plan_stats @pro_plan_stats
+
+  def free(), do: @free_plan_stats
+  def basic(), do: @basic_plan_stats
+  def pro(), do: @pro_plan_stats
+  def enterprise(), do: @enterprise_plan_stats
```
IvanIvanoff: This is called `custom()` n the api module. Let's use the same naming everywhere (I prefer enterprise)

IvanIvanoff:
@tspenov there's one place in `complexity.ex` where we need to change `:custom` to `:enterprise`

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -110,8 +115,9 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
             auth_method: :user_token,
             current_user: current_user,
             san_balance: san_balance(current_user),
-            subscription: Subscription.current_subscription(current_user, @api_product_id)
-          }
+            subscription: Subscription.current_subscription(current_user, @product_sanbase)
```
IvanIvanoff: As a temporary solution I suggest we add here one `or` - if user has API but no sanbase plan - use api plan.

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -131,8 +137,9 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
           auth_method: :user_token,
           current_user: current_user,
           san_balance: san_balance(current_user),
-          subscription: Subscription.current_subscription(current_user, @api_product_id)
-        }
+          subscription: Subscription.current_subscription(current_user, @product_sanbase)
```
IvanIvanoff: As a temporary solution I suggest we add here one `or` - if user has API but no sanbase plan - use api plan.

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -171,8 +179,9 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
           current_user: current_user,
           token: token,
           san_balance: san_balance(current_user),
-          subscription: Subscription.current_subscription(current_user, @api_product_id)
-        }
+          subscription: Subscription.current_subscription(current_user, @product_api)
```
IvanIvanoff: Though we should not fallback to sanbase if api plan is missing, because API plan is more expensive

test/sanbase_web/graphql/blockchain/etherbi_trx_volume_api_test.exs
```diff
@@ -110,6 +110,7 @@ defmodule Sanbase.Etherbi.TransactionVolumeApiTest do
     result =
       context.conn
       |> post("/graphql", query_skeleton(query, "transactionVolume"))
+      |> IO.inspect()
```
IvanIvanoff: remove inspect

lib/sanbase/billing/plan/custom_access.ex
```diff
@@ -27,7 +27,7 @@ defmodule Sanbase.Billing.Plan.CustomAccess do
 
   @metric %{
     metric_name: [:mvrv_ratio, :realized_value],
-    accessible_by_plan: [:free, :basic, :pro, :premium, :custom] |> sort_plans(),
+    accessible_by_plan: [:free, :basic, :pro, :premium, :enterprise] |> sort_plans(),
```
IvanIvanoff: If we remove the AccessContol middleware then this one is not needed, too.

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,72 @@
...
+    external_data_providers: false,
+    sangraphs_access: false
+  }
+
+  @basic_plan_stats %{
+    historical_data_in_days: 2 * 365,
+    realtime_data_cut_off_in_days: 7,
+    signals_limit: 10,
+    external_data_providers: false,
+    sangraphs_access: true
```
IvanIvanoff: why? Won't sangraphs be different product? Also, if this is needed let's add in in the PR that introduces the sangraphs product and plans

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -0,0 +1,72 @@
...
+  def basic(), do: @basic_plan_stats
+  def pro(), do: @pro_plan_stats
+  def enterprise(), do: @enterprise_plan_stats
+
+  def historical_data_in_days(plan, query) when query in @custom_access_queries do
+    Map.get(@custom_access_queries_stats, query)
+    |> Map.get(:plan_access)
+    |> Map.get(plan, %{})
+    |> Map.get(:historical_data_in_days)
+  end
```
IvanIvanoff: I've been thinking how to rework this. One possible idea is to:
```elixir
case Map.get(@custom_access_queries_stats, query) do
 %{plan_access: %{^plan => %{historical_data_in_days: value}}} -> value
_ -> nil
```

lib/sanbase/billing/plan/sanbase_access_checker.ex
tspenov: It is a requirement for the sanbase plans - you can check the card or the google doc. We're not checking anything, it is staying as kinda config now.

lib/sanbase/billing/plan/sanbase_access_checker.ex
tspenov:
```elixir
Map.get(@custom_access_queries_stats, query)
|> get_in([:plan_access, plan, :historical_data_in_days])
```
or 
```elixir
get_in(@custom_access_queries_stats, [query |> String.to_atom(), :plan_access, plan, :historical_data_in_days])
```
:?

tspenov: reworked with the first option

lib/sanbase/billing/plan/sanbase_access_checker.ex
tspenov: I have no further info than what is suggested by the flag name. It is a WIP from product side.

PR #1350, opened 2019-07-30
Add new automated statistics

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -73,7 +73,31 @@ defmodule Sanbase.Auth.Statistics do
 
   def newsletter_subscribed_new_users(subscription_type, datetime) do
     user_settings_with_newsletter_subscription_query(subscription_type)
-    |> where([u], u.inserted_at < ^datetime)
+    |> join(:inner, [u], us in assoc(u, :user))
+    |> where([_us, u], u.inserted_at < ^datetime)
+    |> count()
+    |> Repo.one()
+  end
+
+  def newsletter_subscribed_old_users(
+        subscription_type,
+        registered_datetime,
+        subscription_datetime
+      ) do
+    user_settings_with_newsletter_subscription_query(subscription_type)
+    |> where(
+      fragment(
+        """
+        NOT settings->>'newsletter_subscription_updated_at_unix' IS NULL AND
+        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC >= ? AND
+        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC <= ?
+        """,
+        ^DateTime.to_unix(subscription_datetime),
+        ^DateTime.to_unix(Timex.now())
+      )
+    )
+    |> join(:inner, [u], us in assoc(u, :user))
+    |> where([_us, u], u.inserted_at < ^registered_datetime)
     |> count()
     |> Repo.one()
   end
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -71,6 +71,13 @@ defmodule Sanbase.Auth.Statistics do
     |> Repo.one()
   end
 
+  def newsletter_subscribed_new_users(subscription_type, datetime) do
+    user_settings_with_newsletter_subscription_query(subscription_type)
+    |> where([u], u.inserted_at < ^datetime)
+    |> count()
+    |> Repo.one()
+  end
+
   def newsletter_subscribed_users(subscription_type, from, to) do
     user_settings_with_newsletter_subscription_query(subscription_type)
     |> where(
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -59,18 +59,18 @@ defmodule Sanbase.Auth.Statistics do
     |> Repo.one()
   end
 
-  def newsletter_subscribed_user_count(subscription_type) do
-    user_settings_with_newsletter_subscription_query(subscription_type)
-    |> count()
-    |> Repo.one()
-  end
-
+  @doc ~s"""
+  Return the number of registered users, which use a given subscription type
+  """
   def newsletter_subscribed_users(subscription_type) do
     user_settings_with_newsletter_subscription_query(subscription_type)
     |> count()
     |> Repo.one()
   end
 
+  @doc ~s"""
+  Return the number of newly-registered users, which use a given subscription type
+  """
   def newsletter_subscribed_new_users(subscription_type, datetime) do
     user_settings_with_newsletter_subscription_query(subscription_type)
     |> join(:inner, [u], us in assoc(u, :user))
@@ -79,22 +79,18 @@ defmodule Sanbase.Auth.Statistics do
     |> Repo.one()
   end
 
+  @doc ~s"""
+  Return the number of old registered users, which have subscribed in the given time period
+  """
   def newsletter_subscribed_old_users(
         subscription_type,
         registered_datetime,
         subscription_datetime
       ) do
-    user_settings_with_newsletter_subscription_query(subscription_type)
-    |> where(
-      fragment(
-        """
-        NOT settings->>'newsletter_subscription_updated_at_unix' IS NULL AND
-        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC >= ? AND
-        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC <= ?
-        """,
-        ^DateTime.to_unix(subscription_datetime),
-        ^DateTime.to_unix(Timex.now())
-      )
+    user_settings_newsletter_subscription_added(
+      subscription_type,
+      subscription_datetime,
+      Timex.now()
     )
     |> join(:inner, [u], us in assoc(u, :user))
     |> where([_us, u], u.inserted_at < ^registered_datetime)
@@ -104,17 +100,8 @@ defmodule Sanbase.Auth.Statistics do
 
   def newsletter_subscribed_users(subscription_type, from, to) do
     user_settings_with_newsletter_subscription_query(subscription_type)
-    |> where(
-      fragment(
-        """
-        NOT settings->>'newsletter_subscription_updated_at_unix' IS NULL AND
-        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC >= ? AND
-        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC <= ?
-        """,
-        ^DateTime.to_unix(from),
-        ^DateTime.to_unix(to)
-      )
-    )
+
+    user_settings_newsletter_subscription_added(subscription_type, from, to)
     |> count()
     |> Repo.one()
   end
@@ -130,6 +117,21 @@ defmodule Sanbase.Auth.Statistics do
           """,
           ^subscription_type
         )
+    )
+  end
+
+  defp user_settings_newsletter_subscription_added(subscription_type, from, to) do
+    user_settings_with_newsletter_subscription_query(subscription_type)
+    |> where(
+      fragment(
+        """
+        NOT settings->>'newsletter_subscription_updated_at_unix' IS NULL AND
+        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC >= ? AND
+        (settings->>'newsletter_subscription_updated_at_unix')::NUMERIC <= ?
+        """,
+        ^DateTime.to_unix(from),
+        ^DateTime.to_unix(to)
+      )
     )
   end
 
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -69,39 +69,42 @@ defmodule Sanbase.Auth.Statistics do
   end
 
   @doc ~s"""
-  Return the number of newly-registered users, which use a given subscription type
+  Return the number of users registered after a given datetime, which use a given subscription type
   """
-  def newsletter_subscribed_new_users(subscription_type, datetime) do
+  def newsletter_subscribed_new_users(subscription_type, %DateTime{} = registered) do
     user_settings_with_newsletter_subscription_query(subscription_type)
     |> join(:inner, [u], us in assoc(u, :user))
-    |> where([_us, u], u.inserted_at < ^datetime)
+    |> where([_us, u], u.inserted_at > ^registered)
     |> count()
     |> Repo.one()
   end
 
   @doc ~s"""
-  Return the number of old registered users, which have subscribed in the given time period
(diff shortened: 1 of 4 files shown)


lib/sanbase/auth/statistics.ex
```diff
@@ -71,6 +71,37 @@ defmodule Sanbase.Auth.Statistics do
     |> Repo.one()
   end
 
+  def newsletter_subscribed_new_users(subscription_type, datetime) do
```
IvanIvanoff: Add some documentation to the functions. Search for `@doc` in the codebase to see how

lib/sanbase/auth/statistics.ex
```diff
@@ -71,6 +71,37 @@ defmodule Sanbase.Auth.Statistics do
...
+    |> Repo.one()
+  end
+
+  def newsletter_subscribed_old_users(
+        subscription_type,
+        registered_datetime,
+        subscription_datetime
+      ) do
+    user_settings_with_newsletter_subscription_query(subscription_type)
+    |> where(
```
IvanIvanoff: this `where` fragment is repeated, you should try to abstract it

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -143,8 +173,18 @@ defmodule Sanbase.Statistics do
...
     overall = WatchlistStatistics.watchlists_created(@epoch_datetime, now)
+
     users_with_watchlist_count = WatchlistStatistics.users_with_watchlist_count()
 
+    new_users_with_watchlist_count_14d =
+      WatchlistStatistics.new_users_with_watchlist_count(Timex.shift(now, days: -14))
+
+    old_users_with_watchlist_count_14d =
+      WatchlistStatistics.old_users_with_new_watchlist_count(
+        Timex.shift(now, days: -14),
```
kalkov: Should `now` be `Timex.now()`?

lib/sanbase/auth/statistics.ex
spiderjako: Is it okay now? Added the documentation and the abstraction

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: Yes, it should, fixed it

IvanIvanoff: `now` is defined on the top of the function.. let's make it consistent everywhere (either define `now` once and use it or always call Timex.now)

spiderjako: Did it with 'now', it should be okay now

lib/sanbase/auth/statistics.ex
spiderjako: Did it, thanks

lib/sanbase/auth/statistics.ex
```diff
@@ -59,31 +59,53 @@ defmodule Sanbase.Auth.Statistics do
...
+  """
+  def newsletter_subscribed_users(subscription_type) do
     user_settings_with_newsletter_subscription_query(subscription_type)
     |> count()
     |> Repo.one()
   end
 
-  def newsletter_subscribed_users(subscription_type) do
+  @doc ~s"""
+  Return the number of newly-registered users, which use a given subscription type
```
IvanIvanoff: Replace `newly-registered` and `old registered` with `users registered before/after a given datetime`

lib/sanbase/auth/statistics.ex
```diff
@@ -102,6 +124,21 @@ defmodule Sanbase.Auth.Statistics do
     )
   end
 
+  defp user_settings_newsletter_subscription_added(subscription_type, from, to) do
+    user_settings_with_newsletter_subscription_query(subscription_type)
+    |> where(
```
IvanIvanoff: let's have only the `where` abstracted. In the end it will look like
```
user_settings_with_newsletter_subscription_query(subscription_type)
|> newsletter_updated_filter_query(...)
|> count()
...
```

IvanIvanoff: Also, we kind of have a convention that if the function returns a query it ends with `_query`.

lib/sanbase/auth/statistics.ex
spiderjako: It should be okay now

lib/sanbase/auth/statistics.ex
spiderjako: Did as you've told, now it is only with where abstracted, also changed the name with the _query in the back.

lib/sanbase/auth/statistics.ex
```diff
@@ -59,31 +59,52 @@ defmodule Sanbase.Auth.Statistics do
...
     |> Repo.one()
   end
 
-  def newsletter_subscribed_users(subscription_type, from, to) do
+  @doc ~s"""
+  Return the number of users registered before a given datetime, which have subscribed in the given time period
+  """
+  def newsletter_subscribed_old_users(
+        subscription_type,
+        registered_datetime,
```
IvanIvanoff: not necessary change: you can rename this to `registered_before_datetime`

spiderjako: That seems clever, thanks. Will use that, I see that there are such functions in the file as well

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -94,11 +94,26 @@ defmodule Sanbase.Statistics do
     overall =
       UserStatistics.newsletter_subscribed_users(Sanbase.Auth.Settings.daily_subscription_type())
 
+    newsletter_subscribed_new_users_14d =
+      UserStatistics.newsletter_subscribed_new_users(
+        Sanbase.Auth.Settings.daily_subscription_type(),
```
IvanIvanoff: these 2 seem to be used in a lot of places, you can place on the top of the file:
```elixir
import Sanbase.Auth.Settings, only: [daily_subscription_type: 0, weekly_subscription_type: 0]
```

IvanIvanoff: Usually we order `use -> import -> alias -> require`

IvanIvanoff: After importing change all occurrences in the file, not only your changes

lib/sanbase/user_lists/statistics.ex
```diff
@@ -29,6 +31,30 @@ defmodule Sanbase.UserLists.Statistics do
     |> Repo.all()
   end
 
+  def new_users_with_watchlist_count(from_datetime) do
+    from(
+      ul in UserList,
+      left_join: user in User,
```
IvanIvanoff: `inner_join` is the correct join here

lib/sanbase/user_lists/statistics.ex
```diff
@@ -29,6 +31,30 @@ defmodule Sanbase.UserLists.Statistics do
...
+      where: user.inserted_at > ^from_datetime,
+      select: count(fragment("DISTINCT ?", ul.user_id))
+    )
+    |> Repo.one()
+  end
+
+  def old_users_with_new_watchlist_count(registered_datetime, watchlist_datetime) do
+    from(
+      ul in UserList,
+      left_join: user in User,
```
IvanIvanoff: `inner_join`

test/sanbase/statistics/statistics_test.exs
```diff
@@ -129,7 +133,67 @@ defmodule Sanbase.StatisticsTest do
...
               "watchlist_created_last_30d" => 5,
               "watchlist_created_last_7d" => 5,
-              "watchlist_created_overall" => 5
+              "watchlist_created_overall" => 5,
+              "new_users_with_watchlist_count_14d" => 2,
+              "old_users_with_watchlist_count_14d" => 2
             }} in statistics
   end
+
+  test "newsletter_subscribed_users function" do
```
kalkov: There are a couple of things that are not right with these tests, most of them are because of the shared setup: 
* reading the diff makes you wonder why the asserted result is let's say 1, you don't see any user creation
* you need to constantly scroll in order to understand the test
* probably not all users are needed in every test case, we add confusion to the reader and loose machine performance 

I would create just the minimal needed users + one extra that will not be included in the result in every single test. Yes there will be some repetition, but tests will be easier to read and much faster to execute.

Also here we have two db operations: insert and update which is expensive and not needed
```elixir
u1 = insert(:user)
Ecto.Changeset.change(u1, %{inserted_at: Timex.shift(Timex.now(), days: -600)})
|> Sanbase.Repo.update()
``` 
I may be wrong but adding an `IO.inspect(u1.inserted_at)` prints `~U[2019-08-02 08:52:44Z]` 

Instead we can do:
```elixir
user1 = insert(:user, inserted_at: Timex.shift(Timex.now(), days: -600))
```

test/sanbase/statistics/statistics_test.exs
```diff
@@ -129,7 +133,67 @@ defmodule Sanbase.StatisticsTest do
...
+
+    test_result =
+      Sanbase.Auth.Statistics.newsletter_subscribed_users(
+        Sanbase.Auth.Settings.weekly_subscription_type()
+      )
+
+    assert test_result == 3
+  end
+
+  test "newsletter_subscribed_new_users function" do
```
kalkov: Here a better name would describe what actually the test does, like: 
"returns the number of newly subscribed newsletter users"

lib/sanbase/user_lists/statistics.ex
spiderjako: This the changes that you requested, it should be okay now

test/sanbase/statistics/statistics_test.exs
IvanIvanoff: IMO the setup block is introduced to reduce this repetition. It's name and purpose is so you know that this is exactly what setups your test, so to me it's more readable.

With you suggestion the setup block can be rewritten to:
```elixir
  alias Sanbase.Auth.UserSettings

  setup do
    u1 = insert(:user, %{inserted_at: Timex.shift(Timex.now(), days: -600)})
    _u2 = insert(:user, %{inserted_at: Timex.shift(Timex.now(), days: -20)})
    u3 = insert(:user, %{inserted_at: Timex.shift(Timex.now(), days: -2)})
    u4 = insert(:staked_user, %{inserted_at: Timex.shift(Timex.now(), days: -500)})
    u5 = insert(:staked_user, %{inserted_at: Timex.shift(Timex.now(), days: -15)})
    u6 = insert(:staked_user, %{inserted_at: Timex.shift(Timex.now(), days: -1)})

    UserSettings.change_newsletter_subscription(u1, %{newsletter_subscription: :weekly})
    UserSettings.change_newsletter_subscription(u3, %{newsletter_subscription: :weekly})
    UserSettings.change_newsletter_subscription(u5, %{newsletter_subscription: :daily})

    # Simulate that the user subscribed 170d ago
    with_mock(DateTime, [:passthrough], utc_now: fn -> Timex.shift(Timex.now(), days: -170) end) do
      UserSettings.change_newsletter_subscription(u6, %{
        newsletter_subscription: :weekly
      })
    end

    insert(:watchlist, %{user: u1})
    insert(:watchlist, %{user: u1})
    insert(:watchlist, %{user: u3})
    insert(:watchlist, %{user: u4})
    insert(:watchlist, %{user: u6})
    %{}
  end
```

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: Okay, it is so now

test/sanbase/statistics/statistics_test.exs
```diff
@@ -129,7 +115,67 @@ defmodule Sanbase.StatisticsTest do
...
             }} in statistics
   end
+
+  test "newsletter_subscribed_users function" do
+    test_result =
+      Sanbase.Auth.Statistics.newsletter_subscribed_users(
+        Sanbase.Auth.Settings.daily_subscription_type()
+      )
+
+    assert test_result == 1
```
IvanIvanoff: avoid such unnecessary var bindings. Just write:
```elixir
assert Sanbase.Auth.Statistics.newsletter_subscribed_new_users(
	        Sanbase.Auth.Settings.daily_subscription_type(),
	        Timex.shift(now, days: -16)
	      ) == 1
```

test/sanbase/statistics/statistics_test.exs
spiderjako: I've changed the names, are they okay now?

test/sanbase/statistics/statistics_test.exs
```diff
@@ -129,7 +115,67 @@ defmodule Sanbase.StatisticsTest do
...
+
+    test_result =
+      Sanbase.Auth.Statistics.newsletter_subscribed_users(
+        Sanbase.Auth.Settings.weekly_subscription_type()
+      )
+
+    assert test_result == 3
```
IvanIvanoff: avoid reusing var names where not really needed

test/sanbase/statistics/statistics_test.exs
IvanIvanoff: Where for consistency you place the expected value on the right

spiderjako: okay, will fix this

test/sanbase/statistics/statistics_test.exs
spiderjako: you're right, especially when the vars themselves are not needed, fix is on the way

test/sanbase/statistics/statistics_test.exs
spiderjako: I did as suggested and now I saw, that the order is much more readable, so I've just changed that too.

test/sanbase/statistics/statistics_test.exs
kalkov: Names are perfect

test/sanbase/statistics/statistics_test.exs
kalkov: @IvanIvanoff  In general yes, but not in this particular test. Let's get as example this one:
```elixir
test "returns the number of old users, which have subscribed for the newsletter in a given time period" do
    now = Timex.now()

    assert Sanbase.Auth.Statistics.newsletter_subscribed_old_users(
             daily_subscription_type(),
             Timex.shift(now, days: -5),
             Timex.shift(now, days: -14)
           ) == 1

    assert Sanbase.Auth.Statistics.newsletter_subscribed_old_users(
             weekly_subscription_type(),
             Timex.shift(now, days: -9),
             Timex.shift(now, days: -25)
           ) == 1
  end
```

Scroll to setup, try to memoize the users, get back to the test and try to understand why count is 1.

IvanIvanoff:
@spiderjako I cannot approve as I've created the PR but it looks good to me 👍

PR #1355, opened 2019-08-01
Add custom implementation of the datetime scalar in GraphQL

IvanIvanoff:
#### Summary
Remove the dependency on the Absinthe's implementation of some of the date/datetime types.
We need to truncate the date times to seconds (we don't care for lower precision). GraphQL spec does not define `datetime` type so it's a custom provided type by Absinthe.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


tspenov:
build needs to be fixed - otherwise 👍

PR #1357, opened 2019-08-01
Fix logo fields in project entity

kalkov:
#### Summary
camelCased logo fields work properly now
```
{
  projectBySlug(slug: "santiment") {
    logoUrl
    logo32Url
    logo64Url
  }
}
```


lib/sanbase_web/graphql/schema/types/project_types.ex
```diff
@@ -36,8 +36,8 @@ defmodule SanbaseWeb.Graphql.ProjectTypes do
     end
 
     field(:logo_url, :string)
-    field(:logo_32_url, :string)
-    field(:logo_64_url, :string)
+    field(:logo_32_url, :string, name: "logo32_url")
```
IvanIvanoff: This won't work

IvanIvanoff: This renaming will break at 2 places in AccessControl and TimeframeRestriction where we call `String.to_existing_atom` and there is no `:logo32_url` existing atom

IvanIvanoff: Such discrepancies are better fixed in the root. I imagine you do this so you don't have a custom resolved and fetch it directly from the DB. In that case you can rename the DB columns to `logo32_url` and `logo64_url`

IvanIvanoff: Hm.. This actually won't be checked by these 2 modules

IvanIvanoff: Still, a db migration will be better imo.

kalkov: Well renaming a column in the database because of a bug in one of the [top layers](https://github.com/absinthe-graphql/absinthe/issues/601) seems very wrong to me. What will do when the bug is fixed, we will need to rename the DB fields one more time?

We can add to the `Project` module:
```elixir
def logo32_url do
  "logo_32_url"
end

def logo64_url do
  "logo_64_url"
end

Comment (IvanIvanoff) [reply lib/sanbase_web/graphql/schema/types/project_types.ex]: I don't think we're affected by this bug you've linked, because it's behavior of pure Elixir:
```elixir
"logo_64_url" |> Macro.camelize |> Macro.underscore
"logo64_url"
```

The linked issue is about having only letters (the `v_i_n` example), but we have numbers. And numbers don't have capital letters*, so the transformation is not reversible. You cannot not of `logo64` is really coming from `logo_64` or `logo64`

---
* some fonts support upper and lower digits.

kalkov: @IvanIvanoff Renamed the fields all the way from migration to graphQL

PR #1359, opened 2019-08-02
Implement V2 CH metrics

IvanIvanoff:
#### Summary
Implements generic access to the v2 Clickhouse metrics table.
`@external_resource` is used to mark that changes in the metrics file must trigger recompilation of the module (as data is stored in module attributes)

In the `available_v2_metrics.json` every line represents one metric. Every line contains comma-separated values. The first value is the metric name as it is in Clickhouse. The second value is the aggregation that should be applied if we're querying large intervals.
Different metrics have different default aggregations.

Example: Querying data for 7d periods.
- If the metric is DAA - the values **cannot** be summed - we need to take the average by default
- If the metric is TransactionVolume - we can sum it.

This aggregation is taken if no other aggregation is passed to the function.

Examples of different aggregations:
```elixir
Metric.get("santiment", "daily_active_addresses", Timex.shift(Timex.now, days: -30), Timex.now, "7d", :min)
{:ok,
 [
   %{datetime: ~U[2019-07-04 00:00:00Z], value: 7},
   %{datetime: ~U[2019-07-11 00:00:00Z], value: 3},
   %{datetime: ~U[2019-07-18 00:00:00Z], value: 3},
   %{datetime: ~U[2019-07-25 00:00:00Z], value: 3},
   %{datetime: ~U[2019-08-01 00:00:00Z], value: 0}
 ]}
```
```elixir
Metric.get("santiment", "daily_active_addresses", Timex.shift(Timex.now, days: -30), Timex.now, "7d", :max)
{:ok,
 [
   %{datetime: ~U[2019-07-04 00:00:00Z], value: 24},
   %{datetime: ~U[2019-07-11 00:00:00Z], value: 16},
   %{datetime: ~U[2019-07-18 00:00:00Z], value: 15},
   %{datetime: ~U[2019-07-25 00:00:00Z], value: 18},
   %{datetime: ~U[2019-08-01 00:00:00Z], value: 11}
 ]}
```
```elixir
Metric.get("santiment", "daily_active_addresses", Timex.shift(Timex.now, days: -30), Timex.now, "7d", :avg)
{:ok,
 [
   %{datetime: ~U[2019-07-04 00:00:00Z], value: 11.428571428571429},
   %{datetime: ~U[2019-07-11 00:00:00Z], value: 9.857142857142858},
   %{datetime: ~U[2019-07-18 00:00:00Z], value: 9.142857142857142},
   %{datetime: ~U[2019-07-25 00:00:00Z], value: 9.857142857142858},
   %{datetime: ~U[2019-08-01 00:00:00Z], value: 5.5}
 ]}
```

## API:
Request get timeseries data for mvrv:
```graphql
{
  getMetric(metric: "mvrv_usd_20y") {
    timeseriesData(slug: "ethereum", from: "2019-05-01T00:00:00Z", to: "2019-08-01T00:00:00Z", interval: "1d", aggregation: MAX) {
      datetime
      value
    }
  }
}
```

Request - get metadata for DAA:
```graphql
{
  getMetric(metric: "mvrv_usd_20y") {
    metadata {
      minInterval
      defaultAggregation
    }
  }
}
```

Request - get available metrics:
```graphql
{
  getAvailableMetrics
}

```

Request - get available slugs:
```graphql
{
  getAvailableSlugs
}
```

**TODO**:
- Rework the tests in a future PR because they're messy now: https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-5687

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/metric.ex
```diff
@@ -0,0 +1,123 @@
...
+    field(:computed_at, :utc_datetime)
+  end
+
+  def get(slug, metric, from, to, interval, aggregation \\ nil)
+
+  def get(slug, metric, from, to, interval, aggregation)
+      when aggregation in @aggregations do
+    case metric in @metrics_mapset do
+      false ->
+        close = Enum.find(@metrics_mapset, fn m -> String.jaro_distance(metric, m) > 0.9 end)
```
tspenov: That can be extracted as a helper: `missing_metric_error_message` :?

lib/sanbase/clickhouse/metric.ex
```diff
@@ -0,0 +1,123 @@
...
+  @table "daily_metrics_v2"
+  schema @table do
+    field(:datetime, :utc_datetime, source: :dt)
+    field(:asset_id, :integer)
+    field(:metric_id, :integer)
+    field(:value, :float)
+    field(:computed_at, :utc_datetime)
+  end
+
+  def get(slug, metric, from, to, interval, aggregation \\ nil)
```
tspenov: I think the `metric` be the first argument :?

lib/sanbase_web/graphql/schema/queries/metric_queries.ex
```diff
@@ -0,0 +1,28 @@
...
+  use Absinthe.Schema.Notation
+
+  import SanbaseWeb.Graphql.Cache, only: [cache_resolve: 1]
+
+  alias SanbaseWeb.Graphql.Resolvers.MetricResolver
+  alias SanbaseWeb.Graphql.Complexity
+  alias SanbaseWeb.Graphql.Middlewares.TimeframeRestriction
+
+  object :metric_queries do
+    @desc "Returns a list of slugs of the projects that have a github link"
```
tspenov: leftover

lib/sanbase/clickhouse/metric.ex
IvanIvanoff: done

lib/sanbase_web/graphql/schema/queries/metric_queries.ex
```diff
@@ -0,0 +1,27 @@
+defmodule SanbaseWeb.Graphql.Schema.MetricQueries do
+  use Absinthe.Schema.Notation
+
+  import SanbaseWeb.Graphql.Cache, only: [cache_resolve: 1, cache_resolve: 2]
+
+  alias SanbaseWeb.Graphql.Resolvers.MetricResolver
+
+  object :metric_queries do
+    @desc ~s"""
+    Retrurn data for a given metric.
```
tspenov:
```suggestion
    Return data for a given metric.
```

lib/sanbase_web/graphql/schema/types/metric_types.ex
```diff
@@ -0,0 +1,52 @@
...
+    field(:value, non_null(:float))
+  end
+
+  object :metadata do
+    field(:min_interval, :string)
+  end
+
+  object :metric do
+    @desc ~s"""
+    Return a list of 'datetime' float 'value' for a given metric, slug
```
tspenov:
```suggestion
    Return a list of 'datetime',  float 'value' for a given metric, slug
```

lib/sanbase_web/graphql/schema/types/metric_types.ex
```diff
@@ -0,0 +1,52 @@
...
+    and time period
+    """
```
tspenov: You can add some text for the `aggregation` arg

lib/sanbase_web/graphql/schema/types/metric_types.ex
```diff
@@ -0,0 +1,52 @@
...
+    field :timeseries_data, list_of(:metric_data) do
+      arg(:slug, non_null(:string))
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "1d")
+      arg(:aggregation, :aggregation, default_value: nil)
```
tspenov: we have `aggregation = aggregation || Map.get(@metric_aggregation_map, metric, :last)`. Why don't we just put `:last` as default value :?

IvanIvanoff: If we have a default `:last` we do not really know if the user wants to override the default or is not providing aggregation and the one from the file must be taken. And not all defaults are `last`

lib/sanbase/clickhouse/metric/metric.ex
```diff
@@ -0,0 +1,162 @@
...
+  The metrics are stored in the '#{@table}' clickhouse table where each metric
+  is defined by a `metric_id` and every project is defined by an `asset_id`.
+  """
+
+  use Ecto.Schema
+  require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+
+  # The metrics that are available are described in a json file
+  @external_resource availalbe_metrics_file = Path.join(__DIR__, "available_v2_metrics.json")
+  @metrics_json File.read!(availalbe_metrics_file) |> Jason.decode!()
```
valo:
```suggestion
  @metrics_json File.read!(availabлe_metrics_file) |> Jason.decode!()
```

IvanIvanoff: The var is defined with the wrong name, too, will rename everywhere

lib/sanbase/clickhouse/metric/metric.ex
```diff
@@ -0,0 +1,162 @@
...
+
+  defp metric_query(metric, slug, from, to, interval, aggregation) do
+    query = """
+    SELECT
+      toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS t,
+      #{aggregation(aggregation, "value", "t")}
+    FROM(
+      SELECT
+        dt,
+        argMaxIf(value, computed_at, metric_name = ?2) AS value
```
valo: Is the `argMaxIf` operator needed here? I think this query works on a single metric, right?

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
```diff
@@ -0,0 +1,174 @@
...
+      conn: conn,
+      slug: "ethereum",
+      from: from_iso8601!("2019-01-01T00:00:00Z"),
+      to: from_iso8601!("2019-01-02T00:00:00Z"),
+      interval: "1d"
+    ]
+  end
+
+  test "returns data for an available metric", context do
+    %{conn: conn, slug: slug, from: from, to: to, interval: interval} = context
```
kalkov: I would suggest matching like that in the future, using context all over the place pollutes the tests and doesn't give any value.

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
```diff
@@ -0,0 +1,174 @@
...
+    [
+      conn: conn,
+      slug: "ethereum",
+      from: from_iso8601!("2019-01-01T00:00:00Z"),
+      to: from_iso8601!("2019-01-02T00:00:00Z"),
+      interval: "1d"
+    ]
+  end
+
+  test "returns data for an available metric", context do
```
kalkov: You can group the tests in two describe blocks - "when successful" and "when unsuccessful" and mocks and other stuff can be extracted into setup for the successful case.

test/sanbase/billing/query_access_level_test.exs
```diff
@@ -0,0 +1,262 @@
...
+    test "restricted clickhouse v2 queries" do
+      restricted_queries =
+        Sanbase.Billing.Plan.AccessChecker.Helper.get_metrics_with_access_level(:restricted)
+        |> Enum.sort()
+        |> Enum.reject(&is_atom/1)
+        |> Enum.map(&elem(&1, 1))
+
+      expected_restricted_queries =
+        [
+          "daily_stack_circulation",
```
kalkov: Why at some places we use atoms and in other strings?

test/sanbase/clickhouse/metric/v2_clickhouse_metric_test.exs
```diff
@@ -0,0 +1,164 @@
+defmodule Sanbase.Clickhouse.V2ClickhouseMetricTest do
```
kalkov: A lot of repetition in this test, can some parts be extracted into a setup or functions?

test/sanbase_web/graphql/billing/api_product_access_test.exs
```diff
@@ -26,6 +28,28 @@ defmodule Sanbase.Billing.ApiProductAccessTest do
...
+      assert_called(Metric.get(metric, :_, from, to, :_, :_))
+      assert result != nil
+    end
+
+    test "cannot access RESTRICTED v2 clickhouse metrics for over 3 months", context do
+      from = Timex.shift(Timex.now(), days: -91)
+      to = Timex.shift(Timex.now(), days: -10)
+      metric = v2_restricted_metric()
+      query = metric_query(metric, from, to)
+      result = execute_query(context.conn, query, "getMetric")
```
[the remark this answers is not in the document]
IvanIvanoff: It happens that the function is called but returns an error, so the whole response structure changes, so the way `execute_query` parses the result return `nil`. In such cases `execute_query` is replaced with `execute_query_with_error`

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
IvanIvanoff: I usually do it like:
```elixir
test  "test something", %{conn: conn, project: project} do
```
but in this case we have too many fields and the formatter makes it on multiple lines

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
IvanIvanoff: I totally forgot that you can write a setup block per describe block. Will rework it and it will be cleaner

test/sanbase/billing/query_access_level_test.exs
IvanIvanoff: we use either
`:atom` or `{: atom, str}`
The second one comes from the clickhouse JSON file so we can distinguish between `daily_active_addresses` from the schema and the one from the v2 metrics.

We can go with atoms in the first place because the query name is the atom itself. The query will get rejected if the name is wrong.

With the v2 metrics, we get the metric name as an argument - absinthe now cannot enforce that the correct name is passed. Furthermore, we cannot just convert to atom because this will allow 1 person to crash the whole node (send random strings until the memory for atoms is exhausted).

On the other hand, we can put some checks before converting, but this will not allow us to do the similarity check and include similar metrics in the error response.

--
The whole thing is easy to change and this implementation is not binding

test/sanbase_web/graphql/billing/api_product_access_test.exs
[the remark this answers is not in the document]
kalkov: Then the `result` should also be passed as argument :(

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
kalkov: This is even better in most places

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
kalkov: But we should be careful to not overuse them, two big setup blocks will harm the readability.

lib/sanbase_web/graphql/schema/types/metric_types.ex
tspenov: aha, I got it - want to keep precedence (provided by user > provided by json file > default for all).That `:last` has to go as module attribute `@default_aggregation_all_metrics` IMO

IvanIvanoff: There's no need for default for all because the `aggregation` field is a must in the JSON file.  I like it this way because the logic for choosing the aggregation depends on less stuff

IvanIvanoff: I'll remove this `:last` default arg in the Map.get as it's not necessary because of :point_up:

test/sanbase_web/graphql/billing/api_product_access_test.exs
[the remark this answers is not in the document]
IvanIvanoff: Let's rework the tests in an another PR (linked a favro card in the description)

test/sanbase/clickhouse/metric/v2_clickhouse_metric_test.exs
IvanIvanoff: Let's do that in another PR (linked the favro card in the description) as this PR is huge

lib/sanbase/clickhouse/metric/metric.ex
IvanIvanoff: argMaxIf is for the `computed_at` - when we recompute metrics there's a new `computed_at` and we need to return always the last one, right?

IvanIvanoff: The v2 dashboards have the same `argMaxIf`

test/sanbase_web/graphql/clickhouse/metric_timeseries_data_test.exs
IvanIvanoff: A rule of thumb (I would not follow it always though) is to match in the head only the arguments that you depend on (i.e. if the `key` field has value `max`) (`def foo(%{key: "MAX"}`)
and match the arguments your control flow does not depend on in the body

test/sanbase_web/graphql/billing/api_product_access_test.exs
```diff
@@ -26,6 +28,28 @@ defmodule Sanbase.Billing.ApiProductAccessTest do
...
+      to = Timex.now()
+      metric = v2_free_metric()
+      query = metric_query(metric, from, to)
+      result = execute_query(context.conn, query, "getMetric")
+      assert_called(Metric.get(metric, :_, from, to, :_, :_))
+      assert result != nil
+    end
+
+    test "cannot access RESTRICTED v2 clickhouse metrics for over 3 months", context do
+      from = Timex.shift(Timex.now(), days: -91)
```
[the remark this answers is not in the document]
IvanIvanoff: nope, you cannot (should not, you can put some stuff in module attributes or process dictionary) access anything that's not provided to you as an argument.

lib/sanbase/clickhouse/metric/metric.ex
MilanRusev: Yes, it is to remove the duplicates.

Even safer will be `argMaxIf(value, (version, computed_at), metric_name = ?2) AS value` although it shouldn't matter because we use the same version everywhere now. And probably will remove the version column at some point.

MilanRusev: Yes, it is to remove the duplicates.

However the same thing should be made for the `metric_metadata` table below.

lib/sanbase/clickhouse/metric/metric.ex
```diff
@@ -0,0 +1,219 @@
...
+      FROM #{@table}
+      INNER JOIN (
+        SELECT
```
MilanRusev: Here there may be duplicates as well:
```sql
select
          name AS metric_name,
          argMax(metric_id, (version, computed_at)) as metric_id
        FROM
          metric_metadata
        PREWHERE
          name = ?2
```

The `version` can be skipped - it shouldn't matter because we use the same version everywhere now. And probably will remove the version column at some point.

IvanIvanoff: @MilanRusev version can be skipped means:
```sql
select
          name AS metric_name,
          argMax(metric_id, computed_at) as metric_id
        FROM
          metric_metadata
        PREWHERE
          name = ?2
``` 
?

MilanRusev: Actually Instead of inner join you can add the metric filter in the prewhere clause, like the asset_id. They are similar:
```sql
PREWHERE
...
AND metric_id = (
            SELECT argMax(metric_id, computed_at)
            FROM metric_metadata
            PREWHERE name = ?2
          )
```

lib/sanbase/clickhouse/metric/metric.ex
MilanRusev: Aha now I get what Valio is saying - you don't need argMaxIf but only `argMax(value, computed_at)` because you have only one metric from the filter below.

IvanIvanoff: aargh, now got it - yes, we can remove the `if`

lib/sanbase/clickhouse/metric/metric.ex
```diff
@@ -0,0 +1,219 @@
...
+          name AS metric_name,
+          argMax(metric_id, computed_at) AS metric_id
+        FROM
+          metric_metadata
+        PREWHERE
+          name = ?2
+      ) USING metric_id
+      PREWHERE
+          dt >= toDateTime(?3) AND
+          dt <= toDateTime(?4) AND
```
MilanRusev: Maybe here it should be only `<`

test/sanbase_web/graphql/billing/api_product_access_test.exs
```diff
@@ -260,6 +354,30 @@ defmodule Sanbase.Billing.ApiProductAccessTest do
     end
   end
 
+  # Private functions
+
+  defp v2_free_metric(), do: Metric.free_metrics() |> Enum.random()
+  defp v2_restricted_metric(), do: Metric.restricted_metrics() |> Enum.random()
+
+  defp from_to(from_days_shift, to_days_shift) do
```
kalkov: :100:

PR #1361, opened 2019-08-05
Add sheets access control

tspenov:


lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -136,10 +137,20 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
+    {Product.product_sheets(), SansheetsAccessChecker},
     {Product.product_sangraphs(), SanbaseAccessChecker}
   ]
 
+  def historical_data_in_days(plan, query, _product) when query in @custom_access_queries do
+    Map.get(@custom_access_queries_stats, query)
+    |> get_in([:plan_access, plan, :historical_data_in_days])
+  end
+
+  def realtime_data_cut_off_in_days(plan, query) when query in @custom_access_queries do
```
IvanIvanoff: Should we have different values for the custom metrics in different products? If yes - we need to have SanbaseCustomAccess, etc.

lib/sanbase/billing/plan/sansheets_access_checker.ex
```diff
@@ -0,0 +1,47 @@
...
+  """
+
+  @free_plan_stats %{
+    historical_data_in_days: 3 * 30,
+    realtime_data_cut_off_in_days: 1
+  }
+
+  @basic_plan_stats %{
+    historical_data_in_days: 6 * 30,
+    realtime_data_cut_off_in_days: 0
```
IvanIvanoff: In sheets we'll have much more realtime access?

lib/sanbase/billing/plan/access_checker.ex
tspenov: In docs I see MVRV,RV,TokenAgeConsumed + Some non existing like Mean Age, TB MVRV. Until that is cleared I think this belongs to the main AccessChecker module.

lib/sanbase/billing/plan/sansheets_access_checker.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Sanbase.Billing.Plan.SansheetsAccessChecker do
+  @moduledoc ~s"""
+  Implement the restrictions for the SANSheets product
+  """
+
+  @free_plan_stats %{
+    historical_data_in_days: 3 * 30,
+    realtime_data_cut_off_in_days: 1
```
IvanIvanoff: Why are we giving so much realtime access in `free`? It will be really easy to fake and work around the 30 days realtime data cut off in sanbase/neuro

lib/sanbase/billing/plan/sansheets_access_checker.ex
tspenov: same as API

lib/sanbase/billing/plan/sansheets_access_checker.ex
tspenov: the API is same - the exception are the sanbase plans

test/sanbase_web/graphql/context_plug_test.exs
```diff
@@ -216,5 +217,34 @@ defmodule SanbaseWeb.Graphql.ContextPlugTest do
...
+        |> put_req_header(
+          "user-agent",
+          "Mozilla/5.0 (compatible; Google-Apps-Script)"
+        )
+
+      conn = setup_apikey_auth(conn, apikey)
+
+      conn_context = conn.private.absinthe.context
+
+      assert conn_context.product == 3
```
kalkov: This looks cryptic, I would expect something like `assert conn_context.product == "SANsheets"`
Where I can see the ids of these products?

tspenov: yeah, will change using the public func to `Product.product_sheets()`
here are the products:
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/billing/product.ex#L13
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/billing/product.ex#L28

tspenov:
@kalkov I've renamed here , also in Stripe to `Sheets by Santiment`, though that blocks the PR for a while cause frontend is using the exact name: https://github.com/santiment/sheets-landing/blob/fda6abaec454f2d846e20ae28d20094d1cc3be04/src/utils/plans.js#L24

priv/repo/migrations/20190807074910_update_products_names.exs
```diff
@@ -0,0 +1,28 @@
...
+  alias Sanbase.Billing.Product
+
+  def up do
+    Application.ensure_all_started(:tzdata)
+    Application.ensure_all_started(:prometheus_ecto)
+    Sanbase.Prometheus.EctoInstrumenter.setup()
+
+    %{
+      Product.product_api() => "Neuro by Santiment",
+      Product.product_sanbase() => "Sanabse by Santiment",
```
tspenov:
```suggestion
      Product.product_sanbase() => "Sanbase by Santiment",
```

PR #1362, opened 2019-08-06
Improve error messages for queries available only for ethereum

spiderjako:
With this PR I add proper messages to errors given, when we use any slug instead of ethereum with miner metrics

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
@@ -116,7 +116,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
 
   def mining_pools_distribution(
         _root,
-        %{slug: slug, from: from, to: to, interval: interval},
+        %{slug: slug = "ethereum", from: from, to: to, interval: interval},
         _resolution
       ) do
     case MiningPoolsDistribution.distribution(slug, from, to, interval) do
@@ -130,9 +130,23 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     end
   end
 
+  def mining_pools_distribution(
+        _root,
+        %{slug: slug, from: _from, to: _to, interval: _interval},
+        _resolution
+      ) do
+    log_graphql_error(
+      "Can't fetch Mining Pools Distribution for project with slug: #{slug}",
+      "Currently only ethereum is supported."
+    )
+
+    {:error,
+     "Can't fetch Mining Pools Distribution for project with slug: #{slug}. Currently only ethereum is supported."}
+  end
+
   def miners_balance(
         _root,
-        %{slug: slug, from: from, to: to, interval: interval},
+        %{slug: slug = "ethereum", from: from, to: to, interval: interval},
         _resolution
       ) do
     with {:ok, from, to, interval} <-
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -68,20 +73,31 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
   end
 
+  def gas_used(
+        _root,
+        %{slug: slug, from: _from, to: _to, interval: _interval},
+        _resolution
+      ) do
+    error_msg = "Gas Used for '#{slug}' is not available"
+    error = "Currently only 'ethereum' is supported."
+    log_graphql_error(error_msg, error)
```
IvanIvanoff: Directly put the messages here instead of binding them to variables

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -68,20 +73,31 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
 
   def gas_used(
         _root,
-        %{slug: slug, from: from, to: to, interval: interval},
+        %{slug: "ethereum", from: from, to: to, interval: interval},
```
IvanIvanoff: you can do `slug = "ethereum"` and just use `slug`. This way if we have more projects with gas_used metric in the future you can just change the function definition not to be:
```elixir
def gas_used(_, slug = "ethereum", ...) do
```
but
```elixir
def gas_used(_, slug, ...) when slug in @gas_used_slugs do 
...
```

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: Is it okay the way it is now? We need both of the strings anyways for the log

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: Yes, you're right. Thanks, now it should be alright

test/sanbase_web/graphql/clickhouse/gas_used_test.exs
```diff
@@ -107,7 +107,7 @@ defmodule SanbaseWeb.Graphql.Clickhouse.GasUsedTest do
                result = parse_response(response)
                assert result == nil
              end) =~
-               graphql_error_msg("Gas Used", "unsupported", error)
+               graphql_error_msg_eth_metrics("Gas Used", "unsupported")
```
IvanIvanoff: `graphql_error_msg` was a function used throughout sanbase. `graphql_error_msg_eth_metrics` is not - you can just paste the string here

spiderjako: Did as you said, it's truly better.

IvanIvanoff:
@spiderjako if you'll include only the gas used you can rename the PR to `Improve gas_used error responses`

kalkov:
@IvanIvanoff  Will we keep the matching into the  [Clickhouse modules](https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/gas_used.ex#L38)?

IvanIvanoff:
@kalkov hm... as it was not shown I did not know we had it in the CH module

IvanIvanoff:
Closing in favor of https://github.com/santiment/sanbase2/pull/1373

PR #1364, opened 2019-08-06
Simplify access logic

IvanIvanoff:
#### Summary
- The `basic` and `pro` subscription meta fields are not really meaningful after the latest decision to not have queries that are available only in `pro` and higher plans.
- Rename `subscription` to `access` and give it a few values:
-- free - no time restrictions are applied
-- restricted - time restrictions are applied according to the user's plan
-- forbidden - only basic auth can access these fields
- Remove the `access` field in the plans. It was a huge map of the available metrics that was no longer used but was taking memory and bandwidth reading and sending.
- Remove the `custom_access` subscription field value. There is no corresponding value for the `access` field. Instead, just use the `restricted` value. When a query has custom access it will be automatically caught and the restrictions will be taken from that custom access description.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


test/sanbase/billing/subscription_meta_field_test.exs
```diff
@@ -4,12 +4,12 @@ defmodule Sanbase.Billing.SubscriptionMetaFieldTest do
   # Assert that a query's subscription plan does not change incidentally
```
tspenov: rename this file to `access_meta_field_test`

PR #1367, opened 2019-08-07
Add  cmc project

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -113,9 +121,39 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
...
+    |> Repo.update()
+  end
+
   defp update_local_project(nil, _filepath), do: {:error, "Project not found"}
 
   defp update_local_project(project, fields) do
     Project.changeset(project, fields) |> Repo.update()
   end
+
+  defp image_content_hash(file_path) do
```
kalkov: This should be extracted into a separate module and used here and in `file_resolver`

lib/sanbase/model/cmc_project.ex
```diff
@@ -0,0 +1,46 @@
...
+  alias Sanbase.Repo
+  alias Sanbase.Model.CmcProject
+
+  schema "cmc_project" do
+    belongs_to(:project, Project)
+
+    field(:logos_uploaded_at, :naive_datetime)
+    field(:logo_hash, :string)
+
+    timestamps()
```
IvanIvanoff: This schema looks like `logo_data` and not `cmc_project` to me.

lib/sanbase/model/cmc_project.ex
```diff
@@ -0,0 +1,46 @@
...
+    |> cast(attrs, [:project_id, :logo_hash, :logos_uploaded_at])
+    |> validate_required([:project_id])
+    |> unique_constraint(:project_id)
+  end
+
+  def by_project_id(project_id) do
+    Repo.get_by(CmcProject, project_id: project_id)
+  end
+
+  def insert!(project_id) do
```
IvanIvanoff: I would call this `create` as we use this naming more often

lib/sanbase/model/cmc_project.ex
kalkov: It currently contains the absolute minimum, my idea is to use it to store also metadata for other sync scripts and to know if a local project is on CMC or not.

IvanIvanoff: I'm not sure we'll use it for anything else. 
Also `CmcProject` looks kind of ugly to me. Better be named `CoinmarketcapProject` as we don't have any abbreviations in the db

IvanIvanoff: Also if we use it only to store data for sync scripts it's not really a `project` type but a `data` type

tspenov: I would vote for `project_logo_data` or `coinmarketcap_logo_data` and also think that it might not be used for anything else. We also have `LatestCoinmarketcapData` that stores some info from coinmarketcap.

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -91,6 +97,12 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     end
   end
 
+  defp file_hash_has_changed?(project, filepath) do
+    cmc_project = CmcProject.get_or_insert(project.id)
+    {:ok, image_hash} = image_content_hash(filepath)
```
tspenov: you can return the calculated hash and don't calculate it second time below because it might cpu intensive operation.

lib/sanbase/model/cmc_project.ex
```diff
@@ -0,0 +1,46 @@
...
+    |> Repo.insert!()
+  end
+
+  def get_or_insert(project_id) do
+    {:ok, cmc_project} =
+      Repo.transaction(fn ->
+        by_project_id(project_id)
+        |> case do
+          nil -> insert!(project_id)
+          project_id -> project_id
```
tspenov: that is `project` not `project_id`. I think the anon function should return {:ok, %Project{}} and the  `get_or_insert` should return {:ok, ..} or `{:error, ..}`.Basically:
```elixir
def get_or_insert(project_id) do
      Repo.transaction(fn ->
        by_project_id(project_id)
        |> case do
          nil -> create(project_id)
          project -> {:ok, project_id}
       end
   end)
end
```

kalkov:
Closing in favor of [1374](https://github.com/santiment/sanbase2/pull/1374)

PR #1368, opened 2019-08-07
Updating the headers of the statistics

spiderjako:
The headers had words, which were used repeatedly and could be removed

### DIFF
--- a/lib/sanbase/automatic_statistics/statistics.ex
+++ b/lib/sanbase/automatic_statistics/statistics.ex
@@ -7,8 +7,8 @@ defmodule Sanbase.Statistics do
     "staking_users",
     "registered_users",
     "registered_staking_users",
-    "daily_newsletter_subscriptions",
-    "weekly_newsletter_subscriptions",
+    "users_with_daily_newsletter_subscriptions",
+    "users_with_weekly_newsletter_subscriptions",
     "watchlists",
     "tokens_staked"
   ]
@@ -68,13 +68,20 @@ defmodule Sanbase.Statistics do
     UserStatistics.tokens_staked()
   end
 
-  def get("daily_newsletter_subscriptions") do
+  def get("users_with_daily_newsletter_subscriptions") do
     now = Timex.now()
 
     last_7d =
       UserStatistics.newsletter_subscribed_users(
         daily_subscription_type(),
         Timex.shift(now, days: -7),
+        now
+      )
+
+    last_14d =
+      UserStatistics.newsletter_subscribed_users(
+        daily_subscription_type(),
+        Timex.shift(now, days: -14),
         now
       )
 
@@ -108,22 +115,30 @@ defmodule Sanbase.Statistics do
       )
 
     %{
-      "daily_updates_subscribed_user_count_last_7d" => last_7d,
-      "daily_updates_subscribed_user_count_last_30d" => last_30d,
-      "daily_updates_subscribed_user_count_last_180d" => last_180d,
-      "daily_updates_subscribed_user_count_overall" => overall,
-      "daily_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "daily_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "subscribed_in_the_last_7_days" => last_7d,
+      "subscribed_in_the_last_14_days" => last_14d,
+      "subscribed_in_the_last_30_days" => last_30d,
+      "subscribed_in_the_last_180_days" => last_180d,
+      "subscribed_overall" => overall,
+      "subscribed_new_users" => newsletter_subscribed_new_users_14d,
+      "longtime_users_who_subscribed_in_the_last_14_days" => newsletter_subscribed_old_users_14d
     }
   end
 
-  def get("weekly_newsletter_subscriptions") do
+  def get("users_with_weekly_newsletter_subscriptions") do
     now = Timex.now()
 
     last_7d =
       UserStatistics.newsletter_subscribed_users(
         weekly_subscription_type(),
         Timex.shift(now, days: -7),
+        now
+      )
+
+    last_14d =
+      UserStatistics.newsletter_subscribed_users(
+        weekly_subscription_type(),
+        Timex.shift(now, days: -14),
         now
       )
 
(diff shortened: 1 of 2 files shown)


lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -108,16 +108,16 @@ defmodule Sanbase.Statistics do
...
-      "daily_updates_subscribed_user_count_last_180d" => last_180d,
-      "daily_updates_subscribed_user_count_overall" => overall,
-      "daily_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "daily_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "registered_in_the_last_7_days" => last_7d,
+      "registered_in_the_last_30_days" => last_30d,
+      "registered_in_the_last_180_days" => last_180d,
+      "registered_overall" => overall,
+      "registered_new_users" => newsletter_subscribed_new_users_14d,
+      "old_users_which_registered_in_the_last_14_days" => newsletter_subscribed_old_users_14d
```
IvanIvanoff: The relative pronoun referring to people (users) is `who` and not `which`

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -108,16 +108,16 @@ defmodule Sanbase.Statistics do
...
-      "daily_updates_subscribed_user_count_last_30d" => last_30d,
-      "daily_updates_subscribed_user_count_last_180d" => last_180d,
-      "daily_updates_subscribed_user_count_overall" => overall,
-      "daily_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "daily_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "registered_in_the_last_7_days" => last_7d,
+      "registered_in_the_last_30_days" => last_30d,
+      "registered_in_the_last_180_days" => last_180d,
+      "registered_overall" => overall,
+      "registered_new_users" => newsletter_subscribed_new_users_14d,
```
IvanIvanoff: I'm confused. Why the users that registered 7, 30 and 180 days ago are expressed as such, but the users registered in the last 14 days are `new`?

spiderjako: Registered isn't the most correct word when I think about it. In the first 3, I can change 'registered' to be 'subscribed'?

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: You're right, going to change that.

lib/sanbase/automatic_statistics/statistics.ex
IvanIvanoff: Please do

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -108,16 +108,16 @@ defmodule Sanbase.Statistics do
...
-      "daily_updates_subscribed_user_count_last_180d" => last_180d,
-      "daily_updates_subscribed_user_count_overall" => overall,
-      "daily_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "daily_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "subscribed_in_the_last_7_days" => last_7d,
+      "subscribed_in_the_last_30_days" => last_30d,
+      "subscribed_in_the_last_180_days" => last_180d,
+      "subscribed_overall" => overall,
+      "subscribed_new_users" => newsletter_subscribed_new_users_14d,
+      "old_users_who_subscribed_in_the_last_14_days" => newsletter_subscribed_old_users_14d
```
kalkov: Probably existing is the better word here than old?

spiderjako: Existing is relevant for the new users also. Maybe "longtime" ?

kalkov: Both works for me

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -108,16 +108,16 @@ defmodule Sanbase.Statistics do
...
-      "daily_updates_subscribed_user_count_last_180d" => last_180d,
-      "daily_updates_subscribed_user_count_overall" => overall,
-      "daily_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "daily_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "subscribed_in_the_last_7_days" => last_7d,
+      "subscribed_in_the_last_30_days" => last_30d,
+      "subscribed_in_the_last_180_days" => last_180d,
+      "subscribed_overall" => overall,
+      "subscribed_new_users" => newsletter_subscribed_new_users_14d,
+      "longtime_users_who_subscribed_in_the_last_14_days" => newsletter_subscribed_old_users_14d
```
IvanIvanoff: We can add a complimentary `subscribed_in_the_last_14_days` that counts all of the users so we can see the ratio/difference between old and new subscribers

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -157,12 +157,12 @@ defmodule Sanbase.Statistics do
...
-      "weekly_updates_subscribed_user_count_last_180d" => last_180d,
-      "weekly_updates_subscribed_user_count_overall" => overall,
-      "weekly_updates_subscribed_new_user_count" => newsletter_subscribed_new_users_14d,
-      "weekly_updates_subscribed_old_user_count_last_14d" => newsletter_subscribed_old_users_14d
+      "subscribed_in_the_last_7_days" => last_7d,
+      "subscribed_in_the_last_30_days" => last_30d,
+      "subscribed_in_the_last_180_days" => last_180d,
+      "subscribed_overall" => overall,
+      "subscribed_new_users" => newsletter_subscribed_new_users_14d,
+      "longtime_users_who_subscribed_in_the_last_14_days" => newsletter_subscribed_old_users_14d
```
IvanIvanoff: same comment as above - let's add one new statistic

spiderjako: Alright, it shall be done

PR #1371, opened 2019-08-07
Sanbase plans signals limits

tspenov:


lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -53,11 +63,41 @@ defmodule Sanbase.Billing.Plan.SanbaseAccessChecker do
...
+    |> Map.get(:realtime_data_cut_off_in_days)
+  end
+
+  def signals_limit(plan) do
+    plan_stats(plan)
+    |> get_in([:signals, :limit])
+  end
+
+  def signals_limits_reached?(user, subscription) do
+    created_signsls_count = UserTrigger.triggers_for(user) |> Enum.count()
```
kalkov: typo

test/sanbase/billing/sanbase_product_access_test.exs
```diff
@@ -222,6 +227,82 @@ defmodule Sanbase.Billing.SanbaseProductAccessTest do
...
+
+      query = create_trigger_mutation()
+      result = execute_mutation(context.conn, query, "createTrigger")
+      assert result["trigger"]["id"] != nil
+    end
+
+    test "with BASIC API plan can create new trigger", context do
+      user = insert(:user)
+      insert(:subscription_essential, user: context.user)
+      create_triggers(user, @triggers_limit_count)
```
kalkov: Why are you creating another user here and not using the one from the context?

test/sanbase/billing/sanbase_product_access_test.exs
```diff
@@ -222,6 +227,82 @@ defmodule Sanbase.Billing.SanbaseProductAccessTest do
     end
   end
 
+  describe "Signals limits user has reached signals limit" do
+    test "user with FREE plan cannot create new trigger", context do
```
kalkov: I would group the tests in two blocks "for SANbase", "for SANapi" to improve readability.

test/sanbase/billing/sanbase_product_access_test.exs
```diff
@@ -222,6 +227,82 @@ defmodule Sanbase.Billing.SanbaseProductAccessTest do
...
+  describe "Signals limits user has not reached signals limit" do
+    test "user with FREE plan can create new trigger", context do
+      create_triggers(context.user, @triggers_limit_count - 1)
+      query = create_trigger_mutation()
+      result = execute_mutation(context.conn, query, "createTrigger")
+      assert result["trigger"]["id"] != nil
+    end
+  end
+
+  def create_triggers(user, count) do
```
kalkov: This will create 10 records in the DB just to test the limit? What if we change the limit to 100? It's too expensive.
The better way to take here is to override the limit only in the current test to 2 or 3 and test against it. 
The same way pagination is tested.

kalkov: Do you know of other similar places in the existing test codebase? @IvanIvanoff @tspenov

lib/sanbase_web/graphql/resolvers/signals/user_trigger_resolver.ex
```diff
@@ -15,23 +18,15 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserTriggerResolver do
...
-        {:ok, result}
+    subscription = subscription || @free_subscription
 
-      error ->
-        error
+    if subscription.plan.product_id != Product.product_api() and
+         SanbaseAccessChecker.signals_limits_reached?(current_user, subscription) do
+      {:error, SanbaseAccessChecker.signals_limits_upgrade_message()}
+    else
+      do_create_trigger(current_user, args)
```
IvanIvanoff: Hmm.. Could we somehow avoid spreading the logic of the restrictions in the resolvers?

IvanIvanoff: Why avoid checking if signals limit is reached if product_id is not api?
This check should always apply, moreover because apikeys don't have access to mutations

tspenov: I can add another function that does the check. I am wondering where to place it - I guess main `AccessChecker` is a good place.
We have logic which in `jwt` auth places sanbase || api subscription.
If we have api subscription that mans we don't have sanbase one and we want to not enforce signals limits.

test/sanbase/billing/sanbase_product_access_test.exs
tspenov: leftover

test/sanbase/billing/sanbase_product_access_test.exs
tspenov: @kalkov Good suggestion, though now the limits are in module attribute which is evaluated compile-time. One way I see it to mock `UserTrigger.triggers_for(user)`  or extract this `UserTrigger.triggers_for(user) |> Enum.count()` as public function in triggers and mock it.
Maybe @IvanIvanoff have any suggestion here ?

lib/sanbase_web/graphql/resolvers/signals/user_trigger_resolver.ex
tspenov: extracted the check in a function out of the resolver

test/sanbase/billing/sanbase_product_access_test.exs
kalkov: Using public function which calls the module attribute and mocking it in tests seems the way to go to me.

tspenov: yeah, did it that way

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -165,4 +169,17 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
       unquote(module).realtime_data_cut_off_in_days(plan, query)
     end
   end
+
+  def user_can_create_signal?(user, subscription) do
+    subscription = subscription || @free_subscription
+
+    cond do
+      # If user has API subscription - he has unlimmited signals
```
tspenov:
```suggestion
      # If user has API subscription - he has unlimited signals
```

PR #1372, opened 2019-08-08
Add price chart to trending words signal

spiderjako:


lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -209,6 +209,10 @@ defmodule Sanbase.Signal.Trigger.TrendingWordsTriggerSettings do
       The project **#{project.name}** is in the trending words.
 
       More info here: #{Project.sanbase_link(project)}
+
+      ![Trending Words chart and OHCL price chart for the past 90 days](#{
+        chart_url(project, :trending_words)
```
IvanIvanoff: change to `:volume`

lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -209,6 +209,10 @@ defmodule Sanbase.Signal.Trigger.TrendingWordsTriggerSettings do
       The project **#{project.name}** is in the trending words.
 
       More info here: #{Project.sanbase_link(project)}
+
+      ![Trending Words chart and OHCL price chart for the past 90 days](#{
```
IvanIvanoff: Change to `Volume and OHCL price...`

spiderjako: Done, also added the tests

lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -209,6 +209,8 @@ defmodule Sanbase.Signal.Trigger.TrendingWordsTriggerSettings do
       The project **#{project.name}** is in the trending words.
 
       More info here: #{Project.sanbase_link(project)}
+
+      ![Volume and OHCL price chart for the past 90 days](#{chart_url(project, :volume)})
```
kalkov: Isn't it OHLC?

IvanIvanoff: good catch. we have this issue in chart.ex and in the DAA signal, too

spiderjako: I was looking for it to have resemblance to the other signal settings, for example daily active addresses:
https://github.com/santiment/sanbase2/blob/2d333a9c8ed92fb012b6d99d99c101199515e046/lib/sanbase/signals/trigger/settings/daily_active_addresses_settings.ex#L218

spiderjako: In sanpy README it is also OHCL, in Graphql however it is OHLC, so it is as you say it is, I can do another PR after this one to rename OHCL -> OHLC

kalkov: The term is [Open, High, Low, Close](https://www.google.com/search?q=ohlc&oq=ohlc&aqs=chrome..69i57j0l5.1183j0j7&sourceid=chrome&ie=UTF-8), does not make sense to be Open, High, Close, Low.

Should be fixed eveywhere.

spiderjako: That's right, I will merge this PR and I will make a new one, which will be a fix for the names.

PR #1373, opened 2019-08-08
Reworking the graphql error handling

spiderjako:

### DIFF
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -44,6 +44,18 @@ defmodule Sanbase.Auth.User do
   # User with free subscription that is used for external integration testing
   @sanbase_bot_email "sanbase.bot@santiment.net"
 
+  @derive {Inspect,
+           except: [
+             :salt,
+             :email_token,
+             :email_token_generated_at,
+             :email_token_validated_at,
+             :email_candidate_token,
+             :email_candidate_token_generated_at,
+             :email_candidate_token_validated_at,
+             :consent_id
+           ]}
+
   require Mockery.Macro
   defp mandrill_api, do: Mockery.Macro.mockable(Sanbase.MandrillApi)
 
--- a/lib/sanbase/utils/error_handling.ex
+++ b/lib/sanbase/utils/error_handling.ex
@@ -18,16 +18,12 @@ defmodule Sanbase.Utils.ErrorHandling do
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
 
-  def handle_graphql_error(metric, slug, reason, opts \\ []) do
-    description = Keyword.get(opts, :description, "project with slug")
+  def handle_graphql_error(metric, identifier, reason, opts \\ []) do
+    target = Keyword.get(opts, :description, "project with slug")
 
-    error_msg = "[#{Ecto.UUID.generate()}] Can't fetch #{metric} for #{description}: #{slug}"
+    error_msg = "[#{Ecto.UUID.generate()}] Can't fetch #{metric} for #{target}: #{identifier}"
 
-    error_msg_with_reason =
-      error_msg <>
-        """
-        , Reason: \"#{reason}\"
-        """
+    error_msg_with_reason = error_msg <> ", Reason: #{inspect(reason)}"
 
     Logger.warn(error_msg_with_reason)
 
--- a/lib/sanbase/utils/error_handling.ex
+++ b/lib/sanbase/utils/error_handling.ex
@@ -18,18 +18,23 @@ defmodule Sanbase.Utils.ErrorHandling do
     {:error, "[#{log_id}] Error executing #{query_name}. See logs for details."}
   end
 
-  def graphql_error_msg(metric_name) do
-    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name}"
-  end
-
-  def graphql_error_msg(metric_name, identifier, opts \\ []) do
+  def handle_graphql_error(metric, slug, reason, opts \\ []) do
     description = Keyword.get(opts, :description, "project with slug")
 
-    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name} for #{description}: #{identifier}"
-  end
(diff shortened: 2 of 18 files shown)


lib/sanbase/auth/user.ex
```diff
@@ -41,6 +41,8 @@ defmodule Sanbase.Auth.User do
   @insights_fallback_username "anonymous"
   @insights_fallback_email "anonymous@santiment.net"
 
+  @derive {Inspect, except: [:salt, :email_tokens_fields, :consent_id]}
```
IvanIvanoff: there is no such field as `:email_tokens_field`. This was a shortened version of 
```
  field(:email_token, :string)
    field(:email_token_generated_at, :naive_datetime)
    field(:email_token_validated_at, :naive_datetime)
    field(:email_candidate_token, :string)
    field(:email_candidate_token_generated_at, :naive_datetime)
    field(:email_candidate_token_validated_at, :naive_datetime)
```
so i do not write them by hand

spiderjako: Oh, right :D fixed it

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -352,8 +341,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
         {:ok, result}
 
       {:error, error} ->
-        error_msg = graphql_error_msg("Assets held by address", address, description: "address")
-        log_graphql_error(error_msg, error)
+        error_msg = "Can't fetch Assets held by address for address: #{address}"
+        Logger.warn(error_msg)
```
IvanIvanoff: replace the whole thing with the new function

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -414,8 +401,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
         {:error, error_msg}
 
       {:error, error} ->
-        error_msg = graphql_error_msg("Share of Deposits", slug)
-        log_graphql_error(error_msg, error)
+        error_msg = handle_graphql_error("Share of Deposits", slug, error)
         {:error, error_msg}
```
IvanIvanoff: I would write just 
```elixir
{:error,  handle_graphql_error("Share of Deposits", slug, error)}
```

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: Alright

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -9,7 +9,9 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
   alias SanbaseWeb.Graphql.SanbaseDataloader
 
   import Sanbase.Utils.ErrorHandling,
-    only: [log_graphql_error: 2, graphql_error_msg: 1, graphql_error_msg: 2, graphql_error_msg: 3]
+    only: [
+      handle_graphql_error: 3
```
IvanIvanoff: format to be on one line

spiderjako: Won't that bug the "check-if-formatted" ?

IvanIvanoff: nope, the formatter sometimes accepts your formatting, this is one of these cases

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: It should be okay now

lib/sanbase/utils/error_handling.ex
```diff
@@ -18,18 +18,23 @@ defmodule Sanbase.Utils.ErrorHandling do
...
     description = Keyword.get(opts, :description, "project with slug")
 
-    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name} for #{description}: #{identifier}"
-  end
+    error_msg = "[#{Ecto.UUID.generate()}] Can't fetch #{metric} for #{description}: #{slug}"
+
+    error_msg_with_reason =
+      error_msg <>
+        """
+        , Reason: \"#{reason}\"
```
IvanIvanoff: no need for heredocs. Just 
```elixir
error_msg_with_reason = error_msg <> ", Reason: '#{inspect(reason)'}"
```
will suffice. Also note the `inspect` - if the error is not string (or does not implement String.Chars protocl) - this will fail to transform

lib/sanbase/utils/error_handling.ex
```diff
@@ -18,18 +18,23 @@ defmodule Sanbase.Utils.ErrorHandling do
...
-    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name}"
-  end
-
-  def graphql_error_msg(metric_name, identifier, opts \\ []) do
+  def handle_graphql_error(metric, slug, reason, opts \\ []) do
     description = Keyword.get(opts, :description, "project with slug")
 
-    "[#{Ecto.UUID.generate()}] Can't fetch #{metric_name} for #{description}: #{identifier}"
-  end
+    error_msg = "[#{Ecto.UUID.generate()}] Can't fetch #{metric} for #{description}: #{slug}"
```
IvanIvanoff: Rename `slug` to `identifier` in this case.
Also I'm not sure that `description` is meaningful here. More like `target`?

spiderjako: You're right, target is a bit better than description in this case. Just changed them as you said.

lib/sanbase/utils/error_handling.ex
spiderjako: You're right, changed that

PR #1374, opened 2019-08-08
Persist logo fetching status

kalkov:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -68,6 +75,23 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     end
   end
 
+  defp compare_file_hashes(project, filepath) do
```
IvanIvanoff: usually the functions with `compare` return `:eq`, `:lt` or `:gt`. You can call this `file_hashes_equal?`

lib/sanbase/model/latest_coinmarketcap_data.ex
```diff
@@ -37,9 +39,24 @@ defmodule Sanbase.Model.LatestCoinmarketcapData do
...
       :percent_change_24h,
       :percent_change_7d,
+      :logo_hash,
       :update_time
     ])
     |> validate_required([:coinmarketcap_id, :update_time])
     |> unique_constraint(:coinmarketcap_id)
   end
+
+  def get_or_init(coinmarketcap_id) do
```
IvanIvanoff: let's use more common names:
image

kalkov: But is it actually creating a record?

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -68,6 +75,23 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
...
+    latest_cmc_data = LatestCoinmarketcapData.get_or_init(project.coinmarketcap_id)
+    {:ok, file_hash} = FileHash.calculate(filepath)
+
+    case latest_cmc_data.logo_hash do
+      ^file_hash ->
+        false
+
+      _ ->
+        latest_cmc_data
+        |> LatestCoinmarketcapData.changeset(%{logo_hash: file_hash, update_time: Timex.now()})
```
kalkov: `update_time` is required. It won't be clear how(current script or any other) has set the timestamp. Any suggestion how to resolve this?

IvanIvanoff: Can the changeset automatically add the update_time?

IvanIvanoff: Also, this `update_time` should ultimately be replaced by `timestamps()` which automatically handles the inserted_at and updated_at fields transparently

kalkov: It is actually storing the remote record update time: [here](https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/coinmarketcap2/ticker_fetcher.ex#L109) & [here](https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/coinmarketcap2/ticker.ex#L144).

My suggestions are:
1. Rename `update_time` to `remote_updated_at` and remove not null.
2. Add `logo_updated_at` 
3. Add the classic `updated_at` & `inserted_at`

lib/sanbase/model/latest_coinmarketcap_data.ex
tspenov: I think `create` is for when you actually create/insert in database. Here we should use `build`.
Also we can shorten it: 
```
Repo.get_by(LatestCoinmarketcapData, coinmarketcap_id: coinmarketcap_id) ||
  %LatestCoinmarketcapData{coinmarketcap_id: coinmarketcap_id}
```

lib/sanbase/utils/file_hash.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Sanbase.Utils.FileHash do
+  @hash_algorithm :sha256
+
+  @doc ~s"""
+    Receives a filepath and calculates hash using sha256 algorithm.
```
tspenov: please, remove leading space in @doc for consistency

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -48,17 +49,23 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     case Map.get(local_projects_map, slug) do
       %Project{} = project ->
         with {:ok, local_filepath_64} <- download(url, dir_path_64, filename),
-             {:ok, local_filepath_32} <-
-               resize_image(local_filepath_64, dir_path_32, filename),
+             {:logo_has_changed?, true} <-
+               {:logo_has_changed?, file_hashes_equal?(project, local_filepath_64)},
```
tspenov: should that be called `logo_changed?` instead of `file_hashes_equal?` :?

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
tspenov: I think adding `timestamps` and removing `update_time` will be suffiecient. But I think this migration is for another PR and this one can use the `update_time` as well.

kalkov: If we merge as it is the `update_time` for most of the projects will have wrong value - not when the data has  been updated remotely(its original purpose) but when the logo has been uploaded instead.

IvanIvanoff: Let's just add `logo_updated_at`?

kalkov: And remove NOT NULL from `update_time`?

IvanIvanoff: `update_time` -> `financial_data_updated_at` 🤔

But there should not be a case where you can have `update_time` NULL, only in case of some races (add new project & the logo fetcher runs before the TickerFetcher).

IvanIvanoff: BTW, on a second thought stage breaks this assumption as there we do not scrape CMC data for all projects but only for some of them (so we don't spend more credits)

kalkov: I prefer to keep them separate and depend on assumptions. 
Renaming will not be enough, bigger issues is the NOT NULL...

IvanIvanoff: let's make it not null in that case

IvanIvanoff: We just need to pattern match in `roi.ex` that the update_time is not nil (we dont' use it for anything else()

kalkov: You mean to make it null ;)

lib/sanbase/model/latest_coinmarketcap_data.ex
```diff
@@ -17,6 +18,8 @@ defmodule Sanbase.Model.LatestCoinmarketcapData do
     field(:percent_change_1h, :decimal)
     field(:percent_change_24h, :decimal)
     field(:percent_change_7d, :decimal)
+    field(:logo_hash, :string)
+    field(:logo_updated_at, Ecto.DateTime)
```
IvanIvanoff: `Ecto.DateTime` is deprecated - use `:naive_datetime` instead

lib/sanbase/model/latest_coinmarketcap_data.ex
```diff
@@ -37,9 +40,20 @@ defmodule Sanbase.Model.LatestCoinmarketcapData do
...
       :update_time
     ])
-    |> validate_required([:coinmarketcap_id, :update_time])
+    |> validate_required([:coinmarketcap_id])
     |> unique_constraint(:coinmarketcap_id)
   end
+
+  def get_or_build(coinmarketcap_id) do
+    coinmarketcap_id |> by_coinmarketcap_id ||
+      %LatestCoinmarketcapData{coinmarketcap_id: coinmarketcap_id}
```
IvanIvanoff: It's confusing as you need to know the priorities of `||` and `|>`. More readable would be, IMO:
```elixir
case by_coinmarketcap_id(coinmarketcap_id)
 nil -> %LatestCoinmarketcapData{coinmarketcap_id: coinmarketcap_id}
 data -> data
end
```

tspenov: if that is confusing then just: 
```elixir
by_coinmarketcap_id(coinmarketcap_id) || %LatestCoinmarketcapData{coinmarketcap_id: coinmarketcap_id}
```
That should be more readable than other 2 cases

kalkov: @IvanIvanoff @tspenov [Fixed](https://github.com/santiment/sanbase2/pull/1374/commits/17d3062aeb560832d7e4a6206023c2908870d0cc)

PR #1375, opened 2019-08-08
Fix infrastructure in exchangeWallets. Add allExchangeWallets API

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/exchange_address.ex
+++ b/lib/sanbase/model/exchange_address.ex
@@ -47,12 +47,18 @@ defmodule Sanbase.Model.ExchangeAddress do
 
   def exchange_names_by_infrastructure(_), do: []
 
+  def all_exchange_wallets() do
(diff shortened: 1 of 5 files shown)


lib/sanbase/model/exchange_address.ex
```diff
@@ -47,12 +47,18 @@ defmodule Sanbase.Model.ExchangeAddress do
 
   def exchange_names_by_infrastructure(_), do: []
 
+  def all_exchange_wallets() do
+    from(e in __MODULE__, preload: [:infrastructure], limit: 10_000)
```
IvanIvanoff: there is a limit because in the past when BTC exchange wallets were added they were over 1.7 million. Fetching and loading all of them into memory causes the pod to be killed because it's OOM

PR #1376, opened 2019-08-08
Secret login url

tspenov:
**Before merge needs env var in devops repo**

devops PR: https://github.com/santiment/devops/pull/956

GET `/bot/login/{secret_from_devops}` -> return JWT token


lib/sanbase_web/controllers/bot_login_controller.ex
```diff
@@ -0,0 +1,17 @@
...
+  use SanbaseWeb, :controller
+
+  require Logger
+
+  alias Sanbase.Auth.User
+
+  def index(conn, _params) do
+    user = User |> Sanbase.Repo.get_by(email: User.sanbase_bot_email())
+    {:ok, token, _claims} = SanbaseWeb.Guardian.encode_and_sign(user, %{salt: user.salt})
+    IO.inspect(token)
```
IvanIvanoff: leftover

lib/sanbase_web/router.ex
```diff
@@ -76,6 +80,16 @@ defmodule SanbaseWeb.Router do
     )
   end
 
+  scope "/", SanbaseWeb do
```
IvanIvanoff: should this be the root or it can be something like `/bot` so we do not pollute the `/` and `/login/`

tspenov: maybe `/bot` is best indeed

IvanIvanoff:
@tspenov you can link the devops PR also here

PR #1377, opened 2019-08-09
Add link to the source trigger of the signal notification

spiderjako:
The idea is to link to the source trigger so the user sees what happens

### DIFF
--- a/lib/sanbase/signals/signal.ex
+++ b/lib/sanbase/signals/signal.ex
@@ -28,6 +28,7 @@ defimpl Sanbase.Signal, for: Any do
   end
 
   def send(%{
+        id: id,
         user: user,
         trigger: %{
           settings: %{channel: "telegram", triggered?: true, payload: payload_map}
@@ -36,7 +37,14 @@ defimpl Sanbase.Signal, for: Any do
       when is_map(payload_map) do
     payload_map
     |> Enum.map(fn {slug, payload} ->
-      {slug, Sanbase.Telegram.send_message(user, payload)}
+      {slug, Sanbase.Telegram.send_message(user, extend_payload(payload, id))}
     end)
+  end
+
+  defp extend_payload(payload, user_trigger_id) do
+    """
+    #{payload}
+    The signal was triggered by #{SanbaseWeb.Endpoint.show_signal_url(user_trigger_id)}
+    """
   end
 end
--- a/lib/sanbase_web/endpoint.ex
+++ b/lib/sanbase_web/endpoint.ex
@@ -123,4 +123,8 @@ defmodule SanbaseWeb.Endpoint do
       "/verify_email?" <>
       URI.encode_query(token: email_candidate_token, emailCandidate: email_candidate)
   end
+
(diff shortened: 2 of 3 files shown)


lib/sanbase/signals/signal.ex
```diff
@@ -39,4 +39,11 @@ defimpl Sanbase.Signal, for: Any do
       {slug, Sanbase.Telegram.send_message(user, payload)}
     end)
   end
+
+  defp extend_payload(user_trigger_id, payload) do
```
IvanIvanoff: You also need to apply this function where needed

spiderjako: Yeah, will do it now, thanks :D

lib/sanbase/signals/signal.ex
```diff
@@ -35,8 +35,16 @@ defimpl Sanbase.Signal, for: Any do
       })
       when is_map(payload_map) do
     payload_map
+    |> extend_payload(:user_trigger_id)
```
IvanIvanoff: this is just passing the atom `:user_trigger_id` and not the actual id

spiderjako: I've got it now, it should be okay

IvanIvanoff:
@spiderjako can you add a test that checks that new part of the payload?

lib/sanbase_web/endpoint.ex
```diff
@@ -123,4 +123,8 @@ defmodule SanbaseWeb.Endpoint do
       "/verify_email?" <>
       URI.encode_query(token: email_candidate_token, emailCandidate: email_candidate)
   end
+
+  def show_signal_url(id) do
+    sonar_url() <> "/signal/#{id}/edit"
```
IvanIvanoff: please update to correct link

spiderjako: Alright, it shall be done

lib/sanbase/signals/signal.ex
```diff
@@ -36,7 +37,14 @@ defimpl Sanbase.Signal, for: Any do
       when is_map(payload_map) do
     payload_map
     |> Enum.map(fn {slug, payload} ->
-      {slug, Sanbase.Telegram.send_message(user, payload)}
+      {slug, Sanbase.Telegram.send_message(user, extend_payload(payload, id))}
```
tspenov: putting it directly in `send` won't put the link in the historical activities: https://app.santiment.net/sonar/activity.
I guess we can start like this and change later if that is requested.

IvanIvanoff: hm.. Question is do we need it in the historical activity and if we need to have different text?
We can easily extract `extend_payload` and reuse it

tspenov: I think its is good to link the source trigger there. Though that can be added by frontend with a relative link.

tspenov: Lets leave it like this for now.

PR #1383, opened 2019-08-12
Images for email html templates

haritonasty:


tspenov:
@haritonasty I think (not 100% sure, though) that these should be moved to `priv/static/images/emails/` in order this to work.

tspenov:
@haritonasty force-pushed your first version again because it was working.

IvanIvanoff:
Everything under `priv/static` is served from `/`. Although placing stuff under `assets` should also work because of this part of our dockerfile
```Dockerfile
COPY ./assets /app/assets
RUN cd assets && npm install
RUN cd assets && npm run build:prod
```

webpack should be smart enough to move the images to wherever needed

PR #1385, opened 2019-08-12
Sync subscriptions

tspenov:


config/scheduler_config.exs
```diff
@@ -0,0 +1,52 @@
...
+    ],
+    trending_words_sonar_signal: [
+      schedule: "5-59/5 * * * *",
+      task: {Sanbase.Signal.Scheduler, :run_signal, [Trigger.TrendingWordsTriggerSettings]}
+    ],
+    eth_wallet_signal: [
+      schedule: "5-59/5 * * * *",
+      task: {Sanbase.Signal.Scheduler, :run_signal, [Trigger.EthWalletTriggerSettings]}
+    ],
+    sync_stripe_subscriptions: [
```
IvanIvanoff: This will run in the `signals` pod, is that ok?

IvanIvanoff: Let's separate the schedulers into `Sanbase.Scrapers.Scheduler` and `Sanbase.Signals.Scheduler` and start the appropriate ones in the supervision tree. Also we need to change the schedulers  in the config

tspenov:
devops PR: https://github.com/santiment/devops/pull/962

config/scheduler_config.exs
tspenov: split the signals and scrapers and put in scrapers, though it doesn't matter for this one cause it needs only ecto

PR #1387, opened 2019-08-12
Trending words V2 API

IvanIvanoff:
#### Summary

Request:
```graphql
{
  getTrendingWords(
    from: "2019-08-12T00:00:00Z", 
    to: "2019-08-13T00:00:00Z", 
    interval: "1h", 
    size: 2) {
    datetime
    topWords {
      word
      score
    }
  }
}
```

Response:
```graphql
{
  "data": {
    "getTrendingWords": [
      {
        "datetime": "2019-08-12T00:00:00Z",
        "topWords": [
          {
            "score": 318.5999755859375,
            "word": "simpleledger"
          },
          {
            "score": 231.0018548965454,
            "word": "matic"
          }
        ]
      },
      {
        "datetime": "2019-08-12T01:00:00Z",
        "topWords": [
          {
            "score": 319.0608215332031,
            "word": "simpleledger"
          },
          {
            "score": 240.40638625621796,
            "word": "matic"
          }
        ]
      },
...
```

---
Request:
```graphql
{
  getWordTrendingHistory(word: "matic", from: "2019-08-12T00:00:00Z", to: "2019-08-13T00:00:00Z", interval: "1h", size: 10) {
    datetime
    position
  }
}
```

Response:
```graphql
{
  "data": {
    "getWordTrendingHistory": [
      {
        "datetime": "2019-08-12T00:00:00Z",
        "position": 9
      },
      {
        "datetime": "2019-08-12T01:00:00Z",
        "position": 9
      },
      {
        "datetime": "2019-08-12T02:00:00Z",
        "position": 9
      },
      {
        "datetime": "2019-08-12T03:00:00Z",
        "position": 9
      },
...
```

Request:
```graphql
{
  getProjectTrendingHistory(slug: "bitcoin-cash", from: "2019-08-12T00:00:00Z", to: "2019-08-13T00:00:00Z", interval: "1h", size: 10) {
    datetime
    position
  }
}
```

Response:
```graphql
{
  "data": {
    "getProjectTrendingHistory": [
      {
        "datetime": "2019-08-12T00:00:00Z",
        "position": 1
      },
      {
        "datetime": "2019-08-12T01:00:00Z",
        "position": 2
      },
      {
        "datetime": "2019-08-12T02:00:00Z",
        "position": 5
      },
      {
        "datetime": "2019-08-12T03:00:00Z",
        "position": 1
      },
...
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/social_data/trending_words/trending_words.ex
```diff
@@ -0,0 +1,152 @@
...
+    FROM(
+      SELECT
+        toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS t,
+        argMaxIf(word, dt, dt = t) AS last_word,
+        argMaxIf(score, dt, dt = t) AS last_score
+      FROM(
+        SELECT
+          dt,
+          word,
+          SUM(score) AS score
```
IvanIvanoff: This one is calculating the score for every hour but we might not need them and discard them (if we query with 1 day interval we'll get only the first or last result from every bucket

IvanIvanoff: Pushed a rework that simplifies the query

lib/sanbase/social_data/trending_words/schema.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Sanbase.SocialData.TrendingWords.Schema do
```
tspenov: Lets keep the schema in the trending_words module along with the functionality. Its sort of a documentation cause its not used otherwise I think.

lib/sanbase/social_data/trending_words/trending_words.ex
```diff
@@ -0,0 +1,201 @@
...
+  @spec get_trending_words(DateTime.t(), DateTime.t(), interval, non_neg_integer) ::
+          {:ok, list(trending_word)} | {:error, String.t()}
+  def get_trending_words(from, to, interval, size) do
+    {query, args} = get_trending_words_query(from, to, interval, size)
+
+    ClickhouseRepo.query_reduce(query, args, %{}, fn
+      [dt, word, _project, score], acc ->
+        datetime = DateTime.from_unix!(dt)
+        elem = %{word: word, score: score}
+        Map.update(acc, datetime, [elem], fn words -> [elem | words] end)
```
tspenov: do we need specific ordering in the elements of the list value for the `datetime` key ?

lib/sanbase/social_data/trending_words/trending_words.ex
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  @doc ~s"""
+  Get a list of the currently trending words
+  """
+  @spec get_trending_now(non_neg_integer()) :: {:ok, list(trending_word)} | {:error, String.t()}
+  def get_trending_now(size \\ 10)
+
+  def get_trending_now(size) do
+    case get_trending_words(Timex.shift(Timex.now(), hours: -6), Timex.now(), "1h", size) do
```
tspenov: lets extract `-6` as constant, also use same `now` in from and to and add a comment. BTW why did you chose 6h here ?

lib/sanbase/social_data/trending_words/trending_words.ex
```diff
@@ -0,0 +1,201 @@
...
+        """
+        )
+        GROUP BY t
+        ORDER BY t
+        """
+      ]
+      |> to_string()
+
+    ticker =
+      Sanbase.Model.Project.ticker_by_slug(slug) |> IO.inspect(label: "195", limit: :infinity)
```
tspenov: lefover inspect

lib/sanbase/social_data/trending_words/trending_words.ex
```diff
@@ -0,0 +1,201 @@
...
+    {query, args}
+  end
+
+  defp project_trending_history_query(slug, from, to, interval, size) do
+    IO.inspect(
+      label:
+        "message; #{
+          String.replace_leading("#{__ENV__.file}", "#{File.cwd!()}", "") |> Path.relative()
+        }:#{__ENV__.line()}"
+    )
```
tspenov: leftover

test/sanbase/social_data/trending_words/trending_words_test.exs
```diff
@@ -0,0 +1,218 @@
...
+
+  import Mock
+  import Sanbase.Factory
+  import Sanbase.DateTimeUtils, only: [from_iso8601_to_unix!: 1, from_iso8601!: 1]
+
+  alias Sanbase.SocialData.TrendingWords
+
+  setup do
+    [
+      dt1_str: "2019-01-01T00:00:00Z",
```
kalkov: can we have better names here? 
`datetime1` or `datetime1_str` ...

test/sanbase/social_data/trending_words/trending_words_test.exs
```diff
@@ -0,0 +1,218 @@
...
+            "1h",
+            10
+          )
+
+        assert result == {:error, "Something went wrong"}
+      end
+    end
+  end
+
+  defp word_trending_history_rows(dt1_str, dt2_str, dt3_str) do
```
kalkov: Out of curiosity, these helper functions can be on top of the file close to setup?

lib/sanbase/social_data/trending_words/schema.ex
IvanIvanoff: I was experimenting with this. I think it's a preferred pattern in the Elixir community to have the schema in a file without anything else in it.
Furthermore, us having schemas for CH is kind of weird because we never use them

lib/sanbase/social_data/trending_words/trending_words.ex
IvanIvanoff: We do - we fetched them in ASC order by score and here we're basically reversing them, so the top score is in first place, which is what we want :smiley:

lib/sanbase/social_data/trending_words/trending_words.ex
IvanIvanoff: No particular reason.. Just to ensure that we have something calculated, probably 2-3 hours is enough, too

test/sanbase/social_data/trending_words/trending_words_test.exs
IvanIvanoff: We have them at the end everywhere as we have in all modules the private functions at the bottom. When you want to understand what a module does you don't need to read the private functions - same with the tests

kalkov: Well, you need to read them here. How can you understand what is happening if you don't check the `trending_words_rows`

```elixir
test "clickhouse returns data", context do
      %{dt1_str: dt1_str, dt2_str: dt2_str, dt3_str: dt3_str} = context

      with_mock Sanbase.ClickhouseRepo,
        query: fn _, _ ->
          {:ok, %{rows: trending_words_rows(dt1_str, dt2_str, dt3_str)}}
        end do
        result =
          TrendingWords.get_trending_words(
            from_iso8601!(dt1_str),
            from_iso8601!(dt3_str),
            "1d",
            2
          )

        assert result ==
                 {:ok,
                  %{
                    from_iso8601!(dt1_str) => [
                      %{score: 5, word: "ethereum"},
                      %{score: 10, word: "bitcoin"}
                    ],
                    from_iso8601!(dt2_str) => [
                      %{score: 70, word: "boom"},
                      %{score: 2, word: "san"}
                    ],
                    from_iso8601!(dt3_str) => [
                      %{score: 2, word: "xrp"},
                      %{score: 1, word: "eth"}
                    ]
                  }}
      end
    end
```

lib/sanbase/social_data/trending_words.ex
```diff
@@ -22,59 +18,201 @@ defmodule Sanbase.SocialData.TrendingWords do
...
   @spec get_trending_now(non_neg_integer()) :: {:ok, list(trending_word)} | {:error, String.t()}
   def get_trending_now(size \\ 10)
 
   def get_trending_now(size) do
-    now = Timex.now()
-
-    {from, to, hour} = get_trending_word_query_params(now)
+    case get_trending_words(
+           Timex.shift(Timex.now(), hours: -@hours_back_ensure_has_data),
+           Timex.now(),
```
tspenov: Lets use the same `now` for both from and to

lib/sanbase/social_data/trending_words.ex
```diff
@@ -22,59 +18,201 @@ defmodule Sanbase.SocialData.TrendingWords do
...
+
+    args = [str_to_sec(interval), from, to, size]
+
+    {query, args}
+  end
+
+  defp word_trending_history_query(word, from, to, interval, size) do
+    {query, args} = get_trending_words_query(from, to, interval, size)
+    args_len = length(args)
+    next_pos = args_len + 1
```
tspenov: I think better name is `word_arg_pos` cause it will make it easier to get that here we're calculating the binding position of the `word` arg :?

lib/sanbase/social_data/trending_words.ex
```diff
@@ -22,59 +18,201 @@ defmodule Sanbase.SocialData.TrendingWords do
...
+      |> to_string()
+
+    args = args ++ [word]
+    {query, args}
+  end
+
+  defp project_trending_history_query(slug, from, to, interval, size) do
+    {query, args} = get_trending_words_query(from, to, interval, size)
+    args_len = length(args)
+    next_pos = args_len + 1
```
tspenov: same here `next_pos` -> `ticker_slug_arg_pos`

lib/sanbase_web/graphql/schema/queries/social_data_queries.ex
```diff
@@ -70,6 +70,78 @@ defmodule SanbaseWeb.Graphql.Schema.SocialDataQueries do
...
+      arg(:from, non_null(:datetime))
+      arg(:to, non_null(:datetime))
+      arg(:interval, :string, default_value: "1d")
+      arg(:size, non_null(:integer))
+
+      complexity(&Complexity.from_to_interval/3)
+      middleware(TimeframeRestriction, %{allow_realtime_data: true})
+
+      cache_resolve(&SocialDataResolver.get_project_trending_history/3,
+        ttl: 600,
```
tspenov: trending words are calculated every 1h. I guess we can put even 30 minutes here :?

lib/sanbase_web/graphql/schema/queries/social_data_queries.ex
```diff
@@ -70,6 +70,78 @@ defmodule SanbaseWeb.Graphql.Schema.SocialDataQueries do
...
+        max_ttl_offset: 240
```
tspenov: what is `max_ttl_offset` doing ? I see it is used as some range for the hash func `erlang:phash2` ?

lib/sanbase/social_data/trending_words.ex
IvanIvanoff: Not sure.. we're just calculating what the next position is and it's up to `args` to specify what to put there

lib/sanbase_web/graphql/schema/queries/social_data_queries.ex
IvanIvanoff: If `max_ttl_offset` is 0 the cache will expire exactly after 600 seconds. If the `max_ttl_offset` is 240, the cache will expire after `600 + almost_rand(0, 240)` seconds.
It's almost_rand because the same set of arguments/params will always resolve to the same number.

For example, if you query `allProjects{ ethSpent }` every `ethSpent` is first cached separately and then the whole `allProjects` query is cached.

When `allProjects` expires, it will try re-calculate the `ethSpent`. If we don't have random expiration time we'll need to recalculate all eth spent at the same time. With random TTL only some of the ethSpent will expire and we'll recalculate ~25% of them, then after some time, we'll recalculate another ~25% of them and so on.

lib/sanbase_web/graphql/schema/queries/social_data_queries.ex
IvanIvanoff: We can, but if we cache the value at 12:59, we won't get the new words after 1 minute, but after 29 minutes, so we're introducing a big artificial lag. IMO we can even reduce the TTL here

lib/sanbase/social_data/trending_words.ex
tspenov: ok, lets leave as it is

PR #1392, opened 2019-08-14
Rework eth_wallet signal

IvanIvanoff:
#### Summary
Take the previous balance (time_window time ago) instead of the balance
when the signal was previously triggered

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/metric/available_v2_metrics.json
```diff
@@ -276,6 +276,7 @@
   },
   {
     "metric": "stack_realized_cap_usd_20y",
+    "alias": "realized_value_usd",
```
tspenov: that should be in another PR :?

IvanIvanoff: yes

PR #1396, opened 2019-08-15
Change active subscriptions query

tspenov:
Use Stripe status to determine if a subscription is active. It is active if it isn't cancelled.
Subscription can be cancelled in those 3 cases:

* User cancel from account. It is cancelled by Stripe when `current_period_end` is reached.
* Payment failed and Stripe retry scheme failed. It will be cancelled after max retries reached.
* Manually cancel subscription in Stripe

### DIFF
--- a/lib/sanbase/billing/subscription.ex
+++ b/lib/sanbase/billing/subscription.ex
@@ -24,12 +24,6 @@ defmodule Sanbase.Billing.Subscription do
   Please, contact administrator of the site for more information.
   """
 
-  # After `current_period_end` timestamp passes there is some time until `invoice.payment_succeeded` event is generated
-  # to update the field with new timestamp value. So we add 1 day gratis in which we should receive payment before
-  # we decide that this subscription is not active.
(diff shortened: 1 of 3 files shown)


lib/sanbase/billing/subscription.ex
```diff
@@ -352,13 +346,8 @@ defmodule Sanbase.Billing.Subscription do
     )
   end
 
-  # current_period_end > Timex.now - gratis days
   defp active_subscriptions_query(query) do
-    from(s in query,
-      where:
-        s.status != "canceled" and
-          s.current_period_end > ^Timex.shift(Timex.now(), days: -@subscription_gratis_days)
```
IvanIvanoff: btw, we could have changed to `or` and it will work? Does stripe provide this gratis period?

tspenov: It won't be correct.If it is cancelled there is no way back but to create another subscription. Stripe has gratis period (its configurable) for payment retries until it finally cancels your account. Status goes `active` -> `past_due` -> `cancelled`

PR #1402, opened 2019-08-15
Promo subscription

tspenov:


lib/sanbase/billing/subscription.ex
```diff
@@ -275,11 +278,49 @@ defmodule Sanbase.Billing.Subscription do
 
   def plan_name(subscription), do: subscription.plan.name
 
+  def promo_subscription(user) do
+    unless user.stripe_customer_id do
```
IvanIvanoff: can we handle this with a pattern match in the function head? What bothers me in such situations (all `if`s that are not the last expression) is that if you get an error on line 283, you won't know if this `create` was called or not - which is not the case if we have pattern match

lib/sanbase/billing/subscription.ex
```diff
@@ -275,11 +278,49 @@ defmodule Sanbase.Billing.Subscription do
...
+    end
+
+    promo_subscribe(user)
+  end
+
+  defp promo_subscribe(user) do
+    from(p in Plan, where: p.id in ^@promo_plans)
+    |> Repo.all()
+    |> Enum.map(&promo_subscribe(user, &1))
+    |> Enum.all?(&match?({:ok, _}, &1))
```
IvanIvanoff: What about calling `Enum.filter(&match?(:error, _}, &1)` and pattern match `[]` for success case and `list` for failure, so we can log the stuff the failed?

lib/sanbase/billing/subscription.ex
```diff
@@ -275,11 +278,49 @@ defmodule Sanbase.Billing.Subscription do
...
+      true ->
+        user_subscriptions(user)
+
+      false ->
+        {:error, @generic_error_message}
+    end
+  end
+
+  defp promo_subscribe(user, plan) do
+    data = %{
```
IvanIvanoff: this can be a function with a nice name :?

lib/sanbase/billing/subscription.ex
```diff
@@ -275,11 +278,49 @@ defmodule Sanbase.Billing.Subscription do
...
+    from(p in Plan, where: p.id in ^@promo_plans)
+    |> Repo.all()
+    |> Enum.map(&promo_subscribe(user, &1))
+    |> Enum.all?(&match?({:ok, _}, &1))
+    |> case do
+      true ->
+        user_subscriptions(user)
+
+      false ->
+        {:error, @generic_error_message}
```
IvanIvanoff: Can't we return something better than `generic_error_message`?

lib/sanbase/billing/subscription.ex
```diff
@@ -17,6 +17,9 @@ defmodule Sanbase.Billing.Subscription do
 
   require Logger
 
+  @promo_plans [5, 14, 24]
```
kalkov: What are these numbers? A comment would help.

lib/sanbase_web/graphql/schema/queries/billing_queries.ex
```diff
@@ -80,5 +83,13 @@ defmodule SanbaseWeb.Graphql.Schema.BillingQueries do
 
       resolve(&BillingResolver.renew_cancelled_subscription/3)
     end
+
+    if Config.module_get(Sanbase.Billing.Subscription, :enable_promo_subscription) == "true" do
```
tspenov: Maybe we can make that better with some macro that takes only the `:enable_promo_subscription` and checks

lib/sanbase/billing/subscription.ex
```diff
@@ -148,6 +148,32 @@ defmodule Sanbase.Billing.Subscription do
     end
   end
 
+  @doc """
+  Attach a subscription with a free trial of #{@promo_trial_period_days} days for all highest priced plans for all
+  products to the user. It doesn't require a credit card.
+  """
+  def promo_subscription(%User{stripe_customer_id: stripe_customer_id} = user)
```
IvanIvanoff: This file is starting to get big, how about `Subscription.Promo` module :?

lib/sanbase_web/graphql/schema/queries/billing_queries.ex
IvanIvanoff: If we put all such configs under the same config module `Sanbase.FeatureFlag` we can abstract this, but if every feature flag is part of different module config - not really

IvanIvanoff: We could use https://github.com/tompave/fun_with_flags#boolean-gate under the hood to automate that.
The bigger issue is that we will have to recompile the whole project in order to show/hide the feature. In such cases I'm not sure it's better than just not adding the query?

lib/sanbase/billing/subscription.ex
tspenov: not anything meaningfull - the request will fail if something unexpected like network error or db error.

lib/sanbase_web/graphql/schema/queries/billing_queries.ex
tspenov: yeah we won't be able to change the behaviour without recompiling. I might change it
`enable_promo_subscription: {:system, "ENABLE_PROMO_SUBSCRIPTION", "false"}` -> `enable_promo_subscription: "false"` to remind that changing with env var doesn't change anything but leave the flag just so we have diff behaviour for test and prod env

tspenov: Other way is to put the query in the schema but use the feature flag in the resolver

kalkov:
Not enough tests here, if something goes wrong we can potentially loose money.

I would leave the feature off in test environment like in the others and assert the it is forbidden. Then for testing the happy path would enable the feature just for a single test case.

lib/sanbase/billing/subscription/promo.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Sanbase.Billing.Subscription.Promo do
```
IvanIvanoff: can you add a moduledoc explaining what is a promo subscription

lib/sanbase/billing/subscription/promo.ex
```diff
@@ -0,0 +1,83 @@
...
+        promo_subscribe(user)
+
+      {:error, error} ->
+        Logger.error(
+          "Error creating promotional subscription for user: #{inspect(user)}, reason: #{
+            inspect(error)
+          }"
+        )
+
+        {:error, Subscription.generic_error_message()}
```
IvanIvanoff: why generic error? Can't we be more specific here?

lib/sanbase/billing/subscription/promo.ex
```diff
@@ -0,0 +1,83 @@
...
+  defp promo_subscribe(user) do
+    from(p in Plan, where: p.id in ^@promo_plans)
+    |> Repo.all()
+    |> Enum.map(&promo_subscribe(user, &1))
+    |> Enum.filter(&match?({:error, _}, &1))
+    |> case do
+      [] ->
+        {:ok, Subscription.user_subscriptions(user)}
+
+      errors ->
```
IvanIvanoff: you can add a comment that we do this because we're subscribing for 3 plans and possibly 1, 2 or all 3 could fail

lib/sanbase/billing/subscription/promo.ex
```diff
@@ -0,0 +1,83 @@
...
+  @doc """
+  Attach a subscription with a free trial of #{@promo_trial_period_days} days for all highest priced plans for all
+  products to the user. It doesn't require a credit card.
+  """
+  def promo_subscription(%User{stripe_customer_id: stripe_customer_id} = user)
+      when is_binary(stripe_customer_id) do
+    promo_subscribe(user)
+  end
+
+  def promo_subscription(%User{} = user) do
```
IvanIvanoff: `create_promo_subscription` - let's follow a convention where if there's some action the function name should have a verb

lib/sanbase/billing/subscription/promo.ex
```diff
@@ -0,0 +1,83 @@
...
+  alias Sanbase.Billing.{Subscription, Plan}
+  alias Sanbase.Repo
+
+  @promo_trial_period_days 14
+  @doc """
+  Enterprise plans from API, Sanbase and Sheets products are used for #{@promo_trial_period_days} days
+  promotional free trial of our services.
+  """
+  @promo_plans [5, 14, 24]
+  @promo_metadata %{"promo_event" => "devcon2019"}
```
IvanIvanoff: hardcode 🤔

lib/sanbase_web/graphql/schema/queries/billing_queries.ex
```diff
@@ -80,5 +80,11 @@ defmodule SanbaseWeb.Graphql.Schema.BillingQueries do
 
       resolve(&BillingResolver.renew_cancelled_subscription/3)
     end
+
+    field :promo_subscription, list_of(:subscription_plan) do
```
IvanIvanoff: let's prefix `promo_subscription` with something - activate/get/create?

IvanIvanoff: Shouldn't we provide a promo code with an argument? Or every one can create a promo (trial) account?

lib/sanbase/billing/subscription/promo.ex
tspenov: not anything meaningfull - the request will fail if something unexpected like network error or db error.

tspenov:
This is the final implementation: https://github.com/santiment/sanbase2/pull/1475

PR #1405, opened 2019-08-16
Use diff templates first login

tspenov:


lib/sanbase/auth/user.ex
```diff
@@ -437,13 +437,22 @@ defmodule Sanbase.Auth.User do
     count_other_accounts > 0 or not is_nil(email)
   end
 
-  defp choose_login_template(origin_url) when is_binary(origin_url) do
+  defp choose_login_template(origin_url, first_login?: true) when is_binary(origin_url) do
```
IvanIvanoff: I think it's worth having the template/send_login_email login in a separate module

tspenov: Extracted - it made it easier for testing cause I tested only the pure functions without having to mock and assert that given function is called with given args.

test/sanbase/auth/user_test.exs
```diff
@@ -1,7 +1,8 @@
 defmodule Sanbase.Auth.UserTest do
   use Sanbase.DataCase, async: false
 
-  import Mockery
+  use Mockery
```
tspenov:
```suggestion
 import Mockery
```

PR #1407, opened 2019-08-16
Add handling of negative returned values from historical balance

spiderjako:
There are certain cases, in which the returned values are unexpectedly negative and we want to avoid that.
https://api.santiment.net/graphiql?variables=%7B%7D&query=%7B%0A%20%20assetsHeldByAddress(address%3A%20%220x0000000000000000000000000000000000000000%22)%20%7B%0A%20%20%20%20balance%0A%20%20%20%20slug%0A%20%20%7D%0A%7D%0A

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
@@ -324,6 +324,15 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     HistoricalBalance.assets_held_by_address(address)
     |> case do
       {:ok, result} ->
+        # We do this, because many contracts emit a transfer
+        # event when minting new tokens by setting 0x00...000
+        # as the from address, hence 0x00...000 is "sending"
+        # tokens it does not have which leads to "negative" balance
+
+        result =
+          result
+          |> Enum.reject(fn %{balance: balance} -> balance < 0 end)
+
         {:ok, result}
 
       {:error, error} ->
--- a/test/sanbase_web/graphql/clickhouse/historical_balance/assets_held_by_address_api_test.exs
+++ b/test/sanbase_web/graphql/clickhouse/historical_balance/assets_held_by_address_api_test.exs
@@ -22,7 +22,7 @@ defmodule SanbaseWeb.Graphql.Clickhouse.AssetsHeldByAdderssApiTest do
        assets_held_by_address: fn _ ->
          {:ok,
           [
-            %{balance: 100.0, slug: context.p1.coinmarketcap_id},
+            %{balance: -100.0, slug: context.p1.coinmarketcap_id},
             %{balance: 200.0, slug: context.p2.coinmarketcap_id}
           ]}
        end},
@@ -44,7 +44,6 @@ defmodule SanbaseWeb.Graphql.Clickhouse.AssetsHeldByAdderssApiTest do
                "data" => %{
                  "assetsHeldByAddress" => [
                    %{"balance" => 1.0e3, "slug" => context.eth_project.coinmarketcap_id},
-                   %{"balance" => 100.0, "slug" => context.p1.coinmarketcap_id},
                    %{"balance" => 200.0, "slug" => context.p2.coinmarketcap_id}
                  ]
                }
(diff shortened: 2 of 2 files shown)


IvanIvanoff:
Just a note: The negative balances are not a bug, but rather a consequence of our implementation of the balance calculation. Many contracts emit a transfer event when minting new tokens by setting 0x00...000 as the from address, hence 0x00...000 is "sending" tokens it does not have which leads to "negative" balance

test/sanbase_web/graphql/clickhouse/historical_balance/assets_held_by_address_api_test.exs
```diff
@@ -44,7 +44,6 @@ defmodule SanbaseWeb.Graphql.Clickhouse.AssetsHeldByAdderssApiTest do
                "data" => %{
                  "assetsHeldByAddress" => [
                    %{"balance" => 1.0e3, "slug" => context.eth_project.coinmarketcap_id},
-                   %{"balance" => 100.0, "slug" => context.p1.coinmarketcap_id},
```
IvanIvanoff: can we have a test that is explicitly named "negative balances are discarded" :?

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -324,6 +324,11 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     HistoricalBalance.assets_held_by_address(address)
     |> case do
       {:ok, result} ->
+        result =
+          result
+          |> Enum.reject(fn row -> Map.get(row, :balance) < 0 end)
+          |> Enum.to_list()
```
IvanIvanoff: The functions in the `Enum` module already return lists, no point in calling `to_list` here

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -324,6 +324,11 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     HistoricalBalance.assets_held_by_address(address)
     |> case do
       {:ok, result} ->
+        result =
```
IvanIvanoff: You can add a comment why we do this - check my note in the other comment describing the reason

test/sanbase_web/graphql/clickhouse/historical_balance/assets_held_by_address_api_test.exs
spiderjako: Yes, that's a better idea

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: I was wondering that :D thanks, will be changed

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: Of course, it shall be done

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -324,6 +324,17 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
     HistoricalBalance.assets_held_by_address(address)
     |> case do
       {:ok, result} ->
+        """
```
IvanIvanoff: This is not a comment but a string declaration. For comments you should prepend `#` to every line

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
```diff
@@ -324,6 +324,17 @@ defmodule SanbaseWeb.Graphql.Resolvers.ClickhouseResolver do
...
+        We do this, because many contracts emit a transfer
+        event when minting new tokens by setting 0x00...000
+        as the from address, hence 0x00...000 is "sending"
+        tokens it does not have which leads to "negative" balance
+        """
+
+        result =
+          result
+          |> Enum.reject(fn row -> Map.get(row, :balance) < 0 end)
```
IvanIvanoff:
```suggestion
          |> Enum.reject(fn %{balance: balance} -> balance < 0 end)
```

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: Oh, that's right, thanks

lib/sanbase_web/graphql/resolvers/clickhouse_resolver.ex
spiderjako: That's a better idea really, thanks, it should be changed now

PR #1410, opened 2019-08-19
Add active users statistics

spiderjako:


lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -21,6 +21,17 @@ defmodule Sanbase.Clickhouse.ApiCallData do
     end)
   end
 
+  def api_call_history_all_users(from, to, interval) do
```
IvanIvanoff: `active_users`

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -219,4 +220,14 @@ defmodule Sanbase.Statistics do
       "longtime_users_with_watchlist_count_14d" => old_users_with_watchlist_count_14d
     }
   end
+
+  def get("active_users") do
+    now = Timex.now()
+
+    active_users_in_last_12h = ApiCallData.api_call_history_all_users(Timex.shift(now, hours: -12), now, "1h")
```
IvanIvanoff: You need a single number of the active users, here you get a list of `%{datetime: dt, count: count}`

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -21,6 +21,17 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+    {query, args} = api_call_history_all_users_query(from, to, interval)
+
+    ClickhouseRepo.query_transform(query, args, fn [count, t] ->
```
IvanIvanoff: do not name parameters `t` - use meaningful names

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +58,27 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
     {query, args}
   end
+
+  defp api_call_history_all_users_query(from, to, interval) do
+    interval_sec = Sanbase.DateTimeUtils.str_to_sec(interval)
+
+    query = """
+    SELECT
+      COUNT(user_id),
+      toUnixTimestamp(intDiv(toUInt32(dt), ?1) * ?1) AS t
```
IvanIvanoff: remove this line and select only the count.

spiderjako: Some of the things I realised while I was writing the tests, nonetheless thanks

lib/sanbase/clickhouse/api_call_data.ex
spiderjako: Will change the names, thanks for the heads-up

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +52,26 @@ defmodule Sanbase.Clickhouse.ApiCallData do
 
     {query, args}
   end
+
+  defp api_call_history_active_users_query(from, to, interval) do
+    interval_sec = Sanbase.DateTimeUtils.str_to_sec(interval)
+
+    query = """
+    SELECT
+      COUNT(user_id)
```
IvanIvanoff: Is `count` the function you really need?

spiderjako: Shouldn't the query return the number of people, who made a query in the past 12 hours? If that is the case, then count should work. Do you have something in mind?

IvanIvanoff: It should return that but it won't

spiderjako: Got it, it should be uniqExact, thanks

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -1,5 +1,6 @@
 defmodule Sanbase.Statistics do
   import Sanbase.Auth.Settings, only: [daily_subscription_type: 0, weekly_subscription_type: 0]
+  alias Sanbase.Clickhouse.ApiCallData, as: ApiCallData
```
IvanIvanoff: Do not provide `:as` if you're not really changing the alias name

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -219,4 +221,15 @@ defmodule Sanbase.Statistics do
       "longtime_users_with_watchlist_count_14d" => old_users_with_watchlist_count_14d
     }
   end
+
+  def get("active_users") do
+    now = Timex.now()
+
+    active_users_in_last_12h =
+      ApiCallData.api_call_history_active_users(Timex.shift(now, hours: -12), now, "1h")
```
IvanIvanoff: As you're using `now` in only one place - just provide `Timex.now()` as argument

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -219,4 +221,15 @@ defmodule Sanbase.Statistics do
...
+
+    %{
+      "active_users_in_last_12h" => active_users_in_last_12h
```
IvanIvanoff: let's also add active in the last 24h, 7d, 30d

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -21,6 +21,11 @@ defmodule Sanbase.Clickhouse.ApiCallData do
     end)
   end
 
+  def api_call_history_active_users(from, to, interval) do
+    {query, args} = api_call_history_active_users_query(from, to, interval)
+    ClickhouseRepo.query(query, args)
```
IvanIvanoff: This will return `{:ok, [count]}` (list of one elements) - you should do a transformation (check other places )

test/sanbase/statistics/statistics_test.exs
```diff
@@ -119,6 +120,18 @@ defmodule Sanbase.StatisticsTest do
             }} in statistics
   end
 
+  test "active users statistics" do
+    with_mock Sanbase.ClickhouseRepo,
+      query: fn _, _ -> 3 end do
```
IvanIvanoff: please check the response format of the `query` function

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: Alright, it will be changed

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: It will be in one place for now, but in the future, it could have more than just one statistic trait

lib/sanbase/automatic_statistics/statistics.ex
spiderjako: Okay, this kind of makes the comment above obsolete, it will be done

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -21,6 +21,13 @@ defmodule Sanbase.Clickhouse.ApiCallData do
     end)
   end
 
+  def api_call_history_active_users(from, to, interval) do
+    {query, args} = api_call_history_active_users_query(from, to, interval)
+
+    ClickhouseRepo.query(query, args)
+    |> transform_query_result()
```
IvanIvanoff: Just do
```elixir
 ClickhouseRepo.query_transform(query, args, fn [value] -> value end)
    |> case do
      {:ok, [result]} -> {:ok, result}
      error -> error
    end
```

`transform_query_result` is too generic

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +54,32 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+      to
+    ]
+
+    {query, args}
+  end
+
+  defp transform_query_result(query_result) do
+    elem(query_result, 1)
+    |> Map.get(:rows)
+    |> Enum.at(0)
```
IvanIvanoff: Prefer pattern matching over such code.

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +54,32 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+      from,
+      to
+    ]
+
+    {query, args}
+  end
+
+  defp transform_query_result(query_result) do
+    elem(query_result, 1)
+    |> Map.get(:rows)
```
IvanIvanoff: ClickhouseRepo.query returns `:rows` - `query_transform` does not

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +54,32 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+
+    query = """
+    SELECT
+      uniqExact(user_id)
+    FROM
+      sanbase_api_call_data
+    PREWHERE
+      dt >= toDateTime(?2) AND
+      dt < toDateTime(?3)
+    """
```
tspenov: you have 3 args but only 2 bound variables

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -47,4 +54,32 @@ defmodule Sanbase.Clickhouse.ApiCallData do
 
     {query, args}
   end
+
+  defp api_call_history_active_users_query(from, to, interval) do
```
tspenov: I don't think interval is needed for this case - you just need from and to.
You can have another functions that returns api usage count split in intervals but I don't think that is needed for this task.

spiderjako: You're right, it's something like a leftover, thanks that you pointed that out

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -21,6 +21,13 @@ defmodule Sanbase.Clickhouse.ApiCallData do
     end)
   end
 
+  def api_call_history_active_users(from, to, interval) do
```
IvanIvanoff: Avoid adding too much information that's deducible from the module name - in the module `ApiCallData` we can call this function just `active_users_count` will be good, as it will be used like this: `ApiCallData.active_users_count` which sounds much better than `ApiCallData.api_call...`

spiderjako: You're right. It is truly a complicated name for a function. Just changed it, and changed the stuff above like you said.

lib/sanbase/automatic_statistics/statistics.ex
```diff
@@ -219,4 +221,23 @@ defmodule Sanbase.Statistics do
...
+  def get("active_users") do
+    now = Timex.now()
+
+    active_users_in_last_12h = ApiCallData.active_users_count(Timex.shift(now, hours: -12), now)
+
+    active_users_in_last_24h = ApiCallData.active_users_count(Timex.shift(now, days: -1), now)
+
+    active_users_in_last_7d = ApiCallData.active_users_count(Timex.shift(now, days: -7), now)
+
+    active_users_in_last_30d = ApiCallData.active_users_count(Timex.shift(now, days: -30), now)
```
IvanIvanoff: 180d and overall :? 😃

spiderjako: Alright, will add them. :D

PR #1417, opened 2019-08-22
Remove check for valid notfication channels when getting active triggers

IvanIvanoff:
#### Summary
We need to remove this because even if there's no valid notification
channel the user can see the historical activity in his activity feed.
Furthermore, the check was stopping running triggers with a list of notification
channels

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


GarageInc:
Looks strange(removing validation), but 'Approve' :)

IvanIvanoff:
@GarageInc These checks were done in `get_active_signals`, which is a weird place for them. We have validation for the channel in create/update of triggers

PR #1418, opened 2019-08-22
Add 'all_of' operation

spiderjako:


lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,23 @@ defmodule Sanbase.Signal.Validation.Operation do
...
     do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+
+  defp operator_checker(list, operator) do
+    Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+    |> case do
+      true ->
+        :ok
+
+      false ->
+        {:error, "#{inspect(operator)} is not a valid percent change operation"}
```
tspenov: https://github.com/santiment/sanbase2/pull/1424#discussion_r319839396

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,23 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    end
+
+    Enum.map(list, &Operation.type/1)
```
tspenov: that should be a named helper function - `if all_operations_have_same_type?, do: :ok, else: :error` and maybe a comment why that is required ?

spiderjako: Alright, I will do it so, thanks for the feedback!

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,26 @@ defmodule Sanbase.Signal.Validation.Operation do
 
   defp do_valid_absolute_value_operation?(operation, _),
     do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+
+  defp operator_checker(list, operator) do
```
IvanIvanoff: I'm not sure `operator_checker` is a meaningful name here

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,26 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+    |> case do
+      true ->
+        :ok
+
+      false ->
+        {:error, "#{inspect(operator)} is not a valid operation"}
+    end
```
IvanIvanoff: This case has no effect here

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,26 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+
+    all_operations_have_same_type?(list)
```
IvanIvanoff: This function call is single-handily defining the result of the function

test/sanbase/signals/price/trigger_price_percent_change_test.exs
```diff
@@ -135,6 +135,24 @@ defmodule Sanbase.Signal.TriggerPricePercentChangeTest do
...
+    [triggered | rest] =
+      PricePercentChangeSettings.type()
+      |> UserTrigger.get_active_triggers_by_type()
+      |> Evaluator.run()
+
+    assert rest == []
+    assert Trigger.triggered?(triggered.trigger) == true
+    assert triggered.id == trigger.id
+  end
+
```
IvanIvanoff: add tests for the validations, too - currently, they're not working well but the lack of tests makes the build pass

lib/sanbase/signals/trigger/validation/operation_validation.ex
tspenov: That should be rewritten with `with` checking first `all_operations_have_same_type?` and then if `all_operations_are_valid?` and return corresponding :error tuples if any of these is not.

lib/sanbase/signals/trigger/validation/operation_validation.ex
spiderjako: I've changed that to be more meaningful, it should be more convenient now.

test/sanbase/signals/price/trigger_price_percent_change_test.exs
spiderjako: You're right, I've made a few now

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,30 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+  defp all_operations_valid?(list) do
+    with :ok <- all_operations_have_same_type?(list) do
+      Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+      |> case do
+        true ->
+          :ok
+
+        false ->
+          Enum.reject(list, fn op -> valid_operation?(op) == :ok end)
+          |> Enum.map(fn x -> {:error, "#{inspect(x)} is not a valid operation"} end)
```
tspenov: You are returning a list of :error tuples. I think you can just say that it contains a not valid operation :? (Other option is to return {:error, string_that_concats_all_not_valid_operations}).

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,30 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    |> Enum.uniq()
+    |> case do
+      [_] -> :ok
+      _ -> :error
+    end
+  end
+
+  defp all_operations_valid?(list) do
+    with :ok <- all_operations_have_same_type?(list) do
+      Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
```
tspenov: this also can be handled with `with` - I propose to extract it in a helper function that returns :ok or :error tuples and handle like the first one.

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,30 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    Enum.map(list, &Operation.type/1)
+    |> Enum.uniq()
+    |> case do
+      [_] -> :ok
+      _ -> :error
+    end
+  end
+
+  defp all_operations_valid?(list) do
+    with :ok <- all_operations_have_same_type?(list) do
```
IvanIvanoff: let `all_operations_have_same_type?` return true/false instaed of :ok/:error

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,22 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    |> case do
+      [_] -> true
+      _ -> false
+    end
+  end
+
+  defp all_operations_valid?(list) do
+    with true <- all_operations_have_same_type?(list),
+         true <- Enum.all?(list, fn op -> valid_operation?(op) == :ok end) do
+      true
```
tspenov: shouldn't that return `:ok`

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,22 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+    else
+      false -> {:error, "One of the operations or more are not valid"}
```
tspenov: `Some of the oprations are note valid` ?

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,154 @@
...
+        valid_operation?(operation)
+
+      [op] when op in @absolute_operations ->
+        {:error, "#{inspect(operation)} is an absolute operation, not a percent change one."}
+
+      [op] when op in @combinator_operations ->
+        list = Map.values(operation) |> List.first()
 
-      false ->
+        percent_operations? =
```
tspenov: We can extract this in to a helper that takes the error message and desired type as args

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,140 @@
...
+
+      [op] when op in @absolute_change_operations ->
+        {:error,
+         "#{inspect(operation)} is an absolute change operation, not an absolute value change one."}
+
+      [op] when op in @percent_operations ->
+        {:error, "#{inspect(operation)} is a percent, not an absolute value change one."}
+
+      [op] when op in @combinator_operations ->
+        list = Map.values(operation) |> List.first()
```
tspenov: We have this code 3 times and the only difference is the error message. We can further extract everything after this case: `[op] when op in @combinator_operations`

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,140 @@
...
+
+        if check_operations?(list, @absolute_value_operations) and all_operations_valid?(list) do
```
tspenov: `check_operations?` is not a proper name - it should be more meaningfull. (it checks if it contains only operations of certain type)

tspenov: `all_operations_valid?` returns :ok/:error tuple. If you don't need the message - rewrite to return true/false, otherwise you should use `with`

tspenov: Also `check_operations?` is doing more than what `all_operations_have_same_type?` is doing. So  `all_operations_have_same_type?` becomes obsolete

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,140 @@
...
 
-  def valid_percent_change_operation?(%{some_of: list} = operation) when is_list(list) do
-    Enum.all?(list, fn op -> valid_percent_change_operation?(op) == :ok end)
-    |> case do
-      true ->
-        :ok
+  @channel_operations [:inside_channel, :outside_channel]
+  @combinator_operations [:some_of, :all_of]
+
+  def valid_percent_change_operation?(operation) when is_map(operation) do
```
tspenov: Do we have tests for these new functions ?

lib/sanbase/signals/trigger/validation/operation_validation.ex
spiderjako: It doesn't become all that obsolete, because check_operations (in the near future - list_of_operations_are_from_given_type(_ , _), checks if a list of operations are from a given type, the other just checks if they are from 1 type, there is a slight difference, but it exists.

tspenov: in this case it is not needed, in other cases yes - it is

lib/sanbase/signals/trigger/validation/operation_validation.ex
spiderjako: Added lots of new tests, going to add more if you wish.

lib/sanbase/signals/trigger/validation/operation_validation.ex
spiderjako: Alright, thanks

spiderjako: You're right, it should be better now?

lib/sanbase/signals/trigger/validation/operation_validation.ex
spiderjako: Moved it into another function so it's not so spammed

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,136 @@
...
+  def valid_operation?(op), do: {:error, "#{inspect(op)} is not a valid operation"}
 
-  def valid_absolute_value_operation?(operation),
-    do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+  # Private functions
+  defp list_of_operations_are_from_given_type?(list, operation_type) do
+    if Enum.all?(list, fn elem ->
+         type = Map.keys(elem) |> List.first()
+         type in operation_type
+       end) and Enum.all?(list, fn op -> valid_operation?(op) == :ok end) do
```
tspenov: `Enum.all?(list, fn op -> valid_operation?(op) == :ok end)` that should be removed from here and moved here: https://github.com/santiment/sanbase2/pull/1418/files#diff-71cecc123f40b62575779894dff3bef2R130

spiderjako: You're right, it doesn't make much sense, when it is there, thanks

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,134 @@
...
+  defp all_operations_valid?(list) do
+    list = List.wrap(list)
 
-  defp do_valid_absolute_value_operation?(operation, _),
-    do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+    with true <- all_operations_have_same_type?(list),
+         true <- Enum.all?(list, fn op -> valid_operation?(op) == :ok end) do
+      :ok
+    else
+      false -> {:error, "Some of the operations are not valid"}
```
IvanIvanoff: not true.
All of the operations can be valid, but with different types, so this will return a misleading error

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,134 @@
...
+
+  defp operation_in_combinator_operations_valid?(operation, type) do
+    list = Map.values(operation) |> List.first()
+
+    with true <- list_of_operations_are_from_given_type?(list, type),
+         true <- Enum.all?(list, fn op -> valid_operation?(op) == :ok end) 
+    do
+      :ok
+    else
+      _ -> {:error, "Some of the operations are not valid"}
```
IvanIvanoff: same as above

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,153 @@
...
+    |> Enum.uniq()
+    |> case do
+      [_] -> true
+      _ -> false
+    end
+  end
 
-  defp do_valid_absolute_value_operation?(operation, _),
-    do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+  defp operation_in_combinator_operations_valid?(operation, type) do
```
IvanIvanoff: just `combinator_operation_valid?` will do it?

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -1,66 +1,153 @@
...
-    do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+  def valid_operation?(%{amount_up: value}) when is_number(value), do: :ok
+  def valid_operation?(%{amount_down: value}) when is_number(value), do: :ok
+  def valid_operation?(op), do: {:error, "#{inspect(op)} is not a valid operation"}
 
-  def valid_absolute_change_operation?(%{amount_up: value}) when is_number(value), do: :ok
-  def valid_absolute_change_operation?(%{amount_down: value}) when is_number(value), do: :ok
-  def valid_absolute_change_operation?(%{amount: value}) when is_number(value), do: :ok
+  # Private functions
+  defp list_of_operations_are_from_given_type?(list, operation_type) do
```
IvanIvanoff: just extend `all_operations_have_same_type?` to have a 2-arity version which does this. This function is doing almost the same thing with unnecessarily  strange name

spiderjako: Yeah, you're right

PR #1424, opened 2019-08-26
Add none_of operations signal

spiderjako:


lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,23 @@ defmodule Sanbase.Signal.Validation.Operation do
...
     do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+
+  defp operator_checker(list, operator) do
+    Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+    |> case do
+      true ->
+        :ok
+
+      false ->
+        {:error, "#{inspect(operator)} is not a valid percent change operation"}
```
tspenov: This ties `operator_checker` to percent change signal. If you want it to be generic - maybe pass the error message as param ?

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -15,14 +17,27 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+
+  def valid_absolute_value_operation?(%{some_of: list} = operation) when is_list(list) do
+    operator_checker(list, operation)
+  end
+
+  def valid_absolute_value_operation?(%{none_of: list} = operation) when is_list(list) do
+    operator_checker(list, operation)
+  end
+
+  def valid_absolute_change_operation?(%{some_of: list} = operation) when is_list(list) do
```
tspenov: Maybe that is a question to @IvanIvanoff but what is the difference between `valid_absolute_change_operation?` and `valid_absolute_value_operation?` ?

tspenov: we have also `valid_percent_change_operation?`

lib/sanbase/signals/operation/operation_evaluation.ex
```diff
@@ -14,6 +14,11 @@ defmodule Sanbase.Signal.OperationEvaluation do
     |> Enum.member?(true)
   end
 
+  def operation_triggered?(value, %{none_of: operations}) when is_list(operations) do
+    Enum.map(operations, fn op -> operation_triggered?(value, op) end)
+    |> Enum.all?(fn value -> value == false end)
```
IvanIvanoff: you can also do
```suggestion
    |> Enum.all?(& &1 == false)
```

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,23 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+
+      false ->
+        {:error, "#{inspect(operator)} is not a valid percent change operation"}
+    end
+
+    Enum.map(list, &Operation.type/1)
+    |> Enum.uniq()
+    |> length
+    |> case do
+      1 -> :ok
```
IvanIvanoff:
```elixir
Enum.map(list, &Operation.type/1)
|> Enum.uniq()
|> case do
   [_] -> :ok
   _ -> :error
end
```

If you want to assert that a list has 1 element you can avoid calling `length` which is O(n) and doing unnecessary work to achieve this goal

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -15,14 +17,33 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+  end
+
+  def valid_absolute_change_operation?(%{some_of: list} = operation) when is_list(list) do
+    operator_checker(list, operation)
+    all_operations_have_same_type?(list)
+  end
+
+  def valid_absolute_change_operation?(%{none_of: list} = operation) when is_list(list) do
+    operator_checker(list, operation)
+    all_operations_have_same_type?(list)
```
tspenov: that will skip `operator_checker(list, operation)` and return what `all_operations_have_same_type?(list)` returns.

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +84,25 @@ defmodule Sanbase.Signal.Validation.Operation do
 
   defp do_valid_absolute_value_operation?(operation, _),
     do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+
+  defp operator_checker(list, operator) do
+    Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+    |> case do
+      true ->
+        :ok
```
tspenov: You can put `all_operations_have_same_type?(list)` here and remove from upper calls.

spiderjako: Yeah, that's a bit better I think, thanks :D

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,26 @@ defmodule Sanbase.Signal.Validation.Operation do
 
   defp do_valid_absolute_value_operation?(operation, _),
     do: {:error, "#{inspect(operation)} is not a valid absolute value operation"}
+
+  defp operator_checker(list, operator) do
+    Enum.all?(list, fn op -> valid_operation?(op) == :ok end)
+    |> case do
```
IvanIvanoff: this case has no effect on the outcome of the function

spiderjako: Yes, I totally missed that, thanks

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -63,4 +78,24 @@ defmodule Sanbase.Signal.Validation.Operation do
...
+      true ->
+        all_operations_have_same_type?(list)
```
IvanIvanoff: I think a more readable approach would be to have `all_operations_valid?` and `all_operations_have_some_type?` and just combine them with an `and` or `with`

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -87,6 +87,16 @@ defmodule Sanbase.Signal.Validation.Operation do
...
 
+  def valid_operation?(%{none_of: list}) when is_list(list) do
+    with true <- all_operations_valid?(list),
+         true <- all_operations_have_same_type?(list) do
+      :ok
+    else
+      {:error, message} -> {:error, message}
+      false -> {:error, "Not all operations are from the same type"}
+    end
+  end
```
IvanIvanoff: you can make a function `valid_combinator_operation?(operation, reason)` with the body
```elixir
   with true <- all_operations_valid?(list),
	         true <- all_operations_have_same_type?(list) do
	      :ok
	    else
	      {:error, message} -> {:error, message}
	      false -> {:error,  reason}
	    end
```
and reuse it in the 3 cases

PR #1425, opened 2019-08-28
Renamed logo fetcher urls

spiderjako:
#### Summary
 * Added a migration to rename the logo64_url to just logo_url, thus removing logo_url beforehand, combined with logo32_url's removal

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
+++ b/lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
@@ -20,8 +20,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     Logger.info("#{@log_tag} Started fetching logos from coinmarketcap.")
 
     Temp.track!()
-    dir_path_64 = Temp.mkdir!("logos_64")
-    dir_path_32 = Temp.mkdir!("logos_32")
+    dir_path = Temp.mkdir!("logos")
 
     Map.keys(local_projects_map)
     |> Enum.chunk_every(100)
@@ -29,7 +28,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
       {:ok, remote_projects} = CryptocurrencyInfo.fetch_data(slugs)
 
       Enum.each(remote_projects, fn remote_project ->
-        update_project_logos(remote_project, local_projects_map, dir_path_64, dir_path_32)
+        update_project_logos(remote_project, local_projects_map, dir_path)
       end)
 
       remote_projects
@@ -40,7 +39,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     Logger.info("#{@log_tag} Finished fetching logos from coinmarketcap.")
   end
 
-  defp update_project_logos(remote_project, local_projects_map, dir_path_64, dir_path_32) do
+  defp update_project_logos(remote_project, local_projects_map, dir_path) do
     url = remote_project.logo
     slug = remote_project.slug
     file_extension = Path.extname(url |> String.downcase())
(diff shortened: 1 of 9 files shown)


lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -21,15 +21,14 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
 
     Temp.track!()
     dir_path_64 = Temp.mkdir!("logos_64")
```
kalkov: As there is only one size now, you can remove the `_64` postfix from the names

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -116,16 +112,10 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
     end
   end
 
-  defp resize_image(source_filepath, dest_dir_path, filename) do
-    dest_filepath = dest_dir_path <> "/" <> filename
-    Mogrify.open(source_filepath) |> Mogrify.resize("32x32") |> Mogrify.save(path: dest_filepath)
```
kalkov: Remove the `mogrify` dependency from the mix.exs

kalkov: Also `imagemagick` from Dockerfiles

priv/repo/migrations/20190828092907_remove_logo_url.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Sanbase.Repo.Migrations.RemoveLogoUrl do
```
kalkov: Create 3 migrations in that order:
1. Remove `logo_url`
2. Rename `logo64_url` to `logo_url`
3. Remove `logo32_url`

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
spiderjako: Yes, I was going to ask if I should remove the resizing, when there is no 32x32 anymore

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
spiderjako: You're right

priv/repo/migrations/20190828092907_remove_logo_url.exs
spiderjako: It shall be done.

lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
kalkov: Actually I changed my mind, let's keep the resize but change it to 64x64. If CMC changes the size of the logos or a single logo is with different size that can break things on our side and we will need to change the script. Having a resize step will guarantee us that the logos will always have the same size.

priv/repo/migrations/20190828104909_remove_logo_32_url.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.RemoveLogo32Url do
+  use Ecto.Migration
+
+  def change do
```
kalkov: I don't know about the others but I like to have reversible migrations. Try `mix ecto.rollback --step NUMBER` and you will see it fails. 
Ecto 3 has [`remove/3`](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#remove/3) which is reversible, but we are still not there.
The way to do it in Ecto 2 is with `up/down`:
```elixir
  def up do
    alter table("project") do
      remove(:logo32_url)
    end
  end

  def down do
    alter table("project") do
      add(:logo32_url, :string)
    end
  end
```

priv/repo/migrations/20190828104846_rename_logo_64_url_to_logo_url.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Sanbase.Repo.Migrations.RenameLogo64UrlToLogoUrl do
+  use Ecto.Migration
+
+  def change do
+    rename(table("project"), :logo64_url, to: :logo_url)
```
kalkov: @valo Is this migration reversible?

valo: Should be. Have you tested it?

kalkov: Yes it is, tested it.

PR #1429, opened 2019-09-02
Adding once again the size when uploading a logo

spiderjako:


lib/sanbase/external_services/coinmarketcap2/logo_fetcher.ex
```diff
@@ -51,7 +51,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
              {:logo_has_changed?, true} <-
                {:logo_has_changed?, logo_changed?(project, local_filepath)},
              {:ok, local_filepath} <- resize_image(local_filepath, dir_path, filename),
-             {:ok, uploaded_filepath} <- upload(local_filepath),
+             {:ok, uploaded_filepath} <- upload(local_filepath, 64),
```
kalkov: No need to pass it every time, extract to module attribute and use it also in resize.

PR #1430, opened 2019-09-02
Fix spreadsheet permissions

tspenov:


lib/sanbase/auth/user.ex
```diff
@@ -128,36 +128,32 @@ defmodule Sanbase.Auth.User do
   end
 
   def permissions(%__MODULE__{} = user) do
-    with {:ok, san_balance} <- san_balance(user) do
-      san_balance = san_balance |> Decimal.to_float()
+    active_subscriptions = Subscription.user_subscriptions(user) |> Enum.map(& &1.plan.product.id)
```
IvanIvanoff: It makes sense to optimize this with a special query that returns a list of ids directly as this will be called on every API call

lib/sanbase/billing/subscription.ex
```diff
@@ -235,8 +235,19 @@ defmodule Sanbase.Billing.Subscription do
...
     |> user_subscriptions_query()
     |> active_subscriptions_query()
+    |> join_plan_and_product_query()
+    |> Repo.all()
+  end
+
+  @doc """
+  List active subcriptions' product ids
+  """
+  def list_subscrption_products(user) do
```
kalkov: typo: subscrption

lib/sanbase/billing/subscription.ex
```diff
@@ -235,8 +235,19 @@ defmodule Sanbase.Billing.Subscription do
...
     |> user_subscriptions_query()
     |> active_subscriptions_query()
+    |> join_plan_and_product_query()
+    |> Repo.all()
+  end
+
+  @doc """
+  List active subcriptions' product ids
+  """
+  def list_subscription_products(user) do
```
IvanIvanoff: can we have this in line with the previous function and just call it `user_subscriptions_ids` :?

PR #1437, opened 2019-09-03
Introduce source_slug_mappings table

IvanIvanoff:
#### Summary
In a previous PR (#1435) we changed `coinmarketcap_id` column to `slug`, which should be read as "santiment_id". So, the slug is no longer equal to the coinmarketcap_id. This will allow us to add projects that are not on CMC and to easily handle when a project on CMC gets changed.

So, we're introducing a mapping between a santiment project's slug and the slug of the same project on other sources. 

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/model/project/slug_source_mapping.ex
```diff
@@ -0,0 +1,34 @@
...
+  import Ecto.Changeset
+
+  alias Sanbase.Repo
+  alias Sanbase.Model.Project
+
+  @table "slug_source_mappings"
+
+  schema @table do
+    field(:source, :string)
+    field(:source_slug, :string)
```
tspenov: only `slug` :?

lib/sanbase/model/project/slug_source_mapping.ex
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def changeset(%__MODULE__{} = ssm, attrs \\ %{}) do
+    ssm
+    |> cast(attrs, [:source, :source_slug, :project_id])
+    |> validate_required([:source, :source_slug, :project_id])
+    |> unique_constraint(:source_slug, name: :source_slug_unique_combination)
+  end
+
+  def get_source_slug(%Project{id: project_id}, source) do
```
tspenov: not used

lib/sanbase/web.ex
```diff
@@ -19,7 +19,7 @@ defmodule Sanbase.Application.Web do
       Sanbase.TimescaleRepo,
 
       # Start the Clickhouse Repo
-      start_in({Sanbase.ClickhouseRepo, []}, [:dev, :prod]),
+      start_in({Sanbase.ClickhouseRepo, []}, [:prod]),
```
tspenov: why removed ?

mix.exs
```diff
@@ -84,7 +84,7 @@ defmodule Sanbase.Mixfile do
       {:observer_cli, "~> 1.3"},
       {:earmark, "~> 1.2"},
       {:ecto_enum, "~> 1.3"},
-      {:ex_machina, "~> 2.2", only: :test},
+      {:ex_machina, "~> 2.2", only: [:dev, :test]},
```
tspenov: do we need in :dev ?

priv/repo/migrations/20190903145409_add_slug_source_map_table.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Sanbase.Repo.Migrations.AddSlugSourceMapTable do
+  use Ecto.Migration
+
+  @table "slug_source_mappings"
+  def change do
+    create table(@table) do
+      add(:source, :string)
+      add(:source_slug, :string)
```
tspenov: shouldn't we have non_null for both ?

lib/sanbase/web.ex
IvanIvanoff: I'll add it back, it's annoying while working locally without VPN on

lib/sanbase/model/project/slug_source_mapping.ex
IvanIvanoff: I'll point the `coinmarketcap_id/1` from the project module to use it

mix.exs
IvanIvanoff: It's much easier to test things locally with it rather than writing the changeset + repo inserts by hand

tspenov: then lets allow in dev 👍

lib/sanbase/model/project/slug_source_mapping.ex
IvanIvanoff: I don't like `source_slug` but also just `slug` could introduce some confusion :?

IvanIvanoff: After some thought, I renamed it to `slug` and changed the table name from `slug_source..` to `source_slug...`

PR #1439, opened 2019-09-04
Add coupon support subscriptions

tspenov:
Added optional arg `coupon` to the subscription mutation.

example:
```graphql
    mutation {
      subscribe(card_token: "card_token", plan_id: 5, coupon: "COUPON_CODE") {
        id,
        status
        plan {
          id
          name
          interval
          amount
          product {
            name
          }
        }
      }
    }
```

### DIFF
--- a/lib/sanbase/billing/subscription.ex
+++ b/lib/sanbase/billing/subscription.ex
@@ -23,8 +23,6 @@ defmodule Sanbase.Billing.Subscription do
   Current subscription attempt failed.
   Please, contact administrator of the site for more information.
   """
-  @free_trial_days_for_coupons 14
-  @percent_off_for_free_trial 100
 
   schema "subscriptions" do
     field(:stripe_id, :string)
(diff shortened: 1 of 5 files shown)


IvanIvanoff:
Should we expose the internal plan IDs? Shouldn't we better have `product<String>` and `plan<String>`? So you can use it like this: `subscribe(product: "sanbase", plan: "pro", card_token: "token")

lib/sanbase/billing/subscription.ex
```diff
@@ -333,19 +331,34 @@ defmodule Sanbase.Billing.Subscription do
     end
   end
 
-  defp update_subscription_with_coupon(nil, subscription), do: {:ok, subscription}
+  defp create_stripe_subscription(user, plan, coupon) when not is_nil(coupon) do
```
IvanIvanoff: Should we apply SAN balance percent discounts in such cases (as far as I can tell we do not add this discount)? In most places discounts do not add up, so this should be the way to go

tspenov: Stripe supports only one coupon per subscription. Currently it is implemented this way: if user has supplied coupon we apply it, otherwise we apply SAN stake percent_off if user hass sufficient.

tspenov:
@IvanIvanoff I guess we can ask frontend cause they are the only consumers of this API.

PR #1442, opened 2019-09-09
Documentation for access-plans article

tzanko-matev:
#### Summary
Add reference to documentation for access plans
Related: https://github.com/santiment/articles/pull/3


lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -1,3 +1,4 @@
+# DOCS access-plans/index.md
```
IvanIvanoff: @tzanko-matev should also refer to the repository?

tzanko-matev: Don't think so. If we change the name of the repository we'll have to update everything

PR #1445, opened 2019-09-09
Add migration for inserting market segment from infrastructure

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/migrations/20190909113137_add_more_market_segments.exs
```diff
@@ -0,0 +1,90 @@
...
+      |> Sanbase.Repo.all()
+
+    # Map an infrastructure to the corresponding ID. The name is always equal to
+    # the value, i.e. the name. The key is either the name or the ticker
+    infrastrucutre_to_segment
+    |> Enum.map(fn {k, v} ->
+      segment = Enum.find(market_segments, fn %{name: name} -> name == v || name == k end)
+      {k, segment.id}
+    end)
+    |> Map.new()
```
valo: Shouldn't the mapping between the project and the segments be updated too?

IvanIvanoff: If I understand correctly: The mapping is done through the "project_market_segments" table which holds project_id and market_segment_id, so it is updated.
After the migration a project now has a new market segment: 
image
and the market segment itself has projects:
image

IvanIvanoff: The confusion is probably based on the fact that this PR is not with base master, but with base the PR that introduces this join through table

valo: I was not looking at the right function. The mapping is getting updated indeed

PR #1459, opened 2019-09-13
Fix String.downcase being called on the atom nil

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/project/project.ex
+++ b/lib/sanbase/model/project/project.ex
@@ -329,6 +329,17 @@ defmodule Sanbase.Model.Project do
     not is_erc20?(project)
   end
 
+  def is_trending?(%Project{} = project, trending_words_mapset) do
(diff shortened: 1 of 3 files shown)


lib/sanbase/model/project/project.ex
```diff
@@ -329,6 +329,22 @@ defmodule Sanbase.Model.Project do
     not is_erc20?(project)
   end
 
+  def is_trending?(%Project{} = project, trending_words_mapset) do
```
tspenov: we can simplify this:

```elixir
      [project.ticker, project.name, project.slug]
      |> Enum.reject(&is_nil/1)
      |> Enum.map(&String.downcase/1)
      |> MapSet.new()
      |> MapSet.intersection(trending_words_mapset)
      |> Enum.any?()
```

PR #1460, opened 2019-09-13
Rework access logic

IvanIvanoff:
#### Summary
For a customer, we want to add exchange wallets to their plan. The solution is a generic one: In order to allow access to some queries only when a user has a certain plan (extension plan) you need to do the following:
- Add a new Product and define a Plan with it.
- Add this plan to the customer stripe account.
- The GraphQL query definition must be enriched with the following meta: 
```elixir
meta(access: :extension, product: Product.product_exchange_wallets())
```
where the product is the needed product. In order to pass through the checks for this query, the `AccessControl` middleware must be added, too.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -64,18 +72,38 @@ defmodule SanbaseWeb.Graphql.Middlewares.TimeframeRestriction do
     resolution
   end
 
+  @exchange_wallets_product Product.product_exchange_wallets()
+  @plan_extension_needed [:exchange_wallets]
```
tspenov: Shouldn't we fill this list from query meta tags: `meta(access: :extension)` :?

IvanIvanoff: It's a POC - currently I'm doing a major rework on this part

IvanIvanoff: The end result will be something like this: 
image

IvanIvanoff: @tspenov it's reworked - you can check it again.

lib/sanbase/web.ex
```diff
@@ -19,7 +19,7 @@ defmodule Sanbase.Application.Web do
       Sanbase.TimescaleRepo,
 
       # Start the Clickhouse Repo
-      start_in({Sanbase.ClickhouseRepo, []}, [:dev, :prod]),
```
tspenov: leftover

lib/sanbase/billing/graphql_schema.ex
```diff
@@ -0,0 +1,76 @@
...
+      {{_, _, [module, func]}, _, _} = product_ast
+      product_id = apply(module, func, [])
+      {f, product_id}
+    end)
+    |> Map.new()
+  end
+
+  @doc ~s"""
+  Return all query names that have all `fields` with the values specified in
+  the corresponding position of the `values` lis
```
tspenov:
```suggestion
  the corresponding position of the `values` list
```

PR #1497, opened 2019-09-25
Monitor watchlist

tspenov:
* Allow to toggle watchlist as monitored
* Fetch all insights for user that needs to be send in an email digest:
-- from the last week
-- when insight author is followed or is a SanClan member
-- if the insight has tag that is in User's monitored watchlists assets
* Next step will be to schedule the digest once a week and send to all users that have email and have monitored watchlists
 
Toggle isMonitored flag:

```graphql
    mutation {
      updateWatchlist(
        id: #{watchlist.id},
        isMonitored: true
      ) {
        isMonitored
    }
```


lib/sanbase/auth/user/role.ex
```diff
@@ -0,0 +1,44 @@
...
+    |> cast(attrs, [:name, :code])
+  end
+
+  def san_clan_role_id() do
+    Repo.get_by(__MODULE__, code: "san_clan")
+    |> Map.get(:id)
+  end
+end
+
+defmodule Sanbase.Auth.User.UserRole do
```
IvanIvanoff: I've been thinking how to write such module names where you have duplication:
`User.Role` and `User.UserRole`, same thing with `Settings` and `UserSettings` and I think we can do this by having the modules:
```elixir
Sanbase.Auth.User,
Sanbase.Auth.Role,
Sanbase.Auth.UserRole # and not User.UserRole
```

priv/repo/migrations/20190924060356_add_is_monitored_to_watchlists.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.AddIsMonitoredToWatchlists do
+  use Ecto.Migration
+
+  def change do
+    alter table(:user_lists) do
+      add(:is_monitored, :boolean, default: false)
```
IvanIvanoff: This is one option per user list and it's controlled by its owner, right? If we want to monitor watchlists like stablecoins, top erc20, DEX, etc. this won't be possible because they have different owners?

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
+defmodule Sanbase.UserList.Monitor do
+  @moduledoc """
+    Watchlist can be monitored - which means the creator will receive an email if any
```
IvanIvanoff: nitpick: identation

lib/sanbase/auth/user/role.ex
tspenov: Yeah, that was also bothering me. I'll change the module names but leave them under `user` - what do you think ?

priv/repo/migrations/20190924060356_add_is_monitored_to_watchlists.exs
tspenov: But monitoring here means - send watchlist creator email for the connected insights. So there should be only one user to receive the email.

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
...
+    |> insights_with_asset_in_monitored_watchlist(watchlists)
+  end
+
+  def watchlists_tags(watchlists) do
+    watchlists
+    |> Enum.flat_map(fn watchlist ->
+      watchlist.list_items
+      |> Enum.flat_map(&[&1.project.slug, &1.project.ticker, &1.project.name])
+    end)
+    |> Enum.dedup()
```
IvanIvanoff: You can plug reject is_nil here, too. 
Also, add some docs here as here we're defining what a watchlist tag is.

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
...
+  alias Sanbase.Auth.User
+  alias Sanbase.Auth.User.{UserRole, Role}
+  alias Sanbase.Insight.Post
+  alias Sanbase.Repo
+
+  def insights_to_send(user) do
+    watchlists = monitored_watchlists_for(user)
+
+    week_ago()
+    |> Post.public_insights_after()
```
IvanIvanoff: This is loading too much data (all text for all public insights) - can we simplify it?

Also, I think we can additionally filter the watchlists in the SQL query itself. If we introduce a new query that returns the insight ids of the published insights`where: user.id in ^followed_or_sanclan_user_ids`

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  defp monitored_watchlists_for(%User{id: user_id}) do
+    from(ul in UserList,
+      where: ul.user_id == ^user_id and ul.is_monitored == true,
+      preload: [list_items: [:project]]
+    )
+    |> Repo.all()
+  end
+
+  defp insights_by_followed_users_or_sanclan(insights, user_id) do
```
IvanIvanoff: btw, yesterday we ditched the `sanclan` in favor of `SANfam`

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  defp insights_with_asset_in_monitored_watchlist(insights, watchlists) do
+    watchlists_tags = watchlists_tags(watchlists)
+
+    insights
+    |> Enum.filter(fn %Post{tags: tags} ->
+      tags
+      |> Enum.any?(fn tag ->
+        tag.name in watchlists_tags
```
IvanIvanoff: `watchlist_tags` is potentially a long list for big watchlists. You can make the `watchlist_tags` a MapSet instead of list - the `in` operator works for mapsets, too.

lib/sanbase/user_lists/monitor.ex
```diff
@@ -0,0 +1,79 @@
...
+      end)
+    end)
+  end
+
+  defp users_with_email() do
+    from(u in User, where: not is_nil(u.email))
+    |> Repo.all()
```
IvanIvanoff: Let's move this function to the `User` module

IvanIvanoff: Also, I am not sure we'll ever need the whole list of users with emails - and I do not see this function being used anywhere.

priv/repo/migrations/20190925064207_create_user_roles.exs
```diff
@@ -0,0 +1,32 @@
...
+      add(:role_id, references(:roles), null: false, on_delete: :delete_all, primary_key: true)
+
+      timestamps()
+    end
+
+    create(unique_index(:user_roles, [:user_id, :role_id]))
+
+    execute("""
+    INSERT INTO roles (id, name, code) VALUES
+      (1, 'Santiment Team member', 'san_team'),
```
IvanIvanoff: `Santiment Team Member` :?

lib/sanbase/auth/user/role.ex
```diff
@@ -0,0 +1,44 @@
...
+  alias Sanbase.Repo
+
+  schema "roles" do
+    field(:name, :string)
+    field(:code, :string)
+  end
+
+  def changeset(%__MODULE__{} = role, attrs \\ %{}) do
+    role
+    |> cast(attrs, [:name, :code])
```
IvanIvanoff: |> unique_constraint(:code)

lib/sanbase/user_lists/monitor.ex
tspenov: yeah, we don't need the text - only some data to build the url and maybe title.

lib/sanbase/user_lists/monitor.ex
tspenov: yep, did that

lib/sanbase/user_lists/monitor.ex
tspenov: Removed for now. Will need it probably when I start sending emails.

priv/repo/migrations/20190924060356_add_is_monitored_to_watchlists.exs
IvanIvanoff: But if you want to monitor a widely used public watchlist =? All but 2 of the asset groups here are watchlists: 
image

If you want to monitor that, you need to copy the watchlist and monitor your copy. But this won't work very well because the Top 50 ERC20 projects watchlist is generated by a function + manually curating some ignore list, which can change after you copy it.

So, in order to handle monitoring such watchlists, I think there should be a table "user_watchlist_monitorings" with user_id and watchlist_id pairs

lib/sanbase/auth/user/role.ex
```diff
@@ -0,0 +1,24 @@
...
+    |> unique_constraint(:code)
+  end
+
+  def san_family_role_id() do
+    Repo.get_by(__MODULE__, code: "san_family")
```
IvanIvanoff: the code can be module attribute

IvanIvanoff: btw, another option is to handle the roles like the products - with fixed ids, so we won't even need to call the DB in those cases :?

lib/sanbase/auth/user/user_role.ex
```diff
@@ -0,0 +1,20 @@
...
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  alias Sanbase.Auth.{User, Role}
+
+  @primary_key false
+  schema "user_roles" do
+    belongs_to(:user, User, primary_key: true)
+    belongs_to(:role, Role, primary_key: true)
```
IvanIvanoff: it'll be good to have the corresponding has_one and has_many

priv/repo/migrations/20190924060356_add_is_monitored_to_watchlists.exs
tspenov: @Partysun confirmed that we won't be doing this for now. If we decide later we'll create the new table and migrate existing monitored watchlist and extend to public ones.

PR #1498, opened 2019-09-25
Trending word signal sends all words in one signal

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
+++ b/lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
@@ -112,12 +112,13 @@ defmodule Sanbase.Signal.Trigger.TrendingWordsTriggerSettings do
           %TrendingWordsTriggerSettings{settings | triggered?: false}
 
         [_ | _] = words ->
-          payload =
-            Enum.reduce(words, %{}, fn word, acc ->
-              Map.put(acc, word, payload(settings, word))
-            end)
+          payload = %{words => payload(settings, words)}
 
-          %TrendingWordsTriggerSettings{settings | triggered?: true, payload: payload}
+          %TrendingWordsTriggerSettings{
+            settings
+            | triggered?: true,
+              payload: payload
+          }
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/endpoint.ex
```diff
@@ -87,6 +87,12 @@ defmodule SanbaseWeb.Endpoint do
     |> Path.join(word)
   end
 
+  def trending_word_url(words) when is_list(words) do
+    website_url()
+    |> Path.join("labs/trends/explore")
+    |> Path.join(words |> Enum.join(" OR "))
```
tspenov: should we `URI.encode` the last part or that will work ?

tspenov:
Can you add a test for the new payload ?

lib/sanbase/signals/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -183,14 +184,25 @@ defmodule Sanbase.Signal.Trigger.TrendingWordsTriggerSettings do
...
       """
       The word **#{word}** is in the trending words.
 
       More info here: #{SanbaseWeb.Endpoint.trending_word_url(word)}
       """
     end
 
+    defp payload(%{operation: %{trending_word: true}}, [_, _ | _] = words) do
+      {last, previous} = List.pop_at(words, -1)
+      words_str = (Enum.map(previous, &"**#{&1}**") |> Enum.join(",")) <> " and #{last}"
```
tspenov: `last` is not bold text

lib/sanbase_web/endpoint.ex
IvanIvanoff: true, `Path.join` won't alone handle that properly

lib/sanbase_web/endpoint.ex
```diff
@@ -106,6 +114,7 @@ defmodule SanbaseWeb.Endpoint do
     Config.get(:backend_url)
   end
 
+  @spec api_url :: <<_::64, _::_*8>>
```
tspenov: lets remove this spec

PR #1504, opened 2019-09-26
Support the min suffix for intervals

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


IvanIvanoff:
Now that I think about it - the interval is directly passed to influxdb, supporting `5min` could introduce inconsistencies. 

The best approach would be to implement a custom gql type and use it for intervals instead of `string`

PR #1505, opened 2019-09-26
Simplify the average dev activity dataloader

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/github.ex
+++ b/lib/sanbase/clickhouse/github.ex
@@ -77,9 +77,9 @@ defmodule Sanbase.Clickhouse.Github do
           {:ok, list({String.t(), non_neg_integer()})} | {:error, String.t()}
   def total_dev_activity([], _from, _to), do: {:ok, []}
 
-  def total_dev_activity(organizations, from, to) when length(organizations) > 10 do
+  def total_dev_activity(organizations, from, to) when length(organizations) > 20 do
     total_dev_activity =
-      Enum.chunk_every(organizations, 10)
+      Enum.chunk_every(organizations, 20)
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/github.ex
```diff
@@ -77,9 +77,9 @@ defmodule Sanbase.Clickhouse.Github do
...
   def total_dev_activity([], _from, _to), do: {:ok, []}
 
-  def total_dev_activity(organizations, from, to) when length(organizations) > 10 do
+  def total_dev_activity(organizations, from, to) when length(organizations) > 20 do
     total_dev_activity =
-      Enum.chunk_every(organizations, 10)
+      Enum.chunk_every(organizations, 20)
       |> Sanbase.Parallel.map(&total_dev_activity(&1, from, to),
         timeout: 25_000,
         max_concurrency: 8,
```
tspenov: Isn't it more optimal if `max_concurrency` is divisor of `20` here ?  like `10` so it goes on 2 passes instead of 3

IvanIvanoff: Usually you have a number of CPUs that's a power of 2 (1, 2, 4) - it's 2 on prod I think. This way we can distribute the work more evenly. The default concurrency is `System.schedulers_online`, which by default is the number of cores.

IvanIvanoff: Also, the concurrency is for how many concurrent batches of 20 orgs are queried - it's not about running concurrenly a query for every organization in this lists of length 20

IvanIvanoff:
I had to fix some things in the dataloader and implemented these changes in #1506

PR #1513, opened 2019-09-30
Extend the getMetric API with other metrics

IvanIvanoff:
#### Summary
Extending the `getMetric` API with metrics that are outside of the CH v2 metrics.
This is done by having a dispatch module `Sanbase.Metric` which knows where to get the data from.
To extend the `getMetric` API the following things must be done:
- A module that implements the `Sanbase.Metric.Behaviour` behaviour
- Add this new module to the `@metric_modules` module attribute in the `Sanbase.Metric` module 

New metrics added:
- dev_activity
- github_activity
- telegram_social_volume
- twitter_social_volume
- discord_social_volume
- telegram_social_dominance
- twitter_social_dominance
- discord_social_dominance

[//]: # (#### Related PRs)

#### Screenshots
image

image


test/sanbase/clickhouse/metric/v2_clickhouse_metric_test.exs
```diff
@@ -3,162 +3,83 @@ defmodule Sanbase.Clickhouse.V2ClickhouseMetricTest do
```
tspenov: change also the file name

lib/sanbase/metric/metric.ex
```diff
@@ -0,0 +1,189 @@
...
+  @doc ~s"""
+  Get all available metrics
+  """
+  def available_metrics(), do: @metrics
+
+  @doc ~s"""
+  Get all slugs for which at least one of the metrics is available
+  """
+  def available_slugs_all_metrics() do
+    Sanbase.Cache.get_or_store({:metric_available_slugs_all_metrics, 1800}, fn ->
```
tspenov: Add a named module attribute for 1800

tspenov:
Very nice implementation with a combination of behaviour and dispatch module 👏

lib/sanbase/metric/metric.ex
IvanIvanoff: I've added comments explaining what's this number, I think it's better than just moving it to a module attribute

PR #1519, opened 2019-10-02
Migrate APIs to v2

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/metric/daily_active_addresses.ex
```diff
@@ -0,0 +1 @@
+
```
tspenov: empty

lib/sanbase/clickhouse/metric/description.ex
```diff
@@ -0,0 +1,29 @@
...
+  in circulation is 20 ETH.
+
+  This can be explained as having twenty $1 bills. Alice sends all of them to Bob,
+  Bob sends 10 of the received bills to Charlie and Charlie sends 5 of them to Dean.
+
+  One of the most useful properities of Token Circulation is that this metric is immune
+  to mixers and gives a much better view of the actual amount of tokens that are being
+  transacted
+  """
+end
```
tspenov: Maybe also add the velocity description here

lib/sanbase/utils/transform.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Sanbase.Utils.Transform do
```
tspenov: Add unit tests for these

lib/sanbase/clickhouse/metric/description.ex
IvanIvanoff: Yeah - in case we remove the modules we still need to keep the docs somewhere. I tried to add these to the json file, but json does not support multiline comments and it's ugly. I think I'll migrate the json file to a elixir map and keep the description there?

tspenov: is the json intended to be used from other place or non-elixir devs ? If not we can migrate to elixir map.

PR #1521, opened 2019-10-02
Add access grafana through sanbase

tspenov:

### DIFF
--- a/lib/sanbase/grafana/api.ex
+++ b/lib/sanbase/grafana/api.ex
@@ -1,0 +1,90 @@
+defmodule Sanbase.GrafanaApi do
+  require Sanbase.Utils.Config, as: Config
+  require Mockery.Macro
+  require Logger
+
+  alias Sanbase.Auth.User
+
+  @plan_team_map %{
+    41 => "Sangraphs-Basic",
+    42 => "Sangraphs-Pro"
+  }
+
+  def plan_team_map, do: @plan_team_map
+
+  def get_team_by_name(name) do
+    request_path = "api/teams/search?name=#{name}"
+
+    http_client().get(base_url() <> request_path, headers())
+    |> handle_response()
+    |> Map.get("teams")
+    |> hd()
+  end
+
+  def get_team_members(team_id) do
+    request_path = "api/teams/#{team_id}/members"
+
+    http_client().get(base_url() <> request_path, headers())
+    |> handle_response()
+  end
+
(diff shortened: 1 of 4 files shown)


lib/sanbase/grafana/api.ex
```diff
@@ -0,0 +1,90 @@
...
+      nil -> add_user_to_team(user_id, team_id)
+      _ -> {:ok, "User is already in this team"}
+    end
+  end
+
+  def add_user_to_team(user_id, team_id) do
+    request_path = "api/teams/#{team_id}/members"
+    data = %{"userId" => user_id} |> Jason.encode!()
+
+    http_client().post(base_url() <> request_path, data, headers())
```
IvanIvanoff: We're now using `Mock` more instead of `Mockery` - imo it is good to not pollute the source code with mock preparations

lib/sanbase/grafana/api.ex
```diff
@@ -0,0 +1,90 @@
...
+    request_path = "api/teams/#{team_id}/members"
+
+    http_client().get(base_url() <> request_path, headers())
+    |> handle_response()
+  end
+
+  def add_user_to_team(%User{} = user, plan_id) do
+    team_name = @plan_team_map[plan_id]
+    team_id = get_team_by_name(team_name)["id"]
+    user_id = get_user_by_email_or_metamask(user)["id"]
```
IvanIvanoff: there's no error handling in these 2 HTTP requests - `handle_response` can return an error tuple. You can chain them using `with` and make it return :ok tuple instead of the value itself or :error tuple

lib/sanbase/oauth2/hydra.ex
```diff
@@ -24,19 +25,15 @@ defmodule Sanbase.Oauth2.Hydra do
...
-  def manage_consent(consent, access_token, user, client_id) do
-    user_san_balance = san_balance!(user)
+  def manage_consent(consent, access_token, user) do
+    subscription = Subscription.current_subscription(user, Product.product_sangraphs())
 
-    if has_enough_san_tokens?(
-         user_san_balance,
-         required_san_tokens_by_client(:clients_that_require_san_tokens, client_id)
-       ) do
+    if subscription do
```
IvanIvanoff:
```elixir
case Subscription.current_subscription(user, Product.product_sangraphs()) do
  %Subscription{...} ->
  _ ->  ...
end
```

PR #1522, opened 2019-10-03
Oauth grafana additions

tspenov:

### DIFF
--- a/lib/sanbase/grafana/api.ex
+++ b/lib/sanbase/grafana/api.ex
@@ -16,6 +16,14 @@ defmodule Sanbase.GrafanaApi do
 
   def plan_team_map, do: @plan_team_map
 
+  def get_user_by_email_or_metamask(%User{username: username, email: email}) do
+    token = email || username
+    request_path = "api/users/lookup?loginOrEmail=#{token}"
+
+    http_client().get(base_url() <> request_path, headers())
+    |> handle_response()
+  end
+
   def add_subscribed_user_to_team(%User{} = user, plan_id) do
     team_name = @plan_team_map[plan_id]
 
@@ -98,14 +106,6 @@ defmodule Sanbase.GrafanaApi do
     data = %{"userId" => user_id} |> Jason.encode!()
 
     http_client().post(base_url() <> request_path, data, headers())
-    |> handle_response()
-  end
-
-  def get_user_by_email_or_metamask(%User{username: username, email: email}) do
-    token = email || username
-    request_path = "api/users/lookup?loginOrEmail=#{token}"
-
-    http_client().get(base_url() <> request_path, headers())
     |> handle_response()
   end
 
--- a/lib/sanbase/grafana/api.ex
+++ b/lib/sanbase/grafana/api.ex
@@ -78,12 +78,12 @@ defmodule Sanbase.GrafanaApi do
         _ -> do_remove_user_from_team(user_id, team_id)
       end
     else
-      error ->
+      {:error, error} ->
         Logger.error(
           "Error removing grafana user: #{user_id} from team: #{team_name}: #{inspect(error)}"
         )
 
-        error
+        {:error, error}
     end
   end
 
--- a/lib/sanbase/grafana/api.ex
+++ b/lib/sanbase/grafana/api.ex
@@ -16,11 +16,12 @@ defmodule Sanbase.GrafanaApi do
 
   def plan_team_map, do: @plan_team_map
 
-  def get_user_by_email_or_metamask(%User{username: username, email: email}) do
+  def get_user_by_email_or_username(%User{username: username, email: email}) do
     token = email || username
     request_path = "api/users/lookup?loginOrEmail=#{token}"
 
-    http_client().get(base_url() <> request_path, headers())
+    Path.join(base_url(), request_path)
+    |> http_client().get(headers())
     |> handle_response()
   end
 
@@ -36,7 +37,8 @@ defmodule Sanbase.GrafanaApi do
       }
       |> Jason.encode!()
 
-    http_client().post(base_url() <> request_path, data, headers())
+    Path.join(base_url(), request_path)
+    |> http_client().post(data, headers())
     |> handle_response()
   end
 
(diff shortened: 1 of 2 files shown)


lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
 
   @plan_team_map %{
     41 => "Sangraphs-Basic",
-    42 => "Sangraphs-Pro"
+    42 => "Sangraphs-Pro",
+    43 => "Sangraphs-Premium",
+    44 => "Sangraphs-Basic",
+    45 => "Sangraphs-Pro",
+    46 => "Sangraphs-Premium"
```
IvanIvanoff: Why doubled?

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
   }
 
   def plan_team_map, do: @plan_team_map
 
-  def add_subscribed_user_to_team(%User{} = user, plan_id) do
+  def get_user_by_email_or_metamask(%User{username: username, email: email}) do
+    token = email || username
+    request_path = "api/users/lookup?loginOrEmail=#{token}"
+
+    http_client().get(base_url() <> request_path, headers())
```
IvanIvanoff: You can join those with Path.join to avoid caring if you have a trailing/leading `/` in both parts

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+    data =
+      %{
+        "name" => username || email,
+        "email" => email || username,
+        "login" => username || email,
+        "password" => :crypto.strong_rand_bytes(16) |> Base.encode64() |> binary_part(0, 16)
+      }
+      |> Jason.encode!()
+
+    http_client().post(base_url() <> request_path, data, headers())
```
IvanIvanoff: ditto

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+    43 => "Sangraphs-Premium",
+    44 => "Sangraphs-Basic",
+    45 => "Sangraphs-Pro",
+    46 => "Sangraphs-Premium"
   }
 
   def plan_team_map, do: @plan_team_map
 
-  def add_subscribed_user_to_team(%User{} = user, plan_id) do
+  def get_user_by_email_or_metamask(%User{username: username, email: email}) do
```
IvanIvanoff: `or username`

lib/sanbase/grafana/api.ex
tspenov: these are yearly plans - they go to the same team

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+    |> handle_response()
+  end
+
+  def create_user(%User{username: username, email: email}) do
+    request_path = "api/admin/users"
+
+    data =
+      %{
+        "name" => username || email,
+        "email" => email || username,
```
IvanIvanoff: this will lead to invalid email, let's avoid `|| username`

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+        "login" => username || email,
+        "password" => :crypto.strong_rand_bytes(16) |> Base.encode64() |> binary_part(0, 16)
```
IvanIvanoff: this is a password that will never be used, right?

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
-         {:ok, %{"id" => user_id}} <- get_user_by_email_or_metamask(user),
          {:ok, team_members} <- get_team_members(team_id) do
       team_members
-      |> Enum.find(fn %{"userId" => uid} -> uid == user_id end)
+      |> Enum.find(fn %{"userId" => user_id} -> user_id == grafana_user_id end)
       |> case do
-        nil -> add_user_to_team(user_id, team_id)
-        _ -> {:ok, "User is already in this team"}
+        nil ->
+          remove_user_from_all_teams(grafana_user_id)
```
IvanIvanoff: Let's remove the user only from the product defined teams. For example I'm in some teams that I've setup manually and this will remove me from them?

IvanIvanoff: Ah, it's actually only those teams, not all teams

lib/sanbase/grafana/api.ex
tspenov: yes, but it is required

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+    |> Enum.each(&remove_user_from_team(user_id, &1))
+  end
+
+  def remove_user_from_team(user_id, team_name) do
+    with {:ok, %{"id" => team_id}} <- get_team_by_name(team_name),
+         {:ok, team_members} <- get_team_members(team_id) do
+      team_members
+      |> Enum.find(fn %{"userId" => uid} -> uid == user_id end)
+      |> case do
+        nil -> {:ok, "User is not in this team"}
```
IvanIvanoff:
```suggestion
        nil -> {:ok, "User is not in the #{team_name} team"}
```
:?

lib/sanbase/grafana/api.ex
tspenov: we have users with only username (from metamask login) and we have to fill something here because it is required

lib/sanbase/grafana/api.ex
```diff
@@ -7,36 +7,100 @@ defmodule Sanbase.GrafanaApi do
...
+        _ -> do_remove_user_from_team(user_id, team_id)
+      end
+    else
+      error ->
+        Logger.error(
+          "Error removing grafana user: #{user_id} from team: #{team_name}: #{inspect(error)}"
+        )
+
+        error
```
IvanIvanoff: Let's pattern match on `{:error, error}` and return `{:error, error` here for better logging. This will also assert that you cannot return strange results that are not a tuple

IvanIvanoff:
@tspenov some tests needs to be added, otherwise LGTM

tspenov:
@IvanIvanoff I'll add tests in another PR. I want to test if this flow actually works on stage before that.  (This won't break anything until we ave the login in sanbase button on Prod)

PR #1555, opened 2019-10-16
Allow updating any field of an insight

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/test/sanbase_web/graphql/insight/insight_api_test.exs
+++ b/test/sanbase_web/graphql/insight/insight_api_test.exs
@@ -647,20 +647,23 @@ defmodule SanbaseWeb.Graphql.InsightApiTest do
       assert String.contains?(error["message"], "Cannot update not owned insight")
     end
 
-    test "cannot update title or text published insight", %{conn: conn, user: user, poll: poll} do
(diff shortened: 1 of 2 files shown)


test/sanbase_web/graphql/insight/insight_api_test.exs
```diff
@@ -660,18 +660,28 @@ defmodule SanbaseWeb.Graphql.InsightApiTest do
...
-            id
+            title
+            text
+            updatedAt
           }
         }
       """
 
+      # So we can check that `updatedAt` is actually updated
+      Process.sleep(1000)
```
tspenov: It should be workind without this ?

IvanIvanoff: No, updated_at is with seconds precision and without the sleep it will be updated to the same value 99.9% of the time

tspenov: You can add another value for `updated_at` - `insert(:post, updated_at: Timex.shift(Timex.now, seconds: -1))` and remove the sleep.

PR #1575, opened 2019-10-24
Kafka graphql subscriptions

tspenov:
```graphql
  subscription($source: String!, $symbol: String) {
    exchangeTrades(source: $source, symbol: $symbol) {
      source
      symbol
      timestamp
      amount
      price
      cost
    }
  }
```

```graphql
  subscription($source: String!, $symbol: String) {
    exchangeMarketDepth(source: $source, symbol: $symbol) {
      source
      symbol
      timestamp
      ask
      bid
      asks025PercentDepth
      asks025PercentVolume
      bids025PercentDepth
      bids025PercentVolume
    }
  }
```

All fields for exchange market depth:
```elixir
["ask", "asks025PercentDepth", "asks025PercentVolume", "asks05PercentDepth",
 "asks05PercentVolume", "asks075PercentDepth", "asks075PercentVolume",
 "asks10PercentDepth", "asks10PercentVolume", "asks1PercentDepth",
 "asks1PercentVolume", "asks20PercentDepth", "asks20PercentVolume",
 "asks2PercentDepth", "asks2PercentVolume", "asks30PercentDepth",
 "asks30PercentVolume", "asks5PercentDepth", "asks5PercentVolume", "bid",
 "bids025PercentDepth", "bids025PercentVolume", "bids05PercentDepth",
 "bids05PercentVolume", "bids075PercentDepth", "bids075PercentVolume",
 "bids10PercentDepth", "bids10PercentVolume", "bids1PercentDepth",
 "bids1PercentVolume", "bids20PercentDepth", "bids20PercentVolume",
 "bids2PercentDepth", "bids2PercentVolume", "bids30PercentDepth",
 "bids30PercentVolume", "bids5PercentDepth", "bids5PercentVolume", "source",
 "symbol", "timestamp"]
```


mix.exs
```diff
@@ -61,7 +61,7 @@ defmodule Sanbase.Mixfile do
       {:guardian, "~> 2.0"},
       {:absinthe, github: "absinthe-graphql/absinthe", override: true},
       {:absinthe_ecto, "~> 0.1.0"},
-      {:absinthe_plug, github: "absinthe-graphql/absinthe_plug"},
+      {:absinthe_phoenix, github: "absinthe-graphql/absinthe_phoenix", override: true},
```
IvanIvanoff: Why remove absinthe_plug from here? This will start using absinthe_plug from hex and not from github. We're using github so we have some fixes in master that are still released

lib/sanbase/message_processor.ex
```diff
@@ -0,0 +1,185 @@
+defmodule Sanbase.Kafka.ExchangeMarketDepth do
```
IvanIvanoff: Let's have the modules with only defstruct in a new file. So we have the `kafka` directory and inside it we have the `topics` subdir with the `Sanbase.Kafka.Topic.ExchangeMarketDepth` and ` Sanbase.Kafka.Topic.ExchangeTrade` modules

lib/sanbase/message_processor.ex
```diff
@@ -0,0 +1,185 @@
...
+    |> publish_async(:exchange_trades)
+
+    :ok
+  end
+
+  def handle_message(%{value: value, topic: "exchange_market_depth"}) do
+    value
+    |> Jason.decode!()
+    |> Sanbase.Kafka.ExchangeMarketDepth.format_message()
+    |> publish_async(:exchange_market_depth)
```
IvanIvanoff: Can these dispatches be automated? You can keep a mapping of topic -> module/absinthe_topic_name

lib/sanbase/message_processor.ex
```diff
@@ -0,0 +1,185 @@
...
+
+    :ok
+  end
+
+  defp publish_async(message, topic) do
+    Absinthe.Subscription.publish(
```
IvanIvanoff: I think we should move such functionality inside sanbase_web/graphql directory

lib/sanbase/message_processor.ex
```diff
@@ -0,0 +1,185 @@
...
+    :ok
+  end
+
+  def handle_message(%{value: value, topic: "exchange_market_depth"}) do
+    value
+    |> Jason.decode!()
+    |> Sanbase.Kafka.ExchangeMarketDepth.format_message()
+    |> publish_async(:exchange_market_depth)
+
+    :ok
```
IvanIvanoff: publish_async already returns :ok

lib/sanbase/message_processor.ex
```diff
@@ -0,0 +1,185 @@
...
+  defp format_side(%{side: side} = exchange_trade) do
+    %{exchange_trade | side: String.to_existing_atom(side)}
+  end
+end
+
+defmodule Sanbase.Kafka.MessageProcessor do
+  def handle_messages(messages) do
+    for message <- messages do
+      handle_message(message)
+    end
```
IvanIvanoff: In case of ignoring the result from a for/map I would use 
```elixir
Enum.each(messages, &handle_message/1)
```
as it's dropping the results already and returns a list of `:ok`s

lib/sanbase/web.ex
```diff
@@ -40,6 +56,16 @@ defmodule Sanbase.Application.Web do
...
       Sanbase.Clickhouse.MarkExchanges,
 
+      # Start Kafka consumer supervisor
+      start_in(
+        %{
+          id: Kaffe.GroupMemberSupervisor,
+          start: {Kaffe.GroupMemberSupervisor, :start_link, []},
+          type: :supervisor
+        },
+        [:prod]
```
IvanIvanoff: Can we have `Sanbase.InMemoryKafka.Consumer` similar to `Sanbase.InMemoryKafka.Consumer` so we can easily use it test/dev envs?

IvanIvanoff: This means that there'll be a consumer behaviour similar to `SanExporterEx.ProducerBehaviour`

config/config.exs
```diff
@@ -16,14 +16,23 @@ config :phoenix, :json_library, Jason
...
+config :sanbase, Sanbase, environment: "#{Mix.env()}"
 
 config :sanbase, Sanbase.ApiCallDataExporter,
   kafka_url: {:system, "KAFKA_URL", "blockchain-kafka-kafka"},
   kafka_port: {:system, "KAFKA_PORT", "9092"}
 
+config :kaffe,
+  consumer: [
+    endpoints: [{"blockchain-kafka-kafka", 9092}],
+    topics: ["exchange_trades", "exchange_market_depth"],
```
IvanIvanoff: Can this be configurable with ENV vars?
Part of the reason I reworked the producer was to be able to config this, too.

lib/sanbase/message_processor.ex
tspenov: what do you mean by automated ? You mean without passing `:exchange_market_depth` ?

lib/sanbase/web.ex
tspenov: Saving produced state in memory instead of going to external storage makes sense but how do you imagine an in memory consumer ?

config/config.exs
tspenov: Here is how I change `consumer_group` runtime: https://github.com/santiment/sanbase2/pull/1575/files#diff-7ab4523cf1166d6e1d3f1c247a29ade7R9
I guess I can make a wrapper module `Sanbase.Kafka` - configure some things with env vars and then change default config on init. For the topics I guess we'll need it if stage and prod have diff topic names. We can have comma separated string and parse it.

lib/sanbase/message_processor.ex
IvanIvanoff: I like it when you can add an element to a list and all needed functions will work as expected. For example in order to add a metric to `getMetric` you need just to add the module to this list: https://github.com/santiment/sanbase2/pull/1513/files#diff-a97d09c368460c1f564cfbc2e0b44e7fR9-R12

and the get/metadata/first_datetime, etc. functions will be automatically created, as it's easy to miss one of them.

lib/sanbase/web.ex
IvanIvanoff: I imagine it as a genserver (?) that "consumes" random values at random intervals, i.e. instead of consuming from kafka, it just generates some random data

tspenov: That might be useful only in dev env, not test. But by just adding `:dev` here and you'll easily consume real stage data.

lib/sanbase/kafka/subscription.ex
```diff
@@ -0,0 +1,22 @@
...
+    "exchange_trades" => :exchange_trades,
+    "exchange_market_depth" => :exchange_market_depth
+  }
+
+  def publish(message, kafka_topic) do
+    Absinthe.Subscription.publish(
+      SanbaseWeb.Endpoint,
+      message,
+      Keyword.new([
+        {@kafka_topic_graphql_subscription_map[kafka_topic], message.source <> message.symbol}
```
IvanIvanoff: Using `[]` will return `nil` in case of mistyped/missing topic, maybe we should crash in this case?
Another way is to have `def publish(...) when kafka_topic in @kafka_topics`

lib/sanbase/message_processor.ex
tspenov: Reworked with behaviour and dispatch module

config/config.exs
tspenov: Added ENV variable config support

test/sanbase_web/graphql/kafka/subscription_test.exs
```diff
@@ -87,6 +87,49 @@ defmodule SanbaseWeb.Graphql.Kafka.SubscriptionTest do
...
+              "timestamp" => "2019-09-28T20:53:45Z"
+            }
+          }
+        },
+        subscriptionId: unquote(subscription_id)
+      }
+    end
+  end
+
+  defmacrop expected_exchange_market_depth(subscription_id) do
```
IvanIvanoff: why macro?

config/config.exs
```diff
@@ -16,14 +16,28 @@ config :phoenix, :json_library, Jason
...
   kafka_port: {:system, "KAFKA_PORT", "9092"}
 
+config :sanbase, Sanbase.Kafka,
+  kafka_url: {:system, "KAFKA_URL", "blockchain-kafka-kafka"},
+  kafka_port: {:system, "KAFKA_PORT", "9092"},
+  topics: {:system, "KAFKA_TOPICS_CSV_STRING", "exchange_trades, exchange_market_depth"}
+
+config :kaffe,
+  consumer: [
+    endpoints: [{"blockchain-kafka-kafka", 9092}],
```
IvanIvanoff: Can we have a list of a value like ["configured-in-supervisor-not-here"] or something like this, because it's confusing

config/config.exs
```diff
@@ -16,14 +16,28 @@ config :phoenix, :json_library, Jason
...
+    topics: ["exchange_trades", "exchange_market_depth"],
```
IvanIvanoff: same here - indicate that it's configured in the supervisor, not here

config/config.exs
```diff
@@ -16,14 +16,28 @@ config :phoenix, :json_library, Jason
...
+    consumer_group: "sanbase_kafka_cansumer",
```
IvanIvanoff: ditto

lib/sanbase/application.ex
```diff
@@ -127,6 +127,7 @@ defmodule Sanbase.Application do
     [
       # Start the endpoint when the application starts
       SanbaseWeb.Endpoint,
+      {Absinthe.Subscription, SanbaseWeb.Endpoint},
```
IvanIvanoff: Only in web?

tspenov: yeah, makes sense

PR #1580, opened 2019-10-28
Add getUser API for fetching public data for users

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

#### Screenshots
image

### DIFF
--- a/lib/sanbase/auth/public.ex
+++ b/lib/sanbase/auth/public.ex
@@ -1,0 +1,18 @@
+defmodule Sanbase.Auth.User.Public do
+  alias Sanbase.Auth.User
+  alias Sanbase.Auth.UserSettings
+
+  @sensitive_fields [:email]
+
+  def hide_private_data(%User{} = user) do
+    case UserSettings.settings_for(user) do
+      %{hide_privacy_data: false} ->
+        user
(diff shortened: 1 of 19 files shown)


lib/sanbase/auth/public.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Sanbase.Auth.User.Public do
+  alias Sanbase.Auth.User
+  alias Sanbase.Auth.UserSettings
+
+  @sensitive_fields [:email]
+
+  def hide_private_data(%User{} = user) do
```
tspenov: I think whitelist is better than blacklist in such cases. Otherwise things could sneak up if don't update `@sensitive_fields`

IvanIvanoff: t will be harder as we need to keep also all types so we know how to replace them.
I also plan to rework it via middleware in an upcoming PR as if we start hiding fields that are resolved with a custom resolver we need to manually call the hiding function every time.

There are also things we want to hide that are not part of the ecto user schema, but from the GQL schema

tspenov: ok, lets merge as it is now

PR #1581, opened 2019-10-29
Add more bot users

tspenov:


lib/sanbase/auth/user.ex
```diff
@@ -402,6 +402,11 @@ defmodule Sanbase.Auth.User do
   def insights_fallback_email, do: @insights_fallback_email
   def sanbase_bot_email, do: @sanbase_bot_email
 
+  def sanbase_bot_email(idx) do
+    [left, right] = String.split(@sanbase_bot_email, "@")
+    left <> to_string(idx) <> right
```
IvanIvanoff: a shorter way: `String.replace(@sanbase_bot_email, "@", "#{idx}@")`

PR #1588, opened 2019-10-30
Rework followers to be list of users

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

#### Screenshots
image

image

### DIFF
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -17,7 +17,7 @@ defmodule Sanbase.Auth.User do
   alias Sanbase.Repo
   alias Sanbase.Telegram
   alias Sanbase.Signal.HistoricalActivity
-  alias Sanbase.Following.UserFollower
+  alias Sanbase.Auth.UserFollower
   alias Sanbase.Billing.Subscription
 
   # The Login links will be valid 1 hour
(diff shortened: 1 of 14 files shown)


lib/sanbase/auth/user.ex
```diff
@@ -17,7 +17,7 @@ defmodule Sanbase.Auth.User do
   alias Sanbase.Repo
   alias Sanbase.Telegram
   alias Sanbase.Signal.HistoricalActivity
-  alias Sanbase.Following.UserFollower
+  alias Sanbase.Auth.UserFollower
```
tspenov: I don't think `Auth` is more appropriate than having own `Following` context :?

IvanIvanoff: Well, all user functionality is inside `Auth`

lib/sanbase_web/graphql/schema/queries/user_queries.ex
```diff
@@ -32,6 +32,7 @@ defmodule SanbaseWeb.Graphql.Schema.UserQueries do
       meta(access: :free)
 
       resolve(&UserResolver.get_user/3)
+      # middleware(MaybeHidePrivacyData)
```
tspenov: commented code

PR #1594, opened 2019-10-31
Age Distribution Histogram

IvanIvanoff:
#### Summary

First, refactor the code base to allow easier addition of the histogram data.

---

image

---
Introduce a new field in the `getMetric` API: `histogramData`:
```graphql
{
  getMetric(metric: "age_distribution") {
    histogramData(slug: "santiment", datetime: "2019-01-01T00:00:00Z"){
      labels
      values
    }
  }
}
```

It returns a histogram for a given datetime represented by 2 lists - one of labels and one of values.

---

The `values` is a union type that can be used like this:
image

---

Introduced 2 new fields on the project GQL type to fetch the metrics:
image

---

Checking the data type of the metric, possible values are histogram and timeseries:
image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/metric/metric.ex
```diff
@@ -72,67 +62,77 @@ defmodule Sanbase.Clickhouse.Metric do
...
+  def histogram_data(metric, slug, from, to, interval, limit) do
+    {query, args} = histogram_data_query(metric, slug, from, to, interval, limit)
+    label_template = Map.get(@metrics_label_map, metric)
+
+    ClickhouseRepo.query_reduce(
+      query,
+      args,
+      %{labels: [], values: []},
+      fn [unix, value], %{labels: labels, values: values} ->
+        label =
```
tspenov: Why don't we just return the datetime as label. The API user knows what is the metric so it can make its own label based on the date if needs to. In this way the date value is part of the text so it needs to parsed if there is a need to be used.

IvanIvanoff: As this is not timeseries data but histogram data returning date times won't be very descriptive (or enough). I could add a union type `raw_labels` that will return the raw label value, but it can be other things than date times. For example, we could have a histogram where every column represent a range of amounts and you plot the transactions count based on their amount (transactions transfering 0-1 ETH, 1-10 ETH, 10-100 ETH, etc. )

I think it's better not to hardcode every metrics description in the API clients - in this case sanpy/sansheets/sanbase all need to somehow know the meaning of every histogram

IvanIvanoff: @Partysun what's your opinion on the topic?

Partysun: @IvanIvanoff Agree with you.

tspenov:
@IvanIvanoff the build is failing

lib/sanbase/metric/metric.ex
```diff
@@ -187,13 +252,38 @@ defmodule Sanbase.Metric do
...
 
     case close do
       nil -> {:error, error_msg}
       close -> {:error, error_msg <> " Did you mean '#{close}'?"}
     end
   end
+
+  defp metric_not_available_error_details(metric, :all) do
+    %{
+      close: Enum.find(@metrics_mapset, fn m -> String.jaro_distance(metric, m) > 0.8 end),
```
tspenov: that can be extracted `find_misspelled(metric, @metrics_mapset)` :?

PR #1599, opened 2019-11-05
Write cmc prices to kafka

tspenov:


config/config.exs
```diff
@@ -28,7 +28,15 @@ config :sanbase, Sanbase.Kafka,
...
   consumer_group_basename: {:system, "KAFKA_CONSUMER_GROUP_BASENAME", "sanbase_kafka_consumer"}
 
+config :sanbase, Sanbase.ApiCallDataExporter,
+  kafka_url: {:system, "KAFKA_URL", "blockchain-kafka-kafka"},
+  kafka_port: {:system, "KAFKA_PORT", "9092"}
+
 config :kaffe,
+  producer: [
+    endpoints: [{"blockchain-kafka-kafka", 9092}],
+    topics: ["sanbase_prices"]
```
IvanIvanoff: the producer is patched and should not be configured via config files. Please check the `prepended_children/0` function inside `application.ex`. The new producer should be started in a similar way.

IvanIvanoff: On second thought, there probably should not be a second producer. If you take a look here at `SanExporterEx.Producer` you see that the function takes the topic as a first argument when sending the data.

IvanIvanoff: There is `ApiCallDataExporter` because the apicalls are batched before they are sent (kept in a genserver's state), otherwise we'll send too many separate messages, which is not optimal for kafka. Probably the same should be done for the prices?

lib/sanbase/external_services/coinmarketcap/price_point.ex
```diff
@@ -12,6 +13,21 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.PricePoint do
     :price_btc
   ]
 
+  def to_json(%__MODULE__{datetime: datetime} = point, slug, source \\ @prices_source) do
+    key = source <> "_" <> slug <> "_" <> to_string(DateTime.to_unix(datetime, :nanosecond))
```
IvanIvanoff: let's not used nanoseconds but seconds

IvanIvanoff: Even better would be to keep it in ISO format for easy reading

config/config.exs
tspenov: extracted ApiCallDataExporter module as common kafka exporter module with buffering so both api calls and prices are buffered before going to kafka.

lib/sanbase/application.ex
```diff
@@ -1,5 +1,6 @@
 defmodule Sanbase.Application do
   use Application
+  import Sanbase.ApplicationUtils
```
IvanIvanoff: add some empty lines to separate the groups

lib/sanbase/application.ex
```diff
@@ -101,6 +102,24 @@ defmodule Sanbase.Application do
...
+       [
+         kafka_producer_module: kafka_producer_supervisor_module(),
+         kafka_endpoint: kafka_endpoint()
+       ]},
+      start_in(
+        Supervisor.child_spec(
+          {Sanbase.KafkaExporter, [name: :prices_exporter, topic: kafka_prices_data_topic()]},
+          id: :prices_exporter
+        ),
+        [:dev, :prod]
```
IvanIvanoff: you have an in-memory producer, so it's ok to start in test env. Furthermore, you have to start it in order to test it

tspenov: In test I start it like this so I can control the options: https://github.com/santiment/sanbase2/pull/1599/files#diff-48187da6c205a7cf97828c62c388598dR18

lib/sanbase/external_services/coinmarketcap/price_point.ex
```diff
@@ -12,6 +13,21 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.PricePoint do
     :price_btc
   ]
 
+  def to_json(%__MODULE__{datetime: datetime} = point, slug, source \\ @prices_source) do
```
IvanIvanoff: `to_json` is producing an elixir tuple, not a json value. Maybe just `format` :?

config/test.exs
```diff
@@ -20,10 +20,12 @@ config :tesla, adapter: Tesla.Mock
 # The logger is causing issues with mocking otherwise. Not really sure why
 config :tesla, Tesla.Middleware.Logger, debug: false
 
-config :sanbase, Sanbase.ApiCallDataExporter,
+config :sanbase, Sanbase.KafkaExporter,
   supervisor: Sanbase.InMemoryKafka.Supervisor,
   producer: Sanbase.InMemoryKafka.Producer
 
+config :sanbase, Sanbase.Kafka.Exporter.Prices, producer: Sanbase.InMemoryKafka.Producer
```
IvanIvanoff: I do not see such module - `Sanbase.Kafka.Exporter.Prices`

lib/sanbase/application.ex
IvanIvanoff: In the ApiCallDataExporter, for example, it's started in `:test` env, too.
It also is custom started with different options in some tests cases. This is because you do not want to start it manually every time you test not exactly this functionality but something that is using it

lib/sanbase/external_services/coinmarketcap/graph_data.ex
```diff
@@ -137,6 +137,10 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.GraphData do
       marketcap_total_list
       |> Enum.flat_map(&PricePoint.price_points_to_measurements(&1, measurement_name))
 
+    marketcap_total_list
+    |> Enum.map(&PricePoint.json_kv_tuple(&1, "TOTAL_MARKET"))
+    |> Sanbase.KafkaExporter.persist(:prices_exporter)
```
IvanIvanoff: move the name to a module attribute `@price_exporter :prices_exporter`

lib/sanbase/external_services/coinmarketcap/price_point.ex
```diff
@@ -12,6 +13,21 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.PricePoint do
     :price_btc
   ]
 
+  def json_kv_tuple(%__MODULE__{datetime: datetime} = point, slug, source \\ @prices_source) do
+    key = source <> "_" <> slug <> "_" <> DateTime.to_iso8601(datetime)
+
+    value =
+      %{
+        timestamp: DateTime.to_unix(datetime),
```
IvanIvanoff: Wouldn't this add both `datetime` and `timestamp` to `value`?

test/sanbase/external_services/coinmarketcap/graph_data_test.exs
```diff
@@ -1,31 +1,48 @@
...
-
   import Sanbase.Factory
   import Sanbase.InfluxdbHelpers
+  import Sanbase.TestHelpers
+
+  alias Sanbase.ExternalServices.Coinmarketcap.{GraphData, PricePoint}
+  alias Sanbase.Prices.Store
 
   @total_market_measurement "TOTAL_MARKET_total-market"
+  @topic "asset_prices"
```
IvanIvanoff: Should this be configurable via ENV vars?

lib/sanbase/application.ex
tspenov: Note that this is the implementation for the `prices_exporter`. In the above link of the test it tests scraping and then exporting to kafka so I have to start it  with the name that is used in the coinmarketcap scraper with some options to prevent sleeping in tests and make sure it is not started in test env before.
I renamed `ApiCallDataExporterTest` to `KafkaExporterTest` because it tests more generally the buffering (it uses the api_call data but still it is more general test).

lib/sanbase/external_services/coinmarketcap/price_point.ex
tspenov: Nope. here is a test: https://github.com/santiment/sanbase2/pull/1599/files#diff-003a5ed0796b49362b53e9045ecd72b5R64

PR #1604, opened 2019-11-07
Config cookie key stage

tspenov:


lib/sanbase_web/plug/session_plug.ex
```diff
@@ -13,10 +13,12 @@ defmodule SanbaseWeb.Plug.SessionPlug do
     runtime_opts =
       opts
       |> Keyword.put(:domain, domain())
+      |> Keyword.put(:key, session_key())
```
IvanIvanoff: probably won't affect performance - but as this is executed on every request can we just call the `Config.get` here to avoid additional function calls

PR #1606, opened 2019-11-08
Handle 429 rate limit response from CMC when fetching project page

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap/scraper.ex
+++ b/lib/sanbase/external_services/coinmarketcap/scraper.ex
@@ -17,6 +17,10 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.Scraper do
       {:ok, %Tesla.Env{status: 200, body: body}} ->
         {:ok, body}
 
+      {:ok, %Tesla.Env{status: 429} = resp} ->
+        wait_rate_limit(resp)
+        fetch_project_page(coinmarketcap_id)
+
       {:ok, %Tesla.Env{status: status}} ->
         error_msg = "Failed fetching project page for #{coinmarketcap_id}. Status: #{status}."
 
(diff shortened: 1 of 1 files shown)


lib/sanbase/external_services/coinmarketcap/scraper.ex
```diff
@@ -98,4 +102,15 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.Scraper do
...
+
+  defp wait_rate_limit(%Tesla.Env{status: 429, headers: headers}) do
+    {_, wait_period} =
+      Enum.find(headers, fn {header, _} ->
+        header == "retry-after"
+      end)
+
+    wait_period = String.to_integer(wait_period)
+    wait_until = Timex.shift(Timex.now(), seconds: wait_period)
+    Sanbase.ExternalServices.RateLimiting.Server.wait_until(@rate_limiting_server, wait_until)
```
tspenov: just a question - what is the benefit than just calling `Process.sleep(wait_period * 1000)` ?

IvanIvanoff: Now that I think about - it's probably even wrong to do that, as the sleep is handled by a genserver call. So if we want to sleep for 2 seconds, but there's an ongoing sleep for 1 minute - we'll wait 1 minute

IvanIvanoff: Now that I think about it - this behaviour should be correct.
If we hit multiple rate limits for the same rate limiting server `@rate_limiting_server` then only one of them, then it's ok to wait on other sleeps, so when it's your turn to sleep the time would have already passed.

Additionally we leverage the logging functionality from the same place and don't care about it here

PR #1609, opened 2019-11-11
Allow hide projects

IvanIvanoff:
#### Summary
- Add a field `is_hidden` which is `false` by default.
- Modify the `Project.List` functions to use options list instead of argument per option.
- Modify scrapers to provide `include_hidden_projects?: true` so they scrape data for all projects
- Do not show hidden projects from the API results

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/structure.sql
```diff
@@ -12,10 +12,24 @@ SET client_encoding = 'UTF8';
...
-SET xmloption = ontent;
+SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
+--
+-- Name: timescaledb; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS timescaledb WITH SCHEMA public;
```
tspenov: you can remove timescaledb extension locally and commit structure.sql again to remove these

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+  @preloads [:eth_addresses, :latest_coinmarketcap_data, :github_organizations]
+
+  @moduledoc ~s"""
+  Provide functions for fetching different subsets of projects.
+
+  Provided functions for getting a list of projects based on different filterings.
+
+  ## Shared options
+  Most of the functions accept a keyword options list as the last arguments.
+  Currently two options are supported:
```
tspenov:
```suggestion
  Currently listed below options are supported:
```

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+    - `:preload` (boolean) - Do or do not preload #{inspect(@preloads)}
```
tspenov: `:preload?` :?

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+        from(
+          p in query,
+          inner_join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
+          where: latest_cmc.volume_usd >= ^min_volume,
+          order_by: latest_cmc.rank
+        )
+    end
+  end
+
+  defp maybe_include_hidden_projects?(query, opts) do
```
tspenov: IMO `?` at the end is not needed when starting with `maybe_...`

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+  end
+
+  defp order_by_rank(query) do
+    from(p in query,
+      inner_join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
+      order_by: latest_cmc.rank
+    )
+  end
+
+  defp maybe_order_by_rank_above_volume(query, opts) do
```
tspenov: what happens `:min_volume` is not `nil`  and we have:
```elixir   
    projects_query(opts)
    |> order_by_rank()
```
then we'll have 2 `inner_join: latest_cmc in assoc(p, :latest_coinmarketcap_data)` with diff where - is it working as expected ?

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+    case Keyword.get(opts, :min_volume, nil) do
```
tspenov: if `:min_volume` doesn't exist it defaults to `nil` - so third arg is not needed

lib/sanbase/model/project/list/list.ex
```diff
@@ -0,0 +1,509 @@
...
+  def by_field(values, field, opts \\ [])
+
+  def by_field(values, field, opts) when is_list(values) do
+    projects_query(opts)
+    |> where([p], field(p, ^field) in ^values)
+    |> Repo.all()
+  end
+
+  @doc ~s"""
+  Return a list of projects have their ticker, name and/or slug in the list of
```
tspenov:
```suggestion
  Return a list of projects that have their ticker, name and/or slug in the list of
```

lib/sanbase/model/project/list/list.ex
IvanIvanoff: yeah, I'm just being explicit that `nil` is a supported value for a volume

lib/sanbase/model/project/list/list.ex
IvanIvanoff: `order_by_rank` does not have a where clause - it just adds a `order_by: latest_cmc.rank` which is the same as the `maybe_order_by_rank_above_volume`

tspenov:
Nice work on this 🎉 I also think keyword args as options is a lot more extensible in the future. I left some small comments otherwise 👍

lib/sanbase/model/project/list/list.ex
IvanIvanoff: My tests showed it should be working as expected

priv/repo/structure.sql
IvanIvanoff: oops, thanks

lib/sanbase/model/project/list/list.ex
tspenov: Can you add a test for this case ? Also as these query is used lots by frontend it is better to optimize it (by removing the double join of same table) it even if we should add extra check.
we can do it by passing opts to order_by_rank like this:
```elixir
  defp order_by_rank(query, opts) do
    case Keyword.get(opts, :min_volume) do
      nil ->
        from(p in query,
          inner_join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
          order_by: latest_cmc.rank
        )

      min_volume ->
        query
    end
  end
```

PR #1611, opened 2019-11-12
Refactorings

IvanIvanoff:
#### Summary
- Minimize the usage of the Repo module inside resolvers
- Merge similar resolvers
- Remove the Poll module (the poll table should be dropped later as during our deployment we'll have old & new code running in parallel)

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


priv/repo/structure.sql
```diff
@@ -866,7 +880,8 @@ CREATE TABLE public.project (
     main_contract_address character varying(255),
     long_description text,
     logo_url character varying(255),
-    slug character varying(255)
+    slug character varying(255),
+    is_hidden_from_lists boolean DEFAULT false
```
tspenov: field from other migration

priv/repo/migrations/20191112123819_remove_poll_table.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Sanbase.Repo.Migrations.RemovePollTable do
```
tspenov: maybe RemovePollAssocFromInsights ?

PR #1615, opened 2019-11-15
add images for new story

GarageInc:
**Summary**

Add images for new story

And why do we store images for frontend app in other repository/backend? It looks strange(as saving all icons here) :-)


haritonasty:
I'm agree. So uncomfortable

IvanIvanoff:
I'm not quite sure why you store them here instead of s3?

haritonasty:
how load to s3, @IvanIvanoff ? Through insight?

IvanIvanoff:
@haritonasty currently yes. but through s3 we have a distinction between stage and prod. Let's move forward with this PR and think of a long term solution

PR #1616, opened 2019-11-15
Migrate prices to kafka

tspenov:
image


lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,73 @@
...
+  alias Sanbase.ExternalServices.Coinmarketcap.PricePoint
+
+  @chunk_days 20
+
+  def run() do
+    projects = Project.List.projects()
+    Logger.info("Migrating prices from influxdb for count: #{length(projects)} projects")
+
+    projects
+    |> Enum.map(fn %Project{ticker: ticker, slug: slug} -> ticker <> "_" <> slug end)
```
IvanIvanoff: Enum.map(&Sanbase.Influxdb.Measurement.name_from/1)

lib/sanbase/prices/store.ex
```diff
@@ -22,6 +22,24 @@ defmodule Sanbase.Prices.Store do
     @last_history_price_cmc_measurement
   end
 
+  def first_datetime_multiple_measurements(measurements) when is_list(measurements) do
+    measurements_str = measurements |> Enum.map(fn x -> ~s/"#{x}"/ end) |> Enum.join(", ")
+
+    ~s/SELECT time, price_usd FROM #{measurements_str} ORDER BY time ASC LIMIT 1/
```
IvanIvanoff: why not just `first(price_usd)`?

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,73 @@
...
+    |> Enum.chunk_every(20)
+    |> Enum.flat_map(&first_datetimes/1)
+    |> Enum.each(&get_prices_and_persist_in_kafka/1)
+  end
+
+  def get_prices_and_persist_in_kafka({measurement, first_datetime_iso}) do
```
IvanIvanoff: split into 2 functions, each doing one thing

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,73 @@
...
+      else
+        nil
+      end
+    end)
+    |> Enum.to_list()
+    |> List.insert_at(-1, now)
+    |> Enum.chunk_every(2, 1, :discard)
+  end
+
+  def get_prices(measurement, from, to) do
```
IvanIvanoff: some of those functions should be private

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,73 @@
...
+
+  def run() do
+    projects = Project.List.projects()
+    Logger.info("Migrating prices from influxdb for count: #{length(projects)} projects")
+
+    projects
+    |> Enum.map(fn %Project{ticker: ticker, slug: slug} -> ticker <> "_" <> slug end)
+    |> Enum.chunk_every(20)
+    |> Enum.flat_map(&first_datetimes/1)
+    |> Enum.each(&get_prices_and_persist_in_kafka/1)
```
IvanIvanoff: What if a push to kafka fails? Should we track progress and restart from the last successful one in case of crash

lib/mix/prices/migrate.ex
tspenov: it persist on every prices chunk so it doesn't make sense to split here (though function name contains `and`)

lib/mix/prices/migrate.ex
tspenov: I added better logging. Lets first see if smth will fail and then adjust.

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  def do_work() do
+    projects = Project.List.projects()
+    all_projects_count = length(projects)
+    Logger.info("Migrating prices from influxdb for count: #{all_projects_count} projects")
+
+    projects
+    |> Enum.take(10)
+    |> Enum.map(&Sanbase.Influxdb.Measurement.name_from/1)
+    |> Enum.chunk_every(50)
```
IvanIvanoff: `Enum.take(10)` will take only 10 projects - this `Enum.chunk_every(50)` has no effect

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,107 @@
...
+    |> Enum.flat_map(&first_datetimes/1)
+    |> Enum.reduce(1, fn {measurement, first_datetime_iso}, current_project_count ->
+      Logger.info("Start migrating #{measurement}")
+
+      {time_microsec, _} =
+        :timer.tc(fn -> get_prices_and_persist_in_kafka({measurement, first_datetime_iso}) end)
+
+      Logger.info(
+        "Migrating #{measurement} finished in #{time_microsec / 1_000_000}s. Progress #{
```
IvanIvanoff: You need to check the result(s) of `get_prices_and_persist_in_kafka` in order to detect issues and see if the migration finished

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,107 @@
...
+
+  def do_work() do
+    projects = Project.List.projects()
+    all_projects_count = length(projects)
+    Logger.info("Migrating prices from influxdb for count: #{all_projects_count} projects")
+
+    projects
+    |> Enum.map(&Sanbase.Influxdb.Measurement.name_from/1)
+    |> Enum.chunk_every(50)
+    |> Enum.flat_map(&first_datetimes/1)
```
IvanIvanoff: The `name_from` will only use the ticker and slug and not check if there is such record in influxdb. Should we somewhere plug a `reject` for some missing/nil values?

lib/mix/prices/migrate.ex
```diff
@@ -0,0 +1,107 @@
...
+          current_project_count
+        } / #{all_projects_count}"
+      )
+
+      current_project_count + 1
+    end)
+  end
+
+  defp get_prices_and_persist_in_kafka({measurement, first_datetime_iso}) do
+    slug = String.split(measurement, "_", parts: 2) |> List.last()
```
IvanIvanoff:
```suggestion
    [_ticker, slug] = String.split(measurement, "_", parts: 2)
```

lib/mix/prices/migrate.ex
tspenov: these should be rejected: 
```
[17:19:52.976][debug][] [query] SELECT first(price_usd) FROM "NE_notexist", "ZRX_0x"
[{"ZRX_0x", "2017-08-16T14:09:21Z"}]
```

PR #1627, opened 2019-11-25
Extend signal calculation time

IvanIvanoff:
- #### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/evaluator/evaluator.ex
```diff
@@ -29,8 +29,8 @@ defmodule Sanbase.Signal.Evaluator do
     |> Sanbase.Parallel.map(
       &evaluate/1,
       ordered: false,
-      max_concurrency: 100,
-      timeout: 30_000
+      max_concurrency: 10,
```
IvanIvanoff: reducing the number of concurrent signals being evaluated will also speed the individual evaluation time as they won't fight for resources that much

PR #1631, opened 2019-11-27
Remove unused

tspenov:


lib/sanbase/billing/subscription/promo_coupon.ex
```diff
@@ -1,150 +0,0 @@
...
-    redeem_by: Sanbase.DateTimeUtils.from_iso8601_to_unix!(@promo_end_datetime)
-  }
-  @promo_email_subject %{
-    en: "Get #{@promo_coupon_percent_off}% off ANY Santiment product!",
-    jp: "Santimentの、どのプロダクトも#{@promo_coupon_percent_off}%オフに!"
-  }
-
-  @type send_coupon_args :: %{email: String.t(), message: String.t() | nil}
-
-  schema "promo_coupons" do
```
IvanIvanoff: Do we need to drop this table, too?

tspenov: I'll remove it in subsequent PR

PR #1633, opened 2019-11-27
Wallet signal

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/trigger/settings/metric_trigger_settings.ex
```diff
@@ -83,7 +83,7 @@ defmodule Sanbase.Signal.Trigger.MetricTriggerSettings do
 
     Cache.get_or_store(cache_key, fn ->
       case Metric.timeseries_data(metric, slug, from, to, interval) do
-        {:ok, [_ | _] = result} -> result
+        {:ok, [_ | _] = result} -> result |> Enum.take(-2)
```
tspenov: why taking only last 2 ?

IvanIvanoff: long story short - you are comparing the data now vs the data `time_window` time ago, so the last 2 are enough for this. It's explicitly needed because above we have `3 * time_window_sec` to ensure that there is data. If we do not trim it like this, the percent change computation will be mistaken. 

the way percent change is calculated is to take the last element and take the average of the list without the last element (check ResultBuilder.Transformer)

PR #1649, opened 2019-12-05
Exchange market pairs mapping

tspenov:


lib/sanbase/exchanges/market_pair_mapping.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Sanbase.Exchanges.MarketPairMapping do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "exchange_market_pair_mappings" do
+    field(:exchange, :string)
```
IvanIvanoff: At some point we need to introduce an `exchanges`  table as now we'll use the exchange name as a string both here and in `exchange_addresses`.

lib/sanbase/exchanges/market_pair_mapping.ex
```diff
@@ -0,0 +1,39 @@
...
+    field(:market_pair, :string)
+    field(:from_ticker, :string)
+    field(:to_ticker, :string)
+    field(:from_slug, :string)
+    field(:to_slug, :string)
+    field(:source, :string)
```
IvanIvanoff: do we really need source here? The mapping for a given exchange and pair is unique and there is no need for `source` param to distinguish where we get that from.

priv/repo/migrations/20191205141013_create_exchange_market_pair_mappings.exs
```diff
@@ -0,0 +1,19 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:exchange_market_pair_mappings) do
+      add(:exchange, :string)
+      add(:market_pair, :string)
+      add(:from_ticker, :string)
+      add(:to_ticker, :string)
+      add(:from_slug, :string)
+      add(:to_slug, :string)
```
IvanIvanoff: all should have `null: false` :?

priv/repo/migrations/20191205141411_fill_exchange_market_pairs_from_csv.exs
```diff
@@ -0,0 +1,45 @@
...
+  @source "coinmarketcap"
+
+  def up do
+    setup()
+
+    data =
+      Path.expand("exchange_market_pairs_final.csv", __DIR__)
+      |> File.read!()
+      |> String.split("\n", trim: true)
+      |> Enum.map(fn line ->
```
IvanIvanoff: BTW, you can stream files line-by-line: `File.stream!(path, [], :line)` and directly pipe that into the `Enum.map/2`

priv/repo/migrations/20191205141411_fill_exchange_market_pairs_from_csv.exs
```diff
@@ -0,0 +1,45 @@
...
+          to_slug: to_slug,
+          from_ticker: from_ticker,
+          to_ticker: to_ticker,
+          source: @source,
+          inserted_at: Timex.now(),
+          updated_at: Timex.now()
+        }
+      end)
+
+    Sanbase.Repo.insert_all(MarketPairMapping, data)
```
IvanIvanoff: Add `on_conflict: :do_nothing` param to the insert. On prod there's a high chance the migration will take more than a few seconds and 2 pods will try to execute it and fail on the insert

priv/repo/migrations/20191205141411_fill_exchange_market_pairs_from_csv.exs
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def down do
+    setup()
+
+    Sanbase.Repo.delete_all(MarketPairMapping)
```
IvanIvanoff: Maybe a do nothing down here :? As it can drop some manually inserted records

lib/sanbase/exchanges/market_pair_mapping.ex
tspenov: yes, I was thinking about that too. But lets extend later.

lib/sanbase/exchanges/market_pair_mapping.ex
tspenov: The idea is that here we are mapping only CMC projects to exchange ones but we want to map it to our source + slug.  So if project is not from CMC it will be from other source. The better thing is to map to `project` directly via `slug_source_mapping` instead of `source` + `slug`. But lets start like this for now.

PR #1650, opened 2019-12-06
Migrate price from InfluxDB to ClickHouse

IvanIvanoff:
#### Summary
The GraphQL API is not changed. The only change is that it now will return `null`s for missing data instead of skipping the whole data point.

This PR is big because the price data is used everywhere and we cannot replace its logic without reworking the tests. Big part of the changes come from them. 
The prices are also used in the signals, APIs, charting, etc. - all these places need to be reworked, too.

- Migrate to using prices from ClickHouse instead of InfluxDB.
- Deprecate the `Sanbase.Prices.Store` and `Sanbase.Prices.Utils` modules in favor of `Sanbase.Price` and `Sanbase.Price.Utils`. The SQL for the queries is stored in `Sanbase.Price.SqlQuery`
- The new module has a different API that addresses some bad decisions made when the old pricing module has been written

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


test/sanbase_web/graphql/projects/project_api_combined_stats_test.exs
```diff
@@ -5,143 +5,87 @@ defmodule SanbaseWeb.Graphql.ProjectApiCombinedStatsTest do
...
+                 "projectsListHistoryStats" => [
+                   %{"datetime" => "2017-05-13T00:00:00Z", "marketcap" => 545, "volume" => 220},
+                   %{"datetime" => "2017-05-14T00:00:00Z", "marketcap" => 2000, "volume" => 1400},
+                   %{"datetime" => "2017-05-15T00:00:00Z", "marketcap" => 2600, "volume" => 1600},
+                   %{"datetime" => "2017-05-16T00:00:00Z", "marketcap" => nil, "volume" => nil}
+                 ]
+               }
+             }
+    end
+    |> Sanbase.Mock.with_mock2(&Sanbase.ClickhouseRepo.query/2, {:ok, %{rows: data}})
```
valo: I think this line might be better to be at the top of the lambda. The idea is to see the mocking logic before the code, which will use it. Right now you see the code and later see the mocking logic

IvanIvanoff: I can switch the lambda (assert function) to be the last param and be used like
```elixir
with_mock(&Sanbase.Repo.query/2, response, fn ->

end)
```
but this introduces nesting and removes the ability to pipe. I want to be able to pipe so I can mock multiple functions and end up with code like
```elixir
fn -> code end
|> with_mock2(&StripeApi.create_customer/2, resp1)
|> with_mock2(&StripeApi.create_coupon/1, resp2)
|> with_mock2(&StripeApi.create_subscription/1, resp3)
|> run_with_mocks()
```

which currently is written as passing the mocked functions and data first like this:
```elixir
 with_mocks([
        {StripeApi, [:passthrough],
         [create_customer: fn _, _ -> StripeApiTestReponse.create_or_update_customer_resp() end]},
        {StripeApi, [:passthrough],
         [create_coupon: fn _ -> StripeApiTestReponse.create_coupon_resp() end]},
        {StripeApi, [:passthrough],
         [
           create_subscription: fn _ ->
             {:error, %Stripe.Error{message: "test error", source: "ala", code: "bala"}}
           end
         ]}
      ]) do
<code>
end
```

IvanIvanoff: @valo this is now working:
```elixir
  test "fetch project ROI", context do
    response = last_record_before_fixture(context)

    Sanbase.Mock.init()
    |> Sanbase.Mock.prepare_mock({&Sanbase.Price.last_record_before/2, response})
    |> Sanbase.Mock.run_with_mocks(fn ->
      assert get_roi(context.conn, context.project) == %{"roiUsd" => "2.5"}
    end)
  end
```

valo: I like it :)

lib/sanbase/prices/price.ex
```diff
@@ -0,0 +1,486 @@
...
+  def slugs_with_volume_over(volume, opts \\ [])
+
+  def slugs_with_volume_over(volume, opts) when is_number(volume) do
+    source = Keyword.get(opts, :source, "coinmarketcap")
+    {query, args} = slugs_with_volume_over_query(volume, source)
+
+    ClickhouseRepo.query_transform(query, args, fn [slug] -> slug end)
+  end
+
+  # Get a
```
tspenov: started description/comment

lib/sanbase/prices/price.ex
```diff
@@ -0,0 +1,486 @@
...
+    |> Enum.map(fn list ->
+      %{datetime: datetime} = List.last(list)
+
+      data =
+        Enum.reduce(
+          list,
+          %{volume: 0, volume_usd: 0, marketcap: 0, marketcap_usd: 0},
+          fn elem, acc ->
+            %{
+              marketcap: acc.marketcap + elem.marketcap,
```
tspenov: `elem.marketcap` can be `nil` here because values are passed through `maybe_nullify_values`.

lib/sanbase/prices/price.ex
```diff
@@ -0,0 +1,486 @@
...
+          marketcap_usd: float() | nil,
+          volume: float() | nil,
+          volume_usd: float() | nil
+        }
+
+  @type last_record_before_result :: {:ok, last_record_before_map()} | {:error, error()}
+
+  @type combined_marketcap_and_volume_map :: %{
+          datetime: DateTime.t(),
+          marketcap_usd: float() | nil,
```
tspenov: can it be `nil` here ?

lib/sanbase/prices/price.ex
```diff
@@ -0,0 +1,486 @@
...
+    field(:source, :string)
+    field(:slug, :string)
+    field(:price_usd, :float)
+    field(:price_btc, :float)
+    field(:marketcap_usd, :float)
+    field(:volume_usd, :float)
+  end
+
+  @spec changeest(any(), any()) :: no_return()
+  def changeest(_, _), do: raise("Cannot change the asset_prices table")
```
tspenov: type - `changeset`

lib/sanbase/prices/price.ex
```diff
@@ -0,0 +1,486 @@
...
+              marketcap_usd: acc.marketcap_usd + elem.marketcap_usd,
+              volume: acc.volume + elem.volume,
+              volume_usd: acc.volume_usd + elem.volume_usd
+            }
+          end
+        )
+
+      Map.put(data, :datetime, datetime)
+    end)
+    |> (&{:ok, &1}).()
```
tspenov: Maybe its just me but it looks a bit more cryptic than just match into results and return `{:ok, results}`

lib/sanbase/clickhouse_repo.ex
```diff
@@ -67,6 +67,12 @@ defmodule Sanbase.ClickhouseRepo do
         end
       rescue
         e ->
+          IO.warn("""
```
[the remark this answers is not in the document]
IvanIvanoff: this is a leftover, I should remove it.

lib/sanbase/prices/price_sql_query.ex
```diff
@@ -0,0 +1,360 @@
...
+        #{aggregation(price_aggr, "price_usd", "dt")} AS price_usd,
+        #{aggregation(price_aggr, "price_btc", "dt")} AS price_btc,
+        #{aggregation(marketcap_aggr, "marketcap_usd", "dt")} AS marketcap_usd,
+        #{aggregation(volume_aggr, "volume_usd", "dt")} AS volume_usd,
+        toUInt32(1) AS has_changed
+      FROM #{@table}
+        PREWHERE slug IN (?1) AND
+        dt >= toDateTime(?2) AND
+        dt < toDateTime(?3) AND
+        source = cast(?4, 'LowCardinality(String)')
```
tspenov: why are you casting here but not in `timeseries_data_query` ?

lib/sanbase/prices/price.ex
IvanIvanoff: None of those can be nil

lib/sanbase/prices/price.ex
IvanIvanoff: I'm trying how it looks like with long pipe chains as otherwise everything gets nested with 4 more spaces. We can have a function in the utils module `wrap_ok` :?
```elixir
code
|> code
|> code
|> code
|> wrap_ok()
```

instead of
```elixir
result =
  code
  |> code
  |> code
  |> code

{:ok, result}
```

IvanIvanoff: I'll make a `result = ... ;{:ok result}` now

lib/sanbase/prices/price_sql_query.ex
IvanIvanoff: not sure.. it seems to work but it didn't work in some of the cases. I'll add the explicit casting as I'm not sure if it isn't casting the low cardinality to just string and then compare which would be slower

lib/sanbase/chart/chart.ex
```diff
@@ -88,6 +85,18 @@ defmodule Sanbase.Chart do
   end
 
   defp candlestick_prices(ohlc) do
+    Enum.map(
+      ohlc,
+      fn %{open_price_usd: open, close_price_usd: close, high_price_usd: high, low_price_usd: low} ->
+        [
+          open |> Math.to_float(0.0) |> Float.round(6),
```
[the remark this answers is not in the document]
IvanIvanoff: IMO we should keep them distinct as it's in the Elixir codebase - for example we have `String.from_float/1` which does not allow you to control precision

lib/sanbase_web/graphql/resolvers/price_resolver.ex
```diff
@@ -3,69 +3,62 @@ defmodule SanbaseWeb.Graphql.Resolvers.PriceResolver do
...
+  def history_price(_root, %{slug: @total_market} = args, _resolution) do
+    %{from: from, to: to, interval: interval} = args
+
     with {:ok, from, to, interval} <-
-           calibrate_interval(Store, @total_market_measurement, from, to, interval, 300),
-         {:ok, result} <-
-           Store.fetch_prices_with_resolution(@total_market_measurement, from, to, interval) do
-      {:ok, result |> map_price_data()}
+           calibrate_interval(Price, @total_market_slug, from, to, interval, 300),
+         {:ok, result} <- Price.timeseries_data(@total_market_slug, from, to, interval) do
```
tspenov: we should user `TOTAL_MARKET` for slug when select. Here we save to kafka with `TOTAL_MARKET`: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/external_services/coinmarketcap/web_api.ex#L177

PR #1657, opened 2019-12-11
Sync users with intercom

tspenov:


lib/sanbase/intercom/intercom.ex
```diff
@@ -0,0 +1,171 @@
...
+
+  def get_data_for_user(user_id) do
+    Repo.get(User, user_id)
+    |> fetch_stats_for_user()
+    |> Jason.encode!()
+  end
+
+  def fetch_users() do
+    from(u in User, order_by: u.id)
+    |> Repo.all()
```
IvanIvanoff: Can we avoid using `Repo` with a different module and delegate everything to the `User` module?

lib/sanbase/intercom/intercom.ex
```diff
@@ -0,0 +1,171 @@
...
+
+  def sync_users do
+    fetch_users()
+    |> Stream.map(&fetch_stats_for_user/1)
+    |> Stream.map(&Jason.encode!/1)
+    |> Enum.each(&send_user_stats_to_intercom/1)
+  end
+
+  def get_data_for_user(user_id) do
+    Repo.get(User, user_id)
```
IvanIvanoff: Same here - avoid usign` Repo.get` but `User.by_id`

lib/sanbase/intercom/intercom.ex
```diff
@@ -0,0 +1,171 @@
...
+    stats = %{
+      user_id: id,
+      email: email,
+      name: username,
+      custom_attributes:
+        %{
+          all_watchlists_count: all_watchlists_count(user),
+          all_triggers_count: all_triggers_count(user),
+          all_insights_count: all_insights_count(user),
+          staked_san_tokens: format_balance(san_balance),
```
IvanIvanoff: we have over 4k registered users - running this function in the `map` will make over 16k DB calls. Can we run a few queries to aggregate the data and return a list of tuples {user_id, user_aggergated_stats}

IvanIvanoff: The place for these functions are the statistics module. For example you can use the existing `Sanbase.UserLists.Statistics.users_with_watchlist_count` which will return all users with watchlist. You can build a map from that and replace here with `all_watchlists_count: Map.get(prebuild_map, user.id, 0)` - essentially doing this with a single query.

IvanIvanoff: The same module should be created for the triggers and insights

lib/sanbase/intercom/intercom.ex
tspenov: that will be removed later after testing

lib/sanbase/intercom/intercom.ex
tspenov: Done

PR #1667, opened 2019-12-16
Better mask email in payment discord notif

tspenov:

### DIFF
--- a/lib/sanbase/billing/stripe_event.ex
+++ b/lib/sanbase/billing/stripe_event.ex
@@ -251,7 +251,11 @@ defmodule Sanbase.Billing.StripeEvent do
   end
 
   defp mask_user(%User{email: email}) when is_binary(email) do
-    Regex.replace(~r/(.*@)/, email, "***@")
+    [username, domain] = email |> String.split("@")
+
+    masked_username = Regex.replace(~r|(.{1}).{3}(.*)|, username, "\\1***\\2")
+
+    "#{masked_username}@#{domain}"
   end
 
   defp mask_user(_) do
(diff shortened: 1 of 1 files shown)


lib/sanbase/billing/stripe_event.ex
```diff
@@ -251,7 +251,11 @@ defmodule Sanbase.Billing.StripeEvent do
   end
 
   defp mask_user(%User{email: email}) when is_binary(email) do
-    Regex.replace(~r/(.*@)/, email, "***@")
+    [username, domain] = email |> String.split("@")
+
+    masked_username = Regex.replace(~r|(.{1}).{3}(.*)|, username, "\\1***\\2")
```
IvanIvanoff:
```elixir
Regex.replace(~r|(.{1}).{3}(.*)|, "aaaaaaaaaaaaaaaaaaaaaaaaa", "\\1***\\2")
"a***aaaaaaaaaaaaaaaaaaaaa"
```

Do we want this behavior? I think it would be better to reveal only first and last letter? Something like this:
```elixir
iex(18)> username = "ivan"
"ivan"
iex(19)> String.duplicate("*", String.length(username)) |> 
String.replace_prefix("*", String.at(username, 0)) |> 
String.replace_suffix("*", String.at(username, -1))
```

tspenov: IMO first and last letter will not be sufficient to remind you efficiently who is the user. By filtering only some chars you will and you won't expose whole email.

IvanIvanoff: The idea is to hide the email, not reveal it. In case of my email if it starts with `i***n.ivanoff...` it's just as if nothing was hidden 😃

tspenov: yeah, but it is still unusable if scraped automatically (though the channel is private) and immediately recognizable by human (and especially someone from bussiness if had previous interaction)

PR #1673, opened 2019-12-18
Rework price scraping

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/kafka/kafka_exporter.ex
```diff
@@ -99,34 +99,30 @@ defmodule Sanbase.KafkaExporter do
   @spec handle_call({:persist, data | [data]}, any(), state) :: {:reply, result, state}
         when state: map()
   def handle_call({:persist, data}, _from, state) do
-    data = List.wrap(data)
-    new_messages_length = length(data)
+    data = data = List.wrap(data)
+    can_send_after = Timex.shift(Timex.now(), milliseconds: state.can_send_after_interval)
```
tspenov: I think that here that should be 0 since there is no buffer and its completely synchronious. And we should have: 
```elixir
  defp send_data(data, %{topic: topic, can_send_after:0, size: size}) do
    Logger.info("Sending #{size} events to Kafka topic: #{topic}")
    @producer.send_data(topic, data)
  end
```
without sleep.

test/sanbase/external_services/coinmarketcap/web_api_test.exs
```diff
@@ -0,0 +1,113 @@
...
+          build(:source_slug_mapping, %{source: "coinmarketcap", slug: "bitcoin"})
+        ]
+      })
+
+    insert(:latest_cmc_data, %{coinmarketcap_id: "bitcoin"})
+
+    [project: project]
+  end
+
+  @moduletag capture_log: true
```
tspenov: lets move on top of setup with other module attributes

tspenov:
@IvanIvanoff We can add `PriceScrapingProgress`  to admin for easier overview of last datetime for each slug :?

lib/sanbase/external_services/coinmarketcap/web_api.ex
```diff
@@ -0,0 +1,318 @@
...
+  plug(Sanbase.ExternalServices.ErrorCatcher.Middleware)
+  plug(Tesla.Middleware.BaseUrl, "https://web-api.coinmarketcap.com")
+  plug(Tesla.Middleware.Compression)
+  plug(Tesla.Middleware.Logger)
+
+  @source "coinmarketcap"
+  @prices_exporter :prices_exporter
+  @total_market_measurement "TOTAL_MARKET_total-market"
+  @doc ~s"""
+  Return the first datetime for which a given asset (a projct or the whole market)
```
tspenov:
```suggestion
  Return the first datetime for which a given asset (a project or the whole market)
```

lib/sanbase/external_services/coinmarketcap/web_api.ex
```diff
@@ -0,0 +1,318 @@
...
+  end
+
+  defp store_price_points("TOTAL_MARKET", [], {_, to}) do
+    PriceScrapingProgress.store_progress("TOTAL_MARKET", @source, DateTime.from_unix!(to))
+  end
+
+  defp store_price_points(%Project{slug: slug} = project, price_points, _) do
+    %{datetime: latest_datetime} = Enum.max_by(price_points, & &1.datetime)
+
+    :ok = export_prices_to_kafka(project, price_points)
```
tspenov: These (kafka and influx data storing) are switched for the `TOTAL_MARKET` case

lib/sanbase/kafka/kafka_exporter.ex
IvanIvanoff: There shouldn't be a special case as in the case of `0` it will just not sleep.
If we modify the `can_send_after` it will change the state forever (it's datetime, not timedelta). In this case we should explicitly call `send_data_immediately` function or something similar

IvanIvanoff: reworked

lib/sanbase/external_services/coinmarketcap/web_api.ex
IvanIvanoff: I'll reorder them, although it should make no difference

PR #1674, opened 2019-12-19
Add more stats intercom

tspenov:
Intercom custom attributes:
* used_sanapi
* used_sansheets
* api_calls_count
* weekly_report_watchlist_count

### DIFF
--- a/lib/sanbase/clickhouse/api_call_data.ex
+++ b/lib/sanbase/clickhouse/api_call_data.ex
@@ -2,6 +2,7 @@ defmodule Sanbase.Clickhouse.ApiCallData do
   @moduledoc ~s"""
   Get data about the API Calls that were made by users
   """
+
   require Sanbase.ClickhouseRepo, as: ClickhouseRepo
 
   @doc ~s"""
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -31,6 +32,38 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+      {:error, _error} -> []
+    end
+  end
+
+  def users_used_sansheets() do
+    {query, args} = users_used_sansheets_query()
+
+    ClickhouseRepo.query_transform(query, args, fn value -> value end)
+    |> case do
+      {:ok, result} -> List.flatten(result)
```
IvanIvanoff: this is doing nothing

lib/sanbase/clickhouse/api_call_data.ex
```diff
@@ -31,6 +32,38 @@ defmodule Sanbase.Clickhouse.ApiCallData do
...
+  end
+
+  def api_calls_count_per_user() do
+    {query, args} = api_calls_count_per_user_query()
+
+    ClickhouseRepo.query_transform(query, args, fn [user_id, count] ->
+      {user_id, String.to_integer(count)}
+    end)
+    |> case do
+      {:ok, result} -> result |> Enum.into(%{})
```
IvanIvanoff: you can use `query_reduce` and directly put them into the map

lib/sanbase/clickhouse/api_call_data.ex
tspenov: it returns list of : `[[1], [2]]`, so flatten it

tspenov: flattened during transform

PR #1677, opened 2019-12-20
Sync weekly digest users to mailchimp

tspenov:

### DIFF
--- a/lib/sanbase/email/mailchimp.ex
+++ b/lib/sanbase/email/mailchimp.ex
@@ -1,0 +1,100 @@
+defmodule Sanbase.Email.Mailchimp do
+  require Sanbase.Utils.Config, as: Config
+  require Logger
+
+  @base_url "https://us14.api.mailchimp.com/3.0"
+  @weekly_digest_list_id "41325871aa"
+
+  def run() do
+    if mailchimp_api_key() do
+      {subscribed_in_mailchimp, unsubscribed_from_mailchimp} = get_subscribed_unsubscribed()
+      update_unsubscribed_locally(unsubscribed_from_mailchimp)
+      newly_subscribed = subscribed_in_sanbase() -- subscribed_in_mailchimp
+      add_newly_subscribed_in_mailchimp(newly_subscribed)
+    else
+      :ok
+    end
+  end
+
+  def get_subscribed_unsubscribed() do
+    all_members =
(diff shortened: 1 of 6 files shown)


lib/sanbase/email/mailchimp.ex
```diff
@@ -0,0 +1,100 @@
...
+      |> Enum.map(&Map.get(&1, "email_address"))
+
+    {subscribed, unsubscribed}
+  end
+
+  def update_unsubscribed_locally(unsubscribed) do
+    unsubscribed
+    |> Enum.map(&Sanbase.Repo.get_by(Sanbase.Auth.User, email: &1))
+    |> Enum.reject(&is_nil/1)
+    |> Enum.reject(&(Sanbase.Auth.UserSettings.settings_for(&1).newsletter_subscription == :off))
```
IvanIvanoff: can you alias the User and UserSettings as this is a little bit hard to read

lib/sanbase/email/mailchimp.ex
```diff
@@ -0,0 +1,100 @@
...
+    unsubscribed =
+      Enum.filter(all_members, &(Map.get(&1, "status") == "unsubscribed"))
+      |> Enum.map(&Map.get(&1, "email_address"))
+
+    {subscribed, unsubscribed}
+  end
+
+  def update_unsubscribed_locally(unsubscribed) do
+    unsubscribed
+    |> Enum.map(&Sanbase.Repo.get_by(Sanbase.Auth.User, email: &1))
```
IvanIvanoff: Let's have a `by_email/1` function right after the `by_id/1` in the user module

lib/sanbase_web/controllers/mailchimp_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+  use SanbaseWeb, :controller
+
+  require Logger
+
+  def sync_users(conn, _params) do
+    Task.Supervisor.async_nolink(Sanbase.TaskSupervisor, fn ->
+      Sanbase.Email.Mailchimp.run()
+    end)
+
+    send_response("OK", conn)
```
IvanIvanoff: If the async task fails we'll still report ok so the other side won't retry, is this ok?

tspenov: It will not wait for the sync task to finish. That is for manual testing. I can see that the list in Mailchimp changed and observe the logs which is enough for the manual testing.

PR #1699, opened 2020-01-10
Allow to delete/anonymize comments from the ex_admin board

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/application.ex
```diff
@@ -125,7 +127,9 @@ defmodule Sanbase.Application do
     ]
   end
 
-  def prepended_children(container_type) when container_type in ["web", "all"] do
```
IvanIvanoff: this was never reached with `all` as the previous `prepend_children` matched

priv/repo/migrations/20200110131000_modify_on_delete_comments.exs
```diff
@@ -0,0 +1,24 @@
...
+    execute("ALTER TABLE #{@table} DROP CONSTRAINT comments_root_parent_id_fkey")
+
+    alter table(@table) do
+      modify(:parent_id, references(:comments, on_delete: :delete_all))
+      modify(:root_parent_id, references(:comments, on_delete: :delete_all))
+    end
+  end
+
+  def down do
+    execute("ALTER TABLE #{@table} DROP CONSTRAINT comments_parent_id_fkey")
```
tspenov: `ADD` instead of `DROP` ?

IvanIvanoff: should be `drop` in both cases - destroy the previous constraint and replace it with the other one

PR #1702, opened 2020-01-13
Fix ICO funds raised

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/ico.ex
+++ b/lib/sanbase/model/ico.ex
@@ -107,7 +107,7 @@ defmodule Sanbase.Model.Ico do
   # Private functions
 
   defp funds_raised_ico_end_price_from_currencies(
-         _project,
+         project,
          %Ico{} = ico,
          target_currency,
          date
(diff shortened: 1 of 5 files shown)


lib/sanbase/model/ico.ex
```diff
@@ -116,16 +116,27 @@ defmodule Sanbase.Model.Ico do
...
+          price =
+            Sanbase.Price.Utils.fetch_last_price_before(
+              slug,
+              target_currency,
+              datetime
+            )
+
+          case price != nil and ic.amount != nil do
+            true ->
+              {price, ic.amount}
```
tspenov: this tuple is not needed

IvanIvanoff: yeah, inspect leftover

lib/sanbase/model/project/funds_raised.ex
```diff
@@ -24,6 +24,7 @@ defmodule Sanbase.Model.Project.FundsRaised do
     end
   end
 
+  @spec initial_ico(Sanbase.Model.Project.t()) :: any
```
tspenov: `float | nil` instead of any ?

IvanIvanoff: it's date, not float

lib/sanbase/prices/price_sql_query.ex
```diff
@@ -300,7 +300,7 @@ defmodule Sanbase.Price.SqlQuery do
     """
 
     # Put an artificial lower boundary otherwise the query is too slow
-    from = Timex.shift(datetime, days: -180) |> DateTime.to_unix()
+    from = Timex.shift(datetime, days: -30) |> DateTime.to_unix()
```
tspenov: extract as named module attribute

lib/sanbase/model/project/funds_raised.ex
tspenov: ups,  I was fooled by the diff - I thought it is the result of this 
```
      [] ->
      amounts -> Enum.reduce(amounts, 0, &Kernel.+/2)	        nil

      amounts ->
        Enum.reduce(amounts, 0, &Kernel.+/2)
```

lib/sanbase/prices/price_sql_query.ex
IvanIvanoff: In case this is used only here and is an implementation detail i think the comment is enough

PR #1703, opened 2020-01-13
Add include_incomplete_data bool flag to the getMetric API

IvanIvanoff:
#### Summary
Some of our metrics are computed every 6 hours but are computing the data since the beginning of the day. This means that metrics like `daily_active_addresses` can have a misleading value for today as the day is not completed. In this case the addresses count could seem lower than expected.

We're introducing a flag in the `timeseriesData` field of `getMetric` called `includeIncompleteData` which has a default value of `false`. This flag is used to determine whether or not to include data for the last day.

**NOTE**: This changes the default behaviour. The flag has a default value of `false` whereas the previous behaviour was such as if the flag was set to `true` (which was causing issues to some users that closesly follow the last value)

### Example:
If both `from` and `to` are today and the flag is `false` - return error:

Request:
```graphql
{
  getMetric(metric: "daily_active_addresses") {
    timeseriesData(
      slug: "santiment",
      from: "2020-01-13T00:00:00Z",
      to: "2020-01-13T15:00:00Z",
      interval: "1d",
    	includeIncompleteData: false) {
        datetime
        value
    }
  }
}
```
Response:
```json
{
  "data": {
    "getMetric": {
      "timeseriesData": null
    }
  },
  "errors": [
    {
      "locations": [
        {
          "column": 5,
          "line": 3
        }
      ],
      "message": "[80ea2a49-18bb-4648-9bf6-954eac170c0d] Can't fetch daily_active_addresses for project with slug: santiment, Reason: \"The time range provided [2020-01-13 00:00:00Z - 2020-01-13 15:00:00Z] is contained in today. The metric\\nrequested could have incomplete data as it's calculated since the beginning\\nof the day and not for the last 24 hours. If you still want to see this\\ndata you can pass the flag `includeIncompleteData: true` in the\\n`timeseriesData` arguments\\n\"",
      "path": [
        "getMetric",
        "timeseriesData"
      ]
    }
  ]
}
```

### Example:
If at least `from` is before today (beginning of day) the data for today will not be returned.

### Example:
If the flag is `true` the data will be returned as is. Incomplete data can still be useful:
- Daily Active Addresses are only going to increase during the day. You can use incomplete data if you want to monitor the number of addresses going over a threhsold.
- NVT using circulation will only decrease during the day. You can use incomplete data if you want to monitor the NVT going below a given threhsold. (It is going to decrease as we're dividing by 1 day circulation which is only increasing as the day progresses)

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/helpers/utils.ex
+++ b/lib/sanbase_web/graphql/helpers/utils.ex
@@ -118,6 +118,37 @@ defmodule SanbaseWeb.Graphql.Helpers.Utils do
     {:ok, from, to, interval, ma_interval}
   end
 
+  def calibrate_incomplete_data_params(true, _module, _identifier, from, to) do
+    {:ok, from, to}
+  end
+
+  def calibrate_incomplete_data_params(false, module, identifier, from, to) do
+    case module.has_incomplete_data?(identifier) do
+      true -> rewrite_params_incomplete_data(from, to)
+      false -> {:ok, from, to}
+    end
+  end
+
+  defp rewrite_params_incomplete_data(from, to) do
+    start_of_day = Timex.beginning_of_day(Timex.now())
+
+    case DateTime.compare(from, start_of_day) != :lt do
+      true ->
+        {:error,
+         """
+         The time range provided [#{from} - #{to}] is contained in today. The metric
+         requested could have incomplete data as it's calculated since the beginning
+         of the day and not for the last 24 hours. If you still want to see this
+         data you can pass the flag `includeIncompleteData: true` in the
+         `timeseriesData` arguments
+         """}
+
+      false ->
+        to = if DateTime.compare(to, start_of_day) == :gt, do: start_of_day, else: to
+        {:ok, from, to}
+    end
+  end
+
   def error_details(changeset) do
     changeset
     |> Ecto.Changeset.traverse_errors(&format_error/1)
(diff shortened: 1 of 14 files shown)


lib/sanbase_web/graphql/helpers/utils.ex
```diff
@@ -118,6 +118,37 @@ defmodule SanbaseWeb.Graphql.Helpers.Utils do
     {:ok, from, to, interval, ma_interval}
   end
 
+  def calibrate_incomplete_data_params(true, _module, _identifier, from, to) do
```
tspenov: you can add the boolean flag as last param: `def calibrate_incomplete_data_params(_module, _identifier, from, to, has_incmplete_data?: true)` and `calibrate_incomplete_data_params(module, identifier, from, to, has_incomplete_data?: false)`

IvanIvanoff: I prefer not to pattern match on keyword lists in the function header as this will accept only a single element keyword list - behaviour is not the same as maps where you can partially pattern match

lib/sanbase_web/graphql/schema/types/metric_types.ex
```diff
@@ -95,6 +95,7 @@ defmodule SanbaseWeb.Graphql.MetricTypes do
       arg(:to, non_null(:datetime))
       arg(:interval, :interval, default_value: "1d")
       arg(:aggregation, :aggregation, default_value: nil)
+      arg(:include_incomplete_data, :boolean, default_value: false)
```
tspenov: Can we document this flag here with some examples ? I think that will help both users and devs.

PR #1706, opened 2020-01-13
Merge signal activities in timeline events feed

tspenov:
* Added new event `trigger_fired`
* Only this event contains field `payload` of type `json` which is the same as in `signalHistoricalActivities`
* The `signalHistoricalActivities` also works - will be removed after @santiment/frontend migrates to only `timelineEvents`

```graphql
    {
      timelineEvents(#{args_str}) {
        cursor {
          after
          before
        }
        events {
          eventType,
          insertedAt,
          user {
            id
          },
          userList {
            name
            isPublic
          }
          post {
            id
            tags { name }
          }
          trigger {
            id
            title
            description
          }
          payload
        }
      }
    }
```


IvanIvanoff:
@tspenov can you squash some of the commits so we do not have a lot 'backup` and `merge` in the history

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -55,10 +57,17 @@ defmodule Sanbase.Timeline.TimelineEvent do
               after: DateTime.t()
             }
           }
+  @type fired_triggers_map :: %{
+          user_trigger_id: integer(),
+          user_id: integer(),
```
IvanIvanoff: you can use the more restrictive `non_neg_integer()`

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -83,15 +93,17 @@ defmodule Sanbase.Timeline.TimelineEvent do
         %{limit: limit, cursor: %{type: cursor_type, datetime: cursor_datetime}}
       ) do
     TimelineEvent
-    |> events_by_followed_users_or_sanclan(user_id, min(limit, @max_events_returned))
+    |> events_by_followed_users_or_sanclan_query(user_id, min(limit, @max_events_returned))
```
IvanIvanoff: This could be split into 2 queries merged with `or_where`

PR #1716, opened 2020-01-17
Reorder the tags to match the order they were when created

IvanIvanoff:
#### Summary
Issue is that we need to order them by the id in the join through
posts_tags table. Instead, when queried through the many_to_many
association they are ordered by the tag id. Example: if you create a
post with first tag SAN and second BTC due to the fact that the tag BTC
has lower id it will always come before SAN when fetching it

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/insight/post.ex
+++ b/lib/sanbase/insight/post.ex
@@ -94,7 +94,7 @@ defmodule Sanbase.Insight.Post do
     |> Repo.get(post_id)
     |> case do
       nil -> {:error, "There is no insight with id #{post_id}"}
-      post -> {:ok, post}
+      post -> {:ok, post |> Tag.Preloader.order_tags()}
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/sanbase/insight/post.ex
```diff
@@ -393,4 +408,10 @@ defmodule Sanbase.Insight.Post do
     extract_image_url_from_post(post)
     |> Enum.map(&Sanbase.FileStore.delete/1)
   end
+
+  defp maybe_drop_post_tags(post, %{tags: tags}) when is_list(tags) do
+    Tag.drop_tags(post)
+  end
+
+  defp maybe_drop_post_tags(_, _), do: :oka
```
tspenov:
```suggestion
  defp maybe_drop_post_tags(_, _), do: :ok
```

PR #1721, opened 2020-01-20
Trigger trending words signal only if the trigger_time is close to now

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/utils/datetime/utils.ex
```diff
@@ -1,4 +1,19 @@
 defmodule Sanbase.DateTimeUtils do
+  def time_in_range?(%Time{} = time, %Time{} = from, %Time{} = to) do
```
tspenov: please, add unit test for this one

PR #1730, opened 2020-01-22
Extend get_metric and get_anomaly api call data

IvanIvanoff:
- Rename all `getMetric` queries to `getMetric|<metric>` in the api exported data.
- Rename all `getAnomaly` queries to `getAnomaly|<anomaly>` in the api exported data.

This is done so we can run better analysis on our api calls
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/absinthe_before_send.ex
+++ b/lib/sanbase_web/graphql/absinthe_before_send.ex
@@ -28,7 +28,8 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
 
   @compile :inline_list_funcs
   @compile inline: [
-             cache_result: 3,
+             construct_query_name: 1,
+             cache_result: 2,
              queries_in_request: 1,
              extract_caller_data: 1,
              export_api_call_data: 3,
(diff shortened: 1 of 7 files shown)


tspenov:
maybe a test covering invoking `getMetric` api query and assert that metric is in the query field of api call data ?

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -120,6 +111,19 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     {user_id, san_tokens, auth_method, api_token} =
       extract_caller_data(blueprint.execution.context)
 
+    queries =
+      queries
+      |> Enum.map(fn query ->
+        <<first_char::utf8, rest::binary>> = query |> Macro.camelize()
+        <<first_char + (?a - ?A)::utf8, rest::binary>>
+      end)
```
IvanIvanoff: If we do not do this we'll end up with `get_metric` and `getMetric` in the end results as we support both in the API

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -105,7 +94,7 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     operations
     |> Enum.flat_map(fn %{selections: selections} ->
       selections
-      |> Enum.map(fn %{name: name} -> name end)
+      |> Enum.map(fn %{name: name} -> Inflex.camelize(name, :lower) end)
```
tspenov: Lets use `Absinthe.Utils.camelize`. It is a more stable dependancy and we are in absinthe pipeline module - so it fits better :?

tspenov:
@IvanIvanoff One related test is failing

lib/sanbase_web/graphql/absinthe_before_send.ex
IvanIvanoff: `Absinthe.Utils` sounds to me like some internal/private module and this is not what this dependency is supposed to do, so this function is not guaranteed to be supported/unchanged.

PR #1731, opened 2020-01-23
Signal KV-template payload

IvanIvanoff:
#### Summary
Make the signals payload to be generated by tempalte + KV map. This is the first step to start storing the template and the KV map separately in the DB and to send back the KV map to the frontend.

This PR keeps the old behavior and just introduces the internal structures needed for future changes. The work will be split into multiple PRs for easier reviews.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/operation/operation_text_kv.ex
```diff
@@ -0,0 +1,111 @@
+defmodule Sanbase.Signal.OperationText.KV do
+  @moduledoc ~s"""
+  A module providing a single function to_text/3 which transforms an operation
```
tspenov:
```suggestion
  A module providing a single function to_template_kv/3 which transforms an operation
```

PR #1739, opened 2020-01-29
Order feed

tspenov:
by default ordering is by  `DATETIME` in  `descending` order.
other opts: `orderBy: VOTES`, `orderBy: COMMENTS`, `orderBy: AUTHOR` (last means by author username)

example: `timelineEvents(orderBy: COMMENTS, limit: 3)`

### DIFF
--- a/lib/sanbase/ecto_helper.ex
+++ b/lib/sanbase/ecto_helper.ex
@@ -8,19 +8,23 @@ defmodule Sanbase.EctoHelper do
   alias Sanbase.Repo
 
   @doc """
-  Fetch a list of ids ordered by the count of items in has_many/many_to_many association.
+  Fetch a list of ids ordered by the count of items in has_many association.
+
+  opts:
+    - `order_by` - table field that is used to order by if two counts are the same.
+    Default: `:inserted_at` timestamp.
   """
   @spec fetch_ids_ordered_by_assoc_count(Ecto.Query.t(), atom, Keyword.t()) ::
           list(non_neg_integer())
   def fetch_ids_ordered_by_assoc_count(query, assoc_table, opts \\ []) do
-    order_by_second = Keyword.get(opts, :order_by, :inserted_at)
+    order_same_count_by = Keyword.get(opts, :order_by, :inserted_at)
 
     from(
       entity in query,
       left_join: assoc in assoc(entity, ^assoc_table),
       select: {entity.id, fragment("COUNT(?)", assoc.id)},
       group_by: entity.id,
-      order_by: fragment("count DESC NULLS LAST, ? DESC", field(entity, ^order_by_second))
+      order_by: fragment("count DESC NULLS LAST, ? DESC", field(entity, ^order_same_count_by))
     )
     |> Repo.all()
     |> Enum.map(fn {id, _} -> id end)
--- a/lib/sanbase/ecto_helper.ex
+++ b/lib/sanbase/ecto_helper.ex
@@ -1,0 +1,40 @@
+defmodule Sanbase.EctoHelper do
+  @moduledoc """
+  Module with reusable Ecto helper queries.
+  """
+
+  import Ecto.Query
+
+  alias Sanbase.Repo
+
+  @doc """
+  Fetch a list of ids ordered by the count of items in has_many/many_to_many association.
+  """
+  @spec fetch_ids_ordered_by_assoc_count(Ecto.Query.t(), atom, Keyword.t()) ::
+          list(non_neg_integer())
+  def fetch_ids_ordered_by_assoc_count(query, assoc_table, opts \\ []) do
+    order_by_second = Keyword.get(opts, :order_by, :inserted_at)
+
+    from(
+      entity in query,
+      left_join: assoc in assoc(entity, ^assoc_table),
+      select: {entity.id, fragment("COUNT(?)", assoc.id)},
+      group_by: entity.id,
+      order_by: fragment("count DESC NULLS LAST, ? DESC", field(entity, ^order_by_second))
+    )
+    |> Repo.all()
+    |> Enum.map(fn {id, _} -> id end)
+  end
+
+  @doc """
+  Fetch records of entity by a list of ids and keep the order of the selected records same as the list.
+  """
+  @spec by_id_in_order_query(Ecto.Query.t(), list(non_neg_integer())) :: Ecto.Query.t()
+  def by_id_in_order_query(query, ids) do
+    from(
+      entity in query,
+      where: entity.id in ^ids,
+      order_by: fragment("array_position(?, ?::int)", ^ids, entity.id)
+    )
+  end
+end
--- a/lib/sanbase/timeline/timeline_event.ex
+++ b/lib/sanbase/timeline/timeline_event.ex
@@ -85,19 +85,23 @@ defmodule Sanbase.Timeline.TimelineEvent do
     |> validate_required([:event_type, :user_id])
   end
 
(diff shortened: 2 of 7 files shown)


lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
-  defp events_order_limit_preload_query(query, limit) do
+  defp events_order_limit_preload_query(query, order_by, limit) do
+    query =
+      from(
+        event in query,
+        limit: ^limit
+      )
+      |> order_by_query(order_by)
+
+    from(event in query, preload: [:user_trigger, [post: :tags], :user_list, :user, :votes])
```
IvanIvanoff: Won't
```elixir
query
|> limit(^limit)
|> order_by_query(order_by)
|> preload([:user_trigger, [post: :tags], :user_list, :user, :votes])
```
work?

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+  end
+
+  defp order_by_query(query, :datetime) do
+    from(
+      event in query,
+      order_by: [desc: event.inserted_at]
```
IvanIvanoff: Maybe we should allow ASC/DESC to be a paramter? It would be useful here and with the authors (probably you don't want to sort from lowest number of comments to highest though)

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+    )
+  end
+
+  defp order_by_query(query, :author) do
+    from(
+      event in query,
+      left_join: u in User,
```
IvanIvanoff: inawrner join better? they should be equivalent though as there cannot be an event without an author?

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+    )
+  end
+
+  defp order_by_query(query, :votes) do
+    ids = fetch_ordered_ids(query, Sanbase.Vote)
+    by_id_in_order(query, ids)
+  end
+
+  defp order_by_query(query, :comments) do
+    ids = fetch_ordered_ids(query, Sanbase.Timeline.TimelineEventComment)
```
IvanIvanoff: I think it'll be better to use the following syntax
```elixir
from p in Post,
  left_join: c in assoc(p, :comments),
  select: {p, c}
```
and pass just the tables, not the whole module?

IvanIvanoff: This will make it not necessary to change when we have a similar change like renaming the `Comment` module you did last week.

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+  end
+
+  defp fetch_ordered_ids(query, assoc_module) do
+    from(
+      event in query,
+      left_join: assoc in ^assoc_module,
+      on: assoc.timeline_event_id == event.id,
+      select: {event.id, fragment("COUNT(?)", assoc.id)},
+      order_by: fragment("count DESC NULLS LAST, ? DESC", event.inserted_at),
+      group_by: event.id
```
IvanIvanoff: nitpick: in all other places group by comes before order by (mainly the CH queries)

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+    )
+    |> Repo.all()
+    |> Enum.map(fn {id, _} -> id end)
```
IvanIvanoff: Why `select: {event.id, fragment("COUNT(?)", assoc.id)},` if the second element is always ignored? Is there a way to write the order_by in other way :? At least a comment here would be useful

lib/sanbase/timeline/timeline_event.ex
```diff
@@ -248,12 +256,62 @@ defmodule Sanbase.Timeline.TimelineEvent do
...
+  end
+
+  defp by_id_in_order(query, ids) do
     from(
       event in query,
-      order_by: [desc: event.inserted_at],
-      limit: ^limit,
-      preload: [:user_trigger, [post: :tags], :user_list, :user, :votes]
+      where: event.id in ^ids,
+      order_by: fragment("array_position(?, ?::int)", ^ids, event.id)
```
IvanIvanoff: Please add some comments here what's happening

lib/sanbase/timeline/timeline_event.ex
tspenov: yeah, that is better. Just needed to add `many_to_many` `through` assoc for it to work

lib/sanbase/timeline/timeline_event.ex
tspenov: for `datetime` I think it shouldn't be useful because it will go back to first events. Also for votes and comments. Maybe only for authors. I don't see much the point for introducing another argument and special logic only for this one.

lib/sanbase/timeline/timeline_event.ex
tspenov: yeah, works. I was using `Repo.preload` instead the one from Ecto.Query and wondering why id doesn't work.

lib/sanbase/timeline/timeline_event.ex
tspenov: yes. even will simplify with `join: u in assoc(event, :user)`

lib/sanbase/timeline/timeline_event.ex
tspenov: I see no better way to do this.

lib/sanbase/timeline/timeline_event.ex
IvanIvanoff: yeah, keep in mind that doing `Repo.preload` will do an additional query as it works on results (structs) instead of adding some data to a query

lib/sanbase/timeline/timeline_event.ex
IvanIvanoff: hmm, with many_to_many the order_by becomes harder?

lib/sanbase/ecto_helper.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  alias Sanbase.Repo
+
+  @doc """
+  Fetch a list of ids ordered by the count of items in has_many/many_to_many association.
+  """
+  @spec fetch_ids_ordered_by_assoc_count(Ecto.Query.t(), atom, Keyword.t()) ::
+          list(non_neg_integer())
+  def fetch_ids_ordered_by_assoc_count(query, assoc_table, opts \\ []) do
+    order_by_second = Keyword.get(opts, :order_by, :inserted_at)
```
IvanIvanoff: `order_same_id_by` is more understandable?

tspenov: Actually `order_same_count_by` is better.

lib/sanbase/timeline/timeline_event.ex
tspenov: it is actually `has_many` - `has_many(:comments, through: [:event_comment_mapping, :comment])`

PR #1740, opened 2020-01-29
Add template and KV to the trigger settings

IvanIvanoff:
#### Summary
- Introduces a `template_kv` field to the trigger settings (that is not persisted in the DB)
- Add a layer in the evaluator that takes the `template_kv` and builds the proper string payload (for sending in telegram/email and for tests)
- Add a field `data` in `timeline_events` and `signals_historical_activity` tables and store the KV list there.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/signals/signal.ex
```diff
@@ -53,7 +45,16 @@ defimpl Sanbase.Signal, for: Any do
     end)
   end
 
-  def send_email(_), do: []
+  def send_email(%{
+        user: %Sanbase.Auth.User{
+          id: id,
+          user_settings: %{settings: %{has_telegram_connected: false}}
```
tspenov: why `has_telegram_connected ` for email ?

IvanIvanoff: no idea :smiley:

PR #1748, opened 2020-02-03
Fix leaking private signal anonymous feed

tspenov:


lib/sanbase/timeline/query.ex
```diff
@@ -6,6 +6,20 @@ defmodule Sanbase.Timeline.Query do
   alias Sanbase.Signal.UserTrigger
 
   # Events with public entities and current user private events
+  def events_with_public_entities_query(query) do
+    from(
+      event in query,
+      left_join: ut in UserTrigger,
```
IvanIvanoff: inner joins :?

lib/sanbase/timeline/query.ex
```diff
@@ -6,6 +6,20 @@ defmodule Sanbase.Timeline.Query do
...
+      on: event.user_trigger_id == ut.id,
+      left_join: ul in UserList,
+      on: event.user_list_id == ul.id,
+      where:
+        not is_nil(event.post_id) or
+          ul.is_public == true or
+          fragment("trigger->>'is_public' = 'true'")
```
IvanIvanoff: what's happening here? This means we can have true in the place and false in the other?

IvanIvanoff: ah.. it's different types 😨

tspenov: Means -watchlist is public and trigger is public. Can you explain further because if that doesn't answer your question

IvanIvanoff: I've misunderstood it like a single entity's publicity is defined in two places - but they are actually different entities.

lib/sanbase/timeline/query.ex
tspenov: it should be left_join - we have3 associations where only one is filled and other 2 are empty.
If it is inner join it will return empty always if I am not mistaken

PR #1753, opened 2020-02-05
Fix orderby votes bug

tspenov:


tspenov:
Close - order by VOTES works correctly - the idea is to order by DATE, then VOTES, then DATETIME.

PR #1759, opened 2020-02-06
Fix available metrics APIs

IvanIvanoff:
#### Summary
In the end I am not using the self rehydrating cache. It is still included in the PR as it could be used in the future though.

Known issue: There are exchange metrics written for some coins (binance/bch/ltc) but are only zeroes. This should be fixed by the bigdata team

image
image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/cache/rehydrating_cache.ex
+++ b/lib/sanbase/cache/rehydrating_cache.ex
@@ -1,0 +1,281 @@
+defmodule Sanbase.Cache.RehydratingCache do
+  @moduledoc ~s"""
+  A service that automatically re-runs functions and caches their values at
+  intervals smaller than the TTL so the cache never expires but is just renewed.
+
+  This service is useful when heavy queries need to be cached without any waiting
+  for recalculation when the cache expires.
+
+  Example usage: cache the function `f` under the key `:key` for up to 1 hour
+  but refresh the data every 15 minutes. Under expected conditions the value
+  will be refreshed every 15 minutes and the cache will expire only if the
+  function fails to evaluate for more than 1 hour.
+  """
+  use GenServer
+
+  @name :__rehydrating_cache__
+
+  @run_interval 15_000
+  @purge_timeout_interval 30_000
+
+  defguard are_proper_function_arguments(fun, ttl, refresh_time_delta)
+           when is_function(fun, 0) and is_integer(ttl) and ttl > 0 and
+                  is_integer(refresh_time_delta) and
+                  refresh_time_delta < ttl
+
+  @doc ~s"""
+  Start the self rehydrating cache service.
+
+  Options:
+    functions: A list of function descriptions. A function description is a
+    map with the following keys:
+      - function: Anonymous 0-arity function that computes the value
+      - key: The key the computed value will be associated with
+      - ttl: The maximal time the value will be stored for in seconds
+      - refresh_time_delta: A number of seconds strictly smaller than ttl. Every
+      refresh_time_delta seconds the cache will be recomputed and stored again.
+      The count for ttl starts from 0 again when value is recomputed.
+  """
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: @name)
+  end
+
+  def init(opts) do
+    functions =
+      Keyword.get(opts, :functions, %{})
+      |> Enum.into(
+        %{},
+        fn %{key: key, ttl: ttl, refresh_time_delta: refresh_time_delta, function: fun} = fun_map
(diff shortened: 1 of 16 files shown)


lib/sanbase/cache/rehydrating_cache.ex
```diff
@@ -0,0 +1,182 @@
...
+  defp reply_to_waiting([], _), do: :ok
+
+  defp reply_to_waiting(from_list, value) do
+    Enum.each(from_list, fn from -> GenServer.reply(from, {:ok, value}) end)
+  end
+
+  defp run_function(pid, fun_map, task_supervisor) do
+    Task.Supervisor.async_nolink(task_supervisor, fn ->
+      %{function: fun} = fun_map
+      result = fun.()
```
valo: What is going to happen if the function fails to run?

IvanIvanoff: Depending on how you define fail to run.
- If it returns {:error, error} it will be handled by the :store_result handle_info function.
- If it raises an exception it will be retried on the next iteration (defined by the @run_interval). The only bad thing that will happen is going to happen is that the `get` will wait for the whole `timeout` specified and then fail. There are a few things that can be improved here - maybe add a `yield` function, or `has_cached_value?` functions with which you can test if the value is present.

IvanIvanoff: There could be some race condition in case of a function that directly raises an exception since its first run, though, will investigate deeper.

IvanIvanoff: Reworked

lib/sanbase/cache/rehydrating_cache.ex
```diff
@@ -0,0 +1,281 @@
...
+  function fails to evaluate for more than 1 hour.
+  """
+  use GenServer
+
+  @name :__rehydrating_cache__
+
+  @run_interval 15_000
+  @purge_timeout_interval 30_000
+
+  defguard are_proper_function_argugments(fun, ttl, refresh_time_delta)
```
spiderjako: are_proper_function_argugments -> are_proper_function_arguments

lib/sanbase/cache/rehydrating_cache.ex
```diff
@@ -0,0 +1,281 @@
...
+    GenServer.start_link(__MODULE__, opts, name: @name)
+  end
+
+  def init(opts) do
+    functions =
+      Keyword.get(opts, :functions, %{})
+      |> Enum.into(
+        %{},
+        fn %{key: key, ttl: ttl, refresh_time_delta: refresh_time_delta, function: fun} = fun_map
+           when are_proper_function_argugments(fun, ttl, refresh_time_delta) ->
```
spiderjako: are_proper_function_argugments -> are_proper_function_arguments

lib/sanbase/cache/rehydrating_cache.ex
```diff
@@ -0,0 +1,281 @@
...
+    - key: The key the computed value will be associated with
+    - ttl: The maximal time the value will be stored for in seconds
+    - refresh_time_delta: A number of seconds strictly smaller than ttl. Every
+      refresh_time_delta seconds the cache will be recomputed and stored again.
+      The count for ttl starts from 0 again when value is recomputed.
+  """
+  @spec register_function((() -> any()), any(), pos_integer(), pos_integer()) ::
+          :ok | {:error, String.t()}
+  def register_function(fun, key, ttl, refresh_time_delta)
+      when are_proper_function_argugments(fun, ttl, refresh_time_delta) do
```
spiderjako: are_proper_function_argugments -> are_proper_function_arguments

lib/sanbase/clickhouse/metric/precached_functions.ex
```diff
@@ -0,0 +1,2 @@
+defmodule Sanbase.Clickhouse.Metric.PrecachedFunctions do
```
tspenov: Empty module

lib/sanbase/web.ex
```diff
@@ -24,6 +24,9 @@ defmodule Sanbase.Application.Web do
       # Start the Elasticsearch Cluster connection
       Sanbase.Elasticsearch.Cluster,
 
+      # Self-hydrating cache
+      {Sanbase.Cache.RehydratingCache, task_supervisor: Sanbase.TaskSupervisor},
```
tspenov: should we start it if we don't use it now ?

IvanIvanoff: nope

PR #1761, opened 2020-02-10
Trial on sign up emails

tspenov:


lib/sanbase/billing/subscription/sign_up_trial.ex
```diff
@@ -0,0 +1,128 @@
+defmodule Sanbase.Billing.Subscription.SignUpTrial do
```
IvanIvanoff: Please add a moduledoc here describing the purpose of this table/module

lib/sanbase/billing/subscription/sign_up_trial.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  require Logger
+
+  alias Sanbase.Auth.User
+  alias Sanbase.Billing.Subscription.PromoTrial
+  alias Sanbase.Repo
+
+  @free_trial_days 14
+  # Free trial plans are Sanbase PRO plans
+  @free_trial_plans [13]
```
IvanIvanoff: Let's have one separate module that simply holds these numbers so they can be accessed with a name, what do you think?
So here you would be able to just call `@free_trial_plans [PlanMetadata.sanbase_pro()]`

lib/sanbase/billing/subscription/sign_up_trial.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @day_email_type_map %{
+    3 => :sent_3day_email,
+    11 => :sent_11day_email,
+    14 => :sent_end_trial_email
```
IvanIvanoff: sen**d** or sen**t** ?

lib/sanbase/billing/subscription/sign_up_trial.ex
```diff
@@ -0,0 +1,128 @@
...
+    sent_welcome_email: "welcome_email"
+  }
+
+  schema "sign_up_trials" do
+    belongs_to(:user, User)
+
+    field(:sent_welcome_email, :boolean, dafault: false)
+    field(:sent_3day_email, :boolean, dafault: false)
+    field(:sent_11day_email, :boolean, dafault: false)
+    field(:sent_end_trial_email, :boolean, dafault: false)
```
IvanIvanoff: `has_sent` :?

priv/repo/migrations/20200210084320_update_api_and_sandata_product_names.exs
```diff
@@ -0,0 +1,26 @@
...
+
+  alias Sanbase.Billing.Product
+
+  def up do
+    Application.ensure_all_started(:tzdata)
+    Application.ensure_all_started(:prometheus_ecto)
+    Sanbase.Prometheus.EctoInstrumenter.setup()
+
+    %{
+      Product.product_api() => "SanAPI by Santiment",
```
IvanIvanoff: What's this? Isn't this Neuro?

lib/sanbase/billing/subscription/sign_up_trial.ex
tspenov: Added a few now. Later will create automatic functions based on `product+plan name+interval`

lib/sanbase/billing/subscription/sign_up_trial.ex
tspenov: lets stick to `sent`

priv/repo/migrations/20200210084320_update_api_and_sandata_product_names.exs
tspenov: Yes,  Neuro becomes `SanAPI`

lib/sanbase/billing/subscription/sign_up_trial.ex
IvanIvanoff: I don't understand what's the meaning here, that's why I ask - does it indicate that it should be sent (thus using `sent` - default `true`) or it has been sent (thus uisng `has_sent` - default `false`)

tspenov: hmm, I think both mean the same - it is past time so it indicates whether the email has been sent - which in both cases defaults to false and is ticked when is sent.

lib/sanbase/billing/subscription/sign_up_trial.ex
tspenov: email type could is better fit for `send` - but I use the same fields as the schema so I can toggle them dynamically without using one more map for `send` -> `sent`

PR #1765, opened 2020-02-12
Consider past_due as inactive. Add uniq constraint

tspenov:


lib/sanbase/billing/subscription.ex
```diff
@@ -392,7 +392,7 @@ defmodule Sanbase.Billing.Subscription do
   end
 
   defp active_subscriptions_query(query) do
-    from(s in query, where: s.status != "canceled")
+    from(s in query, where: s.status != "canceled" and s.status != "past_due")
```
IvanIvanoff: better be inclusive and not exclusive - `status == "active" or status = "trial"`?

PR #1766, opened 2020-02-12
Get the availble metrics from the self rehydrating cache

IvanIvanoff:
#### Summary
Also:
- Improve check for available prices
- Return atoms instead of string errors from the rehydrating cache

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/metric/metric.ex
```diff
@@ -257,17 +257,36 @@ defmodule Sanbase.Metric do
   @spec available_metrics_for_slug(any) ::
           {:ok, list(String.t())} | {:nocache, {:ok, list(String.t())}}
   def available_metrics_for_slug(slug) do
-    cache_key =
```
IvanIvanoff: Cache separately where it's called otherwise the `:nocache` result is lost

PR #1772, opened 2020-02-14
Fix available-metrics error (error tuple inside error tuple)

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/project/project_metrics_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/project/project_metrics_resolver.ex
@@ -29,8 +29,6 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectMetricsResolver do
     maybe_register_and_get(cache_key, fun, slug)
   end
 
-  defp get_error_handler(_error, slug), do: {:error, "Cannot fetch available metrics for #{slug}"}
-
   # Get the available metrics from the rehydrating cache. If the function for computing it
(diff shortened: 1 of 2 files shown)


lib/sanbase_web/graphql/resolvers/project/project_metrics_resolver.ex
```diff
@@ -40,10 +38,18 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectMetricsResolver do
...
 
         RehydratingCache.get(cache_key)
-        |> maybe_handle_graphql_error(&get_error_handler(&1, slug))
+        |> maybe_handle_graphql_error(&generate_error_message(&1, slug))
+
+      {:error, :timeout} ->
+        # Try again
+        # Recursively call itself. This is guaranteed to not continue forever
+        # as the graphql request will timeout at some point and stop the recursion
+        maybe_register_and_get(cache_key, fun, slug)
```
tspenov: So we depend on timeout in the caller ? It will be better with max retries or smth like this, but lets leave for later if it is urgent

IvanIvanoff: This executes inside a `cache_resolve` macro which acquires a lock for a limited amount of time - so in some sense the number of retries are bounded as this will just be killed at some point

PR #1776, opened 2020-02-14
Price histogram

IvanIvanoff:
#### Summary
Example of the price histogram for the ethereum moved in the interval `2020-02-16T12:50:00Z - 2020-02-17T12:50:00Z`:
```elixir
%{
   [0.0, 135.34] => 15492.91,
   [135.34, 270.68] => 2631969.94,
   [270.68, 406.02] => 226303.85,
   [406.02, 541.36] => 5540.3,
   [541.36, 676.7] => 8722.71,
   [676.7, 812.04] => 3192.47,
   [812.04, 947.38] => 1036.29,
   [947.38, 1082.72] => 240.57,
   [1082.72, 1218.06] => 268.21,
   [1218.06, 1353.4] => 185.88
 }
```

Graphql API:
image

Request:
```graphql
{
  getMetric(metric: "price_histogram") {
    histogramData(
      slug: "ethereum"
      from: "2020-02-01T00:00:00Z"
      to: "2020-02-02T00:00:00Z"
      interval: "1d",
      limit: 10) {
        values{
           ... on FloatRangeFloatValueList {
            data {
              range
              value
            }
          }
      }
    }
  }
}
```

Response:
```graphql
{
  "data": {
    "getMetric": {
      "histogramData": {
        "values": {
          "data": [
            {
              "range": [
                0,
                135.34
              ],
              "value": 57105.43
            },
            {
              "range": [
                135.34,
                270.68
              ],
              "value": 2289417.31
            },
            {
              "range": [
                270.68,
                406.02
              ],
              "value": 9280.63
            },
            {
              "range": [
                406.02,
                541.36
              ],
              "value": 1895.73
            },
....
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
```diff
@@ -0,0 +1,78 @@
...
+      FROM (
+        SELECT
+            toUnixTimestamp(intDiv(toUInt32(toDateTime(value)), ?4) * ?4) AS t,
+            -sum(measure) AS tokens_amount
+        FROM distribution_deltas_5min FINAL
+        PREWHERE
+          asset_id = ( SELECT argMax(asset_id, computed_at) FROM asset_metadata PREWHERE name = ?1 ) AND
+          dt >= toDateTime(?2) AND
+          dt < toDateTime(?3) AND
+          dt != value
```
valo: I this this condition should be left here. Without this condition the values final sums will be inflated.

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
```diff
@@ -0,0 +1,78 @@
...
+      )
+      ALL LEFT JOIN
+      (
+        SELECT
+          toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?4) * ?4) AS t,
+          avg(price_usd) AS price
+        FROM asset_prices FINAL
+        PREWHERE slug = cast(?1, 'LowCardinality(String)')
+        GROUP BY t
+      ) USING (t)
```
valo: I think we can use the precomputed 5 min interval prices from the intraday metrics. Here is the query I am using in Grafana:

```sql
SELECT
  (intDiv(toUInt32(toDateTime(currentTime)), $Interval * 60 * 60) * $Interval * 60 * 60) * 1000 as t,
  argMax(currentPrice, currentTime) as closePrice,
  sum((currentPrice - acquisitionPrice) * -measure) as profitLoss
FROM(
  SELECT
      toStartOfFiveMinute(dt) as currentTime,
      toStartOfFiveMinute(value) AS acquisitionTime,
      measure,
      acquisitionPrice
  FROM distribution_deltas_5min
  FINAL
  INNER JOIN (
    SELECT
      dt as acquisitionTime,
      value as acquisitionPrice
    FROM intraday_metrics FINAL
    PREWHERE
      asset_id = (SELECT asset_id FROM asset_metadata FINAL PREWHERE name = '$Project') AND
      metric_id = (SELECT metric_id FROM metric_metadata FINAL PREWHERE name = 'price_usd')
  ) USING acquisitionTime
  PREWHERE asset_id = (SELECT asset_id FROM asset_metadata FINAL PREWHERE name = '$Project') AND
    dt > toDateTime($from) AND dt < toDateTime($to)
)
```

valo: One of the problems in computing these deltas will be the speed. I think aggregating the values for intervals bigger than 1 month will be quite slow... We might want to reduce the size of the interval which is requested to be something like less than 1 month

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
IvanIvanoff: I'm not sure I understood?

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
IvanIvanoff: Moved to intraday 5m price.
---
1. 1 year interval for santiment gets computed in less than 2 seconds
2. 1 year interval for ethereum times out (killed after 60 seconds)
3. 30 days interval for ethereum loads in ~30 seconds

It's really about how much transfers there are - some limits could hurt smaller coins
maybe start without any limits, make the frontend use just 1 day intervals and let random big intervals just time out?

lib/sanbase/clickhouse/metric/histogram_metric.ex
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    # Generate limit number of ranges to properly empty ranges as 0
+    ranges_map =
+      Stream.unfold(min, price_ranges)
+      |> Enum.take(limit)
+      |> Enum.into(%{}, fn range -> {range, 0.0} end)
+
+    # Map every price to the proper range
+    price_to_range = fn price ->
```
tspenov: lets extract to private function

lib/sanbase/clickhouse/metric/histogram_metric.ex
```diff
@@ -0,0 +1,94 @@
...
+      bucket = floor((price - min) / bucket_size)
+      lower = min + bucket * bucket_size
+      upper = min + (1 + bucket) * bucket_size
+
+      [Float.round(lower, 2), Float.round(upper, 2)]
+    end
+
+    # Put every amount moved at a given price in the proper bucket
+    data =
+      Enum.reduce(data, ranges_map, fn %{price: price, value: value}, acc ->
```
tspenov: Lets extract this one also, because this function is getting quite big

lib/sanbase/clickhouse/metric/histogram_metric.ex
IvanIvanoff: No need imo, it'll be used only once only here

IvanIvanoff: If it's a separate function I think it should be moved to another module, otherwise the `HistogramMetric` module would have more transform functions than functions for fetching histogram data

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
valo: I think the condition should be removed, as with it the aggregation will be wrong.

IvanIvanoff: The current version of the histogram means how many tokens were last moved at price X - this implies that the amounts are positive numbers.
If `dt != value` is removed we'll have ranges with negative amounts - I think this changes the semantics to mean something different.

lib/sanbase/clickhouse/metric/histogram_metric.ex
```diff
@@ -0,0 +1,94 @@
...
+  def histogram_data("price_histogram" = metric, slug, from, to, interval, limit) do
+    {query, args} = histogram_data_query(metric, slug, from, to, interval, limit)
+
+    ClickhouseRepo.query_transform(query, args, fn [price, amount] ->
+      %{
+        price: Sanbase.Math.to_float(price),
+        value: Sanbase.Math.to_float(amount)
+      }
+    end)
+    |> maybe_transform_into_buckets(limit)
```
tspenov: can we have a test for this transformation ?

lib/sanbase/clickhouse/metric/histogram_metric.ex
tspenov: > HistogramMetric module would have more transform functions than functions for fetching histogram data
That is not a bad thing IMO. I find extracting to functions easier to read, even though transformations are commented and not reusable at this point. But I don't a strong opinion about this so lets leave as it is.

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
valo: After discussing the issue, I am convinced that we can keep the condition. No changes needed.

PR #1780, opened 2020-02-18
Migrate to new api plans

tspenov:


config/scheduler_config.exs
```diff
@@ -56,6 +56,10 @@ config :sanbase, Sanbase.Scrapers.Scheduler,
   global: true,
   timeout: 30_000,
   jobs: [
+    sync_plans_in_stripe: [
+      schedule: "@reboot",
```
IvanIvanoff: :?

tspenov: instead of writing the same migration -> At deploy this will start a task to create in Stripe plans that locally have no `stripe_id` (like the new ones created here).

PR #1786, opened 2020-02-18
Move rehydrating cache to ETS backed cache

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/cache/cache.ex
+++ b/lib/sanbase/cache/cache.ex
@@ -1,7 +1,8 @@
 defmodule Sanbase.Cache do
   @behaviour Sanbase.Cache.Behaviour
   @cache_name :sanbase_cache
-  @max_cache_ttl 300
+  @max_cache_ttl 84600
+
   def name, do: @cache_name
 
   @impl Sanbase.Cache.Behaviour
@@ -84,7 +85,7 @@ defmodule Sanbase.Cache do
     end
   end
 
-  defp cache_item(cache, {_, ttl} = key, value) when is_integer(ttl) and ttl <= @max_cache_ttl do
+  defp cache_item(cache, {key, ttl}, value) when is_integer(ttl) and ttl <= @max_cache_ttl do
     ConCache.put(cache, key, %ConCache.Item{value: value, ttl: :timer.seconds(ttl)})
   end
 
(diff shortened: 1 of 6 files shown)


lib/sanbase/cache/cache.ex
```diff
@@ -84,7 +85,7 @@ defmodule Sanbase.Cache do
     end
   end
 
-  defp cache_item(cache, {_, ttl} = key, value) when is_integer(ttl) and ttl <= @max_cache_ttl do
+  defp cache_item(cache, {key, ttl}, value) when is_integer(ttl) and ttl <= @max_cache_ttl do
```
IvanIvanoff: This is needed so you can later ask for `key` without worrying about what ttl it was stored with.

lib/sanbase/cache/rehydrating_cache/supervisor.ex
```diff
@@ -0,0 +1,44 @@
...
+    task_supervisor_name = Sanbase.Cache.RehydratingCache.TaskSupervisor
+
+    children = [
+      # ETS backed cache
+      Supervisor.child_spec(
+        {ConCache,
+         [
+           name: Store.name(@name),
+           ttl_check_interval: :timer.seconds(15),
+           global_ttl: :infinity,
```
IvanIvanoff: This is ok because all records are forced to supply an integer TTL

lib/sanbase/cache/rehydrating_cache/supervisor.ex
```diff
@@ -0,0 +1,44 @@
...
+      start: {__MODULE__, :start_link, [opts]}
+    }
+  end
+
+  def start_link(opts) do
+    Supervisor.start_link(__MODULE__, opts, name: :"__rehydrating_cache_supervisor_#{@name}__")
+  end
+
+  def init(_opts) do
+    task_supervisor_name = Sanbase.Cache.RehydratingCache.TaskSupervisor
```
tspenov: we don't have that module - we just use the atom here ?

IvanIvanoff: it's just a name - simialr to what we used before that `Sanbase.TaskSupervisor`

PR #1821, opened 2020-03-06
Optimizing logo size

spiderjako:
The idea is to optimise the filesize of the logos, so they aren't that big and at the same time, they don't lose a significant portion of their quality.

https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-6948

Did a little search on morgify and with the help of Ivan, found that I can change the type of the image, so it can use a different colour pattern:
http://www.imagemagick.org/script/command-line-options.php#type

### DIFF
--- a/lib/sanbase/external_services/coinmarketcap/logo_fetcher.ex
+++ b/lib/sanbase/external_services/coinmarketcap/logo_fetcher.ex
@@ -96,6 +96,7 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.LogoFetcher do
 
     Mogrify.open(source_filepath)
     |> Mogrify.resize("#{@size}x#{@size}")
+    |> Mogrify.custom("type", "PaletteAlpha")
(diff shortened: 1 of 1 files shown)


IvanIvanoff:
Can you upload 2-3 logo examples before and after to see the quality (also add the before/after sizes).
Is it clear what `PaletteAlpha` type does? Make sure transparency (alpha channel) is not lost

spiderjako:
Before:
Bitcoin - 3.1kb
logo64_bitcoin

Ethereum - 2.4kb
logo64_ethereum

After:
Bitcoin - 3.0kb
logo64_bitcoin

Ethereum - 2.3kb
logo64_ethereum

IvanIvanoff:
Have you exprimented with other types?

spiderjako:
I actually expretimented with all types and found, that this is the only type, that lowers the filesize

PR #1834, opened 2020-03-12
Fix social volume aligning of datetimes

IvanIvanoff:
#### Summary
Fix issues where the volumes are not summed properly if some of the sources return less or more data

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/social_data/social_volume.ex
+++ b/lib/sanbase/social_data/social_volume.ex
@@ -23,7 +23,7 @@ defmodule Sanbase.SocialData.SocialVolume do
         end,
         max_concurrency: 4
       )
-      |> sum_by_datetime()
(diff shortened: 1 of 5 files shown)


tspenov:
@IvanIvanoff could you add also a test

IvanIvanoff:
@tspenov the tests mock the response from tech_indicators, so the logic passes through this code

tspenov:
@IvanIvanoff ok, but is there anything testing exactly the issue in the description ?

lib/sanbase/social_data/social_volume.ex
```diff
@@ -206,6 +188,21 @@ defmodule Sanbase.SocialData.SocialVolume do
     end)
   end
 
+  defp sum_by_datetime(data) do
```
tspenov: Can you extract this duplicated function

PR #1841, opened 2020-03-18
Restrict api metrics by plan

tspenov:
### These are with free timeframe restriction when customer has `basic` API plan.
* Exchange flow
* % of tokens on exchange
* Daily deposits

which correspond to these in API
```
    {:metric, "active_deposits"},
    {:query, :daily_active_deposits},
    {:metric, "exchange_token_supply"},
    {:metric, "exchange_inflow"},
    {:metric, "exchange_outflow"},
    {:metric, "exchange_balance"},
    {:query, :exchange_volume}
```

### These time-bound go into PRO and up (that means accessing with lower plan will result in `unauthorized` error) :
* circulation_*
* mvrv_usd_*
* realized_value_usd_*
* mean_realized_price_usd_*
but these stay in free: [circulation, mvrv_usd, ...]

### Also in PRO and up:
* active_withdrawals, withdrawal_transactions
* mvrv_long_short_diff_usd

### In custom API plan are:
* Holder distribution ?
* Exchange-specific data (in own extension plan)
* MakerDAO/DeFi data


lib/sanbase/billing/plan/api_access_checker.ex
```diff
@@ -30,6 +30,24 @@ defmodule Sanbase.Billing.Plan.ApiAccessChecker do
...
+  # They are served with the free timeframe restriction.
+  @restricted_in_basic_plan_queries [
+    {:metric, "active_deposits"},
+    {:query, :daily_active_deposits},
+    {:metric, "exchange_token_supply"},
+    {:metric, "exchange_inflow"},
+    {:metric, "exchange_outflow"},
+    {:metric, "exchange_balance"},
+    {:query, :exchange_volume}
+  ]
```
IvanIvanoff: Why are these treated so differently? These metrics are already marked as `restricted`?

lib/sanbase/billing/plan/api_access_checker.ex
```diff
@@ -30,6 +30,24 @@ defmodule Sanbase.Billing.Plan.ApiAccessChecker do
...
+
+  def historical_data_in_days(plan, metric)
+      when plan in [:essential, :basic] and
```
IvanIvanoff: I think we can no longer have `:essential` as name, so this can be just `def historical_data_in_days(:basic, metric) when ..`

tspenov: yes, we can't I have put just in case if it sneaks from somewhere, but will remove.

lib/sanbase/clickhouse/metric/helper.ex
```diff
@@ -9,9 +9,9 @@ defmodule Sanbase.Clickhouse.Metric.Helper do
     |> Enum.filter(fn metric -> String.contains?(metric, str) end)
   end
 
-  def mvrv_metrics(), do: metric_with_name_containing("mvrv") |> wrap()
-  def realized_value_metrics(), do: metric_with_name_containing("realized") |> wrap()
-  def token_age_consumed_metrics(), do: metric_with_name_containing("age_consumed") |> wrap()
+  def mvrv_metrics(), do: wrap(["mvrv_usd"])
+  def realized_value_metrics(), do: wrap(["realized_value_usd"])
+  def token_age_consumed_metrics(), do: wrap(["age_destroyed"])
```
IvanIvanoff: What happens to all timebound metrics in this case?

lib/sanbase/billing/plan/api_access_checker.ex
tspenov: These are some exceptions:
* they should be available in `free` plan
* they should be available in `basic` plan but with `free timeframe restrictions`

lib/sanbase/clickhouse/metric/helper.ex
tspenov: all other timebound metrics go with `min_plan: PRO`. So with lower plan customer gets `unauthrized`

IvanIvanoff: Can you add a comment here with that info here as it's not clear why we're excluding some metrics

IvanIvanoff: Also, in this case can't we just call `wrap(["mvrv_usd"])` where it is used?

lib/sanbase/billing/plan/api_access_checker.ex
IvanIvanoff: If you grep for `essential` in the codebase you get only:
` %{name: "ESSENTIAL"} -> :basic`

lib/sanbase/billing/plan/api_access_checker.ex
```diff
@@ -30,6 +30,24 @@ defmodule Sanbase.Billing.Plan.ApiAccessChecker do
 
   @enterprise_plan_stats @premium_plan_stats
 
+  # Some queries/metrics are further restricted for basic/essential plan.
```
IvanIvanoff: This should probably be moved to `access_checker.ex` as the behaviour for sanbase is not changed now

lib/sanbase/billing/plan/api_access_checker.ex
tspenov: ups, I meant you are right that we can't but wrongly edited

lib/sanbase/billing/plan/api_access_checker.ex
IvanIvanoff: even if we don't have :basic for sanbase it should be there, too

tspenov: we want it only for API

lib/sanbase/clickhouse/metric/helper.ex
tspenov: Removed helper module because it is no longer needed.

lib/sanbase/billing/plan/api_access_checker.ex
tspenov: Also it uses the time restrictions for the API so I think it belongs there

IvanIvanoff:
@tspenov if the min_plan will apply only to one product (I still don't agree with that) let's rename the property in the json file to `min_api_plan` and `min_sanbase_plan` and keep them different?

tspenov:
@IvanIvanoff no, will rework to apply to sanbase too

lib/sanbase_web/graphql/schema/queries/historical_balance_queries.ex
```diff
@@ -45,7 +45,7 @@ defmodule SanbaseWeb.Graphql.Schema.HistoricalBalanceQueries do
     Currently only ETH is supported.
     """
     field :miners_balance, list_of(:miners_balance) do
-      meta(access: :restricted)
+      meta(access: :restricted, min_plan: [SANAPI: :pro, SANBASE: :free])
```
IvanIvanoff, line 48: can we use lowercase atoms like `:sanapi` and `:sanbase`. Starting with an upper letter will make this be treated as a module (modules are just atoms starting with an upper letter) and could be colored as such?

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -104,13 +104,14 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
-    case Map.get(@min_plan_map, query_or_metric, :free) do
+  @spec plan_has_access?(plan, product, query_or_metric) :: boolean()
+        when plan: atom(), product: binary()
+  def plan_has_access?(plan, product, query_or_metric) do
+    case @min_plan_map[query_or_metric][product] || :free do
       :free -> true
-      :essential -> plan != :free
-      :pro -> plan not in [:free, :essential]
-      :premium -> plan not in [:free, :essential, :pro]
+      :basic -> plan != :free
```
IvanIvanoff, line 112: Didn't we decide to rename all `:basic` to the better sounding `:essential` plan in the past?

lib/sanbase/billing/graphql_schema.ex
```diff
@@ -39,17 +39,29 @@ defmodule Sanbase.Billing.GraphqlSchema do
...
     query_min_plan_map =
       get_query_meta_field_list(:min_plan)
       |> Enum.into(%{}, fn
-        {query, nil} -> {{:query, query}, :free}
-        {query, plan} when is_atom(plan) -> {{:query, query}, plan}
+        {query, kw_list} when is_list(kw_list) ->
+          {{:query, query},
+           %{
+             "SANAPI" => Keyword.get(kw_list, :SANAPI, :free),
+             "SANBASE" => Keyword.get(kw_list, :SANBASE, :free)
```
IvanIvanoff, line 51: let's have module attributes `@sanbase_product_id` and `@sanapi_product_id` and use them here as keys. This has a few benefits:
1. using integers over strings for faster access
2. in the access control checker you don't you can directly call this function with the product. RIght now you get the product_id and from it you get the product code and only then you can check the plan with the string value

IvanIvanoff: Also - can we use lower case `:sanapi`  and `:sanbase` atoms (same reason as in the other comments - uppercase atoms are modules)

lib/sanbase/billing/product.ex
```diff
@@ -45,7 +52,13 @@ defmodule Sanbase.Billing.Product do
   end
 
   def code_by_id(id) do
-    Repo.get(__MODULE__, id).code
+    @code_product_id_map
+    |> Enum.into(%{}, fn {k, v} -> {v, k} end)
+    |> Map.get(id)
```
IvanIvanoff, line 57: let's have `@product_code_by_id_map @code_product_id_map |> Enum.into(%{}, fn {k, v} -> {v, k} end)`
and here just use `@product_code_by_id_map |> Map.get(id)`

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -64,10 +62,15 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
...
+       ) do
+    plan = context[:auth][:plan] || :free
+    subscription = context[:auth][:subscription] || @free_subscription
+    product_id = subscription.plan.product_id || context.product_id
+    product = Product.code_by_id(product_id)
 
-    case Plan.AccessChecker.plan_has_access?(plan, query_or_metric) do
+    case Plan.AccessChecker.plan_has_access?(plan, product, query_or_metric) do
       true -> resolution
       false -> Resolution.put_result(resolution, {:error, :unauthorized})
```
IvanIvanoff, line 75: Let's create a much more descriptive message here?

lib/sanbase/billing/graphql_schema.ex
tspenov: Ok, will change to lowercase atoms. Otherwise I'd like to be able to call by "SANAPI" and not setting always @sanbase_product_id and @sanapi_product_id or calling Product.sanapi_product_id().

IvanIvanoff: This one is called in exactly one place (without counting the tests) - the access control middleware.
Plus, all other access check functions accept `product_id` as a parameter, not names

tspenov: I propose to change these places one by one and use these codes everywhere. It will be much easier.
In metrics json is defined `min_plan: {"SANAPI": "pro", "SANBASE": "free"}` - I want to use that directly without converting the key to id.

lib/sanbase/billing/plan/access_checker.ex
tspenov: sandata has basic, sanbase had too (plan is not deleted yet, though users can't subscribe)

PR #1844, opened 2020-03-18
Update rehydrating cache

IvanIvanoff:
#### Summary
- Handle failed tasks - change `:in_progress` to failed and record it for better tracking;
- Handle more correctly running of functions according to their progress - before this PR it could happen that the progress is `:in_progress` without any task running it, but no new task was started
- Improved timeouts and killing of stuck tasks
- Add some datetimes here and there for better observability - inspecting the state is easier with datetimes instead of only unix timestamps

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/cache/rehydrating_cache/rehydrating_cache.ex
+++ b/lib/sanbase/cache/rehydrating_cache/rehydrating_cache.ex
@@ -139,7 +139,7 @@ defmodule Sanbase.Cache.RehydratingCache do
         {:noreply, new_state}
 
       %{progress: :failed} ->
-        # If progress if :failed it will get started on the next run
+        # If progress :failed it will get started on the next run
         new_state = do_fill_waiting_list(state, key, from, timeout)
         {:noreply, new_state}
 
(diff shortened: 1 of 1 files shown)


lib/sanbase/cache/rehydrating_cache/rehydrating_cache.ex
```diff
@@ -126,15 +130,23 @@ defmodule Sanbase.Cache.RehydratingCache do
...
+      %{progress: {:in_progress, _task_pid, _started_time_tuple}} ->
         # If the value is still computing the response will be sent
         # once the value is computed. This will be reached only on the first
         # computation. For subsequent calls with :in_progress progress, the
         # stored value will be available and the previous case will be matched
         new_state = do_fill_waiting_list(state, key, from, timeout)
         {:noreply, new_state}
 
+      %{progress: :failed} ->
+        # If progress if :failed it will get started on the next run
```
valo:
```suggestion
        # If progress :failed it will get started on the next run
```

PR #1855, opened 2020-03-20
Add paywalled insights

tspenov:
* `isPaywallRequired` boolean field
*  if insight is behind paywall and user is not `PRO` and not author of the insight - `text` field is truncated to first 70 words of the insight.


Partysun:
70 is too small. Let's 140 at least ;) @tspenov

lib/sanbase/billing/plan.ex
```diff
@@ -59,6 +59,8 @@ defmodule Sanbase.Billing.Plan do
     end
   end
 
+  def plan_atom_name(_), do: :free
```
IvanIvanoff: Where is this needed? The context plug injects free subscription/plan in case of no authentication

lib/sanbase/billing/plan/sanbase_access_checker.ex
```diff
@@ -67,6 +68,15 @@ defmodule Sanbase.Billing.Plan.SanbaseAccessChecker do
 
   def signals_limits_upgrade_message(), do: @signals_limits_upgrade_message
 
+  def access_paywalled_insights?(nil), do: false
+
+  def access_paywalled_insights?(subscription) do
```
IvanIvanoff: `can_access_paywalled_insights?` :?

lib/sanbase/billing/subscription.ex
```diff
@@ -499,6 +498,15 @@ defmodule Sanbase.Billing.Subscription do
     StripeApi.delete_subscription(stripe_id)
   end
 
+  defp fetch_current_subscription(user_id, product_id) do
+    user_id
+    |> user_subscriptions_query()
+    |> active_subscriptions_query()
+    |> last_subscription_for_product_query(product_id)
+    |> Repo.one()
+    |> Repo.preload(plan: [:product])
```
IvanIvanoff: Let's move the preload as part of the pipeline before `Repo.one()` to reduce the queries to the database as I imagine this function will be called on every api?

lib/sanbase/insight/post.ex
```diff
@@ -252,6 +265,7 @@ defmodule Sanbase.Insight.Post do
   def public_insights_after(datetime, opts \\ []) do
     published_and_approved_insights()
     |> by_is_pulse(Keyword.get(opts, :is_pulse, false))
+    |> by_is_paywall_required(Keyword.get(opts, :is_paywall_required, false))
```
IvanIvanoff: Should the default be `nil` here? As a user you would like to fetch a list of all insights by default, not to only the free ones?

lib/sanbase/insight/post_paywall.ex
```diff
@@ -0,0 +1,55 @@
...
+  alias Sanbase.Insight.Post
+  alias Sanbase.Billing.{Subscription, Product}
+  alias Sanbase.Auth.User
+  alias Sanbase.Billing.Plan.SanbaseAccessChecker
+
+  # Show only first @words_count_shown_as_preview word of content
+  @words_count_shown_as_preview 70
+  @product_sanbase Product.product_sanbase()
+
+  def maybe_filter_paywalled(insights, nil), do: maybe_filter(insights, nil)
```
IvanIvanoff: Can you add a typespec here? By the name it seems like it's a list of insights, but by the code implementation it looks like it can be a single insight, too.
I think a simpler solution would be (we do it like this in other places, too) to accept `insight_or_insights`, call `List.wrap()` and work with lists everywhere

IvanIvanoff: You would also be able to reduce the function chains by removing the `maybe_filter` and just call `Enum.map(insights, &do_filter(&1, querying_user_id))` from the else block

lib/sanbase/insight/post_paywall.ex
```diff
@@ -0,0 +1,55 @@
...
+  defp do_filter(%Post{user_id: user_id} = insight, querying_user_id)
+       when not is_nil(querying_user_id) and user_id == querying_user_id,
+       do: insight
+
+  defp do_filter(%Post{short_desc: short_desc} = insight, _) when is_binary(short_desc) do
+    Map.put(insight, :text, short_desc)
+  end
+
+  defp do_filter(insight, _) do
+    Map.put(insight, :text, text_preview(insight))
```
IvanIvanoff: Should we also filter the comments of paywalled insights? Usually, you cannot access the comments of paywalled content

lib/sanbase/insight/post_paywall.ex
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  defp text_preview(%Post{text: text}) do
+    String.split(text, " ")
+    |> Enum.take(@words_count_shown_as_preview)
+    |> Enum.join(" ")
```
IvanIvanoff: btw, this strips multiple whitespaces and newlines (encoded as \n, not as <br />) - the second part could  be crucial:
```elixir
iex(18)> text = "my name is ivan \n hello \n \n bye"          
"my name is ivan \n hello \n \n bye"
iex(19)> String.split(text) |> Enum.take(70) |> Enum.join(" ")
"my name is ivan hello bye"
```

lib/sanbase_web/graphql/middlewares/past_paywall_filter.ex
```diff
@@ -0,0 +1,44 @@
+defmodule SanbaseWeb.Graphql.Middlewares.PostPaywallFilter do
+  @behaviour Absinthe.Middleware
+  alias Absinthe.Resolution
+
+  alias Sanbase.Insight.{Post, PostPaywall}
+  alias Sanbase.Timeline.TimelineEvent
+
+  def call(%Resolution{errors: errors} = resolution, _opts) when errors != [], do: resolution
```
IvanIvanoff:
```suggestion
  def call(%Resolution{errors: [_|_]} = resolution, _opts), do: resolution
```

lib/sanbase_web/graphql/middlewares/past_paywall_filter.ex
```diff
@@ -0,0 +1,44 @@
...
+    PostPaywall.maybe_filter_paywalled(insights, current_user)
+  end
+
+  defp filter_value(%TimelineEvent{} = event, current_user) do
+    update_event_insight(event, current_user)
+  end
+
+  defp filter_value(%{events: [%TimelineEvent{} | _rest] = events} = timeline, current_user) do
+    %{timeline | events: Enum.map(events, &update_event_insight(&1, current_user))}
+  end
```
IvanIvanoff: As commented above - maybe also filter comments?

lib/sanbase/billing/plan.ex
tspenov: I know but here it comes with current subscription for Sanbase for current user, so subscription is select once more. Maybe I can pass the current subscription from context along with current user but for now I think its ok.

lib/sanbase/insight/post_paywall.ex
tspenov: yeah, reworked a bit.

lib/sanbase/insight/post_paywall.ex
tspenov: `List.wrap()` won't work here because I have to also return `insight_or_insights` based on input

IvanIvanoff: I still think it would be much simpler to handle this on the border:
```elixir
if is_list(data) do
   data |> work
else
   data |> List.wrap |> word |> hd
end
```

lib/sanbase/insight/post.ex
tspenov: yeah, that was the idea. I'll update these to default to `nil`. Though they already should be filled with `nil` from resolver `is_paywall_required: Map.get(args, :is_paywall_required)`. But better to be more explicit here.

lib/sanbase/insight/post_paywall.ex
tspenov: removed comments

lib/sanbase/insight/post_paywall.ex
tspenov: I think function with 2 heads does the same but the result will be cleaner because there will be no nested conditions.

lib/sanbase_web/graphql/middlewares/past_paywall_filter.ex
tspenov: you mean comments for the timeline event or comments for the insight ? Comments for the insight are filtered now.

PR #1862, opened 2020-03-24
Top holders improvements

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/top_holders/metric_adapter.ex
+++ b/lib/sanbase/clickhouse/top_holders/metric_adapter.ex
@@ -47,7 +47,8 @@ defmodule Sanbase.Clickhouse.TopHolders.MetricAdapter do
     aggregation = aggregation || :last
     count = Map.get(selector, :holders_count, @default_holders_count)
 
-    with {:ok, contract, decimals, infr} <- Project.contract_info_infrastructure_by_slug(slug) do
+    with {:ok, contract, decimals, infr} <- Project.contract_info_infrastructure_by_slug(slug),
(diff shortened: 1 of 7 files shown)


lib/sanbase/clickhouse/top_holders/metric_adapter.ex
```diff
@@ -128,8 +125,9 @@ defmodule Sanbase.Clickhouse.TopHolders.MetricAdapter do
   end
 
   @impl Sanbase.Metric.Behaviour
-  def first_datetime(_metric, %{slug: slug}) do
-    with {:ok, contract, _decimals, infr} <- Project.contract_info_infrastructure_by_slug(slug) do
+  def first_datetime(metric, %{slug: slug}) do
+    with {:ok, contract, _decimals, infr} <- Project.contract_info_infrastructure_by_slug(slug),
+         true <- chain_supported?(infr, slug, metric) do
```
tspenov: You can extract in helper function which will basically do the same and return `{:ok, contract, _decimals, infr}` | `{:error, :error_message}` . This will save one line when calling.

tspenov: Another way is to extract like 
```elixir
case Project.contract_info_infrastructure_by_slug(slug) do
  {:ok, contract, _decimals, infr} when infr in @supported_chains_infrastrucutres
    -> body
  {:ok, _contract, _decimals, _infr}
    -> "The metric #{metric} is not supported for #{slug}"
  error
   -> error
end
```

PR #1867, opened 2020-03-25
Add exchange/non-exchange top holders metrics

IvanIvanoff:
#### Summary
Add `amount_in_(non_)exchange_top_holders` metrics. It fetches the amount held by the top N (non-)exchange holders

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/top_holders/top_holders_sql_query.ex
+++ b/lib/sanbase/clickhouse/top_holders/top_holders_sql_query.ex
@@ -5,9 +5,10 @@ defmodule Sanbase.Clickhouse.TopHolders.SqlQuery do
   import Sanbase.Metric.SqlQuery.Helper, only: [aggregation: 3]
 
   def timeseries_data_query(
-        table,
         "amount_in_top_holders",
+        table,
         contract,
+        _blockchain,
         count,
         from,
         to,
@@ -22,14 +23,17 @@ defmodule Sanbase.Clickhouse.TopHolders.SqlQuery do
     FROM (
       SELECT
         toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS dt,
-       #{aggregation(aggregation, "value", "dt")} / #{decimals} AS value
+        #{aggregation(aggregation, "value", "dt")} / #{decimals} AS value
       FROM #{table} FINAL
       PREWHERE
         contract = ?2 AND
         rank <= ?3 AND
         dt >= toDateTime(?4) AND
-        dt < toDateTime(?5)
+        dt < toDateTime(?5) AND
+        rank IS NOT NULL
       GROUP BY dt, address
+      ORDER BY dt, value desc
+      LIMIT ?3 BY dt
     )
     GROUP BY dt
     ORDER BY dt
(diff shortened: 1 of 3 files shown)


lib/sanbase/clickhouse/top_holders/top_holders_sql_query.ex
```diff
@@ -46,6 +50,113 @@ defmodule Sanbase.Clickhouse.TopHolders.SqlQuery do
...
+          FROM blockchain_address_labels
+          PREWHERE blockchain = ?6 AND label IN ('centralized_exchange', 'decentralized_exchange')
+        ) AND
+        contract = ?2 AND
+        dt >= toDateTime(?4) AND
+        dt < toDateTime(?5) AND
+        rank IS NOT NULL
+      GROUP BY dt, address
+      ORDER BY value DESC
+      LIMIT #{count} BY address
```
valo: Isn't `count` the number of top N holders to sum for each `dt`? If that's the case, isn't this limit going to limit the total amount of `dt` records being returned?

IvanIvanoff: `LIMIT N BY expr` applies the limit to each distinct expr (dt) in this case. If we have it here ordered by value in descending order that should take the top N holders for every dt

IvanIvanoff: I've just tested by rewriting the `amount_in_top_holders` to similar query (without the exchange addresses) and it resulted in absolutely the same result, so all group/order/limit here should be correct

valo: Shouldn't this be `LIMIT .. BY dt`? For each `dt` to take the top rows ordered by `value`

IvanIvanoff: oops, yeah, I've reworked and tested with the other query where it's `by dt` (you can also see the string interpolation instead of vars here)

lib/sanbase/clickhouse/top_holders/top_holders_sql_query.ex
```diff
@@ -46,6 +50,113 @@ defmodule Sanbase.Clickhouse.TopHolders.SqlQuery do
...
+      )
+      GROUP BY dt, address
+      ORDER BY dt, value DESC
+      LIMIT ?3 BY dt
+    )
+    GLOBAL ANY INNER JOIN (
+      SELECT address
+      FROM blockchain_address_labels
+      PREWHERE blockchain = ?6 AND label IN ('centralized_exchange', 'decentralized_exchange')
+    ) USING address
```
valo: Is there a reason to use JOIN here and `IN` in the other query? Can't this query be shared for both cases and just a different condition to be plugged in?

IvanIvanoff: The `JOIN` turned out to be faster in my tests, so it's used here. The non-exchange cannot be written nicely with join - it would look like `select address, 1 as isExchange` and filter on `isExchange = 0` to depend on the behaviour where rows with no corresponding value in the right table are filled with default values

IvanIvanoff: The `JOIN` turned out to be faster in my tests, so it's used here. The non-exchange cannot be written nicely with join - it would look like `select address, 1 as isExchange` and filter on `isExchange = 0` to depend on the behaviour where rows with no corresponding value in the right table are filled with default values

valo: ok

PR #1881, opened 2020-03-30
Disallow inspecting the apikey list with apikey authentication

IvanIvanoff:
Sometimes apikeys are shared from a generic user
(free_apikey@santiment.net). This change prevents one apikey from
inspecting the other apikeys

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/user/apikey_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/user/apikey_resolver.ex
@@ -48,6 +48,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ApikeyResolver do
   @doc ~s"""
   Returns a list of all apikeys for the currently JWT authenticated user.
   """
+  def apikeys_list(%User{} = user, _args, %{context: %{auth: %{auth_method: :apikey}}}) do
+    {:error, "Only JWT authenticated users can access their apikeys."}
+  end
+
   def apikeys_list(%User{} = user, _args, _resolution) do
     Apikey.apikeys_list(user)
   end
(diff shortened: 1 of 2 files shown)


lib/sanbase_web/graphql/resolvers/user/apikey_resolver.ex
```diff
@@ -48,6 +48,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ApikeyResolver do
   @doc ~s"""
   Returns a list of all apikeys for the currently JWT authenticated user.
   """
+  def apikeys_list(%User{} = user, _args, %{context: %{auth: %{auth_method: :apikey}}}) do
```
tspenov: the description is says `JWT` but should be apikey

IvanIvanoff: It's correct - in case of `apikey` auth we want to error and say that only JWT auth can access this

tspenov: ok, so better allow only JWT and show this error in catch all head ?
```elixir
def apikeys_list(%User{} = user, _args, %{context: %{auth: %{auth_method: :apikey}}}) do
  Apikey.apikeys_list(user)
end

def apikeys_list(%User{} = user, _args, _resolution) do
  {:error, "Only JWT authenticated users can access their apikeys."}
end
```

IvanIvanoff: We need to also to not throw on basic auth as it allows you to access everything without any restrictions

tspenov: oh, yeah - I forgot about basic auth. In that case all good.

PR #1883, opened 2020-03-30
Add custom queries to social volume

spiderjako:


lib/sanbase/social_data/social_volume.ex
```diff
@@ -123,17 +138,20 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-          }
+    %{"data" => map} = result
+
+    map =
+      Map.to_list(map)
+      |> Enum.map(fn {timestamp, value} ->
+        {:ok, datetime, _status} = DateTime.from_iso8601(timestamp)
+
+        %{
+          datetime: datetime,
```
IvanIvanoff, line 149: You can just use here `datetime: SanbaseDateTimeUtils.from_iso8601!(iso8601_dt)`
Also it's not timestamp but a datetime string

lib/sanbase/social_data/social_volume.ex
```diff
@@ -123,17 +138,20 @@ defmodule Sanbase.SocialData.SocialVolume do
...
   defp social_volume_result(result) do
-    result =
-      result
-      |> Enum.map(fn
-        %{"timestamp" => timestamp, "mentions_count" => mentions_count} ->
-          %{
-            datetime: DateTime.from_unix!(timestamp),
-            mentions_count: mentions_count
-          }
+    %{"data" => map} = result
```
IvanIvanoff, line 141: You can just use `defp social_volume_result(%{"data" => map})`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -123,17 +138,20 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+
+    map =
+      Map.to_list(map)
```
IvanIvanoff, line 144: Don't call that, the Enumerable protocol is implemented for maps as well:
```elixir
iex(1)> %{a: 1, b: 2} |> Enum.map(fn {key, value} -> %{key: key, value: value + 1} end)
[%{key: :a, value: 2}, %{key: :b, value: 3}]
```

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,15 +106,30 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+      case selector do
+        %{slug: slug} ->
+          {"slug=#{slug}", {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"}}
+
+        %{search_text: search_text} ->
+          {"search_text=#{search_text}", {"search_text", search_text}}
+
+        {:error, error} ->
+          error
+      end
```
IvanIvanoff, line 120: A few issues here:
1. The error case will fail
2. Split the whole function into multiple functions. There can be multiple `social_volume_request` functions pattern matching on the selector and another function that accepts these url/option args and uses them

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,15 +106,30 @@ defmodule Sanbase.SocialData.SocialVolume do
   defp source_to_indicator(<<"professional_traders_chat", _::binary>>),
     do: "professional_traders_chat_overview"
 
-  defp social_volume_request(slug, from, to, interval, source) do
-    url = "#{tech_indicators_url()}/indicator/#{source_to_indicator(source |> to_string())}"
+  defp social_volume_request(selector, from, to, interval, source) do
+    {url_arg, option} =
+      case selector do
+        %{slug: slug} ->
+          {"slug=#{slug}", {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"}}
```
IvanIvanoff, line 113: I thought we reworked it so the ticker is not needed anymore in metricshub? Here you're just making a DB call to fetch the ticker only to discard it in metricshub and not use it.

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,15 +106,30 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+        interval
+      }&source=#{source}"
 
     options = [
       recv_timeout: @recv_timeout,
       params: [
-        {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"},
-        {"datetime_from", DateTime.to_unix(from)},
-        {"datetime_to", DateTime.to_unix(to)},
+        option,
```
IvanIvanoff, line 130: what is `option` - rename to a proper name?

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,15 +106,30 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-  defp social_volume_request(slug, from, to, interval, source) do
-    url = "#{tech_indicators_url()}/indicator/#{source_to_indicator(source |> to_string())}"
+  defp social_volume_request(selector, from, to, interval, source) do
+    {url_arg, option} =
+      case selector do
+        %{slug: slug} ->
+          {"slug=#{slug}", {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"}}
+
+        %{search_text: search_text} ->
+          {"search_text=#{search_text}", {"search_text", search_text}}
```
IvanIvanoff, line 116: why the duplication?

lib/sanbase/social_data/social_volume.ex
IvanIvanoff, line 113: As per our discussion, in case of `slug` sanbase should create the query and pass it as `search_text`:
the code should be something like:
```elixir
case slug |> Project.by_slug(only_preload: [:social_volume_query]) do
  %Project{social_volume_query: query} when not is_nil(query) -> query
  %Project{} = project -> SocialVolumeQuery.default_query(project)
  _ -> error
end
```

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -71,7 +71,7 @@ defmodule Sanbase.SocialData.MetricAdapter do
     |> transform_to_value_pairs(:mentions_count)
   end
 
-  def timeseries_data(metric, %{text: text}, from, to, interval, _aggregation)
+  def timeseries_data(metric, %{text: text} = selector, from, to, interval, _aggregation)
```
IvanIvanoff, line 74: why `= selector` and not use it? This results in a compile-time warning

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,34 +107,54 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+      end
+
+    {"slug", query}
+  end
+
+  defp social_volume_selector_handler(%{search_text: search_text}) do
+    {"search_text", search_text}
+  end
+
+  defp social_volume_selector_handler(%{}) do
```
IvanIvanoff, line 125: don't pattern match on maps only, someone could pass a binary slug and this will fail to match any function.

lib/sanbase/social_data/social_volume.ex
```diff
@@ -106,34 +107,54 @@ defmodule Sanbase.SocialData.SocialVolume do
...
 
     options = [
       recv_timeout: @recv_timeout,
       params: [
-        {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"},
-        {"datetime_from", DateTime.to_unix(from)},
-        {"datetime_to", DateTime.to_unix(to)},
-        {"interval", interval}
+        slug_or_search_text_string,
+        {"source", source}
```
IvanIvanoff, line 141: why some params are directly interpolated into the URL and another are here? This inconsistency is confusing

lib/sanbase/social_data/social_volume.ex
```diff
@@ -79,11 +80,11 @@ defmodule Sanbase.SocialData.SocialVolume do
   end
 
   def topic_search(search_text, from, to, interval, source) do
```
IvanIvanoff, line 82: You can also rename the `topic_search` functions and just let the `topic_search` API resolver to call the proper `social_volume` function

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -75,7 +75,7 @@ defmodule Sanbase.SocialData.MetricAdapter do
       when metric in @social_volume_timeseries_metrics do
     "social_volume_" <> source = metric
 
-    Sanbase.SocialData.topic_search(text, from, to, interval, source)
+    Sanbase.SocialData.social_volume(%{text: text}, from, to, interval, source)
     |> transform_to_value_pairs(:mentions_count)
   end
```
IvanIvanoff, line 80: can you unify all these `timeseries_data` functions, as now there is:
-`%{} = selector` and passing the selector
- `%{slug: slug}` and passing it as `%{slug: slug}` (can be assigned to `selector` var and pass it)
- same as above for `%{text: text}`
- getting `%{slug: slug}` and passing just slug - `community_messages_count` should be reworked to accept selector as well

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+  end
 
-  defp source_to_indicator(<<"professional_traders_chat", _::binary>>),
-    do: "professional_traders_chat_overview"
+  defp social_volume_request(selector, from, to, interval, source) do
+    slug_or_search_text_string = social_volume_selector_handler(selector)
 
-  defp social_volume_request(slug, from, to, interval, source) do
-    url = "#{tech_indicators_url()}/indicator/#{source_to_indicator(source |> to_string())}"
+    url = "#{metrics_hub_url()}/social_volume?"
```
IvanIvanoff, line 94: You do not need the `?` here - having non-empty `params` would append the params properly

lib/sanbase_web/graphql/resolvers/social_data_resolver.ex
```diff
@@ -59,7 +59,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.SocialDataResolver do
         %{source: source, search_text: search_text, from: from, to: to, interval: interval},
         _resolution
       ) do
-    case SocialData.topic_search(search_text, from, to, interval, source) do
+    case SocialData.social_volume(search_text, from, to, interval, source) do
```
IvanIvanoff, line 62: Is this the proper way to pass the search text? It should be a selector?

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -75,7 +75,7 @@ defmodule Sanbase.SocialData.MetricAdapter do
       when metric in @social_volume_timeseries_metrics do
     "social_volume_" <> source = metric
 
-    Sanbase.SocialData.topic_search(text, from, to, interval, source)
+    Sanbase.SocialData.social_volume(%{text: text}, from, to, interval, source)
```
IvanIvanoff, line 78: Here you use `%{text: text}` but in the handler function you expect `%{search_text: search_text}`. Please test all functions against stage

config/config.exs
```diff
@@ -121,6 +121,9 @@ config :sanbase, Sanbase.MandrillApi,
 
 config :sanbase, Sanbase.TechIndicators, url: {:system, "TECH_INDICATORS_URL"}
 
+config :sanbase, Sanbase.SocialData,
+  metricshub_url: {:system, "METRICS_HUB_URL", "http://localhost:3000"}
```
IvanIvanoff, line 125: this default path is not nice, set the service name as a default (how you access it with VPN on)

lib/sanbase/social_data/social_volume.ex
spiderjako: My mistake, in the beginning I got confused for a bit with the options stuff, while looking at the old implementation with tech_indicators

lib/sanbase/social_data/community.ex
```diff
@@ -48,6 +48,10 @@ defmodule Sanbase.SocialData.Community do
     end
   end
 
+  def community_messages_count(argument) do
+    {:error, "Invalid argument #{argument}"}
```
IvanIvanoff, line 52: Use `#{inspect(argument)}` as things like tuple/map do not have `to_string` implementation - that would result in a not implemented protocol error:
```elixir
iex(1)> map = %{a: 1}
%{a: 1}
iex(2)> "#{map}"
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{a: 1} of type Map
    (elixir 1.10.0) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.10.0) lib/string/chars.ex:22: String.Chars.to_string/1
```

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -47,11 +47,11 @@ defmodule Sanbase.SocialData.MetricAdapter do
   def has_incomplete_data?(_), do: false
 
   @impl Sanbase.Metric.Behaviour
-  def timeseries_data(metric, %{slug: slug}, from, to, interval, _aggregation)
+  def timeseries_data(metric, %{slug: slug} = selector, from, to, interval, _aggregation)
```
tspenov, line 50: `slug` is not used. You can prefix with underscore to avoid unused warning like this: `%{slug: _slug}`

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -63,19 +63,19 @@ defmodule Sanbase.SocialData.MetricAdapter do
     |> transform_to_value_pairs(:dominance)
   end
 
-  def timeseries_data(metric, %{slug: slug}, from, to, interval, _aggregation)
+  def timeseries_data(metric, %{slug: slug} = selector, from, to, interval, _aggregation)
```
tspenov, line 66: same as above comment about unused `slug`.

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -63,19 +63,19 @@ defmodule Sanbase.SocialData.MetricAdapter do
...
       when metric in @community_messages_count_timeseries_metrics do
     "community_messages_count_" <> source = metric
 
-    Sanbase.SocialData.community_messages_count(slug, from, to, interval, source)
+    Sanbase.SocialData.community_messages_count(selector, from, to, interval, source)
     |> transform_to_value_pairs(:mentions_count)
   end
 
-  def timeseries_data(metric, %{text: text}, from, to, interval, _aggregation)
+  def timeseries_data(metric, %{text: text} = selector, from, to, interval, _aggregation)
```
tspenov, line 74: same as above - here `text` is unused.

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-  defp source_to_indicator(<<"discord", _::binary>>), do: "discord_chats_overview"
-  defp source_to_indicator(<<"telegram", _::binary>>), do: "telegram_chats_overview"
+  defp social_volume_selector_handler(_args) do
+    {:error, "Invalid argument for social_volume, please input a slug or search_text"}
+  end
 
-  defp source_to_indicator(<<"professional_traders_chat", _::binary>>),
-    do: "professional_traders_chat_overview"
+  defp social_volume_request(selector, from, to, interval, source) do
+    slug_or_search_text_string = social_volume_selector_handler(selector)
```
tspenov, line 92: I would suggest you match `{"search_text", search_text} = social_volume_selector_handler(selector)` so it raise a pattern match error here and not try to send a request later (or fail later). Also `search_text` is find - no need for `slug_or_search_text_string `.

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+    url = "#{metrics_hub_url()}/social_volume"
 
     options = [
       recv_timeout: @recv_timeout,
       params: [
-        {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"},
-        {"datetime_from", DateTime.to_unix(from)},
-        {"datetime_to", DateTime.to_unix(to)},
-        {"interval", interval}
+        slug_or_search_text_string,
```
tspenov, line 99: After the above change here it should be `{"search_text", search_text}`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-        topic_search_result(result)
+        _ ->
+          {:error, "Invalid slug"}
+      end
 
-      {:ok, %HTTPoison.Response{status_code: status, body: body}} ->
-        error_result(
-          "Error status #{status} fetching results for search text \"#{search_text}\": #{body}"
-        )
+    {"search_text", query}
```
tspenov, line 80: Here it might return `{"search_text", {:error, "Invalid slug"}}`. You can remove this line and the binding to `query` and return `{"search_text", query_text}` on ln: 71 and `{"search_text", SocialVolumeQuery.default_query(project)}` on 74

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+      |> Project.by_slug(only_preload: [:social_volume_query])
+      |> case do
+        %Project{social_volume_query: %Sanbase.Model.Project.SocialVolumeQuery{query: query_text}}
+        when not is_nil(query_text) ->
+          query_text
 
-    {:ok, result}
-  end
+        %Project{} = project ->
+          SocialVolumeQuery.default_query(project)
```
tspenov, line 74: change to `{"search_text", SocialVolumeQuery.default_query(project)}` as per comments below

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +61,62 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-      )
-      |> Sanbase.Utils.Transform.sum_by_datetime(:mentions_count)
+  defp social_volume_selector_handler(%{slug: slug}) do
+    query =
+      slug
+      |> Project.by_slug(only_preload: [:social_volume_query])
+      |> case do
+        %Project{social_volume_query: %Sanbase.Model.Project.SocialVolumeQuery{query: query_text}}
+        when not is_nil(query_text) ->
+          query_text
```
tspenov, line 71: change to `{"search_text", query_text}` as per below comments.

lib/sanbase/social_data/social_volume.ex
```diff
@@ -28,20 +29,18 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+  def social_volume(selector, from, to, interval, source) do
+    social_volume_request(selector, from, to, interval, source)
     |> case do
       {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
         {:ok, result} = Jason.decode(body)
         social_volume_result(result)
 
       {:ok, %HTTPoison.Response{status_code: status}} ->
-        warn_result("Error status #{status} fetching social volume for project #{slug}")
+        warn_result("Error status #{status} fetching social volume")
```
tspenov, line 40: lets include the selector  instead of slug `... for #{inspect(selector)}`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -28,20 +29,18 @@ defmodule Sanbase.SocialData.SocialVolume do
...
 
       {:error, %HTTPoison.Error{} = error} ->
-        error_result(
-          "Cannot fetch social volume data for project #{slug}: #{HTTPoison.Error.message(error)}"
-        )
+        error_result("Cannot fetch social volume data: #{HTTPoison.Error.message(error)}")
```
tspenov, line 43: same as above

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +65,60 @@ defmodule Sanbase.SocialData.SocialVolume do
...
 
     options = [
       recv_timeout: @recv_timeout,
       params: [
-        {"project", "#{Project.ticker_by_slug(slug)}_#{slug}"},
-        {"datetime_from", DateTime.to_unix(from)},
-        {"datetime_to", DateTime.to_unix(to)},
-        {"interval", interval}
+        {"search_text", search_text},
+        {"from_timestamp", from |> DateTime.to_unix(:second)},
```
IvanIvanoff, line 101: after latest metricshub changes (unix -> iso8601) you need to just change these to `DateTime.to_iso8601` (or |> DateTime.truncate(:second) |> DateTime.to_iso8601() if metricshub does not support bigger precision)

config/config.exs
```diff
@@ -121,6 +121,9 @@ config :sanbase, Sanbase.MandrillApi,
 
 config :sanbase, Sanbase.TechIndicators, url: {:system, "TECH_INDICATORS_URL"}
 
+config :sanbase, Sanbase.SocialData,
+  metricshub_url: {:system, "METRICS_HUB_URL", "http://metricshub.stage.san:31080"}
```
IvanIvanoff, line 125: Such defaults belong to the local configs (either your own .env.dev or in dev.secret.exs). If there is no `METRICS_HUB_URL` env var in the pods (which is the case now) this will work on stage just fine so it would be pushed to prod where it would fail.

IvanIvanoff: Actually it won't work on stage either as this is the wireguard VPN hostname and not the actual service name

lib/sanbase/social_data/community.ex
```diff
@@ -48,6 +48,10 @@ defmodule Sanbase.SocialData.Community do
     end
   end
 
+  def community_messages_count(argument) do
+    {:error, "Invalid argument #{inspect(argument)}"}
```
IvanIvanoff, line 52: `Invalid argument for community_messages_count ...`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -5,20 +5,21 @@ defmodule Sanbase.SocialData.SocialVolume do
   require Sanbase.Utils.Config, as: Config
 
   alias Sanbase.Model.Project
+  alias Sanbase.Model.Project.SocialVolumeQuery
 
   require Mockery.Macro
   defp http_client, do: Mockery.Macro.mockable(HTTPoison)
 
   @recv_timeout 25_000
   @sources [:telegram, :professional_traders_chat, :reddit, :discord]
```
IvanIvanoff, line 14: You need to add the new sources to this list in order for them to be included in `total`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +65,60 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-  def topic_search(search_text, from, to, interval, source) do
-    topic_search_request(search_text, from, to, interval, source)
+  defp social_volume_selector_handler(%{slug: slug}) do
+    slug
+    |> Project.by_slug(only_preload: [:social_volume_query])
     |> case do
-      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
-        {:ok, result} = Jason.decode(body)
-        topic_search_result(result)
+      %Project{social_volume_query: %Sanbase.Model.Project.SocialVolumeQuery{query: query_text}}
```
IvanIvanoff, line 72:
```suggestion
      %Project{social_volume_query: %{query: query_text}}
```
structs also match against maps, you can do that to be short - also you don't expect any other struct here so it should be ok

IvanIvanoff: If you want to keep it, you have it aliased so just `%SocialVolumeQuery` would do

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +65,60 @@ defmodule Sanbase.SocialData.SocialVolume do
...
+    {:error, "Invalid argument for social_volume, please input a slug or search_text"}
+  end
 
-  defp source_to_indicator(<<"professional_traders_chat", _::binary>>),
-    do: "professional_traders_chat_overview"
+  defp social_volume_request(selector, from, to, interval, source) do
+    {"search_text", search_text} = social_volume_selector_handler(selector)
 
-  defp social_volume_request(slug, from, to, interval, source) do
-    url = "#{tech_indicators_url()}/indicator/#{source_to_indicator(source |> to_string())}"
```
IvanIvanoff, line 110: `source_to_indicator` is not used anymore but it should - `social_volume_twitter` won't work as it would attempt to use `twitter` as source while metricshub expects `twitter_crypto`

lib/sanbase/social_data/social_volume.ex
```diff
@@ -62,78 +65,60 @@ defmodule Sanbase.SocialData.SocialVolume do
...
-          }
+  defp social_volume_result(%{"data" => map}) do
+    map =
+      Enum.map(map, fn {datetime, value} ->
+        %{
+          datetime: Sanbase.DateTimeUtils.from_iso8601!(datetime),
+          mentions_count: value
+        }
       end)
+      |> Enum.sort_by(&DateTime.to_unix(&1.datetime))
```
IvanIvanoff, line 119: Let's start using the better way of sorting datetimes introduces in 1.10. This would be:
`Enum.sort(datetimes_list, {:desc, DateTime})` or `Enum.sort(datetimes_list, {:asc, DateTime})`

lib/sanbase/social_data/social_volume.ex
IvanIvanoff: Can you also add some tests that would assert that all available sources in `_<source>` are included in this list?
The easiest way I can think of is to add here a function `def sources(), do: @sources` and then add a test that takes the social metrics' sources:
```elixir
sources = Metric.available_metrics()
|> Enum.filter(
fn 
  "social_volume_" <> _source -> true
  _ -> false
end)
|> Enum.map(fn "social_volume_" <> source -> source end)

assert SocialVolume.sources() |> Enum.sort() == sources |> Enum.sort()
```

lib/sanbase/social_data/social_volume.ex
spiderjako: I've added the twitter handling here:
https://github.com/santiment/metrics-hub/pull/47/files#diff-b89fabd619a158eec84ba1c7a49ba272R61

PR #1884, opened 2020-03-31
Update and add MakerDAO metrics into API

Snedashkovsky:


lib/sanbase/clickhouse/metric/metric_files/makerdao_metrics.json
```diff
@@ -12,10 +12,22 @@
     "has_incomplete_data": true,
     "data_type": "timeseries"
   },
-
   {
-    "human_readable_name": "Multi-Collateral DAI Total Supply",
-    "name": "mcd_total_supply",
+    "human_readable_name": "Single-Collateral WETH Locked",
```
IvanIvanoff: Isn't the full name `Single-collateral DAI` and `Multi-collateral DAI` - excluding `DAI` from the name makes it ambiguous?

IvanIvanoff:
The new metrics should be added to the expected metrics in the tests:
Metrics with `access: free` should be added to this list: https://github.com/santiment/sanbase2/blob/master/test/sanbase/billing/metric_access_level_test.exs#L14

Metrics with `access: restricted` should be added to this list:
https://github.com/santiment/sanbase2/blob/master/test/sanbase/billing/metric_access_level_test.exs#L66

lib/sanbase/clickhouse/metric/metric_files/makerdao_metrics.json
Snedashkovsky: Renamed

PR #1886, opened 2020-03-31
Integrate with first promoter

tspenov:
* Check if user is promoter

```graphql
    {
      currentUser {
        id
        settings {
          isPromoter
        }
      }
    }
```

* Create promoter
```graphql
mutation {
  createPromoter(
    refId:"tsetso"
  ) {
    email
    currentBalance
    earningsBalance
    paidBalance
    promotions {
      refId
      referralLink
      promoCode
      visitorsCount
      leadsCount
      salesCount
      customersCount
      refundsCount
      cancellationsCount
      salesTotal
      refundsTotal
    }
  }
}
```

* Show promoter stats
```graphql
{
  showPromoter {
    email
    currentBalance
    earningsBalance
    paidBalance
    promotions {
      refId
      referralLink
      promoCode
      visitorsCount
      leadsCount
      salesCount
      customersCount
      refundsCount
      cancellationsCount
      salesTotal
      refundsTotal
    }
  }
}
```

### DIFF
--- a/lib/sanbase/promoters/first_promoter_api.ex
+++ b/lib/sanbase/promoters/first_promoter_api.ex
@@ -1,0 +1,120 @@
+defmodule Sanbase.Promoters.FirstPromoterApi do
+  @moduledoc """
+  Wrapper for First Promoter promoters API: https://firstpromoter.com/api/v1/promoters/
+  """
+
+  require Logger
+  require Sanbase.Utils.Config, as: Config
+
+  alias Sanbase.Auth.User
+
+  @promoters_api_base_url "https://firstpromoter.com/api/v1/promoters/"
+
+  @type promoter_args :: %{
+          optional(:ref_id) => String.t(),
+          optional(:coupon_code) => String.t()
+        }
+  @type promoter :: map()
+
+  @spec list() :: {:ok, list(promoter)} | {:error, String.t()}
+  def list() do
+    do_list(promoters: [], page: 1)
+  end
+
+  @spec create(%User{}, promoter_args) :: {:ok, promoter} | {:error, String.t()}
+  def create(user, args \\ %{})
+
+  def create(%User{id: id, email: email}, args) do
+    data = Map.merge(%{email: email, cust_id: id}, args) |> URI.encode_query()
+
+    Path.join(base_url(), "create")
+    |> http_client().post(
+      data,
+      headers() ++ [{"Content-Type", "application/x-www-form-urlencoded"}]
+    )
+    |> handle_response()
+  end
+
+  def create(_user, _),
+    do: {:error, "Can't create promoter account. User doesn't have an email address."}
+
+  @spec show(String.t()) :: {:ok, promoter} | {:error, String.t()}
+  def show(user_id) do
+    Path.join(base_url(), "show?cust_id=#{user_id}")
+    |> http_client().get(headers())
+    |> handle_response()
+  end
+
+  @spec update(String.t(), promoter_args) :: {:ok, promoter} | {:error, String.t()}
+  def update(user_id, args) do
+    data = Map.merge(%{cust_id: user_id}, args) |> URI.encode_query()
+
+    Path.join(base_url(), "update")
+    |> http_client().put(
+      data,
+      [{"Content-Type", "application/x-www-form-urlencoded"} | headers()]
+    )
+    |> handle_response()
+  end
+
+  @spec delete(String.t()) :: {:ok, promoter} | {:error, String.t()}
+  def delete(user_id) do
+    Path.join(base_url(), "delete?cust_id=#{user_id}")
+    |> http_client().delete(headers())
+    |> handle_response()
+  end
+
+  # helpers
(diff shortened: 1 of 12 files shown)


lib/sanbase/promoters/first_promoter_api.ex
```diff
@@ -0,0 +1,120 @@
+defmodule Sanbase.Promoters.FirstPromoterApi do
```
IvanIvanoff: we have `Sanbase.Promoter.FirstPromoter` and here it's `Promoters`. Let's stick to singular?

lib/sanbase/promoters/first_promoter_api.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  defp atomize_keys(key) do
+    String.to_existing_atom(key)
+  rescue
+    _ -> String.to_atom(key)
+  end
+
+  defp http_client(), do: HTTPoison
+
+  defp base_url, do: @promoters_api_base_url
```
IvanIvanoff: I think we should have such functions only if they're public as other modules can't access the module attribute. Let's just use the attribute here?

lib/sanbase/promoters/first_promoter_api.ex
```diff
@@ -0,0 +1,120 @@
...
+  @moduledoc """
+  Wrapper for First Promoter promoters API: https://firstpromoter.com/api/v1/promoters/
+  """
+
+  require Logger
+  require Sanbase.Utils.Config, as: Config
+
+  alias Sanbase.Auth.User
+
+  @promoters_api_base_url "https://firstpromoter.com/api/v1/promoters/"
```
IvanIvanoff: Since Elixir 1.10 there's `Application.compile_env`. The example from the docs is:
```elixir
@db_host Application.compile_env(:my_app, :db_host, "db.local")
```
It was previously discouraged to use `Application.get_env` at module level because it was unexpected for many users that changing config won't have effect as this is injected at compile time. `compile_env` on the other hand is descriptive so we can move such links to the config without breaking any conventions/good practices :?

lib/sanbase/promoters/first_promoter_api.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  @type promoter_args :: %{
+          optional(:ref_id) => String.t(),
+          optional(:coupon_code) => String.t()
+        }
+  @type promoter :: map()
```
IvanIvanoff: are there any required fields that can be described here?

lib/sanbase/promoters/first_promoter_api.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  @spec show(String.t()) :: {:ok, promoter} | {:error, String.t()}
+  def show(user_id) do
+    Path.join(base_url(), "show?cust_id=#{user_id}")
+    |> http_client().get(headers())
+    |> handle_response()
+  end
+
+  @spec update(String.t(), promoter_args) :: {:ok, promoter} | {:error, String.t()}
+  def update(user_id, args) do
```
IvanIvanoff: where is this used?

lib/sanbase/promoters/first_promoter_api.ex
tspenov: What is the benefit of adding to config ? Module attribute in this case seems fine to me.

lib/sanbase/promoters/first_promoter_api.ex
tspenov: Required in what sense ? I can add only the wanted fields when you invoke with graphql query but that means changing on another place if I decide to add new graphql field.
I plan adding example responses in the tests so there it can be seen as a reference.

lib/sanbase/promoters/first_promoter_api.ex
tspenov: update, list and delete are not currently used but I imagine they might get handy later.
I think update will be needed to migrate current promoters because they don't have sanbase id at the moment

lib/sanbase/promoters/first_promoter_api.ex
tspenov: changed to plural cause that is recommended in https://hexdocs.pm/phoenix/contexts.html
> Naming things is hard. If you're stuck when trying to come up with a context name when the grouped functionality in your system isn't yet clear, you can simply use the plural form of the resource you're creating.

lib/sanbase/promoters/first_promoter_api.ex
tspenov: Added example response in tests.

lib/sanbase/promoters/first_promoter_api.ex
IvanIvanoff: Not much other than stating some intentions like `This value intentionally is set during compile time`.
It is a step into the direction of splitting the config into the existing one and config/releases.exs so it's even cleaner what is read compile time and what is read runtime (our current :system tuples) https://groups.google.com/forum/#!topic/elixir-lang-core/jWB-uC7_8UU

PR #1887, opened 2020-04-01
Rework available queries

IvanIvanoff:
#### Summary
- Remove queries available in getMetric
- Remove project fields as they are not queries
- Return better results for historical balance and holders queries

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/project/available_queries.ex
+++ b/lib/sanbase/model/project/available_queries.ex
@@ -31,15 +31,15 @@ defmodule Sanbase.Model.Project.AvailableQueries do
   """
   @spec get(Sanbase.Model.Project.t()) :: [String.t()]
   def get(%Project{} = project) do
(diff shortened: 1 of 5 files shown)


lib/sanbase/model/project/available_queries.ex
```diff
@@ -31,155 +31,87 @@ defmodule Sanbase.Model.Project.AvailableQueries do
   """
   @spec get(Sanbase.Model.Project.t()) :: [String.t()]
   def get(%Project{} = project) do
+    project = project |> Sanbase.Repo.preload([:eth_addresses, :btc_addresses, :infrastructure])
```
tspenov, line 34: do we use `btc_addresses` preload anywhere ?

IvanIvanoff: not anymore, I'll remove them

PR #1895, opened 2020-04-03
Fix metric signal rounding intervals issue

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/test/sanbase/signals/metric/trigger_metric_test.exs
+++ b/test/sanbase/signals/metric/trigger_metric_test.exs
@@ -62,7 +62,7 @@ defmodule Sanbase.Signal.TriggerMetricTest do
           fn -> {:ok, [%{datetime: datetimes |> List.first(), value: 100}]} end,
           fn -> {:ok, [%{datetiem: datetimes |> List.last(), value: 5000}]} end
         ]
-        |> Sanbase.Mock.wrap_consecutives(5)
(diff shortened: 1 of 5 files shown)


test/sanbase/signals/metric/trigger_metric_test.exs
```diff
@@ -54,23 +55,16 @@ defmodule Sanbase.Signal.TriggerMetricTest do
...
-
-      http_resp = %HTTPoison.Response{body: resp, status_code: 200}
+      # Return a fun with arity 5 that will return different results
+      # for consecutive calls
+      mock_fun =
+        [
+          fn -> {:ok, [%{datetime: datetimes |> List.first(), value: 100}]} end,
+          fn -> {:ok, [%{datetiem: datetimes |> List.last(), value: 5000}]} end
+        ]
+        |> Sanbase.Mock.wrap_consecutives(5)
```
tspenov, line 65: I would say invoking like this: `Sanbase.Mock.wrap_consecutives(arity: 5)` reads better but we can change on another iteration.

IvanIvanoff: All these functions are generated compile time. It can be done by generating them as private function and make this one calling it with the proper arguments.

PR #1898, opened 2020-04-06
Add code to products

tspenov:
* Add `code` field in products with codes `SANAPI`, `SANBASE`, `SANDATA`.
the idea is to use that in metrics json `"min_plan": {"SANAPI": "pro", "SANBASE": "free"}` and in graphql metadata  `meta(min_plan: %{"SANAPI" => "pro", "SANBASE => "free")`.
* removing Sansheets product and plans
* Rename Sangraphs -> Sandata


lib/sanbase/auth/user/permissions.ex
```diff
@@ -9,7 +9,7 @@ defmodule Sanbase.Auth.User.Permissions do
       api: Product.product_api() in user_subscriptions_product_ids,
       sanbase: Product.product_sanbase() in user_subscriptions_product_ids,
       spreadsheet: Product.product_sanbase() in user_subscriptions_product_ids,
-      sangraphs: Product.product_sangraphs() in user_subscriptions_product_ids
+      sangraphs: Product.product_sandata() in user_subscriptions_product_ids
```
IvanIvanoff, line 12: why rename function from `sangraphs` to `sandata` but the key is still `sangraphs`?

tspenov: no reason, my search and replace is not in perfect condition today

PR #1904, opened 2020-04-08
Rework price_histogram

IvanIvanoff:
#### Summary
Break the bucket containing the average price for the queried interval
into 2 buckets with that price as divider. This way it can clearly be
seen how many tokens were acquired at lower and higher prices

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/metric/histogram_metric.ex
+++ b/lib/sanbase/clickhouse/metric/histogram_metric.ex
@@ -71,7 +71,7 @@ defmodule Sanbase.Clickhouse.Metric.HistogramMetric do
     # Generate limit number of ranges to properly empty ranges as 0
     ranges_map =
       Stream.unfold(min, price_ranges)
-      |> Enum.take_while(fn [_lower, upper] -> upper > max end)
+      |> Enum.take(limit)
       |> Enum.into(%{}, fn range -> {range, 0.0} end)
 
     # Map every price to the proper range
@@ -110,7 +110,7 @@ defmodule Sanbase.Clickhouse.Metric.HistogramMetric do
   defp break_bucket(bucketed_data, original_data, [low, high], divider) do
     {lower_half_amount, upper_half_amount} =
       original_data
-      |> Enum.reduce({0, 0}, fn %{price: price, value: value}, {acc_lower, acc_upper} ->
+      |> Enum.reduce({0.0, 0.0}, fn %{price: price, value: value}, {acc_lower, acc_upper} ->
         cond do
           price >= low and price < divider -> {acc_lower + value, acc_upper}
           price >= divider and price < high -> {acc_lower, acc_upper + value}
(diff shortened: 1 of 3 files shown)


lib/sanbase_web/graphql/schema/types/metric_types.ex
```diff
@@ -206,12 +206,13 @@ defmodule SanbaseWeb.Graphql.MetricTypes do
...
       arg(:to, non_null(:datetime))
       arg(:interval, :interval, default_value: "1d")
-      arg(:limit, :integer, default_value: 100)
+      arg(:limit, :integer, default_value: 20)
 
       complexity(&Complexity.from_to_interval/3)
-      middleware(AccessControl)
 
-      resolve(&MetricResolver.histogram_data/3)
+      # middleware(AccessControl)
```
tspenov, line 213: commented line

lib/sanbase/clickhouse/metric/sql_query/metric_histogram_sql_query.ex
```diff
@@ -16,10 +24,14 @@ defmodule Sanbase.Clickhouse.Metric.HistogramSqlQuery do
             -sum(measure) AS tokens_amount
         FROM distribution_deltas_5min FINAL
         PREWHERE
+          metric_id = ( SELECT argMax(metric_id, computed_at) FROM metric_metadata PREWHERE name = '#{
```
tspenov, line 27: Before that change it was not filtering by metric. Was that correct ?

IvanIvanoff: for some time yes, then some other metrics were added to the table `distribution_deltas_5min`

lib/sanbase/clickhouse/metric/histogram_metric.ex
```diff
@@ -82,17 +83,44 @@ defmodule Sanbase.Clickhouse.Metric.HistogramMetric do
       low_high_range.(lower, upper)
     end
 
+    # Get the average price for the queried. time range. It will break the [X,Y]
+    # price interval containing that price into [X, price_break] and [price_break, Y]
+    {:ok, %{^slug => price_break}} =
+      Sanbase.Metric.aggregated_timeseries_data("price_usd", %{slug: slug}, from, to, :avg)
```
tspenov, line 89: Can you pass this `price_break` as a param to `maybe_transform_into_buckets` , extract the whole function in utils and add tests to it because it too complex.

IvanIvanoff: not really, because we want to fetch price only in the :ok case of the `maybe_transform...`. The whole function should be reworked in a separate PR as it's too big now.. Let's merge these changes as they introduce important bug fixes

tspenov: sure, lets do

PR #1907, opened 2020-04-09
Update newsletter setting on registration

tspenov:
```graphql
        mutation {
          emailLogin(email: "john@example.com", subscribeToWeeklyNewsletter: true) {
            success
            firstLogin
          }
        }
```
Adds `subscribe_to_weekly_newsletter=true` in the login/register link that is sent to email
`"http://localhost:4000/email_login?email=john%40example.com&subscribe_to_weekly_newsletter=true&token=7eX-..."`

### DIFF
--- a/test/sanbase_web/graphql/user/user_api_test.exs
+++ b/test/sanbase_web/graphql/user/user_api_test.exs
@@ -643,9 +643,8 @@ defmodule SanbaseWeb.Graphql.UserApiTest do
   end
 
   defp to_graphql_args_string(args) do
-    Enum.reduce(args, [], fn {k, v}, acc ->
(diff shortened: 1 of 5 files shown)


test/sanbase_web/graphql/user/user_api_test.exs
```diff
@@ -651,4 +641,12 @@ defmodule SanbaseWeb.Graphql.UserApiTest do
                ]
              }
   end
+
+  defp to_graphql_args_string(args) do
```
IvanIvanoff, line 645: there is such helper function already

IvanIvanoff: `map_to_input_object_str`

tspenov: yes, that will work also

PR #1910, opened 2020-04-10
Filter email comments

tspenov:


lib/sanbase_web/graphql/resolvers/user/user_resolver.ex
```diff
@@ -9,6 +9,11 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserResolver do
   alias Ecto.Multi
   alias Sanbase.Billing.Subscription.SignUpTrial
 
+  def email(%User{} = user, _args, _resolution) do
```
IvanIvanoff, line 12: I suggest we have 3 pattern matches here:
```elixir
def email(%User{email: nil}... => do nothing
def email(%User{id: id}, ... %{context: %{auth: %{current_user: %{id: id}}... => do nothing
def email(...) => hide
```

PR #1916, opened 2020-04-14
Santiment slug min plan exclusion

tspenov:


lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -62,6 +62,11 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
     resolution
   end
 
+  defp check_plan(%Resolution{arguments: %{slug: slug}} = resolution, _)
+       when slug in @allow_access_without_staking do
```
IvanIvanoff, line 66: Can we also rename `@allow_access_without_staking` to `@freely_available_slugs` :?

PR #1918, opened 2020-04-14
Fix random test failure

tspenov:


tspenov:
fixed here: https://github.com/santiment/sanbase2/pull/1917

PR #1928, opened 2020-04-22
Fix metric query

IvanIvanoff:
#### Summary
The aggregation should happen on the original `dt` field and not on `t` after applying the interval rounding. Applying it on `t` would result on applying it on the same value and aggregations like first/last does no longer work properly, but semi-randomly.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
+++ b/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
@@ -32,7 +32,7 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
     query = """
     SELECT
       toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS t,
-      #{aggregation(aggregation, "value", "t")}
+      #{aggregation(aggregation, "value", "dt")}
     FROM(
       SELECT
         dt,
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -41,19 +41,8 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
...
-        ) AND
-        metric_id = (
-          SELECT
-            argMax(metric_id, computed_at) AS metric_id
-          FROM
-            metric_metadata
-          PREWHERE
-            name = ?2
-        )
+        asset_id = ( SELECT argMax(asset_id, computed_at) FROM asset_metadata FINAL PREWHERE name = ?5 ) AND
```
valo, line 44:
```suggestion
        asset_id = ( SELECT asset_id FROM asset_metadata FINAL PREWHERE name = ?5 ) AND
```

lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -41,19 +41,8 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
...
+        metric_id = ( SELECT argMax(metric_id, computed_at) AS metric_id FROM metric_metadata FINAL PREWHERE name = ?2 )
```
valo, line 45:
```suggestion
        metric_id = ( SELECT metric_id AS metric_id FROM metric_metadata FINAL PREWHERE name = ?2 )
```

PR #1929, opened 2020-04-22
Add filter input object to all_projects

IvanIvanoff:
#### Summary
The projects queries now accept a `filters` input object which filters only the project that match the filter.

Available metrics are:
- all on-chain metrics (except top holders)
- price/volume/marketcap metrics
- dev_activity_1d metric

*NOT* available metrics:
- Top Holders
- Social Metrics
- Twitter Metrics
- Github metrics (except dev_activity_1d)

*NOTE*: Intraday metrics could have some tricky behaviour. For example the exchange flow metrics are intraday and available per 5 minutes, so you cannot get the average daily values of them for the past 30 days.
```graphql
{
  healthyProjects: allProjects(
    selector: {
      filters: [
        {
          metric: "volume_usd"
          from: "2020-03-23T00:00:00Z"
          to: "2020-04-23T00:00:00Z"
          aggregation: AVG
          operator: GREATER_THAN
          threshold: 10000000
        }
        {
          metric: "daily_active_addresses"
          from: "2020-03-23T00:00:00Z"
          to: "2020-04-23T00:00:00Z"
          aggregation: AVG
          operator: GREATER_THAN
          threshold: 500
        }
        {
          metric: "dev_activity_1d"
          from: "2020-03-23T00:00:00Z"
          to: "2020-04-23T00:00:00Z"
          aggregation: AVG
          operator: GREATER_THAN
          threshold: 20
        }
      ]
    }
  ) {
    slug
  }
}
```
image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/metric/sql_query_helper.ex
```diff
@@ -6,4 +6,9 @@ defmodule Sanbase.Metric.SqlQuery.Helper do
   def aggregation(:last, value_column, dt_column), do: "argMax(#{value_column}, #{dt_column})"
   def aggregation(:first, value_column, dt_column), do: "argMin(#{value_column}, #{dt_column})"
   def aggregation(aggr, value_column, _dt_column), do: "#{aggr}(#{value_column})"
+
+  def generate_comprasion_string(:less_than, threshold), do: "< #{threshold}"
```
spiderjako, line 10: Shouldn't it be ``generate_comparison_string``?

lib/sanbase_web/graphql/schema/types/project_types.ex
```diff
@@ -22,6 +22,22 @@ defmodule SanbaseWeb.Graphql.ProjectTypes do
...
+  enum :operator_name do
+    value(:less_than)
+    value(:greater_than)
+    value(:greater_than_or_equal_to)
+    value(:less_than_or_equal_to)
+  end
+
+  input_object :project_filter_input_object do
+    field(:metric, non_null(:string))
+    field(:from, non_null(:datetime))
```
tspenov, line 34: Wouldn't `from` and `to` be common between all filters ? That should be depending on the UI mockups.

lib/sanbase_web/graphql/schema/queries/project_queries.ex
```diff
@@ -20,6 +20,8 @@ defmodule SanbaseWeb.Graphql.Schema.ProjectQueries do
     field :all_projects, list_of(:project) do
       meta(access: :free)
 
+      arg(:filters, list_of(:project_filter_input_object))
```
tspenov, line 23: I think that should be rather object, containing global filter options + list of metric filters (what is currently filters). Global opts will be `from`, `to` and maybe some other later. It will be harder to extend if it is just a list.
```
filter: {
  global opts,
  metric_filters: [...] 
}
```

lib/sanbase_web/graphql/schema/types/project_types.ex
IvanIvanoff: No.
For example if you want to filter all projects with average DAA over a threshold currently we use 30d intervals.
But if you want to also filter the current price (or anything with LAST aggregation), you would want to use a smaller interval because fetching the last data point for 30d interval is the same as fetching the last data point for 1d interval but the later is screening much less data.

lib/sanbase_web/graphql/schema/queries/project_queries.ex
IvanIvanoff: As stated above, I don't think we have any global opts for filters.
What can be done though, is to modify the filters not to be a list so they can have a different meaning - currently we take the intersection of all filters, but we could want their union?

tspenov: Possible. And that is a global option. Making it an object with list of filters inside will make changes later a lot easier.

lib/sanbase_web/graphql/schema/types/project_types.ex
tspenov: So you suggest every metric filter to have [from, to] inputs ? That will be more powerful indeed but also harder for users where they in most cases will want to choose these once.
@Partysun what do you think ? ⬆️

IvanIvanoff: I don't think users would use that API - only sanbase. If you use the API you can fetch the data and do the filtering yourself

lib/sanbase_web/graphql/schema/queries/project_queries.ex
IvanIvanoff: I don't think we can nicely make the combinators with pure graphql (as the input is not polymorphic, no unions, etc.). In signals all this is just a JSON, but then there's no type checking, required field errors, etc.

lib/sanbase_web/graphql/schema/types/project_types.ex
tspenov: I am talking for users on sanbase. To be able to have different [from, to] for filter sanbase UI have to give you a way to fill these. But I think users will most like use global [from, to]

IvanIvanoff: in case of `LAST` aggregation there will be an invisible from/to for which the users won't know, that's different than the others (smaller so less data is screened)

lib/sanbase_web/graphql/schema/queries/project_queries.ex
IvanIvanoff: @tspenov reworked

lib/sanbase/metric/metric.ex
```diff
@@ -164,6 +164,41 @@ defmodule Sanbase.Metric do
...
+  Get the aggregated value for a metric, an identifier and time range.
+  The metric's aggregation function can be changed by the last optional parameter.
+  The available aggregations are #{inspect(@aggregations)}. If no aggregation is
+  provided, a default one (based on the metric) will be used.
+  """
+  def slugs_by_filter(metric, from, to, aggregation, operation, threshold)
+
+  for %{metric: metric, module: module} <- @timeseries_metric_module_mapping do
+    def slugs_by_filter(unquote(metric), from, to, aggregation, operation, threshold)
+        when aggregation in unquote(Map.get(@aggregations_per_metric, metric)) do
```
tspenov, line 177: is `Map.get` working in guard ?

IvanIvanoff: No, you should use `:erlang.map_get(key, map)` in guards. But here there's no `Map.get` in a guard as this is expanded to simply `when aggregation in [:avg, :last, :first, :max, ...]` at compile-time.

lib/sanbase/metric/metric.ex
```diff
@@ -164,6 +164,41 @@ defmodule Sanbase.Metric do
     end
   end
 
+  @doc ~s"""
+  Get the aggregated value for a metric, an identifier and time range.
+  The metric's aggregation function can be changed by the last optional parameter.
```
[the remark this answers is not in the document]
IvanIvanoff: No, it's copy-pasted comment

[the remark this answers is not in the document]
IvanIvanoff: I'll rework the params order in the PR for ordering #1936 so I don't need to fix here and rebase and fix conflicts in the other PR

PR #1934, opened 2020-04-24
Properly handle multiple ethSpent aliases

IvanIvanoff:
#### Summary
Using aliases with different arguments should be handled properly - there should be different groups according to the arguments. In the following examples the computation should be split into 3 groups for 1 day, 7 days and 30 days:
```graphql
{
  allErc20Projects {
    slug
    ethSpent30d: ethSpent(days: 30)
    ethSpent7d: ethSpent(days: 7)
    ethSpent1d: ethSpent(days: 1)
  }
}
```

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
+++ b/lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
@@ -3,8 +3,6 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
   alias Sanbase.Model.Project
   alias Sanbase.Metric
 
-  @max_concurrency 8
-
   def data(), do: Dataloader.KV.new(&query/2)
 
   def query(:aggregated_metric, args) do
@@ -59,9 +57,12 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     args = Enum.to_list(args)
 
     Enum.group_by(args, fn %{days: days} -> days end)
-    |> Sanbase.Parallel.map(fn {days, group} ->
-      {days, average_dev_activity(group, days)}
-    end)
+    |> Sanbase.Parallel.map(
+      fn {days, group} ->
+        {days, average_dev_activity(group, days)}
+      end,
+      ordered: false
+    )
     |> Map.new()
   end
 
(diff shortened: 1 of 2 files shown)


lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -142,37 +137,59 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
...
 
     project_eth_spent =
-      for({:ok, value} <- eth_spent_per_address, do: value)
+      for({:ok, value} <- project_addresses_eth_spent, do: value)
       |> Enum.sum()
+      |> IO.inspect(label: "177", limit: :infinity)
 
-    eth_spent_per_address
+    project_addresses_eth_spent
+    |> IO.inspect(label: "179", limit: :infinity)
```
tspenov, line 180: inspect left

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -142,37 +137,59 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
...
     |> Enum.any?(&match?({:error, _}, &1))
-    |> result(project_eth_spent)
+    |> project_eth_spent_result(project_eth_spent)
+    |> IO.inspect(label: "182", limit: :infinity)
```
tspenov, line 183: another one

lib/sanbase_web/graphql/resolvers/project/project_transactions_resolver.ex
```diff
@@ -50,21 +52,29 @@ defmodule SanbaseWeb.Graphql.Resolvers.ProjectTransactionsResolver do
...
     })
-    |> on_load(&eth_spent_from_loader(&1, project))
+    |> on_load(&eth_spent_from_loader(&1, project, days))
   end
 
-  def eth_spent_from_loader(loader, %Project{id: id}) do
+  def eth_spent_from_loader(loader, %Project{id: id}, days) do
     loader
-    |> Dataloader.get(SanbaseDataloader, :eth_spent, id)
+    |> Dataloader.get(SanbaseDataloader, :eth_spent, days)
```
tspenov, line 68: is that working -> changing `id` with `days` ?

IvanIvanoff: The internals are reworked, so the end map is:
%{
  1 => %{<project_id> => ...}
  7 => %{<project_id> => ...}
  30 => %{<project_id> => ...}
}

IvanIvanoff: If this is not done, then you depend on chance that a race condition won't happen and that the `days` you are interested in (3 different days in total here) will be the one fetched - but currently, this is not the case.

PR #1939, opened 2020-04-27
Rework chart configuration queries

IvanIvanoff:
#### Summary
Rework the queries so there are no comparisons with nil.
conf.user_id == ^querying_user_id fails due to security reasons if the
variable holds nil.

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/chart/configuration.ex
+++ b/lib/sanbase/chart/configuration.ex
@@ -57,7 +57,8 @@ defmodule Sanbase.Chart.Configuration do
   end
 
   def configurations(querying_user_id \\ nil) do
-    configurations_query(querying_user_id)
+    __MODULE__
+    |> accessible_by_user_query(querying_user_id)
     |> Repo.all()
   end
 
@@ -91,7 +92,7 @@ defmodule Sanbase.Chart.Configuration do
   end
 
   defp get_chart_configuration(config_id, querying_user_id) do
-    case from(conf in __MODULE__, where: conf.id == ^config_id) |> Repo.one() do
+    case Repo.get(__MODULE__, config_id) do
       %__MODULE__{user_id: user_id, is_public: is_public} = conf
       when user_id == querying_user_id or is_public == true ->
         {:ok, conf}
@@ -105,7 +106,7 @@ defmodule Sanbase.Chart.Configuration do
   end
 
   defp get_chart_configuration_if_owner(config_id, user_id) do
-    case from(conf in __MODULE__, where: conf.id == ^config_id) |> Repo.one() do
+    case Repo.get(__MODULE__, config_id) do
       %__MODULE__{user_id: ^user_id} = conf ->
         {:ok, conf}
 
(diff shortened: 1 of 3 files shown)


lib/sanbase/chart/configuration.ex
```diff
@@ -118,42 +119,39 @@ defmodule Sanbase.Chart.Configuration do
...
-      where:
-        conf.user_id == ^user_id and
-          (conf.user_id == ^querying_user_id or conf.is_public == true)
+      where: conf.user_id == ^user_id
     )
-  end
-
-  defp user_chart_configurations_query(user_id, querying_user_id, nil) do
-    all_user_configurations_query(user_id, querying_user_id)
+    |> chart_configurations_user_can_access(querying_user_id)
```
tspenov, line 127: `chart_configurations_accessible_by_user_query` or only `configurations_accessible_by_user_query` cause the module implies these are chart_configurations.

lib/sanbase/chart/configuration.ex
```diff
@@ -57,7 +57,8 @@ defmodule Sanbase.Chart.Configuration do
   end
 
   def configurations(querying_user_id \\ nil) do
-    configurations_query(querying_user_id)
+    from(__MODULE__)
```
tspenov, line 60: only `__MODULE__` is sufficient

lib/sanbase/chart/configuration.ex
```diff
@@ -118,42 +119,39 @@ defmodule Sanbase.Chart.Configuration do
...
-          (conf.user_id == ^querying_user_id or conf.is_public == true)
+      where: conf.project_id == ^project_id
     )
   end
 
-  defp configurations_query(nil) do
-    from(conf in __MODULE__, where: conf.is_public == true)
+  defp chart_configurations_user_can_access(query, nil) do
+    query
+    |> where([conf], conf.is_public == true)
```
tspenov, line 150: Lets use `from(conf in Query, where: conf.is_public == true)` for consistency

IvanIvanoff: I'll rework the remaining to be like this one as piping is more understandable when you're combining multiple building blocks

lib/sanbase/chart/configuration.ex
```diff
@@ -118,42 +119,35 @@ defmodule Sanbase.Chart.Configuration do
...
 
   defp user_chart_configurations_query(user_id, querying_user_id, project_id) do
-    from(
-      conf in __MODULE__,
-      where:
-        conf.user_id == ^user_id and conf.project_id == ^project_id and
-          (conf.user_id == ^querying_user_id or conf.is_public == true)
-    )
+    filter_by_project_query(project_id)
+    |> where([conf], conf.user_id == ^user_id)
```
tspenov, line 130: lets call it `filter_by_user` :?

lib/sanbase/chart/configuration.ex
```diff
@@ -118,42 +119,35 @@ defmodule Sanbase.Chart.Configuration do
...
+    |> accessible_by_user(querying_user_id)
```
tspenov, line 131: I would rather end all of these that return query with `_query` or none but not some

PR #1942, opened 2020-04-28
Email notifications unread comments

tspenov:


.gitignore
```diff
@@ -19,6 +19,7 @@ erl_crash.dump
 /config/*.secret.exs
 /app/node_modules/
 /app/build/
+/lib/sanbase_test.ex
```
IvanIvanoff, line 22: :?

.iex.exs
```diff
@@ -145,3 +145,11 @@ alias Sanbase.Billing.{
 
 now = fn -> Timex.now() end
 days_ago = fn days -> Timex.shift(Timex.now(), days: -days) end
+
+run_test = fn func ->
```
IvanIvanoff, line 149: `run_postgres_test` :?

lib/sanbase/comments/entity_comment.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Sanbase.Comment.EntityComment do
+defmodule Sanbase.Comments.EntityComment do
```
IvanIvanoff, line 1: The convention is to use singular names for module names and plural for db tables

IvanIvanoff: https://hexdocs.pm/ecto/getting-started.html
```
We've called this schema `Person` because the naming convention in Ecto for schemas is a singularized name.
```

lib/sanbase/comments/notification.ex
```diff
@@ -0,0 +1,146 @@
...
+    end)
+  end
+
+  defp put_event_in_map(notify_users_map, email, comment_id, event) do
+    events = notify_users_map[email][comment_id][:events] || []
+    comment_map = %{comment_id => events ++ [event]}
+
+    Map.update(notify_users_map, email, comment_map, fn current_map ->
+      Map.merge(current_map, comment_map)
+    end)
```
IvanIvanoff, line 97: Lately I find it more readable to do this in such cases:
```elixir
Map.update(notify_users_map, email, comment_map, &Map.merge(&1, &2))
```

lib/sanbase/comments/notification.ex
```diff
@@ -0,0 +1,146 @@
...
+      where: t.inserted_at > ^yesterday,
+      preload: [comment: [:user, parent: :user], timeline_event: :user],
+      order_by: [asc: :inserted_at]
+    )
+  end
+
+  defp previous_comments_query(post_comment, :insight) do
+    from(p in PostComment,
+      where:
+        p.post_id == ^post_comment.post.id and p.comment_id != ^post_comment.comment.id and
```
IvanIvanoff, line 123: if you manually break the line after the first `and` the formatter will keep it this way:
```elixir
defp previous_comments_query(post_comment, :insight) do
  from(p in PostComment,
    where:
      p.post_id == ^post_comment.post.id and
        p.comment_id != ^post_comment.comment.id and
        p.inserted_at < ^post_comment.inserted_at,
    preload: [comment: :user]
  )
end
```

lib/sanbase/comments/notification.ex
```diff
@@ -0,0 +1,146 @@
...
+    from(t in TimelineEventComment,
+      where:
+        t.timeline_event_id == ^timeline_event_comment.timeline_event.id and
+          t.comment_id != ^timeline_event_comment.comment.id and
+          t.inserted_at < ^timeline_event_comment.inserted_at,
+      preload: [comment: :user]
+    )
+  end
+
+  defp last_processed_id(result_map) do
```
IvanIvanoff, line 139: unused

lib/sanbase/comments/notification.ex
```diff
@@ -0,0 +1,146 @@
...
+    result_map
+    |> Map.values()
+    |> Enum.map(&Map.keys/1)
+    |> List.flatten()
```
IvanIvanoff, line 143: you can use `flat_map` insead of map |> flatten

lib/sanbase_web/graphql/cache/cache.ex
```diff
@@ -219,7 +219,7 @@ defmodule SanbaseWeb.Graphql.Cache do
 
     args = args |> convert_values(ttl)
 
-    cache_key = [name, args] |> :erlang.phash2()
+    cache_key = [name, args] |> IO.inspect() |> :erlang.phash2()
```
IvanIvanoff, line 222: inspect left

lib/sanbase/comments/notification.ex
IvanIvanoff: Actually, to avoid bigger intermediate states it could be:
```elixir
Enum.map(fn list -> list |> Map.keys() |> Enum.max end)
|> Enum.max()
```

.gitignore
tspenov: I am using this module to run some basic tests for current feature I am implementing and that require some db setup.
I do this because it is a faster feedback loop while with tests it does lots of recompiling for changing even single character in the feature module :(
I use it like this from iex:
```
iex> run_test.(Test.feature1/0)
```
And in Test.feature I do some db preparation + execute the feature.

.iex.exs
tspenov: lets keep it shorter

lib/sanbase/comments/entity_comment.ex
tspenov: Actually this is context here. Schemas are indeed singular but contexts are plural.
```
Naming things is hard. If you're stuck when trying to come up with a context name when the grouped functionality in your system isn't yet clear, you can simply use the plural form of the resource you're creating. For example, a Users context for managing users. As you grow your application and the parts of your system become clear, you can simply rename the context to a more refined name at a later time.
```
https://hexdocs.pm/phoenix/contexts.html

.gitignore
IvanIvanoff: I think we can do something else:
- Add a different dir, not `/lib`, but `/local_dev` (name doesn't matter) and add that in the gitignore.
- Add a `elixirc_paths(:dev)` to mix.exs with that folder

Put all code you need for running in your console in dev mode inside that dir.

.iex.exs
IvanIvanoff: but it's not descriptive this way - it's doing a postgres transaction and rollback

.gitignore
tspenov: ok, we can go that way.

lib/sanbase/comments/entity_comment.ex
IvanIvanoff: `Sanbase.Comment` in our case is not a context - it is the schema holding module, so here apply the plural table singular module name IMO.

Our true contexts are probably `Sanbase.Auth.Apikey`, `Sanbase.Auth.Statistics` and not many more modules.

.iex.exs
tspenov: Why do we need to know that is a postgres test ? (it might work in mysql too, though not sure). Maybe `run_db_test` ?

IvanIvanoff: `Repo.transaction` is using the postgres adapter: https://github.com/santiment/sanbase2/blob/master/config/config.exs#L50

lib/sanbase/comments/entity_comment.ex
tspenov: image
It exactly context IMO
The module is called `Sanbase.Comment`, which we better rename to `Sanbase.Comments.Comment` to align with the others `Sanbase.Comments.EntityComment` and `Sanbase.Comments.Notification`

.iex.exs
tspenov: If you change the adepter to mysql one - wouldn't the same code be also relevant ?

lib/sanbase/comments/entity_comment.ex
IvanIvanoff: but then there still won't be a context module, as there won't be any `Sanbase.Comments` module at all. In my understanding of contexts, a context module is a module that hides a few other modules (defining schemas) behind it and stitches together their logic.

In the docs example, there is `Accounts` context which stitches together `User` (`users` table) and `Credential` (`credentials` table). In our case I think we should move `PostComment` into this `comments` dir and the `EntityComment` is actually what will emerge as the context module that could be called just `Comments`.

tspenov: Refer to the upper copied text of the phoenix context guide. It doesn't need to have many related things to be called context. It could start with the plural name of the schema - in our case `comments` and the schema file `comment.ex` and then add related things inside (as notification.ex). `post_coment.ex`, `timeline_events_comment.ex` can be part of sub context `entity_comments`

PR #1958, opened 2020-05-11
Add sentiment api

spiderjako:


lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -86,6 +130,38 @@ defmodule Sanbase.SocialData.MetricAdapter do
...
+  end
+
+  def timeseries_data(metric, %{} = selector, from, to, interval, _aggregation)
+      when metric in @sentiment_volume_consumed_timeseries_metrics do
+    "sentiment_volume_consumed_" <> source = metric
+
+    Sanbase.SocialData.sentiment(selector, from, to, interval, source, "volume_consumed")
+    |> transform_to_value_pairs(:value)
+  end
+
```
IvanIvanoff, line 164: All these can be easily packed into 1 function if you have `@sentiment_timeseries_metrics`. Here's a helper function that would do it:
```elixir

defp split_by_source(str) do
  get_first_part = fn binary, splitter ->
    String.split(binary, splitter) |> hd
  end
  
  cond do
    String.ends_with?(str, "_telegram") -> {get_first_part.(str, "_telegram"), "telegram"}
    String.ends_with?(str, "_reddit") ->  {get_first_part.(str, "_reddit"), "reddit"}
    String.ends_with?(str, "_discord") ->  {get_first_part.(str, "_discord"), "discord"}
    String.ends_with?(str, "_twitter") ->  {get_first_part.(str, "_twitter"), "twitter"}
    String.ends_with?(str, "_bitcointalk") ->  {get_first_part.(str, "_bitcointalk"), "bitcointalk"}
    String.ends_with?(str, "_professional_traders_chat") ->  {get_first_part.(str, "_professional_traders_chat"), "bitcointalk"}
  end
end
```

then the function becomes:
```elixir
def timeseries_data(metric, %{} = selector, from, to, interval, _aggregation)
    when metric in @sentiment_consumed_timeseries_metrics do
  "sentiment_" <> type_source = metric
  {type, source} = split_by_source(type_source)
  
  Sanbase.SocialData.sentiment(selector, from, to, interval, source, type)
  |> transform_to_value_pairs(:value)
end
```

IvanIvanoff: You can also reuse this split_by_source function in other parts of the module. This would be nice as the source extraction would happen in a single place.

lib/sanbase/social_data/sentiment.ex
```diff
@@ -0,0 +1,90 @@
...
+          result
+        end,
+        max_concurency: 4
+      )
+      |> Sanbase.Utils.Transform.sum_by_datetime(:value)
+
+    {:ok, result}
+  end
+
+  def sentiment(selector, from, to, interval, source, column) do
```
IvanIvanoff, line 31: rename `column` to a more appropriate name

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -35,9 +35,53 @@ defmodule Sanbase.SocialData.MetricAdapter do
...
+  @sentiment_volume_consumed_timeseries_metrics [
+    "sentiment_volume_consumed_telegram",
+    "sentiment_volume_consumed_discord",
+    "sentiment_volume_consumed_reddit",
+    "sentiment_volume_consumed_twitter",
+    "sentiment_volume_consumed_bitcointalk",
+    "sentiment_volume_consumed_professional_traders_chat",
+    "sentiment_volume_consumed_total"
+  ]
```
IvanIvanoff, line 76: Here we also have a lot of repetition. Let's generate these functions like this:
```elixir
@sentiment_timeseries_metrics for name <- ["sentiment"],
    type <- ["volume_consumed", "positive", "negative", "balance"],
    source <- [
      "telegram",
      "discord",
      "reddit",
      "twitter",
      "bitcointalk",
      "professional_traders_chat",
      "total"
    ],
    do: "#{name}_#{type}_#{source}"
```

lib/sanbase/social_data/metric_adapter.ex
spiderjako: You're right, that there was a lot of boiler plate going around, thanks for the tip! I would rewrite other parts of the code to use split_by_source in another PR, as this one is explicitly for the sentiment API

lib/sanbase/social_data/sentiment.ex
spiderjako: Used "type" from your comment above :point_up:

lib/sanbase/social_data/metric_adapter.ex
spiderjako: That's a lot better now, thanks!

lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -227,4 +242,33 @@ defmodule Sanbase.SocialData.MetricAdapter do
...
+      String.ends_with?(str, "discord") ->
+        {get_first_part.(str, "discord"), "discord"}
+
+      String.ends_with?(str, "twitter") ->
+        {get_first_part.(str, "twitter"), "twitter"}
+
+      String.ends_with?(str, "bitcointalk") ->
+        {get_first_part.(str, "bitcointalk"), "bitcointalk"}
+
+      String.ends_with?(str, "professional_traders_chat") ->
```
tspenov, line 270: I would remove the duplication even further:
```elixir
sources = ["telegram", "discord", ...]
sources 
|> Enum.find(&String.ends_with?(str, &1)) 
|> get_first_part.(str)
```
and swap the args of `get_first_part`

Also this probably will be reused in other social metrics - so I suggest extracting into some social utils and add simple unit tests.

lib/sanbase/social_data/social_helper.ex
```diff
@@ -0,0 +1,53 @@
...
+  def social_metrics_selector_handler(_args) do
+    {:error, "Invalid argument please input a slug or search_text"}
+  end
+
+  def split_by_source(str) do
+    get_first_part = fn splitter, binary ->
+      String.split(binary, splitter) |> hd |> String.trim_trailing("_")
+    end
+
+    sources = [
```
tspenov, line 34: I think we can move
```elixir
  @sources [:telegram, :professional_traders_chat, :reddit, :discord, :twitter, :bitcointalk]

  def sources(), do: @sources
```
from Sentiment And SocisalVolume modules to the Helper module.
Also can we move `total` in the sources.

lib/sanbase/social_data/social_helper.ex
```diff
@@ -0,0 +1,53 @@
...
+      "telegram",
+      "reddit",
+      "discord",
+      "twitter",
+      "bitcointalk",
+      "professional_traders_chat"
+    ]
+
+    source =
+      (sources ++ ["total"])
```
tspenov, line 44: can we move `total` to the sources ?

lib/sanbase/social_data/social_helper.ex
```diff
@@ -0,0 +1,53 @@
...
+      |> Enum.find(&String.ends_with?(str, &1))
+
+    type =
+      source
+      |> get_first_part.(str)
```
tspenov, line 49: That is more readable on one line.
```elixir
type = get_first_part.(source, str)
```

lib/sanbase/social_data/social_helper.ex
spiderjako: In this case, total will be out of this picture

lib/sanbase/social_data/sentiment.ex
```diff
@@ -0,0 +1,94 @@
...
+          |> Enum.flat_map(fn {:ok, data} -> data end)
+          |> Sanbase.Utils.Transform.sum_by_datetime(:value)
+
+        {:ok, result}
+    end
+  end
+
+  def sentiment(selector, from, to, interval, source, type) do
+    sentiment_request(selector, from, to, interval, source, type)
+    |> case do
```
IvanIvanoff, line 39: nitpick: If the clause is short or not in the middle of pipes, prefer 
```elixir
case sentiment_request(selector, from, to, interval, source, type) do
...
```

lib/sanbase/social_data/sentiment.ex
```diff
@@ -0,0 +1,94 @@
...
+
+      {:ok, %HTTPoison.Response{status_code: status}} ->
+        warn_result("Error status #{status} fetching sentiment #{type} for #{inspect(selector)}")
+
+      {:error, %HTTPoison.Error{} = error} ->
+        error_result(
+          "Cannot fetch sentiment #{type} data for #{inspect(selector)}: #{
+            HTTPoison.Error.message(error)
+          }"
+        )
```
IvanIvanoff, line 52: why one is warn another is error?

PR #1959, opened 2020-05-12
Add api for available metrics for given plan

tspenov:
image

* Product defaults to API
image

image

### DIFF
--- a/lib/sanbase/billing/plan/access_checker.ex
+++ b/lib/sanbase/billing/plan/access_checker.ex
@@ -123,6 +123,21 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
     @min_plan_map[query_or_metric][product] || :free
   end
 
+  @spec get_available_metrics_for_plan(product, plan, restriction_type) :: list(binary())
+        when plan: atom(), product: binary(), restriction_type: atom()
+  def get_available_metrics_for_plan(product, plan, restriction_type \\ :all) do
+    case restriction_type do
+      :free -> @free_metrics
+      :restricted -> @restricted_metrics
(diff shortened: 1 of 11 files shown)


lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -123,6 +123,21 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
+  def get_available_metrics_for_plan(product, plan, restriction_type \\ :all) do
+    case restriction_type do
+      :free -> @free_metrics
+      :restricted -> @restricted_metrics
+      :custom -> @custom_access_queries
+      :all -> @all_metrics
+    end
+    |> Enum.filter(&match?({:metric, _}, &1))
+    |> Enum.filter(&plan_has_access?(plan, product, &1))
+    |> Enum.map(fn {_, name} -> name end)
```
IvanIvanoff, line 137: You can make these 3 `Stream` instead of `map`

josefansinger:
@tspenov Can you reconstruct `plan` from the API key if `plan = null`?

tspenov:
@josefansinger yes, but this query is for customers that don't have an api key yet and are trying to evaluate what metrics does this plan has ?
Otherwise It should be smth like this:
```graphql
currentUser {
  availableMetrics
}
```

josefansinger:
@tspenov Ok. I am currently only checking [available slugs](https://github.com/josefansinger/sanapiwrapper/blob/master/R/sanapiwrapper.R#L194) in my wrapper. Hope this list of slugs is empty, if the metric is not available for the current user/plan/API key. But that's unrelated to the PR.

PR #1962, opened 2020-05-13
Distinct burning ntf from regular payment

tspenov:


tspenov:
close die to wrong branch out

PR #1974, opened 2020-05-15
Use precomputed dev activity for market segments query

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
+++ b/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
@@ -26,6 +26,38 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
     field(:computed_at, :utc_datetime)
   end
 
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -26,6 +26,38 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
     field(:computed_at, :utc_datetime)
   end
 
+  def timeseries_data_query(metric, slugs, from, to, interval, aggregation) when is_list(slugs) do
```
tspenov, line 29: Lets not duplicate the function but extract only the condition for `asset_id` based on whether it a slug or list of slugs :?

IvanIvanoff: It's not only that - `group by` is also changed. Interpolating in multiple places would not be much clearer than some small duplication

PR #1978, opened 2020-05-15
Adding social_dominance from metricshub

spiderjako:


lib/sanbase/social_data/social_dominance.ex
```diff
@@ -39,19 +37,13 @@ defmodule Sanbase.SocialData.SocialDominance do
...
-          Enum.reduce(maps_list, %{}, fn map, acc ->
-            Map.merge(map, acc, fn _k, v1, v2 -> v1 + v2 end)
-          end)
-
-        project_mentions = Map.get(map, ticker_slug, 0)
-        total_mentions = Enum.reduce(map, 0, fn {_k, v}, acc -> acc + v end)
-
-        dominance = Sanbase.Math.percent_of(project_mentions, total_mentions) || 0.0
-
+        dominance = Enum.sum(maps_list) / length(@sources)
```
IvanIvanoff, line 40: Why would you want to divide by source count here? Dominance is not defined this way. This leads to big issues if in source A you have a lot of mentions of the project but also the total volume there is big. And in source B you have 10 mentions of the project but since the source is much smaller, this leads to much bigger dominance. It's not fair to take them with the same weight.

PR #1983, opened 2020-05-18
Rework slug validation

spiderjako:


IvanIvanoff:
Would this break existing signals with invalid slugs? Or this check works only with new signals :?

PR #1984, opened 2020-05-18
Upgrade to Ecto 3

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)


lib/sanbase/auth/user.ex
```diff
@@ -1,6 +1,7 @@
 defmodule Sanbase.Auth.User do
   use Ecto.Schema
-  use Timex.Ecto.Timestamps
+
+  # use Timex.Ecto.Timestamps
```
tspenov, line 4: commented code

config/test.exs
```diff
@@ -33,6 +33,11 @@ config :sanbase, Sanbase.Repo,
   database: "sanbase_test",
   pool_size: 5
 
+config :sanbase, Sanbase.ClickhouseRepo,
```
tspenov, line 36: Why is this added if it is not used in tests ?

lib/sanbase/clickhouse_repo.ex
```diff
@@ -46,55 +51,67 @@ defmodule Sanbase.ClickhouseRepo do
...
   end
 
-  defmacro query_reduce(query, args, init, reducer) do
-    quote bind_quoted: [query: query, args: args, init: init, reducer: reducer] do
-      try do
-        require Sanbase.ClickhouseRepo, as: ClickhouseRepo
+  def sanitize_query(query) do
+    query
+    |> IO.iodata_to_binary()
+    |> String.replace(~r/(\?([0-9]+))(?=(?:[^\\"']|[\\"'][^\\"']*[\\"'])*$)/, "?")
```
tspenov, line 88: Can you add a comment + example what is replaced here.

lib/sanbase/insight/post.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Sanbase.Insight.Post do
   use Ecto.Schema
 
-  use Timex.Ecto.Timestamps
+  # use Timex.Ecto.Timestamps
```
tspenov, line 4: commented code

lib/sanbase/insight/post.ex
```diff
@@ -61,7 +61,7 @@ defmodule Sanbase.Insight.Post do
       on_delete: :delete_all
     )
 
-    field(:published_at, :naive_datetime)
+    field(:published_at, :naive_datetime_usec)
```
tspenov, line 64: is that required to fix datetime comparisons in tests ?

lib/sanbase/timeline/cursor.ex
```diff
@@ -4,14 +4,14 @@ defmodule Sanbase.Timeline.Cursor do
   def filter_by_cursor(query, :before, datetime) do
     from(
       event in query,
-      where: event.inserted_at < ^datetime
+      where: event.inserted_at < ^DateTime.to_naive(datetime)
```
tspenov, line 7: We have to convert to naive in order to user comparison with datetime field in query ?

lib/sanbase/voting/vote.ex
```diff
@@ -3,7 +3,8 @@ defmodule Sanbase.Vote do
   Module for voting for insights and timeline events.
   """
   use Ecto.Schema
-  use Timex.Ecto.Timestamps
+
+  # use Timex.Ecto.Timestamps
```
tspenov, line 7: commented code

lib/sanbase_web/admin/auth/user.ex
```diff
@@ -19,14 +17,12 @@ defmodule Sanbase.ExAdmin.Auth.User do
...
-          unless @environment == :dev do
-            column("San balance", fn eth_account ->
-              case EthAccount.san_balance(eth_account) do
-                nil -> ""
-                san_balance -> san_balance |> Decimal.to_string()
-              end
-            end)
-          end
+          column("San balance", fn eth_account ->
+            case EthAccount.san_balance(eth_account) do
```
tspenov, line 21: I think it returns: `san_balance | :error`.
 Also it will make a call to parity service on opening every user in admin in dev env

lib/sanbase/insight/post.ex
IvanIvanoff: I think i'll revert back to `naive_datetime` and change the tests so different insights have a full second difference in creation time

lib/sanbase_web/admin/auth/user.ex
IvanIvanoff: It's very very rare to open user ex_admin from dev. We already need a connection to some services using VPN, so parity is easy to enable as well

config/test.exs
IvanIvanoff: Now it is started in tests. With all updated dependencies I was getting some errors otherwise (maybe they disappeared after all mocks are reworked?). But it works now, I could try to reduce the unneeded deps in tests in another PR

lib/sanbase/clickhouse_repo.ex
IvanIvanoff: Added comments to both order/sanitize functions

lib/sanbase/timeline/cursor.ex
IvanIvanoff: I am not sure and have to double-check

lib/sanbase_web/admin/auth/user.ex
tspenov: the `san_balance ` function returns `:error` instead of `nil`

PR #1987, opened 2020-05-19
Rework social volume selector handling

spiderjako:


lib/sanbase/social_data/social_volume.ex
```diff
@@ -47,6 +47,9 @@ defmodule Sanbase.SocialData.SocialVolume do
             HTTPoison.Error.message(error)
           }"
         )
+
+      {:error, error} ->
+        error_result("Cannot fetch social volume data for #{inspect(selector)}: #{error}")
```
IvanIvanoff, line 52: is error guaranteed to implement the String.Chars protocol here?

spiderjako: The error, which will be returned in there is:
```
"Invalid slug for social volume. Got: #{inspect(slug)}"
```
so it should be okay, though if you insist, I can make it #{inspect(error)}

PR #1992, opened 2020-05-21
Fix availableMetrics to properly include social volume

IvanIvanoff:
#### Summary
Social Volume is now available for all fetachable projects (that have slugs)

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/social_data/metric_adapter.ex
+++ b/lib/sanbase/social_data/metric_adapter.ex
@@ -138,7 +138,7 @@ defmodule Sanbase.SocialData.MetricAdapter do
   def available_aggregations(), do: @aggregations
 
   @impl Sanbase.Metric.Behaviour
-  def available_slugs() do
+  def available_social_slugs() do
     # Providing a 2 element tuple `{any, integer}` will use that second element
     # as TTL for the cache key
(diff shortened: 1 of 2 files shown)


lib/sanbase/social_data/metric_adapter.ex
```diff
@@ -147,7 +147,13 @@ defmodule Sanbase.SocialData.MetricAdapter do
...
 
   @impl Sanbase.Metric.Behaviour
-  def available_slugs(_metric), do: available_slugs()
+  def available_slugs("social_volume_" <> _source),
+    do: {:ok, Project.List.projects_slugs(preload?: false)}
+
+  def available_slugs("social_dominance_" <> _source), do: available_social_slugs()
+
+  def available_slugs("community_messages_count_" <> _source),
+    do: {:ok, Project.List.projects_by_non_null_field(:telegram_link) |> Enum.map(& &1.slug)}
```
tspenov, line 156: does `community_messages_count_total` depend on present :telegram_link in the project ?

IvanIvanoff: No, but it's the closest thing to the truth we can get. Irina will make a metricshub API for fetching the telegram chat ids and we'll keep this map in postgres

PR #1999, opened 2020-05-26
Add price_eth composite metric based on price_usd

tspenov:
image
image
image


lib/sanbase/prices/price.ex
```diff
@@ -160,6 +161,30 @@ defmodule Sanbase.Price do
...
+        Enum.zip(list1, list2)
+        |> Enum.map(fn {item1, item2} ->
+          if item1.datetime == item2.datetime do
+            %{datetime: item1.datetime, value: func.(item1, item2, field)}
+          else
+            %{datetime: item1.datetime, value: 0}
+          end
+        end)
+        |> Enum.reject(&(&1.value == 0))
+      end
```
IvanIvanoff, line 182: `zip` would not work if one of the timeseries has a missing data point somewhere - all pairs after that will mismatch. A more robust solution would be a similar one to the `total` social volume. Here it would be like this:
```elixir
map = list2 |> Enum.into(%{}, fn %{datetime: dt, value: val} -> {dt, val} end)
list1 |> Enum.map(fn %{datetime: datetime, value: value} ->
   value2 = Map.get(map, datetime)
    # do computation. If 1 dt is missing the others won't be misaligned
end)
```

IvanIvanoff: Wе probably should define an `aggregated_timeseries_data` query so it can be used in overview tables

tspenov: I think now the idea is to put it on each coin detailed page.
image

IvanIvanoff: In this case `aggregated_timeseries_data` would be the way as you need to do it as `aggregation: LAST` in some 1d interval

tspenov: Actually in this case the last price is taken from `latest_conimarketcap_data`

tspenov: image

PR #2014, opened 2020-06-03
Add metric growth stats

tspenov:


lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
+defmodule MetricGrowthStats do
```
IvanIvanoff, line 1: `defmodule Sanbase...`

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+  @moduledoc """
+  Module for calculating growth/consecutive changes stats for metrics returning top performers.
+  """
+
+  @doc """
+  * current_to: End datetime till which we should measure the growth, default: start of today
```
IvanIvanoff, line 7: why not just `to`?

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+  * interval_days: Interval in days for measuring the growth, default: 7 days
+  * aggregation: All available agregations for metric, default: :avg
```
IvanIvanoff, line 9: default should be `nil` so the metric's default aggregation applies. For example `:avg` for transaction volume makes no sense.

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+  * limit: number of max elements, default: 10
+  * max_by: :change | :percent_change, default: :percent_change
+  * threshold: Threshold for previous and current value, default: 0
```
IvanIvanoff, line 12: Can you describe how is the threshold applied, not only to what :?

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+  """
+
+  def max_growth(metric, slugs, opts \\ []) do
+    current_to = Keyword.get(opts, :current_to_datetime, Timex.beginning_of_day(Timex.now()))
+    interval_days = Keyword.get(opts, :interval, 7)
```
IvanIvanoff, line 17: You can use the string interval here, as in other APIs and just call `Sanbase.DateTimeUtils.str_to_sec` on it.

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+               %{slug: slug},
+               current_from,
+               current_to,
+               "#{interval_days}d"
+             ) do
+        %{
+          slug => %{
+            previous: p,
+            current: c,
+            change: c - p,
```
IvanIvanoff, line 72: let's use proper full name `previous_value` and `current_value` :?

lib/sanbase/automatic_statistics/metric_growth_stats.ex
```diff
@@ -0,0 +1,125 @@
...
+      interval_days: interval_days,
+      aggregation: aggregation
+    })
+    |> Enum.reduce(%{}, &Map.merge(&1, &2))
+    |> Enum.filter(fn {_slug, data} -> data.previous > threshold && data.current > threshold end)
+    |> Enum.sort_by(fn {_slug, change} -> change[max_by] end, :desc)
+    |> Enum.take(limit)
+  end
+
+  defp calculate_growth("social_volume_total" = metric, slugs, %{
```
IvanIvanoff, line 43: This applies to all social metrics, not just `social_volume_total`. You could do:
```elixir
@social_metrics Sanbase.SocialData.MetricAdapter.available_metrics
defp calculate_growth(metric, ...) when metric in @social_metrics
```

priv/repo/structure.sql
```diff
@@ -4149,5 +4148,5 @@ ALTER TABLE ONLY public.watchlist_settings
...
...0925064207), (20190930150509), (20191001084823), (20191002082331), (20191002102531), (20191017202631), (20191023054309), (20191024145152), (20191025095838), (20191029063113), (20191111151413), (20191112123819), (20191118125110), (20191119123752), (20191121063853), (20191125120729), (20191129145419), (20191202142313), (20191205141013), (20191205141411), (20191209154624), (20191212130812), (20191213141747), (20191218090721), (20191218122743), (20191218145851), (20191220095225), (20200106124357), (20200109094339), (20200110131000), (20200116085403), (20200121094946), (20200130104623), (20200210083533), (20200210084320), (20200211153511), (20200211154147), (20200212090429), (20200213124731), (20200213133314), (20200218075548), (20200218080212), (20200218081027), (20200218081756), (20200218083631), (20200304122654), (20200311142308), (20200312091952), (20200316154107), (20200319161901), (20200406084732), (20200406085205), (20200409053448), (20200421065542), (20200421071844), (20200424074942), (20200424143441), (20200429120144), (20200430110513), (20200507075524), (20200507093029), (20200507093201), (20200507133717), (20200513124741), (20200601122422), (20200601142229), (20200602101130);
```
IvanIvanoff, line 4151: Why deleting `20200504145204` migration? Even if it is removed, it has existed at some point so it's properly marked as being run

lib/sanbase/social_data/social_helper.ex
```diff
@@ -7,19 +7,24 @@ defmodule Sanbase.SocialData.SocialHelper do
...
-      when not is_nil(query_text) ->
-        {:ok, query_text}
-
-      %Project{} = project ->
-        {:ok, SocialVolumeQuery.default_query(project)}
-
-      _ ->
-        {:error, "Invalid slug"}
-    end
+    Sanbase.Cache.get_or_store(
```
IvanIvanoff, line 10: How often is this called so we need to cache it here? I don't see it being used in your code?

tspenov: If you call `social_volume_total` - it goes through all sources and fetches for them. And for every source it queries db to fetch the query that should be used for the slug.  So 6-7 selects for slug.
Other approach can be just fetch the query for slug beforehand and pass it around but I think caching has the added benefit of not using db even between different user requests.

IvanIvanoff: makes sense now

priv/repo/structure.sql
tspenov: Hmm, this is automatically generated and I have run the migrations from master.
This migration seems to be missing the corresponding file in `priv/repo/migrations` so I guess it is something that was applied locally and then commited structure.sql without migration going to master.  @IvanIvanoff

PR #2016, opened 2020-06-03
Fix currently trading projects function

IvanIvanoff:
Usage of or_where was completely undermining the other checks. It had to
be replaced with `AND where(X or Y or Z). This was making projects with
is_hidden=true to be included

#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/model/project/list/list.ex
+++ b/lib/sanbase/model/project/list/list.ex
@@ -399,10 +399,13 @@ defmodule Sanbase.Model.Project.List do
   def by_name_ticker_slug(values, opts) do
     values = List.wrap(values)
 
-    projects_query(opts)
-    |> where([p], fragment("lower(?)", p.name) in ^values)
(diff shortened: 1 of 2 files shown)


lib/sanbase/model/project/list/list.ex
```diff
@@ -411,7 +414,7 @@ defmodule Sanbase.Model.Project.List do
   one or all of its name, ticker and slug are present int he trending words.
   Filtering out projects based on some conditions can be controled by the options.
 
-  See the "Shared options" section at the module documentation for more options.
+  See the "Shared options" sectiorn at the module documentation for more options.
```
spiderjako, line 417: Unnecessary change I presume

IvanIvanoff: fixed

tspenov:
yeah, `or_where` and composed queries doesn't work well together

PR #2022, opened 2020-06-05
Subscribe to weekly mutation for non-logged in users

tspenov:
1. Frontend sends api call:
mutation {
  verifyEmailNewsletter(email:"ex@gmail.com")
}

2. user clicks on email link: /subscribe_email?token=token&email=email
3. Frontend sends api call:
mutation {
  subscribeEmailNewsletter(token:token,email: email)
}


lib/sanbase/email/newsletter_token.ex
```diff
@@ -0,0 +1,92 @@
...
+  def get_by(email, token) do
+    Repo.get_by(__MODULE__, email: email, token: token)
+  end
+
+  def create_email_token(email) do
+    %__MODULE__{}
+    |> change(
+      email: email,
+      token: generate_email_token(),
+      email_token_generated_at: Timex.now(),
```
IvanIvanoff, line 37: To work with ecto 3 changes, can you tweak this to be naive datetieme without microseconds? (NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second))

IvanIvanoff: Actually, I don't see any benefits of using naive over utc datetimes. Maybe you can just switch to utc here? I think this is what we'll aim for in the future

PR #2043, opened 2020-06-16
Add tags/categories to timeline events

tspenov:
tags: `["BY_ME", "BY_SANFAM", "BY_FOLLOWED", "INSIGHT", "PULSE", "ALERT"]`


lib/sanbase/timeline/post_process.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Sanbase.Timeline.PostProcess do
+  @moduledoc """
+  Module with post processing fuctions for the timeline events.
+  """
+
+  # All tags: [:by_me, :by_sanfam, :by_followed, :insight, :pulse, :alert]
```
IvanIvanoff, line 6: let's use `@doc`

lib/sanbase/timeline/post_process.ex
```diff
@@ -0,0 +1,39 @@
...
+
+    events
+    |> Enum.map(fn event ->
+      []
+      |> add_tag.(:by_me, fn -> current_user_id == event.user.id end)
+      |> add_tag.(:by_sanfam, fn -> event.user.id in sanfamily_ids end)
+      |> add_tag.(:by_followed, fn -> event.user.id in followed_users_ids end)
+      |> add_tag.(:insight, fn -> event.post != nil end)
+      |> add_tag.(:pulse, fn -> event.post != nil and event.post.is_pulse end)
+      |> add_tag.(:alert, fn -> event.user_trigger != nil end)
```
IvanIvanoff, line 31: no point in passing functions here as they're always evaluated, you can do just:
```elixir
add_tag = fn
  tags, _tag, false -> tags
  tags, tag, true -> [tag | tags]
end
   events
    |> Enum.map(fn event ->
      []
      |> add_tag.(:by_me, current_user_id == event.user.id)
...
```
and in `add_tags` reverse the tags, otherwise you might rebuild the whole list many times by appending to the end (if order matters).

```elixir
defp add_tags(tags, event) do
    %{event | tags: tags |> Enum.reverse()}
  end
```

```

File: lib/sanbase/timeline/post_process.ex [hunk 3]
```diff
@@ -0,0 +1,39 @@
+defmodule Sanbase.Timeline.PostProcess do
+  @moduledoc """
+  Module with post processing fuctions for the timeline events.
+  """
+
+  # All tags: [:by_me, :by_sanfam, :by_followed, :insight, :pulse, :alert]
+
+  def tag(events, current_user_id \\ nil) do
```
IvanIvanoff, line 8: `add_tags` :?

tspenov: tag as verb means the same IMO

lib/sanbase/timeline/post_process.ex
tspenov: 👍 Yeah, no need to pass func

PR #2057, opened 2020-06-22
Rework total sources in social metrics

spiderjako:


IvanIvanoff:
@spiderjako Where is the corresponding PR in metrics hub that will handle this?
This should be handled by providing a list of all sources instead of `total` - `telegram, reddit, discord, professional_traders_chat, twitter, bitcointalk`, otherwise the sources defined here could be different than the sources defined in metrics hub.

PR #2062, opened 2020-06-23
Do not propagade clickhouse internal errors to API

IvanIvanoff:
#### Summary
Do not propagate clickhouse internal errors to the API. Reuse the UUID in error messages so all errors and API response contain the same UUID.
image

image

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/clickhouse_repo.ex
+++ b/lib/sanbase/clickhouse_repo.ex
@@ -21,46 +21,7 @@ defmodule Sanbase.ClickhouseRepo do
     {:ok, opts}
   end
 
-  defmacro query_transform(query, args) do
-    quote bind_quoted: [query: query, args: args] do
-      try do
-        require Sanbase.ClickhouseRepo, as: ClickhouseRepo
-
-        ordered_params = ClickhouseRepo.order_params(query, args)
-        sanitized_query = ClickhouseRepo.sanitize_query(query)
-
-        ClickhouseRepo.query(query, args)
-        |> case do
-          {:ok, result} ->
-            transform_fn = &ClickhouseRepo.load(__MODULE__, {result.columns, &1})
-
-            result =
-              Enum.map(
-                result.rows,
-                transform_fn
-              )
-
-            {:ok, result}
-
-          {:error, error} ->
-            {:error, error}
-        end
-      rescue
-        e ->
-          log_id = Ecto.UUID.generate()
-
-          Logger.warn(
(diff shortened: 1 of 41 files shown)


lib/sanbase/clickhouse_repo.ex
```diff
@@ -46,7 +47,16 @@ defmodule Sanbase.ClickhouseRepo do
...
+          log_id = Ecto.UUID.generate()
+
+          Logger.warn(
+            "[#{log_id}] Cannot execute ClickHouse query_transform/2. Reason: #{
+              Exception.message(e)
+            }"
+          )
+
+          {:error,
+           "[#{log_id}] Cannot execute database query. If issue persist please contact Santiment Support."}
```
tspenov, line 59: message can be extracted into module attribute

IvanIvanoff: Done. I've removed the query_transform macro so this can easily be done. This also allowed the replacement of `require` with `alias`

lib/sanbase/clickhouse_repo.ex
```diff
@@ -20,49 +21,23 @@ defmodule Sanbase.ClickhouseRepo do
...
+        {:ok, result} ->
+          {:ok, Enum.map(result.rows, transform_fn)}
+
+        {:error, error} ->
+          log_and_return_error(inspect(error), "query_transform/3")
       end
     rescue
-      e -> {:error, "Cannot execute ClickHouse query. Reason: #{Exception.message(e)}"}
+      e ->
+        log_and_return_error(Exception.message(e), "query_transform/3")
```
tspenov, line 40: I think you can pass only the exception and in `log_and_return_error` to get the message and the stacktrace with `Exception.format_stacktrace()`.

IvanIvanoff: done

lib/sanbase/clickhouse_repo.ex
```diff
@@ -74,14 +49,37 @@ defmodule Sanbase.ClickhouseRepo do
...
+        {:error, error} -> log_and_return_error(inspect(error), "query_reduce/4")
       end
     rescue
       e ->
-        {:error, "Cannot execute ClickHouse query. Reason: #{Exception.message(e)}"}
+        log_and_return_error(e, "query_reduce/4")
     end
   end
 
+  defp log_and_return_error(%{} = e, function_executed) do
```
tspenov, line 60: Now that we have the stacktrace I think we can go without the `function_executed `.

PR #2072, opened 2020-06-25
Add support for multiple anomalies and metric name in get anomalies f…

tspenov:
## Changes

Add a new metric `availableAnomaliesPerMetric` in `projectBySlug` which is convenient way to get
all anomalies for given slug along with the metric they are based on. The response is in the format of `[{metric: metric_name, anomalies: []}, ...]`.

## Ticket

https://favro.com/organization/bdbb4f24afc48b29c74f4fa4/cffeb6e3eaf254588cb68a9a?card=San-8393

## Checklist:

## Usage

```graphql
    {
      projectBySlug(slug: "santiment"){
          availableAnomaliesPerMetric {
            metric
            anomalies
          }
      }
    }
```

## Screenshots 

image


test/sanbase_web/graphql/anomaly/api_project_available_metrics_test.exs
```diff
@@ -41,17 +41,43 @@ defmodule Sanbase.Project.AvailableMetricsApiTest do
...
+      if Keyword.get(opts, :anomalies_per_metric) do
+        """
+          availableAnomaliesPerMetric {
+            metric
+            anomalies
+          }
+        """
+      else
+        "availableAnomalies"
+      end
```
IvanIvanoff, line 75: better split into 2 functions

lib/sanbase/anomaly/anomaly.ex
```diff
@@ -97,18 +112,25 @@ defmodule Sanbase.Anomaly do
...
       with anomaly when not is_nil(anomaly) <- Map.get(@metric_and_model_to_anomaly_map, key),
            slug when not is_nil(slug) <- Map.get(asset_map, asset_id) do
-        Map.update(acc, slug, [anomaly], fn list -> [anomaly | list] end)
+        if Keyword.get(opts, :anomalies_per_metric) do
+          Map.update(acc, slug, [%{metric: metric, anomalies: [anomaly]}], fn list ->
+            [%{metric: metric, anomalies: [anomaly]} | list]
+          end)
+        else
+          Map.update(acc, slug, [anomaly], fn list -> [anomaly | list] end)
+        end
```
IvanIvanoff, line 133: I don't think it's a good idea to completely change the response structure based on an argument. These should be different functions

tspenov: Personally I am 50/50 - this way removes duplication but it is a bit harder to reason about.

tspenov: I used 2 separate public functions to avoid returning different result based on argument. But I think it is ok for the private function.

PR #2073, opened 2020-06-25
Fill project contract address from scrapers

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/external_services/project_info.ex
+++ b/lib/sanbase/external_services/project_info.ex
@@ -44,27 +44,28 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
 
   require Logger
 
-  def project_info_missing?(%Project{
-        website_link: website_link,
-        email: email,
-        reddit_link: reddit_link,
-        twitter_link: twitter_link,
-        btt_link: btt_link,
-        blog_link: blog_link,
-        github_link: github_link,
-        telegram_link: telegram_link,
-        slack_link: slack_link,
-        facebook_link: facebook_link,
-        whitepaper_link: whitepaper_link,
-        ticker: ticker,
-        name: name,
-        token_decimals: token_decimals,
-        total_supply: total_supply,
-        main_contract_address: main_contract_address
-      }) do
+  def project_info_missing?(
+        %Project{
+          website_link: website_link,
+          email: email,
+          reddit_link: reddit_link,
+          twitter_link: twitter_link,
+          btt_link: btt_link,
+          blog_link: blog_link,
+          github_link: github_link,
+          telegram_link: telegram_link,
+          slack_link: slack_link,
+          facebook_link: facebook_link,
+          whitepaper_link: whitepaper_link,
+          ticker: ticker,
+          name: name,
+          total_supply: total_supply
+        } = project
+      ) do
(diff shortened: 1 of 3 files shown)


lib/sanbase/external_services/project_info.ex
```diff
@@ -121,13 +123,18 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
...
+
       project
       |> find_or_create_initial_ico()
-      |> Ico.changeset(Map.from_struct(project_info))
+      |> Ico.changeset(project_info_map)
       |> Repo.insert_or_update!()
 
       project
-      |> Project.changeset(Map.from_struct(project_info))
+      |> Project.ContractAddress.add_contract(project_info_map)
```
tspenov, line 134: Maybe use `Map.split` to get only the needed fields for add_contract ?

lib/sanbase/external_services/project_info.ex
```diff
@@ -44,27 +44,29 @@ defmodule Sanbase.ExternalServices.ProjectInfo do
...
+          ticker: ticker,
+          name: name,
+          token_decimals: token_decimals,
+          total_supply: total_supply
+        } = project
+      ) do
     !website_link or !email or !reddit_link or !twitter_link or !btt_link or !blog_link or
       !github_link or !telegram_link or !slack_link or !facebook_link or !whitepaper_link or
-      !ticker or !name or !main_contract_address or !token_decimals or !total_supply
+      !ticker or !name or !Project.has_contract_address?(project) or !token_decimals or
```
tspenov, line 68: Aren't token decimals also taken from the new table ?

lib/sanbase/external_services/project_info.ex
tspenov, line 134: Maybe use `Map.split` to get only needed fields ?

lib/sanbase/model/project/contract_address.ex
```diff
@@ -21,6 +21,26 @@ defmodule Sanbase.Model.Project.ContractAddress do
     |> cast(attrs, [:address, :decimals, :label, :description, :project_id])
   end
 
+  def add_contract(
```
tspenov, line 24: Can you add spec - especially for the `attrs` param

lib/sanbase/external_services/project_info.ex
IvanIvanoff: This seems to introduce not needed complexity

lib/sanbase/external_services/project_info.ex
IvanIvanoff: Let me push some more changes

lib/sanbase/external_services/project_info.ex
IvanIvanoff: The code extracts only the needed fields and the rest are ignored - this pattern is followed throughout sanbase and we haven't used `split` to explicitly do that.

tspenov: Hahah, I wrote it twice cause I thought it was not submitted.
Yeah, my idea is to be easily observable what fields are used - so either use it with a subset and have a typespec for this subset or use pattern matching in the function to get only needed.

PR #2074, opened 2020-06-25
Fix project api funds raised ico test

tspenov:

### DIFF
--- a/test/sanbase_web/graphql/projects/project_api_funds_raised_test.exs
+++ b/test/sanbase_web/graphql/projects/project_api_funds_raised_test.exs
@@ -40,7 +40,7 @@ defmodule SanbaseWeb.Graphql.ProjectApiFundsRaisedTest do
   test "fetch project funds raised", context do
     %{conn: conn, project: project} = context
 
-    Sanbase.Mock.prepare_mock(Sanbase.Price, :last_record_before, fn slug, datetime ->
+    Sanbase.Mock.prepare_mock(Sanbase.Cache, :get_or_store, fn {_, slug, datetime}, _ ->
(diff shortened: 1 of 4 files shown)


IvanIvanoff:
We should not fix the test but the source of the bug. For some strange reason (some race condition) running only the test suite containing this test fails, but running all tests if succeeds.

tspenov:
@IvanIvanoff Turns out that the mocked function is also cached which causes these random failures. The fixes in the tests are actually the correct values. Interestingly when whole suite is run it returned other result probably due to other test being set up at that time and in setup the cache is cleared. Another way of solving is mocking `fetch_last_prices_before` in `Sanbase.Price.Utils` but that didn't succeed for some reason.

test/sanbase_web/graphql/projects/project_api_funds_raised_test.exs
```diff
@@ -40,7 +40,7 @@ defmodule SanbaseWeb.Graphql.ProjectApiFundsRaisedTest do
   test "fetch project funds raised", context do
     %{conn: conn, project: project} = context
 
-    Sanbase.Mock.prepare_mock(Sanbase.Price, :last_record_before, fn slug, datetime ->
+    Sanbase.Mock.prepare_mock(Sanbase.Cache, :get_or_store, fn {_, slug, datetime}, _ ->
```
IvanIvanoff, line 43: The GraphQL cache is cleared in the setup block: https://github.com/santiment/sanbase2/blob/master/test/support/conn_case.ex#L33

We need to add this Sanbase.Cache cleaning there. This should also be added in `DataCase` maybe, as `Sanbase.Cache` can be called from these tests, too

PR #2076, opened 2020-06-26
Add fast available slugs checking

IvanIvanoff:
#### Summary

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/sanbase/available_slugs.ex
+++ b/lib/sanbase/available_slugs.ex
@@ -1,0 +1,55 @@
+defmodule Sanbase.AvailableSlugs do
+  @moduledoc ~s"""
+  Module for fast checking if a slug is existing.
+
+  The slugs are stored in an ETS table and the check is done via :ets.lookup/2.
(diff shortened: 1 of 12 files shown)


lib/sanbase/available_slugs.ex
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  ### Internals
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: Keyword.get(opts, :name, __MODULE__))
+  end
+
+  def init(opts) do
+    ets_table = :ets.new(@ets_table, [:set, :protected, :named_table])
```
tspenov, line 27: should we set also `:write_concurrency` ? And maybe `read_concurrency` ?

IvanIvanoff: There won't be write concurrency, I'll set the read_concurrency to true.

PR #2080, opened 2020-06-29
Filter timeline events by tag

tspenov:
Usage:
* `filterBy: {type: INSIGHT | PULSE | ALERT}`
* `filterBY: {author: ALL | OWN | SANFAM | FOLLOWED}`
* Or combined : `filterBy: {type: INSIGHT,  author: SANFAM}`. Combination is with `AND`

Tags are also changed to: `["OWN", "SANFAM", "FOLLOWED", "INSIGHT", "PULSE", "ALERT"]`


GarageInc:
I thought that you will add filter by tags only :) It is more simple than combine filters by 'type' and 'author' when we got filtered by 'tags'. @tspenov

tspenov:
@GarageInc We already have filters by author. What was missing was to differentiate the type of the event - one of `insight, pulse or alert`. If you use these filters with type and author - will you need these tags at all ?

GarageInc:
No, tags is't necessary. But maybe designers wants to show them in UI.
Im working only with different filters for new tabs.

PR #2087, opened 2020-06-30
Handle atom values in dynamic watchlist and support dynamic datetimes

IvanIvanoff:
#### Summary
- When watchlist arguments are fetched from the DB some values are
transformed from atoms to strings as json does not have the notion of
atom.

- Watchlist settings are stored in DB and are reused. This means that
static from/to datetimes are not going to work. For this, it now
supports 'dynamic_from|to' which accepts either an interval (1d) or the
string 'now'. This data is used to rewrite the from/to dynamically

Now supported:
```graphql
{
  allProjects(selector: {
    filters: [
      {
        aggregation: LAST,
        dynamicFrom: "1d",
        metric: "daily_active_addresses",
        dynamicTo: "now",
      	operator: GREATER_THAN_OR_EQUAL_TO,
      	threshold: 10000}]}) {
    slug
  }
}
```

These new arguments are usable in the dynamic watchlists as well

[//]: # (#### Related PRs)

[//]: # (#### Screenshots)

### DIFF
--- a/lib/map_utils.ex
+++ b/lib/map_utils.ex
@@ -47,13 +47,52 @@ defmodule Sanbase.MapUtils do
 
     iex> %{a: %{b: [%{"name" => ""}]}} |> Sanbase.MapUtils.find_pair_path("not_existing", "ivan")
     []
-
   """
   def find_pair_path(map, key, value) when is_map(map) do
     do_find_pair_path(map, key, value, [])
     |> Enum.map(&(&1 |> List.flatten() |> Enum.reverse()))
     |> Enum.reject(&(&1 == nil || &1 == []))
   end
+
+  @doc ~s"""
+  Atomize the string keys of a map or list of maps.
+
+  #### Examples:
+    iex> %{"a" => %{"b" => %{"name" => "ivan"}}} |> Sanbase.MapUtils.atomize_keys()
+    %{a: %{b: %{name: "ivan"}}}
+
+    iex> [%{"a" => 1}, %{"b" => [%{"c" => %{"d" => 12}}]}] |> Sanbase.MapUtils.atomize_keys()
+    [%{a: 1}, %{b: [%{c: %{d: 12}}]}]
+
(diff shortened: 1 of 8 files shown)


lib/map_utils.ex
```diff
@@ -47,14 +47,53 @@ defmodule Sanbase.MapUtils do
...
+
+
+    iex> %{already: %{atom: :atom}} |> Sanbase.MapUtils.atomize_keys()
+    %{already: %{atom: :atom}}
+  """
+  def atomize_keys(list) when is_list(list) do
+    Enum.map(list, fn elem -> atomize_keys(elem) end)
+  end
+
+  require Logger
```
tspenov, line 81: unused ?

lib/map_utils.ex
```diff
@@ -47,14 +47,53 @@ defmodule Sanbase.MapUtils do
...
-
   """
   def find_pair_path(map, key, value) when is_map(map) do
     do_find_pair_path(map, key, value, [])
     |> Enum.map(&(&1 |> List.flatten() |> Enum.reverse()))
     |> Enum.reject(&(&1 == nil || &1 == []))
   end
 
+  @doc ~s"""
+  Atomize the string keys of a map or list of maps/
```
tspenov, line 58: trailing `/`

lib/sanbase/application.ex
```diff
@@ -194,8 +194,8 @@ defmodule Sanbase.Application do
         {ConCache,
          [
            name: Sanbase.Cache.name(),
-           ttl_check_interval: :timer.seconds(30),
-           global_ttl: :timer.minutes(5),
+           ttl_check_interval: :timer.seconds(2),
+           global_ttl: :timer.seconds(1),
```
tspenov, line 198: what is the idea here ?

IvanIvanoff: oops, I lower the caching ttls locally so i don't have to clear the cache manually

PR #2091, opened 2020-07-02
Enter/Exit screener signal

IvanIvanoff:
## Changes
- Implement `screener_signal` which notifies what projects exited or entered the screener.
- In order not to mark all projects as new, when the screener is created or updated, a `post_create_process/1` or `post_update_process/2` function is called. They properly set the state, so only changes in screener are made into signals.

## Checklist:

### DIFF
--- a/lib/map_utils.ex
+++ b/lib/map_utils.ex
@@ -89,10 +89,8 @@ defmodule Sanbase.MapUtils do
   # Private functions
 
   @compile {:inline, atomize: 1}
-  defp atomize(atom) when is_atom(atom), do: atom
-
-  defp atomize(str) when is_binary(str),
-    do: str |> Inflex.underscore() |> String.to_existing_atom()
+  defp atomize(value) when is_atom(value) or is_binary(value),
+    do: value |> Inflex.underscore() |> String.to_existing_atom()
 
   defp do_find_pair_path(map, key, value, path) when is_map(map) do
     keys = Map.keys(map)
--- a/lib/sanbase/model/project/list/list_selector.ex
+++ b/lib/sanbase/model/project/list/list_selector.ex
@@ -14,6 +14,11 @@ defmodule Sanbase.Model.Project.ListSelector do
     {:ok, Project.List.projects(opts)}
   end
 
+  def slugs(args) do
+    opts = args_to_opts(args)
+    {:ok, Project.List.projects_slugs()}
+  end
+
   @doc ~s"""
   Transform a selector to a keyword list that can be passed to the functions
   in the `Project.List` module to apply filtering/ordering/pagination.
--- a/lib/sanbase/model/project/list/list_selector.ex
+++ b/lib/sanbase/model/project/list/list_selector.ex
@@ -1,7 +1,66 @@
 defmodule Sanbase.Model.Project.ListSelector do
+  import Norm
   import Sanbase.DateTimeUtils
 
   alias Sanbase.Model.Project
+
+  def valid_selector?(args) do
+    args = Sanbase.MapUtils.atomize_keys(args)
+
+    order_by_schema =
+      schema(%{
+        metric: spec(is_binary()),
+        from: spec(&match?(%DateTime{}, &1)),
+        to: spec(&match?(%DateTime{}, &1)),
+        direction: spec(is_atom())
+      })
+
+    pagination_schema =
+      schema(%{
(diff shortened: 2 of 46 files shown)


lib/map_utils.ex
```diff
@@ -89,10 +89,8 @@ defmodule Sanbase.MapUtils do
   # Private functions
 
   @compile {:inline, atomize: 1}
-  defp atomize(atom) when is_atom(atom), do: atom
-
-  defp atomize(str) when is_binary(str),
-    do: str |> Inflex.underscore() |> String.to_existing_atom()
+  defp atomize(value) when is_atom(value) or is_binary(value),
```
tspenov, line 92: Why is this needed ? do we have atoms in camel case ?

lib/sanbase/model/project/list/list_selector.ex
```diff
@@ -1,8 +1,67 @@
...
+        aggregation: spec(is_atom())
+      })
+
+    Enum.map(filters, &conform(&1, filter_schema))
+    |> Enum.find(&match?({:error, _}, &1))
+    |> case do
+      nil ->
+        Enum.map(
+          filters,
+          &([:metric, :from, :to, :operator, :threshold, :aggregation] -- Map.keys(&1))
```
tspenov, line 52: That is because you can't specify required keys in the schema, right ?

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -60,6 +60,8 @@ defmodule Sanbase.Signal.Scheduler do
 
   # returns a tuple {updated_user_triggers, send_result_list}
   defp send_and_mark_as_sent(triggers) do
+    IO.inspect("SEND AND MARK")
```
tspenov, line 63: inspect left

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -74,6 +76,10 @@ defmodule Sanbase.Signal.Scheduler do
           list when is_list(list) ->
             {:ok, updated_user_trigger} = update_last_triggered(user_trigger, list)
 
+            IO.inspect(user_trigger.trigger.settings.state,
```
tspenov, line 79: inspect left

lib/sanbase/signals/trigger/settings/screener_trigger_settings.ex
```diff
@@ -0,0 +1,157 @@
...
+
+      kv = %{
+        type: ScreenerTriggerSettings.type(),
+        operation: settings.operation,
+        added_slugs: added_slugs,
+        removed_slugs: removed_slugs
+      }
+
+      template = """
+      🔔Projects' changes in the screener "Title goes here".
```
tspenov, line 124: I think there should be also a link to the screener  :?

lib/sanbase/signals/trigger/validation/operation_validation.ex
```diff
@@ -90,6 +90,14 @@ defmodule Sanbase.Signal.Validation.Operation do
   def valid_operation?(%{amount_up: value}) when is_number(value), do: :ok
   def valid_operation?(%{amount_down: value}) when is_number(value), do: :ok
 
+  # Validate screener signal
+  def valid_operation?(%{selector: _} = selector) do
+    case Sanbase.Model.Project.ListSelector.valid_selector?(selector) do
+      true -> :ok
+      false -> {:error, "The provided selector is not correct."}
```
tspenov, line 97: `correct` -> `valid` ?

lib/map_utils.ex
IvanIvanoff: When we have JSON in the api in a free format (the `:json` custom type) we cannot enforce the casing. In all other parts of the API there's a built-in transformer which converts camel case to underscore, so the frontend uses `allProjects` but we see it as `all_projects`.

lib/sanbase/model/project/list/list_selector.ex
IvanIvanoff: yes, turned out this library is not supporting that.

lib/sanbase/signals/trigger/settings/screener_trigger_settings.ex
IvanIvanoff: This is automatically added: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/signals/signal.ex#L126-L131

PR #2093, opened 2020-07-06
Implement APIs for reports

tspenov:
## Changes

Added APis to uploading a file report and listing uploaded reports
Admin interface for publishing and controlling if report is only for `PRO` users.

## Ticket

https://www.notion.so/santiment/APIs-to-store-and-retrieve-Alpha-reports-24750f6f15684adba95cc4f3143d0c39

## Checklist:

### Upload report

```bash
curl -X POST \
-F query="mutation { uploadReport(report: \"report\", name: \"Alpha Report title ...\")}" \
-F report=@report.pdf \
-u "user":"pass" \
http://localhost:4000/graphql
```

### List reports
```graphql
    {
      getReports
      {
        url
        name
        description
      }
    }
```

### DIFF
--- a/lib/sanbase/file_store/store.ex
+++ b/lib/sanbase/file_store/store.ex
@@ -3,7 +3,7 @@ defmodule Sanbase.FileStore do
 
   @versions [:original]
   @acl :public_read
-  @extension_whitelist ~w(.jpg .jpeg .gif .png)
+  @extension_whitelist ~w(.jpg .jpeg .gif .png .pdf .csv)
   @max_file_size 5 * 1024 * 1024
 
   @doc ~s"""
--- a/lib/sanbase/report/report.ex
+++ b/lib/sanbase/report/report.ex
@@ -1,9 +1,18 @@
 defmodule Sanbase.Report do
+  @moduledoc """
+  Module that manages .pdf file reports.
+  Files are saved in s3 and the url is stored in database.
+  """
   use Ecto.Schema
+
   import Ecto.Changeset
   import Ecto.Query
 
+  require Logger
+
   alias Sanbase.Repo
+  alias Sanbase.FileStore
+  alias Sanbase.Utils.FileHash
 
   schema "reports" do
     field(:name, :string, null: true)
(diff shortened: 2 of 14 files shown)


lib/sanbase/file_store/store.ex
```diff
@@ -3,7 +3,7 @@ defmodule Sanbase.FileStore do
 
   @versions [:original]
   @acl :public_read
-  @extension_whitelist ~w(.jpg .jpeg .gif .png)
+  @extension_whitelist ~w(.jpg .jpeg .gif .png .pdf .csv)
   @max_file_size 5 * 1024 * 1024
```
IvanIvanoff, line 7: Will this 5MB size be enough for larger PDFs? Maybe we can have different limits for user-uploaded images via `fileUpload` API and admin uploaded files.

lib/sanbase/report/report.ex
```diff
@@ -0,0 +1,80 @@
...
+    |> changeset(params)
+    |> Repo.insert()
+  end
+
+  def save_report(%Plug.Upload{filename: filename} = report) do
+    %{report | filename: milliseconds_str() <> "_" <> filename}
+    |> do_save_report()
+  end
+
+  def list_published_reports(nil) do
```
IvanIvanoff, line 45: `get_published_reports` :?

lib/sanbase/report/report.ex
```diff
@@ -0,0 +1,80 @@
...
+    else
+      {:error, reason} ->
+        Logger.error("Could not save file: #{filename}. Reason: #{inspect(reason)}")
+        {:error, "Could not save file: #{filename}."}
+    end
+  end
+
+  defp milliseconds_str() do
+    DateTime.utc_now()
+    |> DateTime.to_unix(:millisecond)
```
IvanIvanoff, line 77: `:second` would be enough?

priv/repo/migrations/20200706070758_create_reports.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Sanbase.Repo.Migrations.CreateReports do
+  use Ecto.Migration
+
+  def change do
+    create table(:reports) do
+      add(:name, :string)
+      add(:description, :text)
+      add(:url, :string)
```
IvanIvanoff, line 8: can these 3 be null?

lib/sanbase/file_store/store.ex
tspenov: I was thinking the same to introduce separate file store instead of extending this one. For now didn't change because haven't checked how the reports look yet, so they might all be under a `1MB`. There is a size restriction in nginx that should be updated also.

priv/repo/migrations/20200706070758_create_reports.exs
tspenov: I think only name and description. Though I might enforce `not null` also for name otherwise - cause I suppose that is how it will be listed.

IvanIvanoff: `name` should be required mandatory i think.
Also, in the changeset you have
```elixir
    |> validate_required([:url, :is_published, :is_pro])
```
so at application level `url` is enforced, but here it is not.

priv/repo/structure.sql
```diff
@@ -1576,6 +1574,41 @@ CREATE SEQUENCE public.promo_trials_id_seq
...
 
+--
+-- Name: reports; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.reports (
+    id bigint NOT NULL,
+    name character varying(255),
+    description text,
+    url character varying(255),
```
IvanIvanoff, line 1585: I am not sure what these `url`s are going to be, but for example if you open a project page on sanbase right now, it is over 450 chars in some cases

lib/sanbase/report/report.ex
tspenov: Should be fine with both.

priv/repo/migrations/20200706070758_create_reports.exs
tspenov: Added `url` and `name` as not null both in db and app code.

priv/repo/structure.sql
tspenov: `url` is an s3 link to the file. It should be stable ~160 chars + the original file chars. So it would not fit if filename is > 100 chars but we can control this, so it should not be a problem.

PR #2094, opened 2020-07-06
Fix available since

spiderjako:
## Changes

Fix available since when selector not given, gives off internal server error currently

## Ticket

https://www.notion.so/santiment/9b1939e0ccb9494db3b074a6b74787e0?v=1fbf15df65804ae9a76950013b64079b&p=d0fcc05ba3c0416a88f583480da4eed4

## Checklist:

### DIFF
--- a/lib/sanbase/metric/metric.ex
+++ b/lib/sanbase/metric/metric.ex
@@ -314,7 +314,10 @@ defmodule Sanbase.Metric do
 
   for %{metric: metric, module: module} <- @metric_module_mapping do
     def first_datetime(unquote(metric), slug) do
-      unquote(module).first_datetime(unquote(metric), slug)
+      case slug do
+        %{} -> unquote(module).first_datetime(unquote(metric), slug)
+        {:error, msg} -> {:error, msg}
+      end
     end
   end
 
--- a/lib/sanbase/metric/metric.ex
+++ b/lib/sanbase/metric/metric.ex
@@ -313,11 +313,8 @@ defmodule Sanbase.Metric do
   def first_datetime(metric, slug)
 
   for %{metric: metric, module: module} <- @metric_module_mapping do
-    def first_datetime(unquote(metric), slug) do
-      case slug do
-        %{} -> unquote(module).first_datetime(unquote(metric), slug)
-        {:error, msg} -> {:error, msg}
-      end
+    def first_datetime(unquote(metric), selector) do
+      unquote(module).first_datetime(unquote(metric), selector)
     end
   end
 
--- a/lib/sanbase_web/graphql/resolvers/metric_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/metric_resolver.ex
@@ -57,16 +57,20 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
   end
 
   def available_since(_root, args, %{source: %{metric: metric}}) do
-    case valid_selector?(args) do
-      true -> Metric.first_datetime(metric, to_selector(args))
-      false -> {:error, "Please input a slug, word or selector"}
+    selector = to_selector(args)
+
+    case valid_selector?(selector) do
+      true -> Metric.first_datetime(metric, selector)
+      {:error, error} -> {:error, error}
     end
   end
 
   def last_datetime_computed_at(_root, args, %{source: %{metric: metric}}) do
-    case valid_selector?(args) do
-      true -> Metric.last_datetime_computed_at(metric, to_selector(args))
-      false -> {:error, "Please input a slug, word or selector"}
+    selector = to_selector(args)
+
+    case valid_selector?(selector) do
+      true -> Metric.last_datetime_computed_at(metric, selector)
+      {:error, error} -> {:error, error}
     end
   end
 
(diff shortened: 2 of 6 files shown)


lib/sanbase/metric/metric.ex
```diff
@@ -314,7 +314,10 @@ defmodule Sanbase.Metric do
 
   for %{metric: metric, module: module} <- @metric_module_mapping do
     def first_datetime(unquote(metric), slug) do
-      unquote(module).first_datetime(unquote(metric), slug)
+      case slug do
```
IvanIvanoff, line 317: rename to `selector`

lib/sanbase/metric/metric.ex
```diff
@@ -314,7 +314,10 @@ defmodule Sanbase.Metric do
...
+        %{} -> unquote(module).first_datetime(unquote(metric), slug)
+        {:error, msg} -> {:error, msg}
+      end
```
IvanIvanoff, line 320: Let's handle this in the more proper place - the metric resolver. This `{:error, msg}` tuple should not have been propagated here in the first place.

IvanIvanoff: One thing that can be done is to improve the `valid_selector?` function and reuse it in more places. It shouldn't be `  defp valid_selector?(_), do: true` but you should be able to catch the empty map case first and return an error on it.

IvanIvanoff:
@spiderjako Please add tests as well

lib/sanbase_web/graphql/resolvers/metric_resolver.ex
```diff
@@ -252,10 +259,17 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
...
+  defp valid_selector?(_), do: false
   defp to_selector(%{slug: slug}), do: %{slug: slug}
   defp to_selector(%{word: word}), do: %{word: word}
 
   defp to_selector(%{selector: %{} = selector}),
     do: selector
+
+  defp to_selector(_) do
+    {:error, "Missing required slug, word or selector argument"}
+  end
```
IvanIvanoff, line 274: This is too defensive, you have `valid_selector?` to guard for this.

lib/sanbase_web/graphql/resolvers/metric_resolver.ex
```diff
@@ -89,6 +93,9 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
          {:ok, result} <- fit_from_datetime(result, args) do
       {:ok, result |> Enum.reject(&is_nil/1)}
     else
+      false ->
+        {:error, "Cannot use current selector: #{selector}. Please input a correct selector"}
```
IvanIvanoff, line 97: Selector is a map - it does not implement the String.Chars protocol.
 ```suggestion
        {:error, "The provided selector is not valid."}
```

File: lib/sanbase_web/graphql/resolvers/metric_resolver.ex [hunk 5]
```diff
@@ -252,10 +259,17 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
     end
   end
 
-  defp valid_selector?(_), do: true
+  defp valid_selector?(%{slug: list_of_slugs}) when is_list(list_of_slugs), do: true
+  defp valid_selector?(%{selector: args}), do: true
+  defp valid_selector?(%{text: text}), do: true
+  defp valid_selector?(_), do: false
```
IvanIvanoff, line 265: This is a breaking change. Please check what does `valid_selector?` return on errors. (hint: the first vlaid_selector? definition and how this function is used in the `with` statement in timeseries_data for example)

lib/sanbase_web/graphql/resolvers/metric_resolver.ex
```diff
@@ -89,6 +93,9 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
          {:ok, result} <- fit_from_datetime(result, args) do
       {:ok, result |> Enum.reject(&is_nil/1)}
     else
+      false ->
+        {:error, "The provided selector is not valid."}
```
IvanIvanoff, line 97: Just stick to 1 error response in the `valid_selector?` - the error tuple. Handling it this way is error-prone - for example, you don't handle it in all places where `valid_selector?` is used. Currently, it is fixed in 1 out of 3 places where used.

IvanIvanoff: Also this way you're not sure what exactly returned `false` - the error message is best to be written where it originally came from (the `selector_valid?` function in this case).

lib/sanbase_web/graphql/resolvers/metric_resolver.ex
```diff
@@ -252,10 +264,15 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
     end
   end
 
+  defp valid_selector?(%{} = map) when map_size(map) == 0,
+    do: {:error, "The provided selector is not valid."}
```
IvanIvanoff, line 268: "The selector must have at least one field provided. The available selector fields for a metric are listed in the metadata's `availableSelectors` field"

PR #2098, opened 2020-07-07
Full-text insights search

IvanIvanoff:
## Changes

The full-text search will be performed on a specially constructed `document_tokens` tsvector. The document is constructed as all the following combined:
- post title
- post text
- post tags
- post metrics

Request:
```graphql
{
  allInsightsBySearchTerm(searchTerm: "mvrv") {
    title
    text
  }
}
```
Response:
```json
{
  "data": {
    "allInsightsBySearchTerm": [
      {
        "text": "mvrv token age consumed",
        "title": "Token age consumed"
      },
      {
        "text": "mvrv token age consumed",
        "title": "Token age consumed"
      },
      {
        "text": "mvrv token age consumed activity stewardesses",
        "title": "Token age consumed"
      },
      {
        "text": "mvrv token age consumed",
        "title": "What happens now"
      },
      {
        "text": "SOME TEXT",
        "title": "Some MVRV nice article"
      }
    ]
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/insight/search.ex
+++ b/lib/sanbase/insight/search.ex
@@ -1,0 +1,77 @@
+defmodule Sanbase.Insight.Search do
+  import Ecto.Query
+
+  @doc ~s"""
+  Provided a query which has `document_tokens` tsvector field and a search term,
+  filter the results of the query that match the searching conditions only.
+  """
+  def run(query, search_term) do
+    where(
+      query,
+      fragment(
+        """
+        document_tokens @@ plainto_tsquery(?)
+        """,
+        ^search_term
+      )
+    )
+    |> Sanbase.Repo.all()
+  end
+
+  def update_document_tokens(post_id) do
+    {:ok, %{num_rows: 1}} =
+      Sanbase.Repo.query(
+        """
+        UPDATE posts
+        SET document_tokens = (
+          SELECT
+            to_tsvector(
+              posts.title || ' ' ||
+              posts.text ||' ' ||
+              coalesce((string_agg(tags.name, ' ')), '') || ' ' ||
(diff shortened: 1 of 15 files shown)


priv/repo/structure.sql
```diff
@@ -71,8 +71,68 @@ CREATE TYPE public.status AS ENUM (
...
 
+--
+-- Name: increment(); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.increment() RETURNS integer
+    LANGUAGE plpgsql
+    AS $$
+        BEGIN
+                RETURN 1 + 1;
```
tspenov, line 82: 1 + 1 ?

IvanIvanoff: `ecto dump` has included some local tests around function creation

lib/sanbase/insight/search.ex
```diff
@@ -0,0 +1,79 @@
+defmodule Sanbase.Insight.Search do
+  import Ecto.Query
+
+  # Accept query so we can search on all queries/own queries/etc.
+  def run(search_term) do
+    query = from(p in Sanbase.Insight.Post)
+    run(query, search_term)
```
tspenov, line 7: I would rather call that `search_query` which returns query and have:
```elixir
Sanbase.Insight.Post
|> search_query(search_term)
|> Repo.all
```

IvanIvanoff: I think I'll just remove this case without query (it is not used) - it's better to be explicit what insights you are looking at - published, your own, etc.

IvanIvanoff: This would help abstract away the search in the future as well - the search module would only require the existence of `document_tokens` field.

PR #2103, opened 2020-07-10
Fix compile time warnings

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/comments/comment.ex
+++ b/lib/sanbase/comments/comment.ex
@@ -84,8 +84,15 @@ defmodule Sanbase.Comment do
   end
 
   @doc ~s"""
-  Createa a (top-level) comment. When the parent id is nil there is no need to set
-  the parent_id and the root_parent_id - they both should be nil.
+  Createa a (top-level) comment.
+
+  When the parent id is nil:
+  There is no need to set the parent_id and the root_parent_id - they both should be nil.
+
+  When the parent id is not nil:
+  1. In order to properly set the root_parent_id it must be inherited from the parent
+  2. Create the new comment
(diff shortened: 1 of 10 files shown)


lib/sanbase/comments/comment.ex
```diff
@@ -84,8 +84,15 @@ defmodule Sanbase.Comment do
   end
 
   @doc ~s"""
-  Createa a (top-level) comment. When the parent id is nil there is no need to set
-  the parent_id and the root_parent_id - they both should be nil.
+  Createa a (top-level) comment.
```
tspenov, line 87:
```suggestion
  Create a (top-level) comment.
```

lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -37,22 +37,20 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
...
-    dev_7d: averageDevActivity(days: 7)
-    dev_30d: averageDevActivity(days: 30)
-    ...
-    ```
-
-  The `days` key points to a map of results or to an {:error, error} tuple.
-  The map of results has github organizations as key and their average activity
-  as value.
-  """
+  # Returns a map with the average dev activity for every project passed in `args`.
```
tspenov, line 40: isn't that documentation of the function ? In that case I think we should use `@doc` instead of single line comments

IvanIvanoff: We have many `query` functions for completely different metrics. Redefying `@doc` is the same as not having it at all. This way we keep the docs but avoid the warning.

PR #2113, opened 2020-07-16
Get reports by asset tags

tspenov:
## Changes

Adding tags to reports - that should be the assets that are in this report.  
Adding api to fetch all reports that contain one of the provided tags.

## Ticket

https://www.notion.so/santiment/Add-asset-tags-for-reports-Filtering-by-tags-assets-f4e7bb1d43924ba796690da5159e6e29

## Checklist:

## Usage

```graphql
   {
      getReportsByTags(tags: ['t1', 't2'])
      {
        url
        name
        tags
      }
    }
```

## Screenshots

image
image
image

### DIFF
--- a/lib/sanbase/reports/report.ex
+++ b/lib/sanbase/reports/report.ex
@@ -20,19 +20,24 @@ defmodule Sanbase.Report do
     field(:url, :string, null: false)
     field(:is_pro, :boolean, default: false)
     field(:is_published, :boolean, default: false)
+    field(:tags, {:array, :string}, default: [])
 
     timestamps()
   end
 
   @doc false
-  def new_changeset(report, params \\ %{}) do
+  def new_changeset(report, attrs \\ %{}) do
+    attrs = normalize_tags(attrs)
+
     report
-    |> cast(params, [:name, :description, :url, :is_published, :is_pro])
+    |> cast(attrs, [:name, :description, :url, :is_published, :is_pro, :tags])
   end
 
   def changeset(report, attrs) do
+    attrs = normalize_tags(attrs)
+
     report
-    |> cast(attrs, [:name, :description, :url, :is_published, :is_pro])
+    |> cast(attrs, [:name, :description, :url, :is_published, :is_pro, :tags])
     |> validate_required([:url, :name, :is_published, :is_pro])
   end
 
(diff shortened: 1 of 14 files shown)


priv/repo/structure.sql
```diff
@@ -1,8 +1,10 @@
 --
```
IvanIvanoff, line 1: What happened here? Why 4327 lines changed? Can you include only the needed changes from this PR here?

lib/sanbase/reports/report.ex
```diff
@@ -65,22 +77,47 @@ defmodule Sanbase.Report do
...
   end
 
-  def get_published_reports(%{plan: %{name: "FREE"}}) do
-    from(r in __MODULE__, where: r.is_published == true and r.is_pro == false)
-    |> Repo.all()
+  # Helpers
+
+  defp normalize_tags(%{"tags" => tags} = attrs) when is_binary(tags) do
+    Map.put(attrs, "tags", String.split(tags, ~r{\s*,\s*}) |> Enum.map(&String.downcase/1))
   end
```
IvanIvanoff, line 90: Why not just `|> String.split(",")`?

lib/sanbase/reports/report.ex
```diff
@@ -65,22 +77,47 @@ defmodule Sanbase.Report do
     Repo.all(__MODULE__)
   end
 
-  def get_published_reports(nil) do
-    from(r in __MODULE__, where: r.is_published == true and r.is_pro == false)
+  def get_published_reports(subscription) do
+    __MODULE__
+    |> get_published_reports_query(subscription)
```
IvanIvanoff, line 82: You can use `Subscription.plan_name(subscription)` here so you don't pattern match on subscription internals but just on :pro, :free, etc.

lib/sanbase/reports/report.ex
tspenov: To remove whitespaces around

priv/repo/structure.sql
tspenov: Seems`pg_dump` version 12.3 formats the sql. The only non compatible change from postgres 12 is `SET default_table_access_method = 'heap'` which I've removed from `structure.sql`.

IvanIvanoff: well, I should upgrade to postgres 12 or my next PR will have exactly the opposite changes :smiley:

PR #2115, opened 2020-07-16
Rework allProjectsByFunction to return object

IvanIvanoff:
## Changes

- The returned object contains the projects and some stats.
- Rework the return type of the watchlist function to also return the
number of projects that are in the watchlist. In case of present
pagination (the selector function) the total number of projects is
returned so the client knows the right amuount of pages that it can
show.

```graphql
{
  allProjectsByFunction(function: ....) {
    projects {
       slug
    }
   stats {
      projectsCount
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/schema/queries/project_queries.ex
+++ b/lib/sanbase_web/graphql/schema/queries/project_queries.ex
@@ -59,7 +59,7 @@ defmodule SanbaseWeb.Graphql.Schema.ProjectQueries do
       cache_resolve(&ProjectListResolver.all_currency_projects/3)
     end
 
-    field :all_projects_by_function, list_of(:project) do
+    field :all_projects_by_function, :projects_object do
       meta(access: :free)
 
       arg(:function, :json)
--- a/test/sanbase_web/graphql/projects/projects_by_function_test.exs
+++ b/test/sanbase_web/graphql/projects/projects_by_function_test.exs
@@ -11,7 +11,7 @@ defmodule SanbaseWeb.Graphql.ProjectsByFunctionTest do
     coin = insert(:market_segment, %{name: "coin"})
     mineable = insert(:market_segment, %{name: "mineable"})
 
-    insert(:project, %{ticker: "TUSD", slug: "tether", market_segment: stablecoin})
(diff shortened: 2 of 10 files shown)


lib/sanbase_web/graphql/schema/queries/project_queries.ex
```diff
@@ -59,7 +59,7 @@ defmodule SanbaseWeb.Graphql.Schema.ProjectQueries do
       cache_resolve(&ProjectListResolver.all_currency_projects/3)
     end
 
-    field :all_projects_by_function, list_of(:project) do
+    field :all_projects_by_function, :projects_object do
```
tspenov, line 62: The return type is changed, so that is going to break the existing frontend implementation. What is the plan for deployment ?

tspenov: Can we have a new query that returns new way of data. Old that call same context funcs but extracts and returns old way. And when frontend is changed - remove old one ?

test/sanbase_web/graphql/projects/projects_by_function_test.exs
```diff
@@ -59,7 +59,7 @@ defmodule SanbaseWeb.Graphql.ProjectsByFunctionTest do
     function = %{"name" => "market_segment", "args" => %{"market_segment" => "stablecoin"}}
 
     result = execute_query(conn, query(function))
-    projects = result["data"]["allProjectsByFunction"]
+    projects = result["data"]["allProjectsByFunction"]["projects"]
```
tspenov, line 62: Lets add a check for stats also. Especially the pagination case

lib/sanbase_web/graphql/schema/queries/project_queries.ex
IvanIvanoff: The frontend knows about these changes. It is only used in Buidl Heroes page which is really outdated and must be completely reworked, so it is going to be disabled for some time.

test/sanbase_web/graphql/projects/projects_by_function_test.exs
IvanIvanoff: :+1:

PR #2118, opened 2020-07-17
Add filters_combinator in the projects' selector

IvanIvanoff:
## Changes

The `selector` used in `allProjects` and in `function` in dynamic watchlist and `allProjectsByFunction` now has an additional `filtersCombinator: AND | OR` field. It controls how the different filters are combined. The old behavior (and current default) is `AND` - the projects in the result must satisfy **all** filters. The `OR` combination returns the projects that satisfy at least one of the filters. This is used to implement outside channel logic (price less than $1 or more than $10, etc.)

## Checklist:

### DIFF
--- a/lib/sanbase/model/project/list/list_selector.ex
+++ b/lib/sanbase/model/project/list/list_selector.ex
@@ -26,6 +26,7 @@ defmodule Sanbase.Model.Project.ListSelector do
     pagination = args_to_pagination(args) || %{}
 
     with true <- check_filters(filters),
+         true <- valid_filters_combinator?(args),
          {:ok, _} <- conform(order_by, order_by_schema),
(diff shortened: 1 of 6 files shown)


lib/sanbase/model/project/list/list_selector.ex
```diff
@@ -168,6 +170,27 @@ defmodule Sanbase.Model.Project.ListSelector do
...
+        {:error,
+         """
+         Unsupported filter combinator #{inspect(filters_combinator)}.
+         Supported filter combinators are 'and' and 'or'
+         """}
+    end
+  end
+
+  defp args_to_filters_combinator(args) do
+    (get_in(args, [:selector, :filters_combinator]) || "and")
```
tspenov, line 190: isn't `get_in(args, [:selector, :filters_combinator]` returning atom ?

PR #2135, opened 2020-07-28
Votes now can stack up to 20

IvanIvanoff:
## Changes

Rework voting:
- one user can cast between 0 and 20 votes
- remove `totalSanVotes` field from the `votes` field
- introduce `totalVoters` and `currentUserVotes` fields to the `votes` field

```graphql
mutation{
  vote(insightId: 1){
    votes{
      currentUserVotes
      totalVoters
      totalVotes
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/voting/vote.ex
+++ b/lib/sanbase/voting/vote.ex
@@ -88,44 +88,64 @@ defmodule Sanbase.Vote do
     end
   end
 
-  def voted_at(%Post{id: post_id}, %User{id: user_id}) do
-    from(
-      v in __MODULE__,
-      select: v.inserted_at,
-      where: v.post_id == ^post_id and v.user_id == ^user_id
-    )
+  def voted_at(entity, %User{id: user_id}) do
+    entity_user_pair_query(entity, user_id)
+    |> select([v], v.inserted_at)
     |> Repo.one()
   end
 
-  def voted_at(%TimelineEvent{id: timeline_event_id}, %User{id: user_id}) do
-    from(
-      v in __MODULE__,
-      select: v.inserted_at,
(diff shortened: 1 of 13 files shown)


lib/sanbase/voting/vote.ex
```diff
@@ -43,23 +48,104 @@ defmodule Sanbase.Vote do
...
   @spec remove(%__MODULE__{}) ::
           {:ok, %__MODULE__{}} | {:error, Ecto.Changeset.t()}
   def remove(%__MODULE__{} = vote) do
-    Repo.delete(vote)
+    case vote do
+      %__MODULE__{count: 1} ->
+        Repo.delete(vote)
+
+      %__MODULE__{count: count} ->
+        changeset(vote, %{count: count - 1})
```
tspenov, line 86: You can use some optimistic lock on incrementing/decrement votes: https://hexdocs.pm/ecto/Ecto.Changeset.html#optimistic_lock/3 so there are no negatives when there are race condition. Or maybe just a transaction will be sufficient

tspenov: Optimistic locking might not be good for this because it requires keeping one more version field

lib/sanbase/voting/vote.ex
```diff
@@ -43,23 +48,108 @@ defmodule Sanbase.Vote do
...
+  def voted_at(entity, %User{id: user_id}) do
+    entity_query(entity)
+    |> by_user_query(user_id)
+    |> select([v], v.inserted_at)
+    |> Repo.one()
+  end
+
+  def total_votes(entity, user \\ nil)
+
+  def total_votes(entity, nil) do
```
tspenov, line 100: maybe `votes_stats` ?

PR #2144, opened 2020-08-03
Label transaction addresses

tspenov:
## Checklist:

## Usage
```graphql
{
  projectBySlug(slug:"santiment") {
    id
    tokenTopTransactions(
      from: "2020-01-01T00:00:00Z",
      to:"2020-07-20T00:00:00Z"
    ) {
      datetime
      fromAddress {
        address
        isExchange
        labels {
          name
          metadata
        }
      }
      toAddress {
        address
        isExchange
        labels {
          name
          metadata
        }
      }
      trxHash
      trxValue
    }
  }
}
```

## Screenshots

image

### DIFF
--- a/lib/sanbase/clickhouse/labels.ex
+++ b/lib/sanbase/clickhouse/labels.ex
@@ -1,0 +1,105 @@
+defmodule Sanbase.Clickhouse.Labels do
+  @moduledoc """
+  Labeling transaction addresses
+  """
+
+  @type label :: %{
+          name: String.t(),
+          metadata: String.t()
+        }
+
+  @type input_transaction :: %{
+          from_address: %{
+            address: String.t(),
+            is_exhange: boolean
+          },
+          to_address: %{
+            address: String.t(),
+            is_exhange: boolean
+          },
+          trx_value: float,
+          trx_hash: String,
+          datetime: Datetime.t()
+        }
+
+  @type output_transaction :: %{
+          from_address: %{
+            address: String.t(),
+            is_exhange: boolean,
+            labels: list(label)
+          },
+          to_address: %{
+            address: String.t(),
+            is_exhange: boolean,
+            labels: list(label)
+          },
+          trx_value: float,
+          trx_hash: String,
+          datetime: Datetime.t()
+        }
+
+  @spec add_labels(list(input_transaction)) :: {:ok, list(output_transaction)}
+  def add_labels([]), do: {:ok, []}
+
+  def add_labels(transactions) when is_list(transactions) do
+    addresses = get_list_of_addresses(transactions)
+    {query, args} = addresses_labels_query(addresses)
+
+    Sanbase.ClickhouseRepo.query_reduce(query, args, %{}, fn [address, label, metadata], acc ->
+      address = String.downcase(address)
+
+      Map.update(acc, address, [], fn labels ->
+        metadata =
+          case Jason.decode(metadata) do
+            {:ok, _} -> metadata
+            {:error, _} -> ~s|""|
+          end
+
+        label = %{name: label, metadata: metadata}
+        Enum.uniq_by(labels ++ [label], fn %{name: name} -> name end)
+      end)
+    end)
+    |> case do
+      {:ok, address_labels_map} ->
+        {:ok, do_add_labels(transactions, address_labels_map)}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  # helpers
+  defp addresses_labels_query(addresses) do
+    query = """
+    SELECT address, label, metadata
+    FROM blockchain_address_labels FINAL
+    PREWHERE address IN (?1) and sign = 1
+    """
+
+    {query, [addresses]}
+  end
+
+  defp get_list_of_addresses(transactions) do
+    transactions
+    |> Enum.map(fn transaction ->
+      [
+        transaction.from_address.address,
+        transaction.to_address.address
+      ]
+    end)
+    |> List.flatten()
+    |> Enum.uniq()
+    |> Enum.reject(&is_nil/1)
+    |> Enum.map(&String.downcase/1)
+  end
+
+  defp do_add_labels(transactions, address_labels_map) do
+    transactions
+    |> Enum.map(fn %{from_address: from, to_address: to} = transaction ->
+      from = Map.put(from, :labels, Map.get(address_labels_map, from.address, []))
+      to = Map.put(to, :labels, Map.get(address_labels_map, to.address, []))
+
+      %{transaction | from_address: from, to_address: to}
+    end)
+  end
+end
--- a/lib/sanbase/clickhouse/labels.ex
+++ b/lib/sanbase/clickhouse/labels.ex
@@ -46,16 +46,8 @@ defmodule Sanbase.Clickhouse.Label do
     {query, args} = addresses_labels_query(addresses)
 
     Sanbase.ClickhouseRepo.query_reduce(query, args, %{}, fn [address, label, metadata], acc ->
-      metadata =
-        case Jason.decode(metadata) do
-          {:ok, _} -> metadata
-          {:error, reason} -> ~s|""|
-        end
-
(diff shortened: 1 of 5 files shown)


lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,105 @@
+defmodule Sanbase.Clickhouse.Labels do
```
IvanIvanoff, line 1: `Label` as most modules are singular :?

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,105 @@
...
+
+  @spec add_labels(list(input_transaction)) :: {:ok, list(output_transaction)}
+  def add_labels([]), do: {:ok, []}
+
+  def add_labels(transactions) when is_list(transactions) do
+    addresses = get_list_of_addresses(transactions)
+    {query, args} = addresses_labels_query(addresses)
+
+    Sanbase.ClickhouseRepo.query_reduce(query, args, %{}, fn [address, label, metadata], acc ->
+      address = String.downcase(address)
```
IvanIvanoff, line 49: Bitcoin addresses are case sensitive. I think transformations should be done only when needed (sending to DB where case matters) and when sent back to users - not changed.

On the other hand, I think ethereum addresses are stored all lowercased, so you need to lowercase the arguments in the query (only when needed).

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,105 @@
...
+    |> List.flatten()
+    |> Enum.uniq()
+    |> Enum.reject(&is_nil/1)
+    |> Enum.map(&String.downcase/1)
+  end
+
+  defp do_add_labels(transactions, address_labels_map) do
+    transactions
+    |> Enum.map(fn %{from_address: from, to_address: to} = transaction ->
+      from = Map.put(from, :labels, Map.get(address_labels_map, from.address, []))
```
IvanIvanoff, line 99: You lowercase the result from the DB -  here would it work if the list of transactions has an address with a capital letter somewhere? In this case when you get the result back from the map you would not have a match for `Map.get(address_labels_map, from.address, [])`

lib/sanbase/clickhouse/labels.ex
tspenov: For now labels are only for top transactions for erc20/eth. Addresses are made lowercase.

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,104 @@
...
+      metadata =
+        case Jason.decode(metadata) do
+          {:ok, _} -> metadata
+          {:error, reason} -> ~s|""|
+        end
+
+      Map.update(acc, address, [%{name: label, metadata: metadata}], fn labels ->
+        label = %{name: label, metadata: metadata}
+        labels ++ [label]
+      end)
```
IvanIvanoff, line 58:
```suggestion
     label = %{name: label, metadata: metadata}
     Map.update(acc, address, [label], & [label | &1])
```
Is there any reason to append to the back? It is slow. If you need, just call `reverse` on the labels in the end, but there is no `datetime` to require a specific order.

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,104 @@
...
+
+  def add_labels(transactions) when is_list(transactions) do
+    addresses = get_list_of_addresses(transactions)
+    {query, args} = addresses_labels_query(addresses)
+
+    Sanbase.ClickhouseRepo.query_reduce(query, args, %{}, fn [address, label, metadata], acc ->
+      metadata =
+        case Jason.decode(metadata) do
+          {:ok, _} -> metadata
+          {:error, reason} -> ~s|""|
```
IvanIvanoff, line 52: Is there a case where the metadata is a wrong JSON string? If yes - we should error here and not be silent.

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,104 @@
...
+        end
+
+      Map.update(acc, address, [%{name: label, metadata: metadata}], fn labels ->
```
IvanIvanoff, line 55: should this be `label` instead of `name`?

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,104 @@
...
+
+  # helpers
+  defp addresses_labels_query(addresses) do
+    query = """
+    SELECT lower(address) as address, label, metadata
+    FROM blockchain_address_labels FINAL
+    PREWHERE lower(address) IN (?1) and sign = 1
+    """
+
+    {query, [addresses]}
```
IvanIvanoff, line 77: This query would be very slow and would probably time out on prod because the ordering of the blockchain_address_label tables is `ORDER BY (blockchain, asset_id, label, address)`.

Every request would query for addresses of 1 specific blockchain, so in order to make the query faster, we need to include the blockchain the where clause.

Worst case scenario is searching for addresses for which we don't have labels and most of the addresses don't have labels, so the worst case will happen often.

For example:
```sql
SELECT *
FROM blockchain_address_labels
FINAL
WHERE lower(address) IN ('0xjl123j12jlasda', '0xj1ikjashdo', '0xalsdlasdla', '0xjoi1o1asd', '0xajknqo')

Ok.

0 rows in set. Elapsed: 123.312 sec. Processed 71.91 million rows, 12.23 GB (583.15 thousand rows/s., 99.21 MB/s.) 
```
This would run a string transformation (lowercase) and comparison for every single address and it takes over 2 minutes.

IvanIvanoff: hm, my wrong here - using `prewhere` instead of `where` (as in my test) makes the query run under 10 seconds.

lib/sanbase/clickhouse/labels.ex
tspenov: There shouldn't be. I'll remove

lib/sanbase/clickhouse/labels.ex
tspenov: The parent key is also `labels`, so I consider the whole thing + metadata a `label`.

lib/sanbase/clickhouse/labels.ex
tspenov: I'll add the `blockchain` filter. I our case it is only `ethereum`.

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp get_list_of_addresses(transactions) do
+    transactions
+    |> Enum.map(fn transaction ->
+      [
+        transaction.from_address.address,
+        transaction.to_address.address
+      ]
+    end)
+    |> List.flatten()
```
IvanIvanoff, line 80:
```suggestion
    |> Enum.flat_map(fn transaction ->
      [
        transaction.from_address.address,
        transaction.to_address.address
      ]
    end)
```

lib/sanbase/clickhouse/labels.ex
```diff
@@ -0,0 +1,96 @@
...
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp do_add_labels(transactions, address_labels_map) do
+    transactions
+    |> Enum.map(fn %{from_address: from, to_address: to} = transaction ->
+      from =
+        Map.put(from, :labels, Map.get(address_labels_map, String.downcase(from.address), []))
+
+      to = Map.put(to, :labels, Map.get(address_labels_map, String.downcase(to.address), []))
```
IvanIvanoff, line 91:
```suggestion
      from_labels = Map.get(address_labels_map, String.downcase(from.address), [])
      from = Map.put(from, :labels,[]))

      to_labels = Map.get(address_labels_map, String.downcase(to.address), [])
      to = Map.put(to, :labels, to_labels)
```
Avoiding these nestings improves readability

PR #2152, opened 2020-08-04
Add user attributes over time api

tspenov:
## Checklist:

 ## Usage

`days`: default value is `30`
alternatively: `from` and `to` can be provided

```bash
curl -X POST \
-F query="{ getAttributesForUsers(users: [24, 31], days: 40) { userId insertedAt properties} }" \
-u "user":"pass" \
http://localhost:4000/graphql
```

### DIFF
--- a/lib/sanbase_web/graphql/schema/queries/intercom_queries.ex
+++ b/lib/sanbase_web/graphql/schema/queries/intercom_queries.ex
@@ -6,7 +6,7 @@ defmodule SanbaseWeb.Graphql.Schema.IntercomQueries do
   object :user_attributes do
     field(:user_id, :id)
     field(:inserted_at, :datetime)
-    field(:attributes, :json)
+    field(:properties, :json)
   end
 
   object :intercom_queries do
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/graphql/schema/queries/intercom_queries.ex
```diff
@@ -0,0 +1,35 @@
...
+    field :get_attributes_for_users, list_of(:user_attributes) do
+      arg(:users, non_null(list_of(:id)))
+      arg(:from, :datetime)
+      arg(:to, :datetime)
+
+      middleware(BasicAuth)
+
+      resolve(fn _, %{users: users} = args, _ ->
+        from = Map.get(args, :from, Sanbase.DateTimeUtils.days_ago(30))
+        to = Map.get(args, :to, Timex.now())
```
IvanIvanoff, line 29: I would make these params `non_null` to make the queries reproducable

tspenov: I think that is not necessary for this API. I've made it to be able to pass `days` or alternatively `from`+`to`

PR #2155, opened 2020-08-05
Add filtering by asset id

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/labels.ex
+++ b/lib/sanbase/clickhouse/labels.ex
@@ -38,12 +38,12 @@ defmodule Sanbase.Clickhouse.Label do
           datetime: Datetime.t()
         }
 
-  @spec add_labels(list(input_transaction)) :: {:ok, list(output_transaction)}
-  def add_labels([]), do: {:ok, []}
+  @spec add_labels(String.t(), list(input_transaction)) :: {:ok, list(output_transaction)}
+  def add_labels(_, []), do: {:ok, []}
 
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/labels.ex
```diff
@@ -59,14 +59,18 @@ defmodule Sanbase.Clickhouse.Label do
...
+  defp addresses_labels_query(slug, addresses) do
     query = """
     SELECT lower(address) as address, label, metadata
     FROM blockchain_address_labels FINAL
-    PREWHERE blockchain = 'ethereum' and lower(address) IN (?1) and sign = 1
+    PREWHERE
+      blockchain = 'ethereum' AND
+      asset_id = (SELECT asset_id FROM asset_metadata FINAL PREWHERE name = ?1 LIMIT 1) AND
+      lower(address) IN (?2)
+    HAVING sign = 1
```
Snedashkovsky, line 70: Now only the `whale` label use `asset_id`, I think we need to use this query:
```SQL
SELECT lower(address) as address, label, metadata
FROM blockchain_address_labels FINAL
PREWHERE blockchain = 'ethereum'
     AND ((label = 'whale' and asset_id = (SELECT argMax(asset_id, computed_at) FROM asset_metadata FINAL PREWHERE name = ?1))
          OR (label != 'whale' and asset_id = 0))
     AND lower(address) IN (?2)
HAVING sign = 1;
```

PR #2158, opened 2020-08-07
Add new metrics to API

apozdeyev:
## Changes

Add new metrics to API: network_growth_change_1d/7d/30d

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
+++ b/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
@@ -2236,5 +2236,56 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Network Growth Change (1d)",
+    "name": "network_growth_change_1d",
+    "metric": "network_growth_change_1d",
+    "version": "2019-01-01",
+    "access": "free",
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -2236,5 +2236,56 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Network Growth Change (1d)",
+    "name": "network_growth_change_1d",
+    "metric": "network_growth_change_1d",
+    "version": "2019-01-01",
+    "access": "free",
```
tspenov, line 2245: We need to add access `restricted`. That means that `free` user can't have realtime and all historical data.

apozdeyev: ok, but what about `min_plan`, leave it as is?

tspenov: @apozdeyev yes, `min_plan` means that only this plan or higher can access this metric and lower plans can't even see the restricted data.

apozdeyev: Done!

PR #2160, opened 2020-08-10
Stablecoins holders distribution

IvanIvanoff:
## Changes

Add support for `market_segments` selector in the `getMetric` API.
Currently, it is implemented for the daily_metrics and intraday_metrics and label metric tables (not supported for price and social data). 
This is done to allow implementing holders distributions for the whole stablecoins market segment.

If more than 1 market segment is provided, only the projects that have all of required segments are included.

NOTE: All slugs in the market segments are used in the database query at the same time. This might cause troubles for segments that contain hundreds or thousands of projects.

Example:
```graphql
{
  getMetric(metric: "holders_distribution_0.1_to_1"){
    timeseriesData(
      selector: {market_segments: ["Stablecoin"]}
      from: "2020-07-07T10:00:00Z"
      to: "2020-08-06T10:00:00Z"
      interval: "12h"){
        datetime
        value
      }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
+++ b/lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
@@ -26,8 +26,42 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
     field(:computed_at, :utc_datetime)
   end
 
-  def timeseries_data_query(metric, slugs, from, to, interval, aggregation, filters)
-      when is_list(slugs) do
+  # def timeseries_data_query(metric, slugs, from, to, interval, aggregation, filters)
+  #     when is_list(slugs) do
+  #   query = """
+  #   SELECT
+  #     toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS t,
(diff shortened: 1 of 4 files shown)


lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -42,7 +42,7 @@ defmodule Sanbase.Clickhouse.Metric.SqlQuery do
         #{maybe_convert_to_date(:after, metric, "dt", "toDateTime(?3)")} AND
         #{maybe_convert_to_date(:before, metric, "dt", "toDateTime(?4)")} AND
         NOT isNaN(value) AND
-        asset_id IN ( SELECT asset_id FROM asset_metadata FINAL PREWHERE name IN (?5) LIMIT 1 ) AND
+        #{asset_id_filter(slug_or_slugs)} AND
```
tspenov, line 45: We can make it reusable if we pass the position of the param :?

IvanIvanoff: yeah, it should be

PR #2167, opened 2020-08-12
Add new metrics to API

antisaa:
## Changes

Adding new metrics to API:

- transaction_volume_usd
- percent_of_holders_distribution_amount_1e_3

## Checklist:


lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -2304,5 +2304,22 @@
...
+    "name": "transaction_volume_usd",
+    "metric": "transaction_volume_usd",
+    "version": "2019-01-01",
+    "access": "restricted",
+    "selectors": ["slug"],
+    "min_plan": {
+      "SANAPI": "free",
+      "SANBASE": "free"
+    },
+    "aggregation": "avg",
```
tspenov: This one is already added. The only difference I see is the aggregation - here is `avg` and the already added is `sum`. Which of these is correct ?

PR #2168, opened 2020-08-13
Use assetRefId instead of contract in erc20 transfers sql query

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/erc20_transfers.ex
+++ b/lib/sanbase/clickhouse/erc20_transfers.ex
@@ -81,9 +81,9 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
       value / ?1
     FROM #{@table} FINAL
     PREWHERE
-      contract = ?2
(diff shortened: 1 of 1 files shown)


lib/sanbase/clickhouse/erc20_transfers.ex
```diff
@@ -81,9 +81,9 @@ defmodule Sanbase.Clickhouse.Erc20Transfers do
       value / ?1
     FROM #{@table} FINAL
     PREWHERE
-      contract = ?2
-      AND dt >= toDateTime(?3)
-      AND dt <= toDateTime(?4)
+      assetRefId = cityHash64('ETH_' || ?2) AND
```
tspenov, line 84: do you see any performance difference with this change ?

IvanIvanoff: `contract` is not in the order by list, `assetRefId` is first in the that list.

PR #2177, opened 2020-08-21
Add restriction for sansheets

spiderjako:
## Changes

Adding a restriction to sanbase, so that SanSheets is a PRO-only exclusive.

## Ticket

https://www.notion.so/santiment/9b1939e0ccb9494db3b074a6b74787e0?v=e9a9914a6279477981f54dcb3f011320&p=111707de0dd64073968283f9e5aeed5c

## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/context_plug.ex
+++ b/lib/sanbase_web/graphql/context_plug.ex
@@ -62,6 +62,7 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
       context
       |> Map.put(:remote_ip, conn.remote_ip)
       |> Map.put(:origin_url, Plug.Conn.get_req_header(conn, "origin") |> List.first())
+      |> Map.put(:is_sansheets_request, is_sansheets_request(conn))
 
     conn
     |> put_private(:absinthe, %{context: context})
@@ -292,6 +293,13 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
   end
 
   defp get_no_auth_product_id(_), do: @product_id_api
+
+  defp is_sansheets_request(conn) do
+    case Plug.Conn.get_req_header(conn, "user-agent") do
+      [user_agent] -> String.contains?(user_agent, "Google-Apps-Script")
+      _ -> false
+    end
+  end
 
   defp get_apikey_product_id([user_agent]) do
     case String.contains?(user_agent, "Google-Apps-Script") do
--- a/lib/sanbase_web/graphql/middlewares/access_control.ex
+++ b/lib/sanbase_web/graphql/middlewares/access_control.ex
@@ -34,6 +34,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
     # First call `check_from_to_params` and then pass the execution to do_call/2
     resolution
     |> transform_resolution()
+    |> restrict_if_sansheets()
     |> check_plan()
     |> check_from_to_params()
     |> do_call(opts)
@@ -290,6 +291,23 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
   end
 
   defp check_from_to_both_outside(%Resolution{} = resolution), do: resolution
+
+  # Requests from sansheets return `unauthorized` for free users.
+  defp restrict_if_sansheets(
+         %Resolution{
+           context: %{
+             auth: %{subscription: %{plan: %{name: "FREE"}}},
+             is_sansheets_request: true
+           }
+         } = resolution
+       ) do
+    Resolution.put_result(
+      resolution,
+      {:error, "You need to upgrade to Sanbase Pro in order to use SanSheets."}
+    )
+  end
+
+  defp restrict_if_sansheets(resolution), do: resolution
 
   defp update_resolution_from_to(
          %Resolution{arguments: %{from: _from, to: _to} = args} = resolution,
--- a/test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
+++ b/test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
@@ -3,7 +3,9 @@ defmodule SanbaseWeb.Graphql.AccessControlMiddlewareTest do
 
   import Sanbase.Factory
   import SanbaseWeb.Graphql.TestHelpers
-  import Sanbase.DateTimeUtils, only: [from_iso8601!: 1]
+  import Sanbase.DateTimeUtils, only: [from_iso8601_to_unix!: 1, from_iso8601!: 1]
+
+  alias Sanbase.Auth.Apikey
 
   setup do
     contract = "0x132123"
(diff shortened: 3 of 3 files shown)


lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -217,6 +217,13 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
         Subscription.current_subscription(current_user, product_id) ||
           @free_subscription
 
+      product_is_sansheets =
```
tspenov, line 220: That is not correct - it also could be just sanbase product.
You need something - `is_sansheets_request` and it should be true if user-agent contains sheets user agent.

tspenov: Also for future use shorter version is just `product_is_sansheets:  product_id == @product_id_sanbase`

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -37,6 +37,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
     |> check_plan()
     |> check_from_to_params()
     |> do_call(opts)
+    |> check_if_product_is_sansheets()
```
tspenov, line 40: That should be before an `check_plan`

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -291,6 +292,27 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
...
+             auth: %{subscription: subscription},
+             product_is_sansheets: product_is_sansheets
+           }
+         } = resolution
+       ) do
+    %{plan: %{name: name}} = subscription
+
+    case {name, product_is_sansheets} do
+      {"FREE", true} ->
+        Resolution.put_result(resolution, {:error, "Upgrade to Pro in order to use SanSheets."})
```
tspenov, line 307: `You need to upgrade to Sanbase Pro in order to use SanSheets.`

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -291,6 +292,27 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
 
   defp check_from_to_both_outside(%Resolution{} = resolution), do: resolution
 
+  defp check_if_product_is_sansheets(
```
tspenov, line 295:
```elixir
# Requests from sansheets return `unauthorized` for free users.
defp check_sansheets_request( %Resolution{
           context: %{
             auth: %{subscription: %{plan: %{name: "FREE"}}},
             is_sansheets_request:  true
           }
         } = resolution
       ) do
Resolution.put_result(resolution, {:error,  message})
end

defp check_sansheets_request(resolution), do: resolution
```

tspenov: You need also to check - whether we always put `FREE` plan subscription in context plug

test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
```diff
@@ -106,4 +108,37 @@ defmodule SanbaseWeb.Graphql.AccessControlMiddlewareTest do
     assert error ==
              "Cryptocurrencies didn't exist before 2009-01-01 00:00:00Z.\nPlease check `from` and/or `to` parameters values.\n"
   end
+
+  test "returns error when sansheets user is not Pro", context do
+    user = insert(:user)
+    {:ok, apikey} = Apikey.generate_apikey(user)
```
tspenov, line 114: Here we have test only with apikey. Is it possible to use sansheets without apikey as a free user ?
In both cases there should be a test for this.

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -217,6 +217,13 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
...
+        if product_id == @product_id_sanbase do
+          true
+        else
+          false
+        end
```
IvanIvanoff, line 225: This is the equivalent of 
```elixir
if true do
  true
else
  false
end
```

You should just do `product_is_sansheets  = (product_id == @product_id_sanbase)`

lib/sanbase_web/graphql/middlewares/access_control.ex
IvanIvanoff: lets rename to `restrict_if_sansheets` - `check_if` does not imply the action of the function

IvanIvanoff:
@tspenov @spiderjako I suggest we extend the `get_apikey_product_id` function to return a map or a tuple:
```elixir
 defp get_apikey_product_id([user_agent]) do
    case String.contains?(user_agent, "Google-Apps-Script") do
      true -> %{product_id: @product_id_sanbase, is_sansheets_request: true}
      false -> %{product_id: @product_id_api, is_sansheets_request: false}
    end
  end
```
and use that

lib/sanbase_web/graphql/middlewares/access_control.ex
IvanIvanoff:
```
tspenov 3 hours ago Member
You need also to check - whether we always put FREE plan subscription in context plug
```
We use `anon_user_base_context` for anon users in the context plug, so in all cases there should be a subscription to check

IvanIvanoff: One more thing - you want to check not only the plan, but also the product. I am not entirely sure if only sanbase pro has access to sansheets, or sanapi/sandata has too.

test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
spiderjako: Yes, it could've been used, there should def be a test for this, thanks

test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
```diff
@@ -106,4 +108,69 @@ defmodule SanbaseWeb.Graphql.AccessControlMiddlewareTest do
...
+    }
+    """
+
+    result =
+      conn
+      |> post("/graphql", query_skeleton(query))
+
+    error = List.first(json_response(result, 200)["errors"])["message"]
+
+    assert error == "You need to upgrade to Sanbase Pro in order to use SanSheets."
```
IvanIvanoff, line 174: It's easier to use the `json_response/2` function directly when building `result` as you might want to reuse it later.

```suggestion
    result =
      conn
      |> post("/graphql", query_skeleton(query))
      |> json_response(200)
     
    error_msg = result["errors"] |> hd() |> Map.get("message")
    assert error_msg == "You need to upgrade to Sanbase Pro in order to use SanSheets."
```

test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
```diff
@@ -106,4 +108,69 @@ defmodule SanbaseWeb.Graphql.AccessControlMiddlewareTest do
...
+      transactionVolume(
+        slug: "santiment",
+        from: "#{Timex.shift(Timex.now(), days: -10)}",
+        to: "#{Timex.now()}",
+        interval: "1d") {
+          datetime
+          transactionVolume
+      }
+    }
+    """
```
IvanIvanoff, line 166: This api is deprecated, let's not use it in new tests.

spiderjako: Okay

lib/sanbase_web/graphql/middlewares/access_control.ex
tspenov: Lets make it only for `Sanbase PRO` cause I think that is what is visible everywhere and what business expects. But also I  don't might making it accessible also for not free API plans. I think both should be fine.

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -58,10 +58,17 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
...
+      case Plug.Conn.get_req_header(conn, "user-agent") do
+        [user_agent] -> String.contains?(user_agent, "Google-Apps-Script")
+        _ -> false
+      end
+
     context =
       context
       |> Map.put(:remote_ip, conn.remote_ip)
       |> Map.put(:origin_url, Plug.Conn.get_req_header(conn, "origin") |> List.first())
+      |> Map.put(:is_sansheets_request, is_sansheets_request)
```
IvanIvanoff, line 71: Let's make this a private function and just call it here: `|> Map.put(:is_sansheets_request, is_sansheets_request(conn))`

IvanIvanoff (approved):

tspenov (approved):

PR #2186, opened 2020-08-26
Properly restrict all API calls coming from sansheets

IvanIvanoff:
## Checklist:


tspenov:
@IvanIvanoff Why do you change and halt in `context_plug` ? I think `AccessControl` was the right place to handle this business logic. It is true that halting in the plug is faster than in middlewares/resolvers but IMO we should halt there only based on wrong things with request not on other logic.

spiderjako (approved):

IvanIvanoff:
@tspenov The `AccessControl` won't apply to all queries - for free queries/mutations without from-to we do not apply the access control middleware. We'll proceed with this change as it's needed on prod but it can be placed in a more appropriate place after that.

PR #2194, opened 2020-08-31
Add labeled metrics

ninjascant:
## Changes

Add new label-based metrics to Sanbase

## Checklist:


IvanIvanoff:
are these really labeled metrics? if yes - they should have the proper selectors list. also why aren't they in the label tables?

Snedashkovsky:
@IvanIvanoff These metrics are based on labels, but depend on several metrics at once. Therefore, it is not possible to put them in the `intraday_label_based_metrics` table.
Similar metrics were already in the `intraday_metrics` table.

IvanIvanoff (approved):

PR #2195, opened 2020-08-31
Adding a check whether the label is main or not

spiderjako:
## Checklist:

### DIFF
--- a/lib/sanbase/model/project/project.ex
+++ b/lib/sanbase/model/project/project.ex
@@ -376,6 +376,10 @@ defmodule Sanbase.Model.Project do
     end
   end
 
+  def has_main_contract_addresses(project) do
+    project.contract_addresses |> Enum.findby(&(&1.label == "main"))
+  end
+
   defp preload_query(query, opts) do
     case Keyword.get(opts, :preload?, true) do
       false ->
--- a/lib/sanbase/model/project/project.ex
+++ b/lib/sanbase/model/project/project.ex
@@ -376,8 +376,17 @@ defmodule Sanbase.Model.Project do
     end
   end
 
-  def has_main_contract_addresses(project) do
-    project.contract_addresses |> Enum.findby(&(&1.label == "main"))
+  def has_main_contract_addresses?(project) do
+    preloaded_project =
+      project
+      |> Repo.preload([:contract_addresses])
+
+    preloaded_project.contract_addresses
+    |> Enum.find(&(&1.label == "main"))
+    |> case do
+      nil -> false
+      _ -> true
+    end
   end
 
   defp preload_query(query, opts) do
--- a/lib/sanbase/model/project/project.ex
+++ b/lib/sanbase/model/project/project.ex
@@ -377,11 +377,9 @@ defmodule Sanbase.Model.Project do
   end
 
   def has_main_contract_addresses?(project) do
-    preloaded_project =
-      project
-      |> Repo.preload([:contract_addresses])
-
-    preloaded_project.contract_addresses
+    project
+    |> Repo.preload([:contract_addresses])
+    |> Map.get(:contract_addresses)
     |> Enum.find(&(&1.label == "main"))
     |> case do
       nil -> false
--- a/test/sanbase/external_services/coinmarketcap/scraper_test.exs
+++ b/test/sanbase/external_services/coinmarketcap/scraper_test.exs
@@ -25,22 +25,34 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.ScraperTest do
 
   test "scraping a project with a main contract doesn't make the new contract main" do
     main_contract = build(:contract_address, label: "main")
-    ordinary_contract = build(:contract_address)
-    third_contract = build(:contract_address)
+    second_contract = build(:contract_address)
 
     project =
       insert(:random_project,
-        contract_addresses: [main_contract, ordinary_contract]
+        contract_addresses: [main_contract]
       )
 
     project_info_map = %ProjectInfo{
-      slug: "santiment",
-      main_contract_address: third_contract.address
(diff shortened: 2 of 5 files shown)


lib/sanbase/model/project/project.ex
```diff
@@ -376,6 +376,10 @@ defmodule Sanbase.Model.Project do
     end
   end
 
+  def has_main_contract_addresses(project) do
```
IvanIvanoff, line 379:
```suggestion
  def has_main_contract_addresses?(project) do
```

lib/sanbase/model/project/project.ex
```diff
@@ -376,6 +376,10 @@ defmodule Sanbase.Model.Project do
...
+    project.contract_addresses |> Enum.findby(&(&1.label == "main"))
```
IvanIvanoff, line 380: You have to do it like this:
```
project
|> Repo.preload([:conract_addresses])
...
```
This is no-op when the contracts are preloaded, but makes it work when they're not, too.

IvanIvanoff: Also what is `Enum.findby`? There's no` Enum.findby` and there's neither `Enum.find_by`.
Lastly, you want this function to return `true/false` (answers the question has?) and not `nil` or the contract address element.

tspenov:
@spiderjako please include a test case for this change

lib/sanbase/model/project/project.ex
```diff
@@ -376,6 +376,19 @@ defmodule Sanbase.Model.Project do
     end
   end
 
+  def has_main_contract_addresses?(project) do
+    project
+    |> IO.inspect()
```
IvanIvanoff, line 381: inspect left

lib/sanbase/model/project/project.ex
```diff
@@ -376,6 +376,19 @@ defmodule Sanbase.Model.Project do
...
+    |> Repo.preload([:contract_addresses])
```
IvanIvanoff, line 382: Also think what this code is actually doing and why it's not correct:
```elixir
project
|> |> Repo.preload([:contract_addresses])
```

test/sanbase/external_services/project_info_test.exs
```diff
@@ -137,7 +137,12 @@ defmodule Sanbase.ExternalServices.ProjectInfoTest do
     |> Repo.insert!()
 
     project =
-      %Project{slug: "santiment", name: "Santiment", ticker: "OLD_TICKR"}
+      %Project{
+        slug: "santiment",
+        name: "Santiment",
+        ticker: "OLD_TICKR",
+        contract_addresses: ["0x0000000", "0x0000001"]
+      }
```
IvanIvanoff, line 145: Use the factory to create a project with 2 contracts

IvanIvanoff: Hm, this test is for insertion of tags, not related to this PR.
You should add a test to the scraper tests where:
1. First insert a project and a `main` contract for it
2. Run the scraper
3. Assert the new contract is not labeled as main

lib/sanbase/model/project/project.ex
```diff
@@ -376,6 +376,19 @@ defmodule Sanbase.Model.Project do
     end
   end
 
+  def has_main_contract_addresses?(project) do
+    preloaded_project =
+      project
+      |> Repo.preload([:contract_addresses])
+
+    preloaded_project.contract_addresses
+    |> Enum.find(&(&1.label == "main"))
```
IvanIvanoff, line 385:
```suggestion
    project
    |> Repo.preload([:contract_addresses])
    |> Map.get(:contract_addresses)
    |> Enum.find(&(&1.label == "main"))
```

test/sanbase/external_services/coinmarketcap/scraper_test.exs
```diff
@@ -19,4 +22,25 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.ScraperTest do
...
+    ordinary_contract = build(:contract_address)
+    third_contract = build(:contract_address)
+
+    project =
+      insert(:random_project,
+        contract_addresses: [main_contract, ordinary_contract]
+      )
+
+    project_info_map = %ProjectInfo{
+      slug: "santiment",
```
tspenov, line 37: here `santiment` slug is used whereas above a random project is build.

test/sanbase/external_services/coinmarketcap/scraper_test.exs
```diff
@@ -19,4 +22,25 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.ScraperTest do
...
+      main_contract_address: third_contract.address
+    }
+
+    File.read!(Path.join(__DIR__, "data/project_page.html"))
+    |> Scraper.parse_project_page(project_info_map)
```
tspenov, line 42: `parse_project_page` only updates `%ProjectInfo` struct - it doe not update the project and its contract addresses

test/sanbase/external_services/coinmarketcap/scraper_test.exs
```diff
@@ -19,4 +22,25 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.ScraperTest do
...
+
+    assert Enum.find(project.contract_addresses, &(&1.address == third_contract.label != "main"))
```
tspenov, line 44: I've never seen that  x==y==z used but seems not correct - `address` is compared with `label`and with string `main`. I guess the correct should be `&1.address == third_contract.address and &1.label != "main"`

tspenov (commented):
One of the tests is not testing the behaviour correctly. I see that next text is almost the same so it also should be fixed.

IvanIvanoff (approved):

PR #2196, opened 2020-09-01
Add sent signals per day limit for users

IvanIvanoff:
## Changes
Limit the number of signal notifications one user can receive per calendar day. This is done to restrict an accidental spammy signal. The number of notifications is configured in the user settings and can be changed.
API for changing:
```graphql
mutation{
  updateUserSettings(settings: {signalsPerDayLimit: "{\"email\": 100, \"telegram\": 200}"}){
    signalsPerDay
  }
}
```

## Checklist:

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -34,8 +34,7 @@ config :sanbase, Sanbase.ExternalServices.RateLimiting.Server,
 config :sanbase, Sanbase.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   database: "sanbase_test",
-  pool_size: 5,
-  ownership_timeout: 200_000
+  pool_size: 5
 
 config :sanbase, Sanbase.ClickhouseRepo,
   pool: Ecto.Adapters.SQL.Sandbox,
--- a/lib/sanbase/signals/evaluator/scheduler.ex
+++ b/lib/sanbase/signals/evaluator/scheduler.ex
@@ -82,11 +82,9 @@ defmodule Sanbase.Signal.Scheduler do
             {user_trigger, []}
 
           list when is_list(list) ->
-            {:ok,
-             %{last_triggered: last_triggered, total_sent_succesfully: total_sent_succesfully}} =
+            {:ok, %{last_triggered: last_triggered}} =
               handle_send_results_list(user_trigger, list)
 
-            update_user_signals_sent_per_day(user_trigger.user, total_sent_succesfully)
             user_trigger = update_trigger_last_triggered(user_trigger, last_triggered)
             {user_trigger, list}
         end
@@ -96,12 +94,6 @@ defmodule Sanbase.Signal.Scheduler do
       map_type: :map
     )
     |> Enum.unzip()
-  end
-
-  defp update_user_signals_sent_per_day(user, count) do
-    %{signals_fired: signals_fired} = Sanbase.Auth.UserSettings.settings_for(user)
-    signals_fired = Map.update(signals_fired, Date.utc_today(), count, &(&1 + count))
-    Sanbase.Auth.UserSettings.update_settings(user, %{signals_fired: signals_fired})
   end
 
   # Note that the `user_trigger` that came as an argument is returned with
@@ -214,18 +206,18 @@ defmodule Sanbase.Signal.Scheduler do
   end
 
   defp log_sent_messages_stats([], type) do
-    Logger.info("There were no signals triggered of type #{type}")
+    Logger.info("There were no #{type} signals triggered.")
   end
 
   defp log_sent_messages_stats(list, type) do
-    successful_messages = list |> Enum.count(fn {_elem, status} -> status == :ok end)
+    successful_messages_count = list |> Enum.count(fn {_elem, status} -> status == :ok end)
 
     for {_, {:error, error}} <- list do
-      Logger.warn("Cannot send a signal. Reason: #{inspect(error)}")
+      Logger.warn("Cannot send a #{type} signal. Reason: #{inspect(error)}")
     end
 
     Logger.info(
-      "In total #{successful_messages}/#{length(list)} #{type} signals were sent successfully"
+      "In total #{successful_messages_count}/#{length(list)} #{type} signals were sent successfully."
     )
   end
 
--- a/lib/sanbase/signals/evaluator/scheduler.ex
+++ b/lib/sanbase/signals/evaluator/scheduler.ex
@@ -20,6 +20,20 @@ defmodule Sanbase.Signal.Scheduler do
 
   defguard is_non_empty_map(map) when is_map(map) and map != %{}
 
+  @doc ~s"""
+  Processfor all active signals with the given type. The processing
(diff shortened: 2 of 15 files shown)


config/test.exs
```diff
@@ -34,7 +34,8 @@ config :sanbase, Sanbase.ExternalServices.RateLimiting.Server,
 config :sanbase, Sanbase.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   database: "sanbase_test",
-  pool_size: 5
+  pool_size: 5,
+  ownership_timeout: 200_000
```
tspenov, line 38: Do we still need this ?

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -20,6 +20,20 @@ defmodule Sanbase.Signal.Scheduler do
 
   defguard is_non_empty_map(map) when is_map(map) and map != %{}
 
+  @doc ~s"""
+  Processfor all active signals with the given type. The processing
```
tspenov, line 24:
```suggestion
  Process for all active signals with the given type. The processing
```

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -90,12 +96,25 @@ defmodule Sanbase.Signal.Scheduler do
...
   end
 
-  defp update_triggered(
-         %{
-           user: user,
-           id: trigger_id,
-           trigger: %{last_triggered: last_triggered, settings: settings}
-         },
+  # Note that the `user_trigger` that came as an argument is returned with
+  # modofied `last_triggered`.
```
tspenov, line 100:
```suggestion
  # modified `last_triggered`.
```

lib/sanbase/signals/signal.ex
```diff
@@ -3,123 +3,105 @@ defprotocol Sanbase.Signal do
...
+           trigger: %{settings: %{payload: payload_map}}
+         },
+         max_signals_to_send
+       )
+       when is_binary(email) and is_map(payload_map) do
+    fun = fn _identifier, payload ->
+      do_send_email(email, payload, id)
+    end
+
+    send_or_limit(payload_map, max_signals_to_send, fun)
```
tspenov, line 69: @IvanIvanoff I can't find where you send email after limit is reached

IvanIvanoff: I don't currently. The proper behaviour would be to track for every user the channels used and send the limit reached notification to all channels. The reason behind that is simple - if a telegram message hits the limit, you would also want to receive an email for that if you're receiving signals there, too.

IvanIvanoff: On a second thought, the better behaviour would be to track the channels limits separately and just send the notification to whatever channel triggered it. The current behaviour has the limitation that if you enable 2 channels for 1 signal it would bump the count by 2

config/test.exs
IvanIvanoff: Nope

lib/sanbase/signals/signal.ex
```diff
@@ -139,4 +154,131 @@ defimpl Sanbase.Signal, for: Any do
...
+
+    signals_fired_now =
+      Enum.into(sent_list_per_channel, %{}, fn {channel, sent_list} ->
+        count = Enum.count(sent_list, fn {_, result} -> result == :ok end)
+        {channel, count}
+      end)
+
+    signals_fired_today = Map.get(signals_fired, map_key, %{})
+
+    maybe_send_limit_reached_notification(
```
tspenov, line 215: I am wondering why notification is sent here . We have `send_or_limit` before that is executed for each channel before that. What is the idea here ?

lib/sanbase/signals/signal.ex
```diff
@@ -130,6 +122,29 @@ defimpl Sanbase.Signal, for: Any do
...
+    %{
+      signals_fired: signals_fired,
+      signals_per_day_limit: signals_per_day_limit
+    } = user_settings
+
+    # A map of "channel" => list pairs
+    notifications_sent_today = Map.get(signals_fired, Date.utc_today() |> to_string(), %{})
+
+    Enum.reduce(@channels, %{}, fn channel, map ->
+      channel_limit = Map.get(signals_per_day_limit, channel, @default_limit_per_day)
```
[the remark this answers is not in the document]
IvanIvanoff: yes, it does

lib/sanbase/signals/signal.ex
```diff
@@ -139,4 +154,131 @@ defimpl Sanbase.Signal, for: Any do
...
+          remaining ->
+            elem = {identifier, fun.(identifier, payload)}
+            {[elem | list], remaining - 1}
+        end
+      end)
+
+    result
+  end
+
+  defp update_user_signals_sent_per_day(user, sent_list_per_channel) do
```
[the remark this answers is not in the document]
IvanIvanoff: I should rename the function as it is checking for limits and sends notifications

lib/sanbase/signals/signal.ex
```diff
@@ -3,123 +3,115 @@ defprotocol Sanbase.Signal do
...
+    # updated. This can happen because the sending of signals happens with a
+    # concurrency of 20, so 2+ processes can be sending notifications to a user
+    # at the same time and compute the same `max_signals_to_send` which would
+    # lead to exceeded the given limit. Without the lock the notification for
+    # exceeded the limit can be sent more than once as well.
+    lock = Mutex.await(Sanbase.SignalMutex, {:user, user.id}, 30_000)
+
+    max_signals_to_send = max_signals_to_send(user_trigger)
+
+    # Returns a list of 2-element tuples where the first element is the channel
```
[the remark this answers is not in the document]
IvanIvanoff: We don't have a public function to typespec here

lib/sanbase/signals/signal.ex
IvanIvanoff: The `send_or_limit` does not know about channels. I haven't put any extra effort into splitting these functions - maybe just don't overload one function with doing more than one thing (send signal vs send limit reach notification)

IvanIvanoff: This one does not sit very pretty here indeed though - I can think how to make it more tidy.

lib/sanbase/signals/signal.ex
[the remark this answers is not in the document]
IvanIvanoff: PS: Now I remember - it must be here so it's inside the mutex, otherwise you might run into race conditions fetching and updating the number of fired signals per user.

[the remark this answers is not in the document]
IvanIvanoff: Removed the sending of limit reached notifications from here, so the name is now properly describing the function.

lib/sanbase/signals/signal.ex
IvanIvanoff: Done.

tspenov (approved):

PR #2206, opened 2020-09-04
BTC Top Transactions

tspenov:
## Changes
Add bitcoin top transactions.
**Note: `fromAddress` for BTC top transactions will be `null`**

## Checklist:

## Usage

```graphql
{
  projectBySlug(slug:"bitcoin") {
    id
    tokenTopTransactions(
      from: "2020-08-11T19:00:00.000Z",
      to:"2020-09-11T18:59:59.999Z",
      limit:2
    ) {
      datetime
      fromAddress {
        address
        isExchange
        labels {
          name
          metadata
        }
      }
      toAddress {
        address
        isExchange
        labels {
          name
          metadata
        }
      }
      trxHash
      trxValue
    }
  }
}
```

## Screenshots

image

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
+++ b/lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
@@ -126,4 +126,27 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.UtxoSqlQueries do
     args = [interval, span, addresses, from_unix, to_unix]
     {query, args}
   end
+
+  # type: :received | :send
+  def btc_top_transactions_query(from, to, limit, opts \\ []) do
+    to_unix = DateTime.to_unix(to)
+    from_unix = DateTime.to_unix(from)
+
+    balance_change_expr =
+      case Keyword.get(opts, :type, :funds_recv) do
+        :funds_recv -> "balance - oldBalance"
+        :funds_sent -> "-(balance - oldBalance)"
+      end
+
+    query = """
+    SELECT toUnixTimestamp(dt), address, #{balance_change_expr} as amount, txID
+    FROM btc_balances
+    PREWHERE  amount > 1000 AND dt >= toDateTime(?1) AND dt < toDateTime(?2)
+    ORDER BY amount DESC
+    LIMIT ?3
+    """
+
+    args = [from_unix, to_unix, limit]
+    {query, args}
+  end
 end
--- a/lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
+++ b/lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
@@ -131,12 +131,12 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.UtxoSqlQueries do
     to_unix = DateTime.to_unix(to)
     from_unix = DateTime.to_unix(from)
 
-    # only > 500 BTC transfers if range is > 1 week, otherwise only bigger than 100
-    amount_filter = if Timex.diff(to, from, :days) > 7, do: 500, else: 100
+    # only > 100 BTC transfers if range is > 1 week, otherwise only bigger than 20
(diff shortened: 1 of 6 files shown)


lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
```diff
@@ -126,4 +126,23 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.UtxoSqlQueries do
     args = [interval, span, addresses, from_unix, to_unix]
     {query, args}
   end
+
+  def btc_top_transactions_query(from, to, limit) do
+    to_unix = DateTime.to_unix(to)
+    from_unix = DateTime.to_unix(from)
+
+    # only > 500 BTC transfers if range is > 1 week, otherwise only bigger than 100
```
IvanIvanoff, line 134: 500/100 BTC is a huge transaction (over 5/1 million dollars at the current price) - it should be decreased imo

lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
```diff
@@ -126,4 +126,23 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.UtxoSqlQueries do
...
+    amount_filter = if Timex.diff(to, from, :days) > 7, do: 500, else: 100
+
+    query = """
+    SELECT toUnixTimestamp(dt), address, balance - oldBalance as amount, txID
+    FROM btc_balances
```
IvanIvanoff, line 139:
```suggestion
    FROM btc_balances FINAL
```
to avoid duplicates :?

lib/sanbase/clickhouse/labels.ex
```diff
@@ -51,17 +51,28 @@ defmodule Sanbase.Clickhouse.Label do
...
+    HAVING sign = 1
+    """
+
+    {query, [addresses]}
+  end
+
   defp addresses_labels_query(slug, addresses) do
     query = """
-    SELECT lower(address) as address, label, metadata
+    SELECT address as address, label, metadata
```
IvanIvanoff, line 75: This would make it work for BTC but break it for ETH as we have issues with different casing (in some tables they are lowercased and in others they are not)

lib/sanbase_web/graphql/schema/types/transaction_types.ex
```diff
@@ -8,7 +8,7 @@ defmodule SanbaseWeb.Graphql.TransactionTypes do
   end
 
   object :address do
-    field(:address, non_null(:string))
+    field(:address, :string)
     field(:is_exchange, non_null(:boolean))
     field(:labels, non_null(list_of(:label)), default_value: [])
```
IvanIvanoff, line 13: Is this correct - `address` can be null but the `is_exchange` and label cannot?

lib/sanbase/clickhouse/labels.ex
tspenov: Yeah, good catch. I introduced separate query for bitcoin where it makes different query but forgot to bring `lower` back the eth/erc20 one

lib/sanbase_web/graphql/schema/types/transaction_types.ex
tspenov: Address can be null in the bitcoin case. `label` has default value and `is_exchange` - it is `false` by default also though not in the graphql schema.

lib/sanbase/clickhouse/historical_balance/fetchers/utxo_sql_queries.ex
tspenov: Looks good with the current data. But I can relax further to 100/20 BTC to be safe, though it will also probably slow down the query.

lib/sanbase_web/graphql/schema/types/transaction_types.ex
tspenov: make the parent `from_address` object nullable.

PR #2211, opened 2020-09-08
Add short URL graphql API

IvanIvanoff:
## Changes

- Add API for creating and fetching short urls:
```graphql
mutation { createShortUrl(fullUrl: "full url goes here") }
```
```graphql
{ getFullUrl(shortUrl: "short url goes here") }
```

## Checklist:


priv/repo/migrations/20200908092849_add_short_urls_table.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Sanbase.Repo.Migrations.AddShortUrlsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:short_urls) do
+      add(:short_url, :string, null: false)
+      add(:full_url, :text, null: false)
```
tspenov, line 7: I think we should we also add an optional user - if url is created from logged in user.
Also some counter for analytics - how many times this url is used.

lib/sanbase/short_url/short_url.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Sanbase.ShortUrl do
+  use Ecto.Schema
+
+  import Ecto.Query
+  import Ecto.Changeset
+
+  alias Sanbase.Repo
+
+  @short_url_length 8
```
tspenov, line 9: I think 4 chars should be enough = 64 ^ 4 ~ 16 mil

priv/repo/migrations/20200908092849_add_short_urls_table.exs
```diff
@@ -0,0 +1,12 @@
...
+    end
```
tspenov, line 8: Missing create timestamp

priv/repo/structure.sql
```diff
@@ -4827,3 +4879,5 @@ INSERT INTO public."schema_migrations" (version) VALUES (20200728105033);
 INSERT INTO public."schema_migrations" (version) VALUES (20200804093238);
 INSERT INTO public."schema_migrations" (version) VALUES (20200813141704);
 INSERT INTO public."schema_migrations" (version) VALUES (20200826101751);
+INSERT INTO public."schema_migrations" (version) VALUES (20200826114101);
```
tspenov, line 4882: this migration is missing

lib/sanbase_web/graphql/schema/queries/short_url_queries.ex
```diff
@@ -0,0 +1,27 @@
...
+    field :get_full_url, :string do
+      meta(access: :free)
+      arg(:short_url, non_null(:string))
+
+      resolve(&ShortUrlResolver.get_full_url/3)
+    end
+  end
+
+  object :short_url_mutations do
+    @desc "Change the user's avatar."
```
tspenov, line 20: description is not for this mutation

lib/sanbase/short_url/short_url.ex
IvanIvanoff: The collision chance with 4 is too high. You can test by running the following code in iex:
```elixir
 (fn ->
             Stream.unfold(0, &{&1, &1 + 1})
             |> Enum.reduce(MapSet.new(), fn x, acc ->
               short_url =
                 :crypto.strong_rand_bytes(4)
                 |> Base.url_encode64(padding: false)
                 |> binary_part(0, 4)

               case short_url in acc do
                 true ->
                   IO.inspect("FOUND COLLISION ON #{x}")

                 false ->
                   :ok
               end

               MapSet.put(acc, short_url)
             end)
           end).()
```

You'll find out that with time, the collisions will start to appear very often:
image

I have run the same test with 8 (tinyurl also uses 8) and it did not result in any collisions for a long time (I stopped it around ~20-30 million).

priv/repo/structure.sql
IvanIvanoff: https://github.com/santiment/sanbase2/blob/master/priv/repo/migrations/20200826114101_fill_market_segments_type.exs

IvanIvanoff: It was missing from the structure.sql file, `ecto.dump` has properly added it here.

lib/sanbase/short_url/short_url.ex
tspenov: I don't think we'll have any collisions in practice. As I said we have ~ 16mil diff strings with 64 chars or ~14mil with 62.
I doubt we will have more than 10000 links for this internal service. If we want to be very conservative we can put 5 chars - then we will have 916 mil. Also we can always retry when hitting the uniq constraint.

IvanIvanoff: One of the circulating ideas is to constantly generate short URLs and never store page settings in the URL itself (always show the short url). In this case we would have a lot of urls. The difference between 4 and 8 characters is so small that I don't think it's worth risking and trying to optimize further. Opening a chart right now leads to URLs with length of ~600 and bringing this down to ~40 (https://app.santiment.net/short/12345678) is already a huge improvement.

tspenov: ok, lets go with this approach 👍

tspenov (approved):

PR #2218, opened 2020-09-09
Add preview fields for reports for non pro users

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/report_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/report_resolver.ex
@@ -24,6 +24,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ReportResolver do
     {:ok, reports}
   end
 
+  def get_reports(_root, _args, _resolution) do
+    {:ok, Report.get_published_reports(nil)}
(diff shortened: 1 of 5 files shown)


lib/sanbase_web/graphql/resolvers/report_resolver.ex
```diff
@@ -24,6 +24,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.ReportResolver do
     {:ok, reports}
   end
 
+  def get_reports(_root, _args, _resolution) do
+    {:ok, Report.get_published_reports(nil)}
```
IvanIvanoff, line 28: I would use `:free` - much clearer than `nil`

tspenov: I have to distinguish - :free logged in user, from :free not logged in. The free logged in will have access to url for reports that are not `pro`, whereas the non logged in won't have.

tspenov: If we show urls to both groups - which is possible also. Then using just :free will be ok.

PR #2263, opened 2020-09-28
Add balance to the top claimers

IvanIvanoff:
## Changes
Extend the uniswap top claimers with current balance field:
```graphql
{
getMetric(metric: "uniswap_top_claimers"){
  histogramData(
    selector: {slug: "uniswap"}
    from: "2020-09-18T00:00:00Z"
    to:"2020-09-19T00:00:00Z"
    limit: 10){
      values {
        ... on StringAddressFloatValueList{
          data{
            address
            value
            balance
            labels
          }
        }
    }
  }
}}
```

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/uniswap/metric_adapter.ex
+++ b/lib/sanbase/clickhouse/uniswap/metric_adapter.ex
@@ -18,6 +18,8 @@ defmodule Sanbase.Clickhouse.Uniswap.MetricAdapter do
 
   @default_complexity_weight 0.3
 
+  @contract "0x1f9840a85d5af5bf1d1762f925bdaddc4201f984"
+
   @impl Sanbase.Metric.Behaviour
   def has_incomplete_data?(_), do: false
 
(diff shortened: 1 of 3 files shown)


lib/sanbase/clickhouse/uniswap/metric_adapter.ex
```diff
@@ -44,7 +46,7 @@ defmodule Sanbase.Clickhouse.Uniswap.MetricAdapter do
       dt >= toDateTime(?1) AND
       dt < toDateTime(?2) AND
       from = '0x090d4613473dee047c3f2706764f49e0821d256e' AND
```
spiderjako, line 48: Can't this 'from' variable be taken out as a const and not hardcoded, like 'contract' below?

IvanIvanoff: i've left it as is because `from` is not reused but the contract is

PR #2270, opened 2020-10-01
Improve the speed of erc20 transers queries

IvanIvanoff:
## Changes
Add usage of erc20 transfers table ordered by dt in some of the queries.
Switch from contract to assetRefId as it is first in the order by, while contract is not present.

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/fetchers/erc20_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/fetchers/erc20_balance.ex
@@ -108,7 +108,7 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.Erc20Balance do
     FROM #{@table} FINAL
     PREWHERE
       address IN (?1) AND
-      contract = ?2
+      assetRefId = cityHash64('ETH_' || ?2)
     GROUP BY address
     """
 
@@ -128,7 +128,7 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.Erc20Balance do
     FROM #{@table}
     PREWHERE
       address IN (?1) AND
-      contract = ?2 AND
+      assetRefId = cityHash64('ETH_' || ?2) AND
       dt >= toDateTime(?3) AND
       dt < toDateTime(?4) AND
       sign = 1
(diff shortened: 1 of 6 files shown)


tspenov:
@IvanIvanoff when should we use address ordered and when dt ordered ?

IvanIvanoff:
@tspenov address order is used when the where clause is `from = ...` - this is used for computing metrics for a single address or list of addresses. 

dt order is used when the where clause is `dt >= ... and dt < ...` - this is used for computing metrics for the whole contract/asset - for example the top transactions. 

In case the where clause has both `from` and `dt` restrictions I've chosen the address order. I think this way more data is cut off when reading.

lib/sanbase/clickhouse/historical_balance/fetchers/erc20_balance.ex
```diff
@@ -108,7 +108,7 @@ defmodule Sanbase.Clickhouse.HistoricalBalance.Erc20Balance do
     FROM #{@table} FINAL
     PREWHERE
       address IN (?1) AND
-      contract = ?2
+      assetRefId = cityHash64('ETH_' || ?2)
```
tspenov, line 111: is assetRefId also first in `erc20_balances`. From what I see here: https://github.com/santiment/clickhouse-tables/blob/6ecfd3374ba908dae929f689d7168066629086cd/db/20190219164136_create_erc20_balances.sql#L32 the `contract` is in first place - though not sure it is the current create statement.

IvanIvanoff: This is a very old and no longer (I suspect) used migration to create the tables. Now they're defined with yaml and a k8s operator. But indeed it is still `contract` first. 

PS: I do such checks by executing `select * from system.tables where name = 'erc20_balances_shard_v10'\G` (or whatever the proper shard name is) and it shows the currently running version.

IvanIvanoff: It should be assetRefId in the first place for consistency. Let's wait until it is properly switched before merging this PR.

PR #2297, opened 2020-10-15
Replace zero address eth/erc20

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/utils/transform.ex
+++ b/lib/sanbase/utils/transform.ex
@@ -110,4 +110,9 @@ defmodule Sanbase.Utils.Transform do
     end)
     |> Enum.sort_by(&DateTime.to_unix(&1[:datetime]))
   end
+
+  def maybe_transform_from_address("0x0000000000000000000000000000000000000000"), do: "mint"
+  def maybe_transform_from_address(address), do: address
+  def maybe_transform_to_address("0x0000000000000000000000000000000000000000"), do: "burn"
+  def maybe_transform_to_address(address), do: address
 end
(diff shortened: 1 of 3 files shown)


lib/sanbase/utils/transform.ex
```diff
@@ -110,4 +110,9 @@ defmodule Sanbase.Utils.Transform do
     end)
     |> Enum.sort_by(&DateTime.to_unix(&1[:datetime]))
   end
+
+  def maybe_transform_from_address("0x0000000000000000000000000000000000000000"), do: "mint"
+  def maybe_transform_from_address(address), do: address
+  def maybe_transform_to_address("0x0000000000000000000000000000000000000000"), do: "burn"
+  def maybe_transform_to_address(address), do: address
```
IvanIvanoff, line 117: One thing we might do is rewrite like this (last 16 characters are ignored):
```elixir
  def maybe_transform_from_address("0x0000000000000000000000000" <> _rest), do: "mint"
```
this will catch all `0x00...dead`, `0x00...0000`, etc. addresses
@Snedashkovsky should we rewrite these other addresses starting with many zeroes as well

Snedashkovsky: @IvanIvanoff We need to rewrite only the zero address "0x0000000000000000000000000000000000000000"

Snedashkovsky (approved):

lib/sanbase/utils/transform.ex
IvanIvanoff: Or even better so you don't allow addresses with wrong size:
```elixir
"0x000000000000000000000000" <> <<_::binary-size(16)>>
```

IvanIvanoff (approved):

PR #2313, opened 2020-10-19
Remove old metric_anomaly API and related modules

IvanIvanoff:
## Changes
Deprecate the old `metricAnomaly` API in favor of `getSignal`. As of writing this (23.04.2021) the API has been used once for the last 30 days. This seems like some exploration/accidental execution of the query and no real usage.
```graphql
{
  metricAnomaly(
    from: "utc_now-7d"
    to: "utc_now"
    metric: DAILY_ACTIVE_ADDRESSES
    slug: "santiment") {
      datetime
      metricValue
  }
}
```

## Checklist:


tspenov:
I guess we're waiting for frontend to stop using the old API. Seems still used here: https://github.com/santiment/app/blob/ef4a8552c4729be23b425916ef0648e4a60994d1/src/ducks/Studio/timeseries/fetcher.js#L33

tspenov (approved):

PR #2318, opened 2020-10-21
Improve error handling when sending signals

IvanIvanoff:
## Changes
Extend the reported errors count. Now when signal sending fails the following format will be reproted:
```
In total 5/35 metric_signal signals were sent successfully.
5 failed with the reason channel_disabled
5 failed with the reason no_email
10 failed with the reason no_telegram
10 failed with the reason unspecified
0 failed with the reason telegram_bot_blocked
```

## Checklist:

### DIFF
--- a/lib/sanbase/signals/evaluator/scheduler.ex
+++ b/lib/sanbase/signals/evaluator/scheduler.ex
@@ -211,26 +211,29 @@ defmodule Sanbase.Signal.Scheduler do
 
   defp log_sent_messages_stats(list, type) do
     successful_messages_count = list |> Enum.count(fn {_elem, status} -> status == :ok end)
+    errors = for {_, {:error, error}} <- list, do: error
 
-    for {_, {:error, error}} <- list do
+    Enum.each(errors, fn error ->
       Logger.warn("Cannot send a #{type} signal. Reason: #{inspect(error)}")
-    end
+    end)
 
-    telegram_bot_blocked_count =
-      Enum.count(list, fn
-        {_identifier, {:error, error}} when is_binary(error) ->
-          String.contains?(error, "blocked the telegram bot")
-
-        {_identifier, _status} ->
-          false
+    errors_to_count_map =
+      errors
+      |> Enum.group_by(fn
+        %{reason: reason} -> reason
+        _ -> :unspecified
       end)
-
-    disabled_count = Enum.count(list, &match?({_, :channel_disabled}, &1))
+      |> Map.new(fn {reason, list} -> {reason, length(list)} end)
+      |> Enum.reject(fn {_reason, count} -> count == 0 end)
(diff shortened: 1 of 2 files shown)


lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -225,12 +226,20 @@ defmodule Sanbase.Signal.Scheduler do
...
+      |> Map.put(:telegram_bot_blocked, telegram_bot_blocked_count)
+      |> Map.update(:unspecified, 0, &(&1 - telegram_bot_blocked_count))
 
     Logger.info("""
     In total #{successful_messages_count}/#{length(list)} #{type} signals were sent successfully.
-    #{disabled_count} failed because the user has disabled the notification channel.
-    #{telegram_bot_blocked_count} failed because the user has blocked the telegram bot.
+    #{
+      Enum.map(errors_to_count_map, fn {reason, count} ->
+        "#{count} failed with the reason #{reason}\n"
```
tspenov, line 240: Can we also log the trigger id in case we want to dig deeper ?

lib/sanbase/signals/evaluator/scheduler.ex
```diff
@@ -225,12 +226,20 @@ defmodule Sanbase.Signal.Scheduler do
           false
       end)
 
-    disabled_count = Enum.count(list, &match?({_, :channel_disabled}, &1))
+    errors_to_count_map =
+      errors
+      |> Enum.group_by(fn error -> Map.get(error, :reason, :unspecified) end)
+      |> Map.new(fn {reason, list} -> {reason, length(list)} end)
+      |> Map.put(:telegram_bot_blocked, telegram_bot_blocked_count)
+      |> Map.update(:unspecified, 0, &(&1 - telegram_bot_blocked_count))
```
tspenov, line 234: Any idea what could go to `unspecified` ?

IvanIvanoff: It is all that are not manually handled - if telegram/email/webhook returns some other error that is not explicitly handled maybe - nxdomain, some connection issue, etc.

lib/sanbase/signals/evaluator/scheduler.ex
IvanIvanoff: All errors are explicitly logged a few lines above here. In this message only aggregated data is included so it can guide you what to look at in the logs above it.

IvanIvanoff: What should be extended is the error map returned from `signal.ex`

IvanIvanoff: Added

tspenov (approved):

PR #2329, opened 2020-10-26
API Call Rate Limiting

IvanIvanoff:
## Changes

Implement API call rate limiting per month.

The API Call Rate Limit (ACRL) applies only to the Sanapi product and not Sanbase.

## Limits

The ACRL has 3 main branches:

### Basic authentication

When basic authentication is used no limits apply. This is used by internal services that query the API. No external user will ever use basic authentication.

### Authenticated user

The rate limits of authenticated users depend on their SanAPI plan. The plans have the following limits per month (can be changed):
- free: 1,000 (1k)
- basic: 300,000 (300k)
- pro: 600,000 (600k)
- enterprise and custom: infinity

### Anonymous user

In case of anonymous API calls, the limit is applied to the IP address. This can cause troubles with dynamic IPs or if you're behind NAT - in both cases you might be assigned an IP that has already exhausted its API calls.

## Implementation details

The API calls limit metadata is stored in 2 places:
- In a central Postgres DB, which is the source of truth.
- In-memory of every web pod (as of the time of writing this we run 3 web pods).

When a API call is made, we first check in a in-memory ETS table if there are API calls left to be made for the user/remote_ip. The data is stored as a pair (API calls left, quota). In the beginning both values are equal to `quota` and subsequent API calls reduce the first value until it gets less or equal to zero. When this happens, the web pod updates the API calls made for that user/ip in the central DB and gets a new quota (default and only value of quotas is 100).

### Pros
- It is implemented this way so not every API call makes a slow DB request but falls back to much faster in-memory function calls.
- No distributed stores are needed.
- No communication between nodes is needed

### Cons
- This approach has one issue that is easily ignorable - the actual number of allowed API calls can be slightly higher than the theoretical limit by about `(N-1)*quota_size` where `N = #web pods`. This can happen when there are exactly `quota_size` number of API calls left and all web pods get a quota. All web pods will get the `quota_size` quotas, so the value that goes over the limit is exactly (N-1)*quota_size`

> Notes: Getting a quota **does not** update the number of API calls made in the DB. The update happens only when `update_usage_db` function is called. This can happen for 2 reasons:
- The quota is exhausted.
- Forcing an update - when shutting down the node, the API calls made are stored so their count is not lost.

####

## Checklist:


lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,191 @@
...
+    changeset(acl, %{api_calls: new_api_calls})
+    |> Repo.update()
+  end
+
+  defp user_has_limits?(%User{is_superuser: true}), do: false
+  defp user_has_limits?(%User{}), do: true
+
+  defp remote_ip_has_limits?(remote_ip) do
+    with false <- Sanbase.Utils.IP.is_san_cluster_ip?(remote_ip),
+         false <- Sanbase.Utils.IP.is_localhost?(remote_ip) do
```
tspenov, line 185: Instead of checking for localhost we can check some config option and ?

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -41,6 +42,11 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
     &ContextPlug.basic_authentication/1
   ]
 
+  @should_halt_methods [
+    &ContextPlug.halt_sansheets_request?/2,
+    &ContextPlug.halt_api_call_limit_reached?/2
```
tspenov, line 47: Lets add a feature flag and for start to not halt but just observe admin how things go for a while :?

lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -69,6 +73,24 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     |> maybe_create_or_drop_session(blueprint.execution.context)
   end
 
+  defp maybe_update_api_call_limit_usage(
```
tspenov, line 76: `maybe_update_api_usage` ? and why maybe - is there a case when the usage should not be updated ?

lib/sanbase/auth/user.ex
```diff
@@ -73,6 +73,7 @@ defmodule Sanbase.Auth.User do
     field(:first_login, :boolean, default: false, virtual: true)
     field(:avatar_url, :string)
     field(:is_registered, :boolean, default: false)
+    field(:is_superuser, :boolean, default: false)
```
tspenov, line 76: `is_superuser` should be also editable through admin

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -0,0 +1,119 @@
...
+        true = :ets.insert(@ets_table, {user_id, quota, quota})
+
+        {:ok, quota}
+
+      [{^user_id, api_calls_left, _}] ->
+        {:ok, api_calls_left}
+    end
+  end
+
+  def get_quota(:remote_ip, remote_ip, _auth_method) do
```
tspenov, line 60: this is identical with the upper one - you can use `type` for `:user | :remote_ip` and `id` for `user | remote_ip` and reuse the code.

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -0,0 +1,119 @@
...
+      [{^user_id, :infinity, :infinity}] ->
+        :ok
+
+      [{^user_id, api_calls_left, _quota}] ->
+        true = :ets.update_element(@ets_table, user_id, {2, api_calls_left - count})
+        :ok
+    end
+  end
+
+  def update_usage(:remote_ip, remote_ip, count, _auth_method) do
```
tspenov, line 103: Same - the code can be reused

lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,191 @@
...
+    "basic" => 300_000,
+    "pro" => 600_000,
+    "enterprise" => :infinity,
+    "custom" => :infinity
+  }
+
+  @limit_per_hour_per_plan %{
+    "free" => 100,
+    "basic" => 3000,
+    "pro" => 6000,
```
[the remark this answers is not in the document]
IvanIvanoff: 6000 * 720 = 4.3 million which is ~7 times bigger

lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,191 @@
+defmodule Sanbase.ApiCallLimit do
+  use Ecto.Schema
+
+  import Ecto.{Query, Changeset}
+
+  alias Sanbase.Repo
+  alias Sanbase.Auth.User
+  alias Sanbase.Billing.{Product, Subscription}
+
+  @compile :inline_list_funcs
```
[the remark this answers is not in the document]
IvanIvanoff: I haven't given it much thought. Now that I think about it I had some wrong assumptions - it is in no way transitive. Furthermore, this is added at the start of the `Enum` and `List` modules so it is applied everywhere where we use these funcs.

lib/sanbase/api_call_limit/api_call_limit.ex
IvanIvanoff: Not sure what you mean by that?
`san_cluster` is needed so we check requests coming from our cluster (this might **not** work right now because of ngninx) and `localhost` is needed when you run it locally so you don't block your localhost:4000/graphiql requests

priv/repo/migrations/20201021132548_add_api_call_limits_table.exs
```diff
@@ -0,0 +1,17 @@
...
+  use Ecto.Migration
+
+  @table :api_call_limits
+  def change do
+    create table(@table) do
+      add(:user_id, references(:users, on_delete: :delete_all))
+      add(:remote_ip, :string, default: nil)
+      add(:has_limits, :boolean, default: true)
+      add(:api_calls_limit_plan, :string, default: "free")
+      add(:api_calls, :jsonb)
```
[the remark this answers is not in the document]
IvanIvanoff: This is the migration - here it is :jsonb, in the schema module it is :map

lib/sanbase_web/graphql/absinthe_before_send.ex
IvanIvanoff: It's not only `api_usage`as `api_usage` applies also to the kafka exporter, so it is indeed api call limiting here.
It does not apply if the product is not sanapi - we don't limit the sanbase calls.

lib/sanbase/api_call_limit/ets.ex
IvanIvanoff: I didn't do it as I thought that at some point a further abstraction would be too complex to understand.
But now I see if we make the entity with 2 params and not 1 it would work great - `entity` and `entity_key` (as the key cannot be obtained in the same way for remote_ip and user)

priv/repo/migrations/20201021132548_add_api_call_limits_table.exs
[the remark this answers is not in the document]
IvanIvanoff: Hm, :map actually maps to :jsonb but allows a default value, I'll change it

lib/sanbase/api_call_limit/api_call_limit.ex
tspenov: yeah I mean we can disable rate limiting locally in dev.exs or with env var instead of checking for localhost

IvanIvanoff: I should add this on/off switch but I think we should keep this one here as well. This way we'll reduce the things you need to setup in order to properly develop locally

tspenov: it could have a default value to not limit in dev.exs, so no setup will be needed

lib/sanbase/billing/subscription.ex
```diff
@@ -124,15 +124,15 @@ defmodule Sanbase.Billing.Subscription do
...
+         {:ok, _canceled_sub} <- sync_with_stripe_subscription(stripe_sub, sub),
+         %__MODULE__{user: user} <- Repo.preload(sub, [:user]),
          {:ok, _} <- Sanbase.ApiCallLimit.update_user_plan(user) do
       Sanbase.Billing.StripeEvent.send_cancel_event_to_discord(sub)
 
       {:ok,
        %{
          is_scheduled_for_cancellation: true,
-         scheduled_for_cancellation_at: canceled_sub.current_period_end
+         scheduled_for_cancellation_at: sub.current_period_end
```
tspenov, line 135: why change this one ? We first call cancel in Stripe which updates some fields on the existing subscription in Stripe and then we call `sync_with_stripe_subscription` - which syncs the updates in db and returns the updated record. This record will have updated `current_period_end`.

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -73,7 +73,9 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
       conn
       |> put_private(:absinthe, %{context: context})
 
-    case should_halt?(conn, context, @should_halt_methods) do
+    rate_limiting_enabled? = Config.get(:rate_limiting_enabled?)
+
+    case rate_limiting_enabled? and should_halt?(conn, context, @should_halt_methods) do
```
tspenov, line 78: That controls also sansheets :? I think it should be only for api.

lib/sanbase/api_call_limit/api_call_limit.ex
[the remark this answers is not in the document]
tspenov: ups - my bad missed one 0

lib/sanbase_web/graphql/context_plug.ex
IvanIvanoff: true

lib/sanbase/billing/subscription.ex
IvanIvanoff: Actually I am not sure what happened here.. With `canceled_sub` the test for checking the cancel date failed for some reason

IvanIvanoff: Actually this comment of yours is made against an old commit - indeed now here we **do not** replace `canceled_sub` with `sub` - it was a bug introduced and later fixed.

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -34,15 +41,18 @@ defmodule Sanbase.ApiCallLimit.ETS do
   A special case is when the authentication is Basic Authentication. It is used
   exclusievly from internal services and there will be no limit imposed.
 
-  Returns {:ok, number} or {:error, error}
+  Returns {:ok, infinity}, {:ok, number} or {:error, map}
```
tspenov, line 44: This is redundant - there is spec right below

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -34,15 +41,18 @@ defmodule Sanbase.ApiCallLimit.ETS do
...
+  @spec get_quota(entity_type, entity, atom()) ::
+          {:ok, :infinity} | {:ok, integer()} | {:error, map()}
   def get_quota(_type, _entity, :basic), do: {:ok, :infinity}
   def get_quota(:user, %User{} = user, _auth_method), do: do_get_quota(:user, user, user.id)
   def get_quota(:remote_ip, ip, _auth_method), do: do_get_quota(:remote_ip, ip, ip)
 
   @doc ~s"""
-  Updates the 
+  Updates the number of api calls made by a user or an ip address. The number of
+  API calls is tracked in-memory in an ETS table and after a certain number of
```
tspenov, line 54: `API calls are tracked`

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -34,15 +41,18 @@ defmodule Sanbase.ApiCallLimit.ETS do
...
+  API calls is made, the number is updated in the centralized database.
```
tspenov, line 55: `API calls are made`

lib/sanbase_web/graphql/context_plug.ex
```diff
@@ -156,6 +160,23 @@ defmodule SanbaseWeb.Graphql.ContextPlug do
...
+      api_calls_left_this_hour: left_hour,
+      api_calls_left_this_minute: left_minute
+    } = error_map
+
+    [
+      {"X-RateLimit-Reset", blocked_for_seconds},
+      {"X-RateLimit-Remaining-Month", left_month},
+      {"X-RateLimit-Remaining-Hour", left_hour},
+      {"X-RateLimit-Remaining-Minute", left_minute},
+      {"X-RateLimit-Remaining", left_minute}
```
tspenov, line 176: Should we also add the "X-RateLimit-Limit-Month|Hour|Minute" ?

lib/sanbase/api_call_limit/ets.ex
IvanIvanoff: `is` refers to `number`, not `api calls`.

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -0,0 +1,154 @@
...
+        end
+
+      [{^entity_key, :infinity, :infinity}] ->
+        {:ok, :infinity}
+
+      [{^entity_key, api_calls_left, quota}] when api_calls_left <= 0 ->
+        :ok = ApiCallLimit.update_usage_db(entity_type, entity, quota + -api_calls_left)
+
+        with {:ok, quota} <- ApiCallLimit.get_quota_db(entity_type, entity) do
+          true = :ets.insert(@ets_table, {entity_key, quota, quota})
```
tspenov, line 101: can we reuse `get_quota_db_and_update_ets` here ?

lib/sanbase_web/graphql/context_plug.ex
IvanIvanoff: Added. 
I also added the limits in the OK case, which required some big rework (all across the in-memory ETS to context plug handling)

lib/sanbase/api_call_limit/ets.ex
IvanIvanoff: `is` refers to `number`, not to `api calls`

lib/sanbase/utils/datetime/utils.ex
```diff
@@ -1,4 +1,10 @@
 defmodule Sanbase.DateTimeUtils do
+  def seconds_to_human_readable(seconds) do
+    seconds
+    |> Timex.Duration.from_seconds()
+    |> Elixir.Timex.Format.Duration.Formatters.Humanized.format()
```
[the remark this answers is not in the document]
IvanIvanoff: Nope, copy-paste from Timex documentation

lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,257 @@
...
+  defdelegate get_quota(type, entity, auth_method), to: __MODULE__.ETS
+  defdelegate update_usage(type, entity, count, auth_method), to: __MODULE__.ETS
+
+  def get_quota_db(type, entity) do
+    case get_by(type, entity) do
+      nil ->
+        {:ok, acl} = create(type, entity)
+        do_get_quota(acl)
+
+      %{has_limits: false} ->
```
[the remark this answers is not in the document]
IvanIvanoff: We don't apply this rule (use module if it's module) in many places, too

lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,257 @@
...
+
+    %{
+      month_str: now |> Timex.beginning_of_month() |> to_string(),
+      hour_str: %{now | :minute => 0, :second => 0, :microsecond => {0, 0}} |> to_string(),
+      minute_str: %{now | :second => 0, :microsecond => {0, 0}} |> to_string()
+    }
+  end
+
+  defp user_to_plan(%User{} = user) do
+    api_subscription = Subscription.current_subscription(user, @product_api_id)
```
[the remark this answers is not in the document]
IvanIvanoff: Yes, looks better.

lib/sanbase/api_call_limit/ets.ex
IvanIvanoff: Yes, and not only here. After a slight tweak to the function (to properly return the result) it can be reused in 3 places in this function.

Additional to that, that particular case is wrong - it should have an additional `metadata` fourth param.

lib/sanbase/api_call_limit/api_call_limit.ex
```diff
@@ -0,0 +1,257 @@
...
+
+  defp create(:user, %User{} = user, count) do
+    %{month_str: month_str, hour_str: hour_str, minute_str: minute_str} = get_time_str_keys()
+
+    api_calls = %{month_str => count, hour_str => count, minute_str => count}
+
+    plan = user_to_plan(user)
+
+    has_limits =
+      user_has_limits?(user) and
```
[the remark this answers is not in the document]
IvanIvanoff: I would keep them separate for now - `user_has_limits?` works only on the user struct, not user/subscription pairs.

lib/sanbase/api_call_limit/ets.ex
```diff
@@ -0,0 +1,142 @@
...
+
+        :ok
+
+      [{^entity_key, api_calls_remaining, _quota, _metadata}] ->
+        true = :ets.update_element(@ets_table, entity_key, {2, api_calls_remaining - count})
+        :ok
+    end
+  end
+
+  defp get_quota_db_and_update_ets(entity_type, entity, entity_key) do
```
tspenov, line 121: This function returns {:ok, metadata} in one case and `true` in the other.  And sometimes we use the result, sometimes we return :ok after invoking it.
Can we unify somehow ?

tspenov (approved):

lib/sanbase/api_call_limit/ets.ex
IvanIvanoff: fixed

PR #2331, opened 2020-10-28
Improve how data is fetched for the metric signal

IvanIvanoff:
## Changes
With the old approach it can happen that the data is split into 2 buckets, so part of the data gets ignored.
With the new approach this is no longer possible, making the signal more correct.

## Checklist:

### DIFF
--- a/lib/sanbase/signals/trigger/settings/metric_trigger_settings.ex
+++ b/lib/sanbase/signals/trigger/settings/metric_trigger_settings.ex
@@ -91,11 +91,21 @@ defmodule Sanbase.Signal.Trigger.MetricTriggerSettings do
     last = now
 
     Cache.get_or_store(cache_key, fn ->
-      with {:ok, [_ | _] = first} <-
-             Metric.timeseries_data(metric, selector, first, middle, time_window),
-           {:ok, [_ | _] = second} <-
-             Metric.timeseries_data(metric, selector, middle, last, time_window) do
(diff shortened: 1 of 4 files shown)


lib/sanbase/signals/trigger/settings/metric_trigger_settings.ex
```diff
@@ -91,11 +91,21 @@ defmodule Sanbase.Signal.Trigger.MetricTriggerSettings do
...
-             Metric.timeseries_data(metric, selector, middle, last, time_window) do
-        [first |> List.last(), second |> List.last()]
+      with {:ok, data1} <- Metric.aggregated_timeseries_data(metric, selector, first, middle),
+           {:ok, data2} <- Metric.aggregated_timeseries_data(metric, selector, middle, last) do
+        # NOTE: Rework when the aggregated_timeseries_data function
+        # starts to return the result in the same format
+
+        to_value = fn
+          nil -> 0
+          %{} = map -> Map.values(map) |> hd()
```
tspenov, line 101: why do we take the first value ? In the previous code we take the last of the timeseries values ?

IvanIvanoff: It contains only 1 value here, so this is basically unpacking the list of 1 elements.

tspenov (approved):

tzanko-matev (approved):

tzanko-matev (approved):

PR #2352, opened 2020-11-10
Migrate wrong wallet movement signals. Add validation

tspenov:
## Changes

* Migrate `selector` with `%{currency: currency}` -> `%{slug: currency}` when infrastructure is not Ripple
* Migrate `Own` -> `ETH` infrastructure
* Add proper validation to the wallet movement signal selector

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
@@ -115,7 +115,8 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   end
 
   # erc20 case
-  @spec selector_to_args(selector) :: {Atom.t(), String.t(), non_neg_integer()}
+  @spec selector_to_args(selector) ::
+          {module(), String.t(), non_neg_integer()} | {:error, String.t()}
   def selector_to_args(%{infrastructure: "ETH", slug: slug})
       when is_binary(slug) and slug != "ethereum" do
     with {:ok, contract, decimals} <- Project.contract_info_by_slug(slug),
--- a/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
@@ -115,41 +115,46 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   end
 
   # erc20 case
-  defp selector_to_args(%{infrastructure: "ETH", slug: slug})
-       when is_binary(slug) and slug != "ethereum" do
+  @spec selector_to_args(selector) :: {Atom.t(), String.t(), non_neg_integer()}
+  def selector_to_args(%{infrastructure: "ETH", slug: slug})
+      when is_binary(slug) and slug != "ethereum" do
     with {:ok, contract, decimals} <- Project.contract_info_by_slug(slug),
          do: {Erc20Balance, contract, decimals}
   end
 
-  defp selector_to_args(%{infrastructure: "ETH"}) do
+  def selector_to_args(%{infrastructure: "ETH"}) do
     with {:ok, contract, decimals} <- Project.contract_info_by_slug("ethereum"),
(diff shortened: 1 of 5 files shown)


lib/sanbase/clickhouse/historical_balance/historical_balance.ex
```diff
@@ -115,41 +115,46 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   end
 
   # erc20 case
-  defp selector_to_args(%{infrastructure: "ETH", slug: slug})
-       when is_binary(slug) and slug != "ethereum" do
+  @spec selector_to_args(selector) :: {Atom.t(), String.t(), non_neg_integer()}
```
IvanIvanoff, line 118: It can return an error as well if the `with` does not match. 
Also, you can use `module()` instead of `Atom.t()`

lib/sanbase/clickhouse/historical_balance/historical_balance.ex
```diff
@@ -115,41 +115,47 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   end
 
   # erc20 case
-  defp selector_to_args(%{infrastructure: "ETH", slug: slug})
-       when is_binary(slug) and slug != "ethereum" do
+  @spec selector_to_args(selector) ::
+          {module(), String.t(), non_neg_integer()} | {:error, String.t()}
```
IvanIvanoff, line 119: The spec on the `contract_info_by_slug` function itself is wrong. On error it returns a tuple as the second parameter, not a string.

lib/sanbase/clickhouse/historical_balance/historical_balance.ex
```diff
@@ -115,41 +115,47 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
...
-  defp selector_to_args(%{infrastructure: "BNB"} = selector) do
+  def selector_to_args(%{infrastructure: "BNB"} = selector) do
     slug = Map.get(selector, :slug, "binance-coin")
 
     with {:ok, contract, decimals} <- Project.contract_info_by_slug(slug),
          do: {BnbBalance, contract, decimals}
   end
+
+  def selector_to_args(selector) do
+    {:error, "Invalid selector: #{inspect(selector)}"}
```
IvanIvanoff, line 159: Invalid historical balance selector

lib/sanbase/signals/trigger/validation/target_validation.ex
```diff
@@ -67,11 +67,9 @@ defmodule Sanbase.Signal.Validation.Target do
...
 
   def valid_historical_balance_selector?(selector) when is_map(selector) do
-    keys = Map.keys(selector)
-
-    case Enum.all?(keys, fn key -> key in [:infrastructure, :currency, :slug] end) do
-      true -> :ok
-      false -> {:error, "#{inspect(selector)} is not a valid selector - it has unsupported keys"}
+    case Sanbase.Clickhouse.HistoricalBalance.selector_to_args(selector) do
+      {_, _, _} -> :ok
+      {:error, _error} -> "#{inspect(selector)} is not a valid selector."
```
IvanIvanoff, line 72: is not a valid historical balance selector

priv/repo/migrations/20201110115647_migrate_wrong_wallet_movements_signals.exs
```diff
@@ -0,0 +1,68 @@
...
+        settings: new_settings
+      })
+  end
+
+  defp maybe_update(
+         %{
+           trigger: %{
+             settings:
+               %{
+                 selector: %{infrastructure: "Own"}
```
IvanIvanoff, line 47: Do we have `own` written with different casing anywhere?

tspenov: No - it is the same case as in db.

PR #2360, opened 2020-11-17
Add blockchain addresses table and watchlists API

IvanIvanoff:
## Changes
Add `blockchain_address` tables in postgres.
Extend the watchlists API to support blockchain addresses watchlists.
Add a `type` field that is used to distinguish the different watchlists types. Currently `PROJECT` and `BLOCKCHAIN_ADDRESS` are the supported types.

Note: `type` has `PROJECT` as default so the old behaviour is kept.

Fetch watchlists of current user. 
```graphql
{

  currentUser{
    blockchainAddressWatchlists: watchlists(type: BLOCKCHAIN_ADDRESS)
    projectWatchlists: watchlists(type: PROJECT)
  }
}
```
Fetch a watchlists - after the `id` is known, the type should not be provided as arugment
```graphql
{
  watchlist(id: 515) {
    listItems{
       blockchainAddress{ address }
    }
  }
}
```
The same applies to fetching watchlists by slug
```graphql
{
  watchlistBySlug(slug: "stablecoins")
     projects {
       	project{ id slug}
    }
  }
}
```

Create a watchlist
```graphql
mutation {
  createWatchlist(
    name: "haha"
    type: BLOCKCHAIN_ADDRESS
    isPublic: true
    listItems: [
      {
        blockchainAddress: {
          address: "0x123a"
          labels: []
          notes: "hahahahahahha"
          infrastructure: "ETH"
        }
      }
      {
        blockchainAddress: {
          address: "0x123b"
          labels: []
          notes: "hahahahahahha"
          infrastructure: "BTC"
        }
      }
      {
        blockchainAddress: {
          address: "0x123c"
          labels: []
          notes: "hahahahahahha"
          infrastructure: "XRP"
        }
      }
    ]
  ) {
    id
  }
}
```

Update a watchlist:
```graphql
mutation {
  updateWatchlist(
    id: 69
    listItems: [
      {
        blockchainAddress: {
          address: "0x123a"
          labels: ["HAHO", "PE"]
          notes: "hahahahahahha"
          infrastructure: "ETH"
        }
      }
      {
        blockchainAddress: {
          address: "0x123b"
          notes: "hahahahahahha"
          infrastructure: "BTC"
        }
      }
      {
        blockchainAddress: {
          address: "0x123c"
          labels: []
          notes: "hahahahahahha"
          infrastructure: "XRP"
        }
      }
    ]
  ) {
    id
    listItems {
      blockchainAddress {
        address
        labels {
          name
        }
      }
    }
  }
}
```

## Checklist:


lib/sanbase/blockchain_address/blockchain_address.ex
```diff
@@ -0,0 +1,49 @@
...
+    |> Sanbase.Repo.insert(on_conflict: :nothing)
+  end
+
+  def maybe_create(list) when is_list(list) do
+    changesets = list |> Enum.map(&changeset(%__MODULE__{}, &1)) |> Enum.with_index()
+
+    Enum.reduce(
+      changesets,
+      Ecto.Multi.new(),
+      fn {changeset, offset}, multi ->
```
tspenov, line 34: I think the struct from the changeset is more descriptive than just a number as a second param of `Multi.insert`.

lib/sanbase/blockchain_address/blockchain_address.ex
```diff
@@ -0,0 +1,49 @@
...
+          on_conflict: {:replace, [:address]},
+          conflict_target: [:address, :infrastructure_id],
+          returning: true
+        )
+      end
+    )
+    |> Sanbase.Repo.transaction()
+    |> case do
+      {:ok, result} -> {:ok, Map.values(result)}
+      {:error, error} -> {:error, error}
```
tspenov, line 46: I think it returns different tuple in the error case - `{:error, failed_operation, failed_value, changes_so_far}`

lib/sanbase/blockchain_address/blockchain_address_label.ex
```diff
@@ -0,0 +1,39 @@
...
+          on_conflict: {:replace, [:name]},
+          conflict_target: [:name],
+          returning: true
+        )
+      end
+    )
+    |> Sanbase.Repo.transaction()
+    |> case do
+      {:ok, result} -> {:ok, Map.values(result)}
+      {:error, error} -> {:error, error}
```
tspenov, line 36: Error tuple is different as mentioned above

lib/sanbase/user_lists/user_list.ex
```diff
@@ -24,21 +24,27 @@ defmodule Sanbase.UserList do
...
   alias Sanbase.Repo
   alias Sanbase.Timeline.TimelineEvent
+  alias Sanbase.BlockchainAddress.{BlockchainAddressUserPair, BlockchainAddressLabel}
 
   schema "user_lists" do
-    field(:name, :string)
-    field(:slug, :string)
-    field(:description, :string)
-    field(:is_public, :boolean, default: false)
+    field(:type, WatchlistType, default: "project")
```
tspenov, line 30: Is it working with string, I see we use atom 2 lines below ?

IvanIvanoff: this default is a leftover from the time `type` was a string. I'll change to atom here as well

lib/sanbase/blockchain_address/blockchain_address.ex
IvanIvanoff: Yeah, I had the feeling I am missing something. Indeed with the change of using `offset` to just using the changeset as key it will lead to much nicer error here

IvanIvanoff: Actually, on a second though, we don't need to do that - the error itself will be a changset error and it will contain the values that caused the error.

lib/sanbase/blockchain_address/blockchain_address.ex
IvanIvanoff: I'll keep the offset otherwise the order of insertion is broken. The order of insertion is useful to keep the order of elements to be the same as the provided ones. I am not entirely sure this is documented behaviour, but changing from offset to changeset here a few tests broke because the order is different.

PR #2376, opened 2020-12-01
Introduce Twitter OAuth login

IvanIvanoff:
## Changes

Hitting the `/auth/twitter` path will redirect to the twitter oauth server and then redirect to sanbase.

## Permissions granted
Twitter allows 3 levels of permissions that can be granted. The one with least amount of permissions is called `read-only` and looks like this and we're using it:
image

## Checklist:


lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -43,4 +43,22 @@ defmodule SanbaseWeb.AuthController do
...
   end
+
+  def callback(%{assigns: %{ueberauth_auth: %{provider: :twitter} = auth}} = conn, _params) do
+    twitter_id_str = auth.uid
+
+    with true <- is_binary(twitter_id_str),
+         {:ok, user} <- Sanbase.Auth.User.find_or_insert_by_twitter_id(twitter_id_str),
+         {:ok, token, _claims} <- SanbaseWeb.Guardian.encode_and_sign(user, %{salt: user.salt}) do
+      conn
+      |> put_flash(:info, "Successfully authenticated.")
```
tspenov, line 54: Is that flash really visible ?

lib/sanbase/auth/user/user_twitter.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  def find_or_insert_by_twitter_id(twitter_id, username \\ nil) do
+    case Repo.get_by(User, twitter_id: twitter_id) do
+      nil ->
+        %User{
+          twitter_id: twitter_id,
+          username: username,
+          salt: User.generate_salt(),
+          first_login: true
+        }
```
tspenov, line 13: Shouldn't we use the changeset here ?

lib/sanbase/auth/user/user_twitter.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Sanbase.Auth.User.Twitter do
+  alias Sanbase.Repo
+  alias Sanbase.Auth.User
+
+  def find_or_insert_by_twitter_id(twitter_id, username \\ nil) do
```
tspenov, line 5: I think currently we rely on having one of `email` or `username` (though we don't have constraint to require at least one of them).

lib/sanbase/auth/user/user_twitter.ex
IvanIvanoff: This is copied in many places. We mainly use `change` instead of `changeset` if we modify fields that are otherwise not allowed, so they cannot be accidentally passed via some API/function.

lib/sanbase/auth/user/user_twitter.ex
tspenov, line 5: username param is not user - so lets remove it

lib/sanbase/grafana/api.ex
```diff
@@ -16,23 +16,26 @@ defmodule Sanbase.GrafanaApi do
...
+    user_unique_str = User.get_unique_str(user)
     request_path = "api/admin/users"
 
     data =
       %{
-        "name" => username || email,
-        "email" => email || username,
-        "login" => username || email,
+        "name" => user_unique_str,
+        "email" => email,
```
tspenov, line 36: Not sure if email is not a required param: https://grafana.com/docs/grafana/latest/http_api/admin/#global-users

lib/sanbase_web/controllers/auth_controller.ex
IvanIvanoff: no, removed.

lib/sanbase/auth/user/user_twitter.ex
IvanIvanoff: This is done like this to be consistent with the other find_or_insert functions in the user module

lib/sanbase/grafana/api.ex
IvanIvanoff: Reworked.

tspenov (approved):

PR #2379, opened 2020-12-02
Use eth_labels_final table in labels query

ninjascant:
## Changes

Use `eth_labels_final` table instead of `blockchain_address_labels` to obtain address labels. This will allow to speed up query

### DIFF
--- a/lib/sanbase/clickhouse/metric/metric_files/eth2_metrics.json
+++ b/lib/sanbase/clickhouse/metric/metric_files/eth2_metrics.json
@@ -1,7 +1,7 @@
 [
   {
-    "human_readable_name": "ETH2.0 Staker Count",
-    "name": "eth2_staker_count",
+    "human_readable_name": "ETH 2.0 Stakers Count",
+    "name": "eth2_stakers_count",
     "metric": "eth2_staker_count",
     "version": "2019-01-01",
     "access": "restricted",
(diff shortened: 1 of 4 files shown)


tspenov:
For some reason I get `DB::Exception: Table default.eth_labels_final doesn't exist` when testing locally. I see there is such table when I connect to clickhouse. The only difference from querying from sanbase is that we use `clickhouse-proxy` and other user. We can deploy and test on stage with this query:
```
{
  projectBySlug(slug:"santiment") {
    tokenTopTransactions(from: "utc_now-1w", to: "utc_now", limit:100) {
      fromAddress {
        address
        labels {
          name
        }
      }
      toAddress {
        address
        labels {
          name
        }
      }
      trxValue
      trxHash
    }
  }
}
```
to see whether we have proper labels.

tspenov:
After adding `eth_labels_final` to user facing cluster - it works.

tspenov (approved):

lib/sanbase/clickhouse/metric/metric_files/eth2_metrics.json
```diff
@@ -0,0 +1,19 @@
+[
+  {
+    "human_readable_name": "ETH2.0 Staker Count",
+    "name": "eth2_staker_count",
```
IvanIvanoff, line 4: Let's expose it as `eth2_stakers_count` (plural) to be consistent with other `*_count` metrics

Snedashkovsky: Updated

PR #2387, opened 2020-12-04
Use the email from the twitter oauth. Refactor login code

IvanIvanoff:
## Changes
- Link the twitter account to an existing account if the email addresses match
- Rework the `find_or_insert_by_*` user functions to accept a map of attributes. This is mainly used to properly set `is_registered` when using twitter/google oauth as it does not need a second verification step.

## Checklist:

### DIFF
--- a/lib/sanbase/auth/user/user_twitter.ex
+++ b/lib/sanbase/auth/user/user_twitter.ex
@@ -2,19 +2,29 @@ defmodule Sanbase.Auth.User.Twitter do
   alias Sanbase.Repo
   alias Sanbase.Auth.User
 
-  def find_or_insert_by_twitter_id(twitter_id, username \\ nil) do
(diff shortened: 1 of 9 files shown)


lib/sanbase/auth/user/user_twitter.ex
```diff
@@ -2,19 +2,29 @@ defmodule Sanbase.Auth.User.Twitter do
   alias Sanbase.Repo
   alias Sanbase.Auth.User
 
-  def find_or_insert_by_twitter_id(twitter_id, username \\ nil) do
+  def find_or_insert_by_twitter_id(twitter_id, attrs \\ %{}) do
```
tspenov, line 5: find_or_insert_by_twitter_id identical to find_or_insert_by_email - we can have `find_or_insert_by(field, filed_value, attrs)` ?

tspenov (approved):

PR #2388, opened 2020-12-07
Support state OAuth parameter for after-auth redirects

IvanIvanoff:
## Changes
`state` should be a base64 encoded json map with 2 keys - `success_redirect_url` and `fail_redirect_url`
https://www.loom.com/share/0eb59a9668f8492580d49a3abce5f275

## Checklist:

### DIFF
--- a/lib/sanbase_web/controllers/auth_controller.ex
+++ b/lib/sanbase_web/controllers/auth_controller.ex
@@ -8,11 +8,6 @@ defmodule SanbaseWeb.AuthController do
   plug(Ueberauth)
 
   alias Sanbase.Auth.User
-  alias Ueberauth.Strategy.Helpers
-
-  def request(conn, _params) do
-    render(conn, "request.html", callback_url: Helpers.callback_url(conn))
-  end
 
   def delete(conn, _params) do
     conn
@@ -25,7 +20,9 @@ defmodule SanbaseWeb.AuthController do
     |> redirect(to: "/")
   end
 
-  def callback(%{assigns: %{ueberauth_auth: %{provider: :google} = auth}} = conn, _params) do
+  def callback(%{assigns: %{ueberauth_auth: %{provider: :google} = auth}} = conn, params) do
+    redirect_urls = get_redirect_urls(params)
+
     email = auth.info.email
 
     with true <- is_binary(email),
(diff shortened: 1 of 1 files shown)


lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -73,4 +73,21 @@ defmodule SanbaseWeb.AuthController do
   defp twitter_login(_email, twitter_id) do
     User.find_or_insert_by(:twitter_id, twitter_id, %{is_registered: true})
   end
+
+  defp get_redirect_urls(%{"state" => state}) do
+    {:ok, json} = Base.decode64(state)
+    {:ok, map} = Jason.decode(json)
```
tspenov, line 79: you can use bang! version - that will raise instead of pattern match error :?

lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -73,4 +73,21 @@ defmodule SanbaseWeb.AuthController do
...
+
+    %{
+      success: Map.get(map, "success_redirect_url", SanbaseWeb.Endpoint.website_url()),
```
tspenov, line 82: Lets allow only with domain `santiment.net` ?

IvanIvanoff: I'll put some restrictions in place but as we put the token in httpOnly cookie and only our backend knows the signing keys, there's no security risk when redirecting - we won't send anything a malicious party can extract.

lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -73,4 +72,36 @@ defmodule SanbaseWeb.AuthController do
...
+      |> Jason.decode!()
+
+    %{
+      success: Map.get(map, "success_redirect_url", website_url),
+      fails: Map.get(map, "fail_redirect_url", website_url)
+    }
+    |> Enum.into(%{}, fn {key, url} ->
+      # Use the state provided redirects only if they're form the santiment host
+      with %URI{host: host} <- URI.parse(url),
+           "santiment" <- host |> String.split(".") |> Enum.at(".") do
```
tspenov, line 91: There is typo in `Enum.at`.  
I think `String.ends_with?(host, "santiment.net")` is sufficient check here

IvanIvanoff: `String.ends_with?` is returning true for many invalid addresses like `fake-santiment.net`

PR #2392, opened 2020-12-09
Remove invalid labels for DEX & CEX addresses

ninjascant:
## Changes

Add new rules for addresses with DEX and CEX labels:
* if an address has CEX label - remove `deposit` and `withdrawal` labels
* if an address has DEX label - remove the `dex_trader` label

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/labels.ex
+++ b/lib/sanbase/clickhouse/labels.ex
@@ -124,10 +124,16 @@ defmodule Sanbase.Clickhouse.Label do
                    splitByChar(',', owners) as owner_arr,
                    arrayZip(label_arr, owner_arr) as labels_owners,
                    multiIf(
+                       -- if there is the `system` label for an address, we exclude other labels
                        has(label_arr, 'system'), arrayFilter(x -> x.1 = 'system', labels_owners),
+                       -- if an address has a `centralized_exchange` label and at least one of the `deposit` and
+                       -- `withdrawal` labels, we exclude the `deposit` and `withdrawal` labels.
                        has(label_arr, 'centralized_exchange') AND hasAny(label_arr, ['deposit', 'withdrawal']), arrayFilter(x -> x.1 NOT IN ('deposit', 'withdrawal'), labels_owners),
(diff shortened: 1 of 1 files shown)


tspenov:
Question not related to current changes:
Isn't the `PREWHERE` clause redundant because you have `INNER JOIN` with filter by these addresses ?
```sql
            ANY INNER JOIN (
                SELECT cityHash64(address) as address_hash,
                       address
                FROM (
                    SELECT lower(arrayJoin([?2])) as address
                )
            )
            USING address_hash
            PREWHERE address_hash IN (
                SELECT cityHash64(address)
                FROM (
                    SELECT lower(arrayJoin([?2])) as address
                )
            )
```

IvanIvanoff (approved):

lib/sanbase/clickhouse/labels.ex
```diff
@@ -125,9 +125,10 @@ defmodule Sanbase.Clickhouse.Label do
                    arrayZip(label_arr, owner_arr) as labels_owners,
                    multiIf(
                        has(label_arr, 'system'), arrayFilter(x -> x.1 = 'system', labels_owners),
+                       has(label_arr, 'centralized_exchange') AND hasAny(label_arr, ['deposit', 'withdrawal']), arrayFilter(x -> x.1 NOT IN ('deposit', 'withdrawal'), labels_owners),
```
tspenov, line 128: Can you write a sql comment in English on all these filters because they are really obscure and tied to how our labels work ?

Snedashkovsky:
@tspenov You are right, the result of the query will be the same, but using `PREWHERE` speeds up the query by 5 times.

lib/sanbase/clickhouse/labels.ex
Snedashkovsky: Added. 
Please check if queries with added comments will be executed

tspenov: yes, works

tspenov (approved):

PR #2411, opened 2020-12-15
Add active subscriptions intercom property

tspenov:
## Changes

Add active subscriptions Intercom property

## Checklist:

### DIFF
--- a/lib/sanbase/auth/statistics.ex
+++ b/lib/sanbase/auth/statistics.ex
@@ -4,6 +4,8 @@ defmodule Sanbase.Auth.Statistics do
   alias Sanbase.Repo
   alias Sanbase.Math
   alias Sanbase.Auth.{User, UserSettings}
+  alias Sanbase.UserList
+  alias Sanbase.Billing.Subscription
 
   def tokens_staked() do
     san_balances =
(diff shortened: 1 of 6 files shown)


lib/sanbase/auth/statistics.ex
```diff
@@ -116,6 +118,58 @@ defmodule Sanbase.Auth.Statistics do
     |> Repo.one()
   end
 
+  # Resource coud be watchlist, insight, user_trigger struct or any other struct which belongs to User
```
IvanIvanoff, line 121: typo coud -> could

lib/sanbase/auth/statistics.ex
```diff
@@ -116,6 +118,58 @@ defmodule Sanbase.Auth.Statistics do
...
+      Enum.reduce(queries, resource_query, fn query_func, acc ->
+        query_func.(acc)
+      end)
+
+    query
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
+
+  def screeners_user_count_map do
```
IvanIvanoff, line 143: wouldn't `user_screeners_count_map` be better?

lib/sanbase/auth/statistics.ex
```diff
@@ -116,6 +118,58 @@ defmodule Sanbase.Auth.Statistics do
...
+      queries: [
+        fn query ->
+          from(r in query, where: fragment("?.function->>'name' != 'empty'", r))
+        end
+      ]
+    )
+  end
+
+  def active_subscriptions_map() do
+    from(s in Subscription, where: s.status == "active", preload: [plan: [:product]])
```
IvanIvanoff, line 154: Let's make this a separate function in the Subscription module - query/Repo calls from outside the schema module should be avoided.

lib/sanbase/auth/statistics.ex
```diff
@@ -116,6 +118,58 @@ defmodule Sanbase.Auth.Statistics do
...
+    |> Sanbase.Repo.all()
+    |> Enum.map(fn s -> %{user_id: s.user_id, sub: "#{s.plan.product.name}/#{s.plan.name}"} end)
```
IvanIvanoff, line 156: `product_name` instead of `sub` would be more understandable?

lib/sanbase/auth/statistics.ex
```diff
@@ -116,6 +118,58 @@ defmodule Sanbase.Auth.Statistics do
...
+    |> Enum.group_by(fn s -> s.user_id end)
+    |> Enum.map(fn {user_id, subs} ->
+      {user_id, Enum.map(subs, & &1.sub) |> Enum.join(",")}
```
IvanIvanoff, line 159:
```elixir
Enum.join(", ")
```
Let's add one interval for better formatting.

IvanIvanoff: One idea: You can do all the 3 steps in 1 pass (not sure if it's clearer):
```elixir
|> Enum.reduce(%{}, fn {user_id, sub}, acc ->                             
  Map.update(acc, user_id, sub, & "#{&1}, #{sub}")  
end)
```

tspenov: I'll go with the current version here - the transformations seem more clear to me

PR #2413, opened 2020-12-15
Change social dominance old API

spiderjako:
## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/social_data_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/social_data_resolver.ex
@@ -6,6 +6,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.SocialDataResolver do
 
   alias Sanbase.{SocialData, TechIndicators}
   alias SanbaseWeb.Graphql.SanbaseDataloader
+  alias SanbaseWeb.Graphql.Resolvers.MetricResolver
 
   @context_words_default_size 10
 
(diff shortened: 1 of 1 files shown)


lib/sanbase_web/graphql/resolvers/social_data_resolver.ex
```diff
@@ -155,10 +156,26 @@ defmodule SanbaseWeb.Graphql.Resolvers.SocialDataResolver do
 
   def social_dominance(
         _root,
-        %{slug: slug, from: from, to: to, interval: interval, source: source},
+        %{slug: _slug, from: _from, to: _to, interval: _interval, source: source} = args,
         _resolution
       ) do
-    SocialData.social_dominance(%{slug: slug}, from, to, interval, source)
+    source =
```
IvanIvanoff, line 162:
```suggestion
    source = if source == :all, do: :total, else: :source
```
you can do it on 1 row also

spiderjako: ``mix format`` makes it 5 lines of code, wouldn't that be a problem if it were to change?

IvanIvanoff: `mix format` cannot change between the 2 formats - the one with do..end, and the one with a kw-list (`,do: ... , else: ...`)

spiderjako: ah, didn't see that it was a different format, thanks

IvanIvanoff (approved):

PR #2414, opened 2020-12-15
Add blockchain address labels

IvanIvanoff:
## Changes

Exponse both santiment-defined labels (`origin: "santiment"`) and user-defined labels (`origin: "user"`) in the blockchain address graphql type.

User-defined labels are defined only in the context of a watchlist. A user can define labels that will be additionally displayed along the santiment provided labels.

Request
```graphql
{
  watchlist(id: 1173) {
    listItems {
      blockchainAddress {
        address
        labels {
          name
          metadata
          origin
        }
      }
    }
  }
}
```

Response:
```json
{
  "data": {
    "watchlist": {
      "listItems": [
        {
          "blockchainAddress": {
            "address": "0x04291180af677efd464432138d2a5d766a3898d4",
            "labels": [
              {
                "metadata": "{\"owner\": \"bitfinex\"}",
                "name": "CEX Deposit",
                "origin": "santiment"
              }
            ]
          }
        },
        {
          "blockchainAddress": {
            "address": "0x00e6a95f6f438cbe873d0465e1a2533052e155f0",
            "labels": [
              {
                "metadata": "{\"owner\": \"bittrex\"}",
                "name": "CEX Trader",
                "origin": "santiment"
              }
            ]
          }
        },
        {
          "blockchainAddress": {
            "address": "0x003b9fb45362194a35a30cf076c937b5f483dc73",
            "labels": [
              {
                "metadata": null,
                "name": "CEX",
                "origin": "user"
              },
              {
                "metadata": "{\"owner\": \"bittrex\"}",
                "name": "CEX Trader",
                "origin": "santiment"
              },
              {
                "metadata": "{\"owner\": \"\"}",
                "name": "Whale, token:santiment",
                "origin": "santiment"
              }
            ]
          }
        }
      ]
    }
  }
}
```

---

Request:
```graphql
{
  blockchainAddress(
    selector: {
      address: "0x019efa7ae5950a3d0ebccb78ca04b861c91b6114"
      infrastructure: "ETH"
    }) {
      address
      labels {
        name
        origin
      }
  }
}
```

Response:
```json
{
  "data": {
    "blockchainAddress": {
      "address": "0x019efa7ae5950a3d0ebccb78ca04b861c91b6114",
      "labels": [
        {
          "name": "Whale, token:1world",
          "origin": "santiment"
        },
        {
          "name": "CEX Trader",
          "origin": "santiment"
        }
      ]
    }
  }
}
```

## Checklist:


lib/sanbase_web/admin/blockchain_address/blockchain_address_user_pair_label.ex
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def changeset(post_tag, attrs \\ %{}) do
+    post_tag
+    |> cast(attrs, [:blockchain_address_user_pair_id, :label_id])
+    |> validate_required([:blockchain_address_user_pair_id, :label_id])
+  end
+end
+
+defmodule SanbaseWeb.ExAdmin.BlockchainAddressUserPairLabel do
```
tspenov, line 19: Why admin module is here ?

tspenov:
@IvanIvanoff please, add tests

lib/sanbase_web/graphql/resolvers/blockchain_address_resolver.ex
```diff
@@ -64,6 +64,24 @@ defmodule SanbaseWeb.Graphql.Resolvers.BlockchainAddressResolver do
...
+  def labels(%{address: address} = root, _args, %{context: %{loader: loader}}) do
+    address = BlockchainAddress.to_internal_format(address)
+
+    loader
+    |> Dataloader.load(SanbaseDataloader, :address_labels, address)
+    |> on_load(fn loader ->
+      santiment_labels = Dataloader.get(loader, SanbaseDataloader, :address_labels, address) || []
+
+      user_labels =
+        Map.get(root, :labels, [])
```
tspenov, line 76: Why user labels are in the root ?

IvanIvanoff: It depends where does the `root` come from. We have 2 cases:
- From the `blockchainAddress` query - an address does not belong to a user and the blockchain address schema does not even have labels as we store our labels in CH. In this case this key is missing and  the root is built on top of a "pure" address. This is why we use `Map.get`
- From watchlists - in this case the root is built on top of a `blockchainAddressUserPair` as seen [here](https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex#L210-L221). In this case we have some additional fields, not present in the "pure" address that can be added - for example the labels.

lib/sanbase_web/admin/blockchain_address/blockchain_address_user_pair_label.ex
IvanIvanoff: This is following the steps done to implement the post tags. We don't need a special schema module for the join-through tables for the API to work but in order to implement the `Add New Label` button, some routes/pages need to exist (see the button link) in order to work. For this we need to define the schema module and the ex-admin module, so it's done in 1 place.

lib/sanbase_web/graphql/resolvers/blockchain_address_resolver.ex
tspenov: So in the second case we we'll have `labels(%UserList{address: address}` ?
Can we make `labels` with two heads then - I think that will be more clear :?

IvanIvanoff: The argument is not typed when watchlist is used. In the case of watchlist we get as root something like:
```elixir
%{
  address: "0x04291180af677efd464432138d2a5d766a3898d4",
  id: 1,
  infrastructure: "ETH",
  labels: [],
  notes: nil
}
```

So, in my opinion, having this Map.get/3 with a short comment on top is better than making 2 functions with almost identical body (copy-paste of the Dataloader type)

tspenov: ok, lets go that way

lib/sanbase_web/admin/blockchain_address/blockchain_address_user_pair_label.ex
tspenov: Aha it is actually in `lib/sanbase_web/admin` path. IMO even if it used only for admin - the schema module  is better to be in its own file in`lib/sanbase/...` path.

tspenov (approved):

PR #2432, opened 2021-01-04
Evaluate watchlist functions during validation

IvanIvanoff:
## Changes

After all manual checks are done, finish the watchlist function validation by evaluating the function.
Creating a watchlist with a function that cannot be evaluated will run to constant runtime errors, so this is used as a final check. 
The drawback of this approach is that in case all the manual validations pass, we do not really know what is the reason for the function being broken and a reason cannot be reported.

## Checklist:

### DIFF
--- a/lib/sanbase/user_lists/watchlist_function.ex
+++ b/lib/sanbase/user_lists/watchlist_function.ex
@@ -115,7 +115,7 @@ defmodule Sanbase.WatchlistFunction do
   @doc ~s"""
   Checks if function evaluates. This is used as a last resort to checking if a
   function is valid as some edge cases can be missed. Creating a watchlist
-  with a function that cannot be evaulated will cause constant errors on runtime.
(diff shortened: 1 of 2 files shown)


lib/sanbase/user_lists/watchlist_function.ex
```diff
@@ -112,6 +112,23 @@ defmodule Sanbase.WatchlistFunction do
 
   def valid_function?(%__MODULE__{name: "empty"}), do: true
 
+  @doc ~s"""
+  Checks if function evaluates. This is used as a last resort to checking if a
+  function is valid as some edge cases can be missed. Creating a watchlist
+  with a function that cannot be evaulated will cause constant errors on runtime.
```
tspenov, line 118:
```suggestion
  with a function that cannot be evaluated will cause constant errors on runtime.
```

tspenov (approved):

PR #2434, opened 2021-01-05
Make tests easier to run

IvanIvanoff:
## Changes
Remove some tests that use influxdb. Add a flag `influx_store_enabled`
and set it to false in test env so the exporters do not push data to
influx

## Checklist:


test/sanbase_web/graphql/social_data/twitter_api_test.exs
```diff
@@ -1,196 +0,0 @@
-defmodule Sanbase.Github.TwitterApiTest do
```
tspenov, line 1: Don't we still need these tests ?

lib/sanbase/external_services/coinmarketcap/ticker_fetcher.ex
```diff
@@ -67,9 +67,11 @@ defmodule Sanbase.ExternalServices.Coinmarketcap.TickerFetcher do
     |> Enum.each(&store_latest_coinmarketcap_data!(&1, cmc_id_to_slugs_mapping))
 
     # Store the data in Influxdb
-    tickers
-    |> Enum.flat_map(&Ticker.convert_for_importing(&1, cmc_id_to_slugs_mapping))
-    |> Store.import()
+    if Application.get_env(:sanbase, :influx_store_enabled, true) do
```
tspenov, line 70: I guess that is a temporary thing ?

IvanIvanoff: Until we remove completely influxdb

test/sanbase_web/graphql/social_data/twitter_api_test.exs
IvanIvanoff: We cannot run these tests without Influx and the goal of this PR is to remove Influxdb as a test dependency.
I'll be moving twitter to CH in the next few days though, so these tests will be deleted either way.

tspenov: We can just mock `all_records_for_measurement` and `last_record_for_measurement` and not use influx but keep the tests.
After migrating to CH the api interface will be the same so tests can be reused. (maybe just the function names will change)

IvanIvanoff: Added the tests with mocks.

PR #2438, opened 2021-01-06
Sync user emails in sanbase and stripe

tspenov:
## Changes

User might change his email in Sanbase, so we want keep Sanbase in sync with Stripe.

## Checklist:

### DIFF
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -338,6 +338,24 @@ defmodule Sanbase.Auth.User do
     |> Repo.update()
   end
 
+  @doc """
+  Sync users' emails in Stripe.
+  User might change his email in Sanbase, so we want keep Sanbase in sync with Stripe.
+  """
+  def sync_subscribed_users_with_changed_email() do
+    from(u in __MODULE__,
+      where:
+        not is_nil(u.email) and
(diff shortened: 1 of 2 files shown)


lib/sanbase/auth/user.ex
```diff
@@ -338,6 +338,24 @@ defmodule Sanbase.Auth.User do
...
+        not is_nil(u.email) and
+          not is_nil(u.email_candidate_token_validated_at) and
+          not is_nil(u.stripe_customer_id),
+      select: %{email: u.email, stripe_customer_id: u.stripe_customer_id}
+    )
+    |> Repo.all()
+    |> Enum.each(fn %{email: email, stripe_customer_id: stripe_customer_id} ->
+      Stripe.Customer.update(stripe_customer_id, %{email: email})
+    end)
+  end
```
IvanIvanoff, line 357: Should we place this in another module?
Also, how many such users there are? Can we benefit by using `Sanbase.Parallel.map` instead?

tspenov: There are ~100 users. I can put a filter on `email_candidate_token_validated_at` > Timex.beginning_of_the_day(now) so
it takes only new ones since older are migrated.

IvanIvanoff: 100 is good and should be executed fast

PR #2443, opened 2021-01-07
Rework the README.md

IvanIvanoff:
## Checklist:


.env.example
```diff
@@ -15,3 +17,43 @@ HYDRA_CONSENT_URI=/oauth2/consent/requests
...
+# CLICKHOUSE_DATABASE_URL="ecto://default@clickhouse.stage.san:30901/default"
+# CLICKHOUSE_DATABASE_URL="ecto://default@clickhouse-facing.stage.san:31901/default"
+# METRICS_HUB_URL="http://metricshub.stage.san:31080/"
+# TECH_INDICATORS_URL="http://techindicators.stage.san:31080"
+# PARITY_URL="parity.stage.san:30954"
+# INFLUXDB_HOST="influxdb-influxdb.default.svc.cluster.local"
+# ETHAUTH_URL="http://ethauth"
+
+############
+##  Prod  ##
```
tspenov, line 39: I think the stage section is enough for the README.

.env.example
```diff
@@ -1,9 +1,11 @@
...
-PARITY_URL=https://parity.stage.internal.santiment.net/
-PARITY_BASIC_AUTH_USERNAME=
-PARITY_BASIC_AUTH_PASSWORD=
+DATABASE_URL="ecto://postgres:postgres@localhost:5432/sanbase_dev"
+CLICKHOUSE_DATABASE_URL="ecto://default@clickhouse.stage.san:30901/default"
 
+## Enable/Disable scrapers
 COINMARKETCAP_PRICES_ENABLED=true
 COINMARKETCAP_TICKERS_ENABLED=true
 ETHERSCAN_CRAWLER_ENABLED=true
```
tspenov, line 11: we can set these to false by default

README.md
```diff
@@ -1,27 +1,130 @@
...
+- Elasticsearch
+- Parity
+
+### Running outside Docker
+
+This is the preferred way to run the project if you are going to work on the
+project regularly because this way the development process and build times are
+much faster.
+
+[Running sanbase locallylocally](docs/sanbase-local-development.md)
```
tspenov, line 104:
```suggestion
[Running sanbase locally](docs/sanbase-local-development.md)
```

.env.example
IvanIvanoff: This is not the readme, but the `.env.example` - it can contain all non-secret links

.env.example
IvanIvanoff: Actually almost all of these are not used anymore. I'll remove them completely.

.env.example
```diff
@@ -1,17 +1,53 @@
...
-PARITY_URL=https://parity.stage.internal.santiment.net/
-PARITY_BASIC_AUTH_USERNAME=
-PARITY_BASIC_AUTH_PASSWORD=
-
-COINMARKETCAP_PRICES_ENABLED=true
-COINMARKETCAP_TICKERS_ENABLED=true
-ETHERSCAN_CRAWLER_ENABLED=true
-TWITTER_SCRAPER_ENABLED=true
+DATABASE_URL="ecto://postgres:postgres@localhost:5432/sanbase_dev"
+CLICKHOUSE_DATABASE_URL="ecto://default@clickhouse.stage.san:30901/default"
```
tspenov, line 6: Lets use `chproxy` and `sanbase` as in prod. 
The url should be: `ecto://sanbase@clickhouse-proxy.stage.san:30901/default`

.env.example
tspenov: Yes they are not secret but we should not encourage using them from local dev.

PR #2461, opened 2021-01-13
Improve signals payload

IvanIvanoff:
## Changes

Extend the metrics signal payload with some short data details like:
```elixir
# When aggregation is first or last
"Generated by the value of the metric at 10 Jan 2021 15:00 UTC"
```

```elixir
# When aggregation is average. It is the same for min/max
"Generated by the average value of the metric in the interval: 10 Jan 2021 15:00 UTC  11 Jan 2021 15:00 UTC"
```

```elixir
# When aggregation is sum
"Generated by the sum of all metric values in the interval: 10 Jan 2021 15:00 UTC  11 Jan 2021 15:00 UTC"
```

## Checklist:


lib/sanbase/utils/datetime/utils.ex
```diff
@@ -1,4 +1,12 @@
 defmodule Sanbase.DateTimeUtils do
+  @doc ~s"""
+  Return a human readable representation of a datetime
+  """
+  def to_human_readable(datetime) do
+    datetime
+    |> Timex.format!("{0D} {Mshort} {YYYY} {h24}:{m} Z")
```
tspenov, line 7: I think `UTC` is more human readable than `Z`

IvanIvanoff: `Z` is more widely used to represent UTC datetime in many RFCs.
But now that you mention it, our users are not programmers, so the decision making should not be lead by this. I'll change to `UTC` in this case.

PR #2462, opened 2021-01-13
Add balance in usd in assets_held_by_address

tspenov:
## Changes
Added `balanceUsd` in assets held by address api

```graphql
      {
        assetsHeldByAddress(
          selector: {infrastructure: "#{infrastructure}", address: "#{address}"}
        ){
            slug
            balance
            balanceUsd
        }
      }
```

## Checklist:


IvanIvanoff:
@tspenov This implementation is ok, but I think ultimately it should be resolved via the Dataloader only when it's requested. I imagine we won't always need the `usd` balance but with this implementation it will always fetch it. Can you add a `# TODO` comment on the `balance_usd` field so we don't forget?

IvanIvanoff (approved):

PR #2467, opened 2021-01-15
Support CLICKHOUSE_REPO_ENABLED env var

IvanIvanoff:
## Changes

Support `CLICKHOUSE_REPO_ENABLED` env var that is used in the local development process. Currently, if sanbase is started without a CH connection it just crashes as the CH Ecto Repo cannot be started. The `MarkExchanges` service also needs CH connection in its initiation process.

Now sanbase can be started as `CLICKHOUSE_REPO_ENABLED=false iex -S mix phx.server` and it won't crash.
This can be used while the current task does not make any CH queries - all the postgres related data. This also makes it easy to work on something without internet connection.

## Checklist:

### DIFF
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -173,7 +173,10 @@ defmodule Sanbase.Application do
       SanbaseWeb.Telemetry,
 
       # Start the Clickhouse Repo
-      start_in(Sanbase.ClickhouseRepo, [:prod, :dev]),
+      start_if(
+        fn -> start_in(Sanbase.ClickhouseRepo, [:prod, :dev]) end,
+        fn -> System.get_env("CLICKHOUSE_REPO_ENABLED", "true") |> String.to_existing_atom() end
+      ),
 
       # Star the API call service
       Sanbase.ApiCallLimit.ETS,
--- a/lib/sanbase/application.ex
+++ b/lib/sanbase/application.ex
@@ -174,8 +174,11 @@ defmodule Sanbase.Application do
 
       # Start the Clickhouse Repo
       start_if(
-        fn -> start_in(Sanbase.ClickhouseRepo, [:prod, :dev]) end,
-        fn -> System.get_env("CLICKHOUSE_REPO_ENABLED", "true") |> String.to_existing_atom() end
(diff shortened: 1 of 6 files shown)


lib/sanbase/application.ex
```diff
@@ -173,7 +173,10 @@ defmodule Sanbase.Application do
       SanbaseWeb.Telemetry,
 
       # Start the Clickhouse Repo
-      start_in(Sanbase.ClickhouseRepo, [:prod, :dev]),
+      start_if(
+        fn -> start_in(Sanbase.ClickhouseRepo, [:prod, :dev]) end,
+        fn -> System.get_env("CLICKHOUSE_REPO_ENABLED", "true") |> String.to_existing_atom() end
```
tspenov, line 178: Can we reuse `Sanbase.ClickhouseRepo.enabled?()` here ?

IvanIvanoff: Actually I introduced it in order to be used here, but forgot to replace it.

IvanIvanoff: One more thing: It looks weird to have part of the check in the first function and the second check in the second function. I've reworked it.

lib/sanbase/clickhouse/mark_exchanges.ex
```diff
@@ -118,27 +118,30 @@ defmodule Sanbase.Clickhouse.MarkExchanges do
...
-          Process.send_after(self(), :set_state, 5_000)
-
-          Map.put(%{}, :exchange_wallets_set, MapSet.new())
-          |> Map.put(:updated_at, Timex.now())
-      end
+  defp fill_state() do
+    ch_repo_enabled? = Sanbase.ClickhouseRepo.enabled?()
+    env = Application.get_env(:sanbase, :env)
+
+    case ch_repo_enabled? and env != "test" do
```
tspenov, line 125: we compare with atoms above -> `Application.get_env(:sanbase, :env) in [:dev, :prod]`.

IvanIvanoff: I accidentally merged the PR, fixing it here: https://github.com/santiment/sanbase2/pull/2472

PR #2474, opened 2021-01-18
Add free subscription for staked users

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -170,6 +170,7 @@ defmodule Sanbase.Auth.User do
   defdelegate san_balance!(user), to: __MODULE__.SanBalance
 
   # Uniswap San Staking functions
+  defdelegate fetch_all_staked_users(), to: __MODULE__.UniswapStaking
   defdelegate update_all_san_staked_users(), to: __MODULE__.UniswapStaking
   defdelegate fetch_san_staked_user(user), to: __MODULE__.UniswapStaking
 
@@ -371,6 +372,21 @@ defmodule Sanbase.Auth.User do
       distinct: true
     )
     |> Repo.all()
+  end
+
+  @doc """
+  Mark user as registered.
+  Sign up with email has 2 phases.
+  1. Creating user object and sending token to user email.
+  2. Verifying the token from email.
+  Only after the token is verified the user is marked as registered.
+  """
+  def mark_as_registered(%User{is_registered: true} = user), do: {:ok, user}
+
+  def mark_as_registered(%User{is_registered: false} = user) do
+    user
+    |> User.changeset(%{is_registered: true})
+    |> Repo.update()
   end
 
   defp can_remove_eth_account?(%User{id: user_id, email: email}, address) do
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -170,9 +170,9 @@ defmodule Sanbase.Auth.User do
   defdelegate san_balance!(user), to: __MODULE__.SanBalance
 
   # Uniswap San Staking functions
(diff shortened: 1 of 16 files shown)


lib/sanbase/auth/user/uniswap_staking.ex
```diff
@@ -40,22 +50,47 @@ defmodule Sanbase.Auth.User.UniswapStaking do
...
+    end)
+    |> case do
+      {:ok, result} ->
+        Logger.info("Finished update_all_san_staked_users ok: #{inspect(result)}")
+        {:ok, result}
+
+      {:error, reason} ->
+        Logger.error("Finished update_all_san_staked_users error: #{inspect(reason)}")
+        {:error, reason}
+    end
```
IvanIvanoff, line 71: This could be done more idiomatically with Ecto.Multi

lib/sanbase/auth/user/uniswap_staking.ex
```diff
@@ -40,22 +50,47 @@ defmodule Sanbase.Auth.User.UniswapStaking do
...
   end
 
   @doc """
   Fetch the total SAN tokens staked as liquidity provider in Uniswap
   pair contracts that we follow for all connected user wallets.
   """
-  @spec fetch_san_staked_user(%User{}) :: float()
-  def fetch_san_staked_user(user) do
+  @spec fetch_san_staked_user(%User{} | non_neg_integer()) :: float()
+  def fetch_san_staked_user(%User{} = user) do
```
IvanIvanoff, line 79: We are using `staked_used` as a user who has a connected wallet with SAN tokens (in code and in tests). Here I think it would be better to indicate `uniswap` either in the name of the function or through a parameter.

lib/sanbase/auth/user/uniswap_staking.ex
```diff
@@ -40,22 +50,47 @@ defmodule Sanbase.Auth.User.UniswapStaking do
...
+
+  def fetch_san_staked_user(user_id) do
+    {:ok, user} = User.by_id(user_id)
+    user = user |> Repo.preload(:eth_accounts)
+
+    calc_san_staked_user(user)
+  end
+
+  # Helpers
+  defp calc_san_staked_user(user) do
```
IvanIvanoff, line 93: You can pass directly the `eth_accounts` here instead of the full user and have a pattern for `[]` to directly return

lib/sanbase/auth/user.ex
```diff
@@ -226,7 +227,7 @@ defmodule Sanbase.Auth.User do
         %User{}
         |> User.changeset(user_create_attrs)
         |> Repo.insert()
-        |> maybe_create_sign_up_trial(attrs)
+        |> maybe_create_free_or_trial_subscription(attrs)
```
IvanIvanoff, line 230: Should this be 2 separate functions? Or the name should be more generic like:
```elixir
maybe_give_free_access()
```

tspenov: It can't be 2 separate funcs because if you give free subscription you can't give trial. 
`maybe_give_free_access()` is shorter and more generic doesn't really describe it well so I prefer the longer one.

lib/sanbase/auth/user/uniswap_staking.ex
tspenov: This function is more probably to be called with a `user` than with user's eth_accounts.

lib/sanbase/auth/user/uniswap_staking.ex
tspenov: It is but here here it will be more verbose to use and don't see any benefits. The code after the case is just for logging 
so I can see if the job has finished correctly but it might be removed.

PR #2475, opened 2021-01-18
Update min interval to average_fees_usd to 15 min

tspenov:
## Checklist:


IvanIvanoff (changes requested):
This metric is fetched from daily_metrics_v2. The `dt` column type there is `Date`, so we don't have any intraday values.

tspenov:
close - we have the right intraday table represented by separate metric

PR #2478, opened 2021-01-19
Add seeds

IvanIvanoff:
## Changes

Add better seeds. The seeding process generates the same data as the seed to `:rand` is hardcoded.

## Checklist:


README.md
```diff
@@ -138,6 +138,14 @@ If you want to access an IEX console attached to the running server, run:
 bin/console.sh
 ```
 
+## Seeding data
+
+To seed example data run:
+
+```console
+INSIGHTS_DISCORD_NOTIFICATION_ENABLED=false CLICKHOUSE_REPO_ENABLED=false mix run lib/mix/seeds/seeds.exs
```
tspenov, line 146: Why do we need this ? BTW that should be false by default in dev.exs

IvanIvanoff: Because it will try to send a notification in Discord during seeding. I'll move it to dev.exs

PR #2480, opened 2021-01-20
Add discord link to projects

atanasormanow:
## Changes
- add a discord link column in the project table
- copy all discord links from the slack link column to the discord one 

## Checklist:


priv/repo/migrations/20210120090715_migrate_slack_link_to_discord.exs
```diff
@@ -0,0 +1,40 @@
...
+      |> Repo.update!()
+    end)
+  end
+
+  defp transform_discord_link(link) do
+    long_prefix = "https://discordapp.com/invite/"
+    short_prefix = "https://discord.gg/"
+
+    String.replace_prefix(link, long_prefix, short_prefix)
+  end
```
IvanIvanoff, line 39: For such cases using pattern matching looks very nice as well:
```elixir
defp transform_link("https://discordapp.com/invite/" <> rest), do: "https://discord.gg/" <> rest
defp transform_link(link), do: link
```

atanasormanow: Looks better to be honest, ill do it your way

test/sanbase_web/graphql/projects/project_api_test.exs
```diff
@@ -265,6 +265,48 @@ defmodule SanbaseWeb.Graphql.ProjectApiTest do
...
+        redditLink
+      }
+    }
+    """
+
+    result =
+      context.conn
+      |> post("/graphql", query_skeleton(query, "projectBySlug"))
+      |> json_response(200)
+      |> get_in(["data", "projectBySlug"])
```
tspenov, line 299: Just FYI we have a helper in `SanbaseWeb.Graphql.TestHelpers` for the above 4 lines because they are so common in tests.
They can be written as:
```elixir
result = execute_query(context.conn, query, "projectBySlug")
```

priv/repo/migrations/20210120090715_migrate_slack_link_to_discord.exs
```diff
@@ -0,0 +1,36 @@
...
+
+  defp migrate_discord_links() do
+    Project.List.projects_by_non_null_field(
+      :slack_link,
+      include_hidden_projects?: true
+    )
+    |> Enum.filter(fn %Project{slack_link: link} ->
+      String.contains?(link, "discord")
+    end)
+    |> Enum.map(fn %Project{slack_link: link} = project ->
```
tspenov, line 27: You can use `Enum.each` instead of `Enum.map` cause you don't care about the result here

priv/repo/migrations/20210120090715_migrate_slack_link_to_discord.exs
```diff
@@ -0,0 +1,36 @@
...
+      Project.changeset(project, %{discord_link: link})
+      |> update_change(:discord_link, &transform_discord_link/1)
```
tspenov, line 29: I think `update_change` is redundant. You can transform discord link on the previous line ?

tspenov, line 29: Also `transform_discord_link` doesn't exist - there is  `transform_link`

tspenov:
@atanasormanow I see you've ran the migration locally but probably you didn't have any records so the code path with missing function `transform_discord_link` was not executed. I suggest you insert a record in projects and rerun the migration and then verify in database that the change really happened.

priv/repo/migrations/20210120090715_migrate_slack_link_to_discord.exs
atanasormanow: Fair, will do

priv/repo/migrations/20210120090715_migrate_slack_link_to_discord.exs
atanasormanow: Yeah it is more performance costly, i was wondering if it is worth using in this case. I will just do it on the previous line as you said.

IvanIvanoff (approved):

tspenov (approved):

PR #2485, opened 2021-01-21
Make pagination work with erc20 filter

IvanIvanoff:
## Changes
Fix the following query:
```graphql
{
  allProjectsByFunction(function: "{\"args\":{\"filters\":[],\"pagination\":{\"page\":1,\"page_size\":10}},\"name\":\"selector\"}"){ 
    stats{	
      projectsCount
    }
    projects{
      id
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/model/project/list/list.ex
+++ b/lib/sanbase/model/project/list/list.ex
@@ -474,8 +474,13 @@ defmodule Sanbase.Model.Project.List do
   end
 
   defp erc20_projects_query(opts) do
+    projects_query(opts)
+    |> filter_erc20_projects()
+  end
+
+  defp filter_erc20_projects(query) do
     from(
-      p in projects_query(opts),
+      p in query,
       inner_join: infr in assoc(p, :infrastructure),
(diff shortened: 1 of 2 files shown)


lib/sanbase/model/project/list/list.ex
```diff
@@ -474,8 +474,13 @@ defmodule Sanbase.Model.Project.List do
   end
 
   defp erc20_projects_query(opts) do
+    projects_query(opts)
+    |> filter_erc20_projects()
```
tspenov, line 478: `project_query(opts)` - might also call `filter_erc20_projects()`, so this might be called twice.
What is the the ecto query in this case ?

IvanIvanoff: Good catch. I've pushed a fix - it's not the prettiest but it works fine.
Also, I think it won't be the case that you call this function and provide `included_slugs: :erc20` but it's good to have a guard against this.

tspenov (approved):

PR #2488, opened 2021-01-22
Rework balance usd with dataloader

atanasormanow:
## Changes
- Use a custom resolver for the `balance_usd` field
- Remove the old function for getting the data - `add_balance_usd`
- Sort the assets returned from `assets_held_by_address` by balance
- Do not expect `balance_usd` in the tests for `assets_held_by_address`
- Rewrite the api tests concerning `balance_usd` with `Sanbase.Mock`

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
@@ -70,7 +70,7 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
                {:ok, ethereum_assets} <- EthBalance.assets_held_by_address(address) do
       sorted_assets =
         (erc20_assets ++ ethereum_assets)
-        |> Enum.sort_by(&Map.get(&1, :balance), &>=/2)
+        |> Enum.sort_by(& &1.balance, :desc)
 
       {:ok, sorted_assets}
     end
--- a/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
+++ b/lib/sanbase/clickhouse/historical_balance/historical_balance.ex
@@ -67,9 +67,12 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   @spec assets_held_by_address(map()) :: {:ok, list(map())} | {:error, String.t()}
   def assets_held_by_address(%{infrastructure: "ETH", address: address}) do
     async with {:ok, erc20_assets} <- Erc20Balance.assets_held_by_address(address),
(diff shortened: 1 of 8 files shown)


lib/sanbase_web/graphql/dataloader/price_dataloader.ex
```diff
@@ -42,6 +42,24 @@ defmodule SanbaseWeb.Graphql.PriceDataloader do
     |> Map.new()
   end
 
+  def query(:last_price_usd, slugs) do
+    now = Timex.now()
+    yesterday = Timex.shift(Timex.now(), days: -1)
```
tspenov, line 47: Since we need only the last value - the `from` param can be even `now - 1 hour`

tspenov:
@atanasormanow There are 2 failing tests that are related

lib/sanbase_web/graphql/dataloader/price_dataloader.ex
IvanIvanoff: I think we should leave it as 1 day so we have time to scrape some missing prices. This can happen if Kafka has some issues for example or the price is lagging in the source where we're getting it.

lib/sanbase/clickhouse/historical_balance/historical_balance.ex
```diff
@@ -67,9 +67,12 @@ defmodule Sanbase.Clickhouse.HistoricalBalance do
   @spec assets_held_by_address(map()) :: {:ok, list(map())} | {:error, String.t()}
   def assets_held_by_address(%{infrastructure: "ETH", address: address}) do
     async with {:ok, erc20_assets} <- Erc20Balance.assets_held_by_address(address),
-               {:ok, ethereum} <- EthBalance.assets_held_by_address(address) do
-      assets = add_usd_balance(ethereum ++ erc20_assets)
-      {:ok, assets}
+               {:ok, ethereum_assets} <- EthBalance.assets_held_by_address(address) do
+      sorted_assets =
+        (erc20_assets ++ ethereum_assets)
+        |> Enum.sort_by(&Map.get(&1, :balance), &>=/2)
```
IvanIvanoff, line 73: For clarity, you can use `:desc` instead of providing a function:
```elixir
[%{a: 1}, %{a: 3}, %{a: 2}] |> Enum.sort_by( & &1.a, :desc)
```

atanasormanow: Done

PR #2491, opened 2021-01-25
Atomize only subset of promoter fields

tspenov:
## Checklist:


IvanIvanoff (approved):

IvanIvanoff:
Do we have a test passing through this?

We can a have a test that passes all the fields (including the excluded ones), so if we add a new field in this filter without having it in the ecto schema it can break on the `String.to_existing_atom/1` call

tspenov:
It is actually in the graphql schema.
The tests here pass the whole promoter response to `extract_and_atomize_needed_fields`: https://github.com/santiment/sanbase2/blob/master/test/sanbase_web/graphql/promoter/promoter_api_test.exs
Adding a new field to the filter will actually break the test but I'll add a comment in the filter.

PR #2493, opened 2021-01-25
Rework access checker and access restrictions

IvanIvanoff:
## Changes

- Fix issue when `realtime_days_cut_off` is `0` to properly put `null` in `restrictedTo`.
- Fix function args inconsistencies (use `plan, product, metric` order in most places)
- Add caching to get_access_restrictions

## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/middlewares/access_control.ex
+++ b/lib/sanbase_web/graphql/middlewares/access_control.ex
@@ -72,9 +72,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
            resolution
        ) do
     plan = context[:auth][:plan] || :free
-    subscription = context[:auth][:subscription] || @free_subscription
-    product_id = subscription.plan.product_id || context.product_id
-    product = Product.code_by_id(product_id)
+    product = Product.code_by_id(context[:product_id]) || "SANAPI"
 
     case Plan.AccessChecker.plan_has_access?(plan, product, query_or_metric) do
       true ->
@@ -275,7 +273,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
         %{restricted_from: restricted_from, restricted_to: restricted_to} =
           Sanbase.Billing.Plan.Restrictions.get(
             context[:__query_or_metric_atom_name__],
-            context[:auth][:subscription],
+            context[:auth][:plan],
             context[:product_id]
           )
 
--- a/lib/sanbase_web/graphql/middlewares/access_control.ex
+++ b/lib/sanbase_web/graphql/middlewares/access_control.ex
@@ -47,7 +47,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
   end
 
   # The name of the query/mutation can be passed in snake case or camel case.
-  # Here we transform the name to an altom in snake case for consistency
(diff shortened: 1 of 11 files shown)


lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -189,10 +187,10 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
     product_id = subscription.plan.product_id || context.product_id
 
     historical_data_in_days =
-      Subscription.historical_data_in_days(subscription, query, product_id)
+      Subscription.historical_data_in_days(subscription, product_id, query)
```
tspenov, line 190: We can use `AccessChecker.historical_data_in_days(subscription.plan, product_id, query)` and remove the one from `Subscription`. Same for `realtime_data_cut_off_in_days`

lib/sanbase_web/graphql/resolvers/access_control_resolver.ex
```diff
@@ -1,14 +1,20 @@
...
-  alias Sanbase.Billing.Subscription
+  alias Sanbase.Billing.Product
+  alias SanbaseWeb.Graphql.Cache
 
-  def get_access_restrictions(_root, _args, %{
-        context: %{product_id: product_id} = context
-      }) do
-    subscription = context[:auth][:subscription] || Subscription.free_subscription()
+  def get_access_restrictions(_root, args, %{context: context}) do
+    plan = Map.get(args, :plan) || context[:auth][:plan] || :free
```
tspenov, line 6: I don't think we pass any args to `get_access_restrictions` so `Map.get(args, :plan)` won't work. But I think we do have it always here `context[:auth][:plan]` :?

lib/sanbase_web/graphql/resolvers/access_control_resolver.ex
```diff
@@ -1,14 +1,20 @@
...
 
-    restrictions = Sanbase.Billing.Plan.Restrictions.get_all(subscription, product_id)
+    product_id =
+      Product.id_by_code(Map.get(args, :product)) || context[:product_id] || Product.product_api()
```
tspenov, line 9: Same here.

lib/sanbase_web/graphql/resolvers/access_control_resolver.ex
tspenov: Ok, I didn't bother to scroll below. I saw that we provide plan and product as optional args

lib/sanbase_web/graphql/middlewares/access_control.ex
IvanIvanoff: Done

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -41,7 +47,7 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
   end
 
   # The name of the query/mutation can be passed in snake case or camel case.
-  # Here we transform the name to an atom in snake case for consistency
+  # Here we transform the name to an altom in snake case for consistency
```
tspenov, line 50:
```suggestion
  # Here we transform the name to an atom in snake case for consistency
```

tspenov (approved):

PR #2495, opened 2021-01-27
Switch asset_prices to asset_prices_v3

apozdeyev:
## Changes

To optimize requesting data from asset_prices table on facing CH, we created a new table with replicated engine instead of distributed one.

## Checklist:

### DIFF
--- a/bin/clickhouse/tables_used
+++ b/bin/clickhouse/tables_used
@@ -1,7 +1,7 @@
 anomalies_v2
 api_call_data
 asset_metadata
-asset_prices
+asset_prices_v3
 blockchain_address_labels
 bnb_balances
 btc_balances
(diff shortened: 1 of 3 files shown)


IvanIvanoff:
@apozdeyev Is this table present only the facing cluster or is it also in the default cluster? If yes - this might be first such 
case. This might have some impact on the development process and would force devs to work only with the facing cluster. This could be an issue when someone wants to work on a table that has still not be added to the facing cluster and uses the default one (this was the case adding the eth2 metrics)

apozdeyev:
> @apozdeyev Is this table present only the facing cluster or is it also in the default cluster? If yes - this might be first such
> case. This might have some impact on the development process and would force devs to work only with the facing cluster. This could be an issue when someone wants to work on a table that has still not be added to the facing cluster and uses the default one (this was the case adding the eth2 metrics)

it presents on both clusters. all tables which we add to facing CH should have the same name and the same data.

IvanIvanoff (approved):

PR #2496, opened 2021-01-27
Add Kaiko price scraping for a subset of the slugs

IvanIvanoff:
## Changes
```graphql
{
  getMetric(metric: "price_usd") {
    timeseriesData(
      selector: {slug: "bitcoin", source: "kaiko"}
      from: "utc_now-30m"
      to: "utc_now"
      interval: "30s"
      cacheParams: { baseTtl: 5, maxTtlOffset: 5 } ) {
        value
        datetime
    }
  }
}
```

## Checklist:


lib/sanbase/scrapers/kaiko.ex
```diff
@@ -0,0 +1,128 @@
...
+      fn %{price: price, datetime: datetime} ->
+        key = @source <> "_" <> slug <> "_" <> DateTime.to_iso8601(datetime)
+
+        value =
+          %{
+            timestamp: DateTime.to_unix(datetime),
+            source: @source,
+            slug: slug,
+            price_usd: (quote_asset == "usd" && price) || nil,
+            price_btc: (quote_asset == "btc" && price) || nil,
```
tspenov, line 110: I think it is better to export both prices with one call ?

lib/sanbase/scrapers/kaiko.ex
```diff
@@ -0,0 +1,128 @@
...
+    Logger.info("Scraping Kaiko prices for #{length(pairs())} pairs")
+
+    Sanbase.Parallel.map(
+      pairs(),
+      fn {base_asset, quote_asset, slug} ->
+        identifier = base_asset <> "/" <> quote_asset
+
+        timerange = %{
+          start_time: last_datetime_scraped(identifier),
+          end_time: Timex.now(),
```
tspenov, line 27: Should `end_time` be the same for all assets scraped ?

lib/sanbase/scrapers/kaiko.ex
IvanIvanoff: Fixed

lib/sanbase/scrapers/kaiko.ex
IvanIvanoff: Fixed.
Although, this is not an issue because depending on the `end_time` the last datetime stored in postgres will be different. This makes sure that we're always scraping from where we stopped the previous time.

lib/sanbase/scrapers/kaiko.ex
```diff
@@ -0,0 +1,177 @@
...
+      %{
+        price_usd: nil,
+        price_btc: nil,
+        marketcap_usd: nil,
+        volume_usd: nil
+      }
+      |> Map.merge(map)
+      |> Map.put(:datetime, datetime)
+    end)
+    |> Enum.reject(fn elem -> is_nil(elem.price_usd) and is_nil(elem.price_btc) end)
```
tspenov, line 146: We store even if one of the prices is nil ? Should we not store at all if one of the prices is nil ?

IvanIvanoff: Because the prices are based on trades alone, I think it is highly possible that the price is provided only when such a trade happens.
Not 100% sure though

tspenov (approved):

PR #2497, opened 2021-01-28
Introduce getSignal API with the data from signals CH table

IvanIvanoff:
#### Add getSignal API, recycling most of the old anomaly API
Since the `getAnomaly` API is not used and the data for anomalies hasn't been updated in almost a year, it had to be removed. The signals table has a similar structure, so a big part of the old functionality was recycled. 

- Add Signal behaviour.
- Add `signal_adapter.ex`, implementing the Signal behaviour.
- Add `signal.ex` as a dispatch module, i.e. only calls the above adapter. (there will be further changes to make this work similar to `Sanbase.Metric.Helper`)
- Do a mass renaming `anomaly -> signal` for the old, but still usable logic for anomalies.
- Move/Rename some files/folders to match the new structure.
- Reuse and fix anomaly tests
- Fix some typoos in a couple of comments.

## Checklist:

---
  The `timeseries_data` and `aggregated_timeseries_data` support `selector` with `slug` or `slugs` as argument.
```graphql
{
  getSignal(signal: "dai_mint") {
    availableSince(slug: "multi-collateral-dai")
  }
}
```

<details>
<summary>Result </summary>

```graphql
{
  "data": {
    "getSignal": {
      "availableSince": "2019-11-21T11:25:06Z"
    }
  }
}
```
</details>

```graphql
{
  getSignal(signal: "dai_mint") {
    metadata {
      availableAggregations
      availableSlugs
      dataType
      defaultAggregation
      minInterval
      signal
    }
  }
}
```

<details>
<summary>Result</summary>

```graphql
{
  "data": {
    "getSignal": {
      "metadata": {
        "availableAggregations": [
          "ANY",
          "SUM",
          "AVG",
          "MIN",
          "MAX",
          "LAST",
          "FIRST",
          "MEDIAN",
          "COUNT"
        ],
        "availableSlugs": [
          "multi-collateral-dai"
        ],
        "dataType": "TIMESERIES",
        "defaultAggregation": "COUNT",
        "minInterval": "1m",
        "signal": "dai_mint"
      }
    }
  }
}
```
</details>

```graphql
{
  getSignal(signal: "dai_mint") {
    timeseriesData(slug: "multi-collateral-dai", from: "2020-12-01T00:00:00Z", to: "2021-01-01T00:00:00Z", aggregation: COUNT) {
      value
      datetime
    }
  }
}
```

<details>
<summary>Result</summary>

```graphql
{
  "data": {
    "getSignal": {
      "timeseriesData": [
        {
          "datetime": "2020-12-01T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-04T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-16T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-17T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-20T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-23T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-28T00:00:00Z",
          "value": 1
        },
        {
          "datetime": "2020-12-30T00:00:00Z",
          "value": 2
        },
        {
          "datetime": "2020-12-31T00:00:00Z",
          "value": 1
        }
      ]
    }
  }
}
```

</details>

```graphql
{
  getSignal(signal: "dai_mint") {
    aggregatedTimeseriesData(selector: {slugs: ["multi-collateral-dai"]}, from: "2020-01-01T00:00:00Z", to: "2021-02-01T00:00:00Z", aggregation: COUNT)
  }
}
```

<details>
<summary>Result </summary>

```graphql
{
  "data": {
    "getSignal": {
      "aggregatedTimeseriesData": 42
    }
  }
}
```
</details>


lib/sanbase/signal/file_handler.ex
```diff
@@ -0,0 +1,75 @@
...
+  #    if the data is queried with interval bigger than 'min_interval'
+  #  - min_interval - the minimal interval the data is available for
+  #  - table - the table name in ClickHouse where the signal is stored
+
+  @signals_file "signal_files/available_signals.json"
+  @external_resource available_signals_file = Path.join(__DIR__, @signals_file)
+  @signals_json File.read!(available_signals_file) |> Jason.decode!()
+  @aggregations Sanbase.Metric.SqlQuery.Helper.aggregations()
+
+  @metric_map Helper.name_to_field_map(@signals_json, "metric")
```
IvanIvanoff, line 45: `metric` => `signal`. This rename should be done in the JSON file and the function names and module attribute names.

lib/sanbase/signal/file_handler.ex
```diff
@@ -0,0 +1,75 @@
...
+  @access_map Helper.name_to_field_map(@signals_json, "access", &String.to_atom/1)
+  @table_map Helper.name_to_field_map(@signals_json, "table")
+  @aggregation_map Helper.name_to_field_map(@signals_json, "aggregation", &String.to_atom/1)
+  @min_interval_map Helper.name_to_field_map(@signals_json, "min_interval")
+  @human_readable_name_map Helper.name_to_field_map(@signals_json, "human_readable_name")
+  @data_type_map Helper.name_to_field_map(@signals_json, "data_type", &String.to_atom/1)
+  @metric_to_signal_map Helper.fields_to_name_map(@signals_json, [
+                          "metric"
+                        ])
```
IvanIvanoff, line 54: There should no longer be such a mapping - the function that returns this module attribute is unused as well.

lib/sanbase/signal/signal.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Sanbase.Signal do
```
IvanIvanoff, line 1: Add a moduledoc similar to the one in Sanbase.Metriv

lib/sanbase/signal/signal.ex
```diff
@@ -0,0 +1,35 @@
...
+  alias Sanbase.Signal.SignalAdapter
```
IvanIvanoff, line 2: Add a comment here:
```suggestion
  # As there is a single signal adapter now, the dispatching is done directly.
  # After a second signals source is introduced, a dispatching logic similar to the one
  # found in Sanbase.Metric should be implemented.
  alias Sanbase.Signal.SignalAdapter
```

lib/sanbase/signal/signal.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  def metadata(signal) do
+    SignalAdapter.metadata(signal)
+  end
+
+  def first_datetime(signal, slug) do
+    SignalAdapter.first_datetime(signal, slug)
+  end
+
+  def timeseries_data(signal, slug, from, to, interval, aggregation) do
```
IvanIvanoff, line 28: The second param should be a `selector` - this is a map and one of its possible values is `%{slug: slug}`. You can pattern match in the function headers (for example in `first_datetime(signal, %{slug: slug})`) where you explicitly need a single slug provided.

The last parameter should be `opts` - a keyword list that has a default value of empty list. One of its keys should be `aggregation`. One obvious benefit is that this way when you want to use the default aggregation you need to pass `nil` as the last argument, which has ambiguous meaning to the readers. 

Both these suggestions are implemented in the metric module and can be seen there. So, in the end, the function definition should look like this:
```elixir
def timeseries_data(signal, selector, from, to, interval, opts \\ [])

def timeseries_data(signal, selector, from, to, interval, opts) do
  SignalAdapter.timeseries_data(signal, selector, from, to, interval, opts)
end
```

And the code in the signal adapter:
```elixir
  def timeseries_data(_signal, %{slug: []}, _from, _to, _interval, _opts), do: {:ok, []}

  def timeseries_data(signal, %{slug: slug}, from, to, interval, opts) do
    aggregation = Keyword.get(opts, :aggregation) || Map.get(@aggregation_map, signal)
    ....
```

(The first case with `slug: []` is some micro-optimization so we don't pass through much of the logic in case of empty list for slugs. This is mostly the case when we use for alerts or any other logic where a filter is applied that can lead to empty list)

lib/sanbase/signal/signal_adapter.ex
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  def available_aggregations(), do: @aggregations
+
+  @impl Sanbase.Signal.Behaviour
+  def available_signals(), do: @signals
+
+  @impl Sanbase.Signal.Behaviour
+  def available_signals(slug) do
```
IvanIvanoff, line 30: This is one example where the above mentioned `selector` object (instead of `slug`) has meaning only when it holds a single slug. So, this can be rewritten as:
```elixir
def available_signals(%{slug: slug}) when is_binary(slug) do
```

lib/sanbase/signal/signal_adapter.ex
```diff
@@ -0,0 +1,119 @@
...
+  @impl Sanbase.Signal.Behaviour
+  def metadata(signal) do
+    default_aggregation = Map.get(@aggregation_map, signal)
+
+    {:ok,
+     %{
+       signal: signal,
+       metric: Map.get(@metric_map, signal),
+       min_interval: Map.get(@min_interval_map, signal),
+       default_aggregation: default_aggregation,
```
IvanIvanoff, line 51: You can just do the `Map.get` here as in the previous 2 cases instead of binding the value to a variable.

lib/sanbase/signal/signal_adapter.ex
```diff
@@ -0,0 +1,119 @@
...
+    |> case do
+      {:ok, [result]} -> {:ok, result}
+      {:error, error} -> {:error, error}
+    end
+  end
+
+  @impl Sanbase.Signal.Behaviour
+  def timeseries_data(signal, slug, from, to, interval, aggregation) do
+    default_aggregation = Map.get(@aggregation_map, signal)
+    aggregation = aggregation || default_aggregation
```
IvanIvanoff, line 71: The comment https://github.com/santiment/sanbase2/pull/2497/files#r576153976 refers to this place

lib/sanbase_web/graphql/resolvers/signal_resolver.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  alias Sanbase.Signal
+
+  require Logger
+
+  @datapoints 300
+
+  def get_signal(_root, %{signal: signal}, _resolution) do
+    case Signal.has_signal?(signal) do
+      true -> {:ok, %{signal: signal}}
```
tspenov, line 18: I think `signal` here is a string. Shouldn't we return signal object

lib/sanbase_web/graphql/schema/types/signal_types.ex
```diff
@@ -80,20 +73,40 @@ defmodule SanbaseWeb.Graphql.AnomalyTypes do
...
+    @desc ~s"""
+    A derivative of the `timeseriesData` - read its full descriptio if not
+    familiar with it.
+
+    `aggregatedTimeseriesData` returns a single float value instead of list
+    of datetimes and values. The single values is computed by aggregating all
+    of the values in the specified from-to range with the `aggregation` aggregation.
+    """
+    field :aggregated_timeseries_data, :float do
+      arg(:slug, non_null(:string))
```
IvanIvanoff, line 88: You should allow `selector`, too.
And because of this, both should not have `non_null(:string)` as either of them can be omitted if the other is present. This applies to all queries here that have `:slug`

lib/sanbase_web/graphql/resolvers/signal_resolver.ex
atanasormanow: Not sure, but i think metrics work this way too.

IvanIvanoff: This resolver has a somewhat strange behaviour/meaning that should probably be documented.

Long story short - it does not matter what it returns.

The `get_signal` returns a `signal` object. The top-level resolver should fill all fields that do not have a custom resolver. All values not present in the object definition are ignored.

This said - the `get_signal` resolver should return a map that fills the fields without resolvers but there are no such fields. So it would be ok if it just returns an empty map. The `%{signal: signal}` return is serving only one purpose - easily distribute the value to the other resolvers.

I can illustrate that with a few functions from the metric resolver:
```elixir

  def get_available_slugs(_root, _args, %{source: %{metric: metric}}),
    do: Metric.available_slugs(metric)

  def get_human_readable_name(_root, _args, %{source: %{metric: metric}}),
    do: Metric.human_readable_name(metric)

  def get_metadata(%{}, _args, %{source: %{metric: metric}} = resolution) do
  ```
  
The result of the top level resolver can be access easier and much cleaner than [root object arguments](https://github.com/santiment/sanbase2/blob/master/lib/sanbase_web/graphql/helpers/utils.ex#L42-L51).

Now that I think about it, both this `source` value and the `_root` should be the same, but in order to make sure I have to research it again.

File: lib/sanbase/signal/signal_adapter.ex [hunk 11]
```diff
@@ -0,0 +1,119 @@
...
+  def first_datetime(signal, %{slug: slug}) when is_binary(slug) do
+    {query, args} = first_datetime_query(signal, slug)
+
+    ClickhouseRepo.query_transform(query, args, fn [datetime] ->
+      DateTime.from_unix!(datetime)
+    end)
+    |> case do
+      {:ok, [result]} -> {:ok, result}
+      {:error, error} -> {:error, error}
+    end
```
IvanIvanoff, line 66: This code is is seen very often, so there's a helper function:
```elixir
  import Sanbase.Utils.Transform, only: [maybe_unwrap_ok_value: 1]
```
and you can replace the whole case with just `|> maybe_unwrap_ok_value()`

lib/sanbase/signal/sql_query/signal_sql_query.ex
```diff
@@ -0,0 +1,144 @@
...
+    )
+    INNER JOIN (
+      SELECT asset_id, name
+      FROM asset_metadata FINAL
+      PREWHERE name IN (?4)
+    ) USING (asset_id)
+    GROUP BY slug
+    """
+
+    # signal_id = ( SELECT signal_id FROM #{@metadata_table} FINAL PREWHERE name = ?1 LIMIT 1 )
```
IvanIvanoff, line 134: Seems like you've commented out this and forgot to return it.

lib/sanbase_web/graphql/resolvers/signal_resolver.ex
```diff
@@ -0,0 +1,80 @@
+defmodule SanbaseWeb.Graphql.Resolvers.SignalResolver do
+  import SanbaseWeb.Graphql.Helpers.Utils
+  import SanbaseWeb.Graphql.Helpers.CalibrateInterval
+  import Sanbase.Utils.ErrorHandling, only: [handle_graphql_error: 3]
+  import Sanbase.Metric.Selector, only: [args_to_selector: 1, args_to_raw_selector: 1]
+
+  import Sanbase.Utils.ErrorHandling,
+    only: [handle_graphql_error: 3, maybe_handle_graphql_error: 2]
```
IvanIvanoff, line 8: You are importing the module twice. If I am not mistaken, here the first import is overwritten and not used.

tspenov (approved):

PR #2501, opened 2021-02-01
Rename signals to alerts

IvanIvanoff:
## Changes

There are two main reasons for this change:
- The frontend exposes the signals as `alerts`. Changing the internal names used makes sense for consistency reasons. `Alert` also makes much more sense instead of `signals` for the use case.
- We'll introduce the `signals` CH table as a `getSignal` API and it will make sense to have a `Sanbase.Signal` module dedicated to that.

## Checklist:


lib/sanbase/alerts/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -1,24 +1,24 @@
...
+  Trigger settings for trending words alert.
 
-  The signal supports the following operations:
+  The alert supports the following operations:
 
   1. Send the list of trending words at predefined time every day
-  2. Send a signal if some word enters the list of trending words.
-  3. Send a signal if some project enters the list of trending words
-  4. Send a signal if some project from a watchlist enters the list
+  2. Send a alert if some word enters the list of trending words.
```
tspenov, line 8:
```suggestion
  2. Send an alert if some word enters the list of trending words.
```

tspenov (approved):

PR #2512, opened 2021-02-04
Fix login issues when email is not lowercase.

IvanIvanoff:
## Changes

In case the user is already registered and the login query contains not
lowercased email, this leads to a problem as the value is not normalized
before checking it.

## Checklist:

### DIFF
--- a/lib/sanbase/auth/user.ex
+++ b/lib/sanbase/auth/user.ex
@@ -141,7 +141,7 @@ defmodule Sanbase.Accounts.User do
       :twitter_id,
       :username
     ])
-    |> normalize_username(attrs)
+    |> normalize_username(attrs[:username], :username)
     |> normalize_email(attrs[:email], :email)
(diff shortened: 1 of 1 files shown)


lib/sanbase/auth/user.ex
```diff
@@ -252,21 +254,22 @@ defmodule Sanbase.Accounts.User do
...
   end
 
-  defp normalize_username(changeset, %{username: username}) when not is_nil(username) do
-    put_change(changeset, :username, String.trim(username))
-  end
+  defp normalize_username(changeset, nil, _), do: changeset
 
-  defp normalize_username(changeset, _), do: changeset
+  defp normalize_username(changeset, username, field) do
+    put_change(changeset, field, normalize_login_field_value(username))
```
tspenov, line 260: We should not downcase username IMO

tspenov:
We should add a test covering this change.

PR #2517, opened 2021-02-05
Remove filter and group by source

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/social_data/trending_words.ex
+++ b/lib/sanbase/social_data/trending_words.ex
@@ -195,25 +195,25 @@ defmodule Sanbase.SocialData.TrendingWords do
     SELECT
       t,
       word,
-      project,
-      total_score AS score
+      any(project) AS project,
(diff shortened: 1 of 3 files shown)


lib/sanbase/social_data/trending_words.ex
```diff
@@ -195,25 +195,25 @@ defmodule Sanbase.SocialData.TrendingWords do
...
+      sum(score) / 4 as total_score
     FROM(
         SELECT
            toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), ?1) * ?1) AS t,
            word,
            any(project) AS project,
-           SUM(score) / 4 AS total_score
-        FROM #{Config.get(:trending_words_table)}
+           argMax(score, dt) as score
+        FROM trending_words_v3_top_500
```
IvanIvanoff, line 206: We need to keep the configuration here instead of hardcoding

IvanIvanoff (approved):

PR #2522, opened 2021-02-09
Add data to the short url type

IvanIvanoff:
## Changes
image
image

## Checklist:


Partysun:
@IvanIvanoff How to update existed/created shortUrl

IvanIvanoff:
@tspenov This PR does not aim to automatically update existing short urls, as it's not really clear how to do it.
Currently, you have to do it manually. When we merge this and the frontend is fully migrated and using `data`, we can do a migration that splits the `fullUrl` by the first `?` (where the params start) and store the first part in the url and the second part in the data field. But this cannot be done now as all existing short urls will be broken.

PR #2529, opened 2021-02-10
Add warnings-as-errors in prod/test Dockerfile

IvanIvanoff:
## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ RUN cd assets && npm run build:prod
 COPY . /app
 RUN mix format --check-formatted
 
-RUN mix compile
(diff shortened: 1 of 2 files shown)


tspenov:
@IvanIvanoff What is the goal here - to not push code that has warnings. 
Shouldn't we put it also in directly `elixirc_options: [warnings_as_errors: true]` - so we catch it even earlier ?

IvanIvanoff:
@tspenov It's more for testing purposes. We can't really do this because of some of the dependencies we have. For example we can't directly upgrade instream and now it emits warnings. Ideally, we would have this enabled so we can catch adding new warnings to the code as there is no benefit in not doing that.

PR #2531, opened 2021-02-11
Rework Kaiko price scraping

IvanIvanoff:
## Changes

- Run the Kaiko price scraping more often with higher granularity (every 5 seconds and get the data with 1 second granularity)
- Rework the postgres communication and reduce it to 2 queries - fetch all `last_scraped` with 1 query and record back all the new last scraped with a single query as well.

## Checklist:

### DIFF
--- a/lib/sanbase/scrapers/kaiko.ex
+++ b/lib/sanbase/scrapers/kaiko.ex
@@ -9,38 +9,69 @@ defmodule Sanbase.Kaiko do
   @source "kaiko"
   @url "https://eu.market-api.kaiko.io"
   @recv_timeout 30_000
-  @interval "10s"
-  @interval_sec 10
+  @interval "1s"
+  @interval_sec 1
+  @rounds_per_minute 12
 
+  # Scraping is started every round minute. We want to scrape every 5 seconds.
+  # Because of this, every minute scrape consists of many scrapres.
   def run(opts \\ []) do
-    # cronjobs cannot use sub-minute intervals.
-    # The cronjob runs every 30 seconds by starting it twice, the second time
-    # with a sleep: 30_000 argument, so it sleeps for 30 seconds before running
-    if milliseconds = Keyword.get(opts, :sleep), do: Process.sleep(milliseconds)
+    rounds = Keyword.get(opts, :rounds_per_minute, @rounds_per_minute)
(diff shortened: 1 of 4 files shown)


lib/sanbase/scrapers/kaiko.ex
```diff
@@ -91,32 +122,33 @@ defmodule Sanbase.Kaiko do
       URI.encode_query(opts)
   end
 
-  defp last_datetime_scraped(slug) do
-    case Progress.last_scraped(slug, @source) do
-      nil -> Timex.shift(Timex.now(), minutes: -10)
-      %DateTime{} = dt -> dt
+  defp last_datetime_scraped(slug, last_dt_map) do
+    case Map.get(last_dt_map, slug) do
+      nil -> Timex.shift(Timex.now(), minutes: -1)
```
tspenov, line 127: why we set it to this one ?

lib/sanbase/scrapers/kaiko.ex
```diff
@@ -9,38 +9,69 @@ defmodule Sanbase.Kaiko do
...
+          prices = combine_prices(usd_prices, btc_prices)
+
+          :ok = export_to_kafka(prices, slug)
+
+          [
+            get_last_datetime(usd_prices, base_asset, "usd"),
+            get_last_datetime(btc_prices, base_asset, "btc")
+          ]
+        end,
+        max_concurrency: 50,
```
tspenov, line 58: How many pairs for kaiko do we have ?
Are kaiko able to handle such concurrency ?

lib/sanbase/scrapers/kaiko.ex
IvanIvanoff: This scraper is for the purposes of realtime scraping and not historical data. I've reduced the value here 10 times as I've reduced the interval 10 times (from 10s to 1s). This is so the result will still fit in one page (iirc the max values per page is 100)

lib/sanbase/scrapers/kaiko.ex
IvanIvanoff: I hope so. I've made it this big so we can execute all the necessary queries before the next rescrape starts in a few seconds.

tspenov (approved):

PR #2532, opened 2021-02-11
Add price in eth metric to API

antisaa:
## Changes

Adding new metrics to API
- price in eth
- labeled holders metrics
## Checklist:


tspenov (approved):

tspenov:
Although we have to remove the current `price_eth` before merging

antisaa:
> Although we have to remove the current `price_eth` before merging

I can't find `price_eth` in available metrics, how is it being computed?

tspenov:
The timeseries metric is here: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/prices/price.ex#L173
It now takes the price of asset in usd and the price of eth in usd and divides them.
I'll take care of removing and testing it. It should be also replaced per project with the new metric.

tspenov:
The tests are failing because you are using the `name`field  instead of `metric` field in tests. @antisaa

tspenov:
@antisaa my bad. The .json file was missing from the file handler. Pushed a fix.

lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -2081,5 +2081,22 @@
...
+    "access": "free",
+    "selectors": ["slug"],
+    "min_plan": {
+      "SANAPI": "free",
+      "SANBASE": "free"
+    },
+    "aggregation": "last",
+    "min_interval": "5m",
+    "table": "intraday_metrics",
+    "has_incomplete_data": true,
```
IvanIvanoff, line 2099: This should be `false`

PR #2533, opened 2021-02-11
Improve GraphQL caching

IvanIvanoff:
## Changes

A few changes:
- If there are `cachingParams` - propagate them (not in the nicest way right now) to the Absinthe before_send function.
- Do not round datetimes that much, only truncate seconds.  This will improve the time new data appears in the API. With the old implementation, if the provided `to` datetime is 17:34:50, it will be rounded down to 17:30:00 and the fact that now the price updates much more is not visible in the API even if there is no caching.
  - This might have some slight negative effects in some computations (in the dataloader) but this should occur rarely and have no negative effects.
 

## Checklist:

### DIFF
--- a/lib/sanbase/model/project/list/selector/transform.ex
+++ b/lib/sanbase/model/project/list/selector/transform.ex
@@ -54,8 +54,9 @@ defmodule Sanbase.Model.Project.ListSelector.Transform do
   def transform_from_to(%{from: from, to: to} = map) do
     %{
       map
-      | from: if(is_binary(from), do: from_iso8601!(from) |> round_datetime(), else: from),
-        to: if(is_binary(to), do: from_iso8601!(to) |> round_datetime(), else: to)
+      | from:
+          if(is_binary(from), do: from_iso8601!(from) |> DateTime.truncate(:second), else: from),
+        to: if(is_binary(to), do: from_iso8601!(to) |> DateTime.truncate(:second), else: to)
     }
   end
 
@@ -99,8 +100,8 @@ defmodule Sanbase.Model.Project.ListSelector.Transform do
         to = Timex.shift(now, seconds: -shift_to_by)
 
         map
-        |> Map.put(:from, from |> round_datetime())
-        |> Map.put(:to, to |> round_datetime())
+        |> Map.put(:from, from |> DateTime.truncate(:second))
+        |> Map.put(:to, to |> DateTime.truncate(:second))
     end
   end
 
--- a/lib/sanbase/model/project/list/selector/transform.ex
+++ b/lib/sanbase/model/project/list/selector/transform.ex
@@ -55,8 +55,8 @@ defmodule Sanbase.Model.Project.ListSelector.Transform do
     %{
       map
       | from:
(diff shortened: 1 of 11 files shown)


lib/sanbase_web/graphql/absinthe_before_send.ex
```diff
@@ -103,8 +103,19 @@ defmodule SanbaseWeb.Graphql.AbsintheBeforeSend do
     all_queries_cachable? = queries |> Enum.all?(&Enum.member?(@cached_queries, &1))
 
     if all_queries_cachable? do
+      cache_key =
+        case Process.get(:__change_absinthe_before_send_caching_ttl__) do
```
tspenov, line 107: These process globals makes it very hard to reason about what happens. I am not sure we can avoid them but we can document somehow where they might be set and where they are used :?

lib/sanbase/model/project/list/selector/transform.ex
```diff
@@ -99,8 +100,8 @@ defmodule Sanbase.Model.Project.ListSelector.Transform do
         to = Timex.shift(now, seconds: -shift_to_by)
 
         map
-        |> Map.put(:from, from |> round_datetime())
-        |> Map.put(:to, to |> round_datetime())
+        |> Map.put(:from, from |> DateTime.truncate(:second))
```
tspenov, line 103: Won't changing from 5s -> ~1s result in many more computations ?
Can we somehow introduce it only for this new more granular prices ?

lib/sanbase_web/graphql/absinthe_before_send.ex
IvanIvanoff: I want to rework everything to not use it. But the problem is not trivial - the document provider phase happens before any of the resolution phases, so this is the only way I found to work with this. This approach is actually not a solution at all, because if we have many `cachingParams` in a single query, it will persist only one of them.

lib/sanbase/model/project/list/selector/transform.ex
IvanIvanoff: Is this the proper PR for this comment? Maybe you refer to #2531 

The issue here is that when we round time to 5 minute, 17:34 will be rounded down to 17:30 and we'll not serve the last 4 minutes of data at all.

tspenov: What was the initial idea of rounding to `5m` ? Isn't it because most of the metrics were with `5m` lowest granularity and it results in fetching more data from cache.
Now we introduce more granular prices, so my question is isn't it better to keep old rounding and update the behaviour only for the more granular data. Maybe make `DateTime.truncate(:second)` based on additional supplied user params :?

IvanIvanoff: This rounding will have effect only in the `included_slugs_by_filters` function where it is used for caching , so the rounding can be brought back. There are other, more important places where the datetime rounding places.

On a side note: Maybe the dt rounding can be changed slightly to round up, not down? So 17:34 will be rounded to 17:35 and we won't exclude that much data.

IvanIvanoff: I've change the `round_datetime` to support this type of rounding.

PR #2534, opened 2021-02-12
Format properly amount in discord payments notifications

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/billing/stripe_event.ex
+++ b/lib/sanbase/billing/stripe_event.ex
@@ -136,7 +136,7 @@ defmodule Sanbase.Billing.StripeEvent do
        })
        when amount > 1 do
     message = """
-    Failed card charge for $#{amount / 100}.
+    Failed card charge for #{format_cents_amount(amount)}.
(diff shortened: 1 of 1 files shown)


lib/sanbase/billing/stripe_event.ex
```diff
@@ -373,6 +377,10 @@ defmodule Sanbase.Billing.StripeEvent do
     |> Kernel.+(1)
   end
 
+  defp format_cents_amount(amount) do
+    "$" <> :erlang.float_to_binary(amount / 100, decimals: 0)
```
IvanIvanoff, line 381: You can use:
```elixir
import Sanbase.Math, only: [to_float: 1, round_float: 1]

amount |> to_float() |> round_float()
```

tspenov: That does not do the trick. I want to avoid scientific notation.
image

IvanIvanoff: Ah, yeah.. In this case we use the `Number` library: `Number.Delimit.number_to_delimited`:
image

PR #2537, opened 2021-02-15
Rework contract addresses handling

IvanIvanoff:
## Changes

- Clean the contract addresses handling a little bit.
- Introduce `opts` keyword param that controls which contract address to return.
- Introduce a new contract label `latest_onchain_contract` that points to the latest real on-chain contract address. This is used together with the custom contracts like `reputation_contract` which is needed when a project has more than 1 contract or migrates to a new one.
  - After deploying everything will work 1:1 as before. The only change occurs after the new label is added to any project.

## Checklist:

### DIFF
--- a/lib/sanbase/model/project/contract_address.ex
+++ b/lib/sanbase/model/project/contract_address.ex
@@ -50,4 +50,10 @@ defmodule Sanbase.Model.Project.ContractAddress do
   def list_to_main_contract_address(list) when is_list(list) do
     Enum.find(list, &(&1.label == "main")) || List.first(list)
   end
+
+  def list_to_latest_onchain_contract_address(list) when is_list(list) do
+    Enum.find(list, &(&1.label == "latest_onchain_contract")) ||
+      Enum.find(list, &(&1.label == "main")) ||
+      List.first(list)
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/sanbase/model/project/contract_address.ex
```diff
@@ -50,4 +50,10 @@ defmodule Sanbase.Model.Project.ContractAddress do
   def list_to_main_contract_address(list) when is_list(list) do
     Enum.find(list, &(&1.label == "main")) || List.first(list)
   end
+
+  def list_to_latest_onchain_contract_address(list) when is_list(list) do
+    Enum.find(list, &(&1.label == "latest_onchain_contract")) ||
```
tspenov, line 55: How do we label a contract as `latest_onchain_contract` - through admin panel ?

tspenov: One thing that I saw in master is that we are putting `main` by default when adding contract which can lead to more than one contract labeled `main`: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/model/project/contract_address.ex#L41

tspenov: But actually if we have `label` key with nil value it will work ok.

IvanIvanoff: Yes, I am not 100% sure who and how puts `nil`, but:
image
image

tspenov: Should we add a select box in admin - nil, main, latest ?

IvanIvanoff: The label is free-text, not necessary one of these.

tspenov (approved):

PR #2542, opened 2021-02-17
Add metadata to access restrictions

IvanIvanoff:
## Checklist:


lib/sanbase/billing/plan/restrictions.ex
```diff
@@ -107,4 +110,36 @@ defmodule Sanbase.Billing.Plan.Restrictions do
...
+              "priceVolumeDiff",
+              "socialDominance",
+              "githubActivity",
+              "minersBalance",
+              "ohlc",
+              "getProjectTrendingHistory"
+            ],
+       do: "5m"
+
+  defp min_interval("query", _query), do: nil
```
tspenov, line 144: What queries go here ? What happens on frontend side if min_interval is nil ?

IvanIvanoff: This is everything without interval - `projectBySlug`, `allExchanges`, `watchlist`, etc.

PR #2543, opened 2021-02-19
Use the eth_top_holders_daily_union table

IvanIvanoff:
## Checklist:


lib/sanbase/clickhouse/top_holders/top_holders_sql_query.ex
```diff
@@ -177,12 +177,19 @@ defmodule Sanbase.Clickhouse.TopHolders.SqlQuery do
...
       contract = ?1
     """
 
     args = [contract]
 
     {query, args}
   end
+
+  defp table_to_where_keyword(table) do
+    case String.contains?(table, "union") do
```
tspenov, line 190: Can we write a comment why we can't use prewhere on this union table (or is it on all union tables) ?

IvanIvanoff: I am not exactly sure why this is needed, it just does not work with prewhere.

PR #2545, opened 2021-02-19
Return price_eth from intraday metrics after fix

tspenov:
## Changes

Bring back `price_eth` from intraday metrics after it is historically backfilled

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
+++ b/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
@@ -2084,8 +2084,8 @@
   },
   {
     "human_readable_name": "Price in ETH",
-    "name": "price_eth_5m",
-    "metric": "price_eth_5m",
+    "name": "price_eth",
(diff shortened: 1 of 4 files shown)


lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -2084,8 +2084,8 @@
   },
   {
     "human_readable_name": "Price in ETH",
-    "name": "price_eth_5m",
-    "metric": "price_eth_5m",
+    "name": "price_eth",
+    "metric": "price_eth",
```
IvanIvanoff, line 2088: This is a breaking change. Looking at the API usage it seems like this metric is not used that much and there is a pattern in the usage, so maybe this are the API tests: 
image

In order to not break things you can add just a new record that is the same except for the name.

tspenov: yeah, we should introduce some metric `aliases` field

PR #2548, opened 2021-02-22
Add new field for favorite metrics

atanasormanow:
## Checklist:

```graphql
{
  currentUser {
    settings {
      favoriteMetrics
    }
  }
}
```

<details>
<summary> Result </summary>

```graphql
{
  "data": {
    "currentUser": {
      "settings": {
        "favoriteMetrics": []
      }
    }
  }
}
```
</details>


lib/sanbase/auth/settings.ex
```diff
@@ -23,6 +23,7 @@ defmodule Sanbase.Accounts.Settings do
     field(:newsletter_subscription_updated_at_unix, :integer, default: nil)
     field(:is_promoter, :boolean, default: false)
     field(:paid_with, :string, default: nil)
+    field(:favourite_metrics, {:array, :string}, default: [])
```
spiderjako, line 26: I think we use american english, so ``favorite..`` would be better I think

atanasormanow: Yeah, i was wondering about that spelling. In notion was written as "favourite" but i like the american word better tbh :smiley:

IvanIvanoff: We're not strongly following neither British nor American English. 
Elixir itself uses British (`@behaviour` instead of `@behavior`) but in our API use the American`color` instead of `colour`. In the metrics we have the British `labelled` instead of `labeled`. So - it's a mess :smiley:

test/sanbase_web/graphql/user/user_settings_test.exs
```diff
@@ -21,6 +21,26 @@ defmodule SanbaseWeb.Graphql.UserSettingsTest do
...
+    query = change_favorite_metrics_query(["price_eth"])
+    result = conn |> execute(query, "updateUserSettings")
+
+    assert result == %{"favoriteMetrics" => ["price_eth"]}
+
+    assert UserSettings.settings_for(user, force: true) |> Map.get(:favorite_metrics) == [
+             "price_eth"
+           ]
+
+    query = change_favorite_metrics_query(["price_btc"])
```
IvanIvanoff, line 34: You can test here with more than 1 metric - `["price_btc", "nvt", "mvrv_usd"]` for example. This would actually test that we're storing the data as an array instead of allowing only a single element here.

lib/sanbase/auth/settings.ex
```diff
@@ -45,6 +46,7 @@ defmodule Sanbase.Accounts.Settings do
       :hide_privacy_data,
       :is_promoter,
       :paid_with,
+      :favorite_metrics,
```
IvanIvanoff, line 49: You could also add a validation in the changeset that the list of metrics actually contain valid metrics.
Hint: Do not use `Enum.all?(metrics, &Metric.has_metric?(&1)` but do it with an `Enum.find/2` or `Enum.filter/reject`  so you can actually report back the unsupported/mistyped metric(s)

lib/sanbase/auth/settings.ex
```diff
@@ -69,6 +72,21 @@ defmodule Sanbase.Accounts.Settings do
...
+    type
+    |> Stream.map(&Sanbase.Metric.has_metric?/1)
+    |> Enum.find(&(&1 != true))
+    |> case do
+      nil ->
+        []
+
+      {:error, error} ->
+        [favorite_metrics: "Invalid metric found in the favorite metrics list. #{error}"]
+    end
```
IvanIvanoff, line 87: A more simple way would be:
```suggestion
    metrics
    |> Enum.find_value([], fn metric ->
      case Sanbase.Metric.has_metric?(metric) do
        true -> false
        {:error, error} -> [favorite_metrics: "Invalid metric found in the favorite metrics list. #{error}"]
      end
    end)
```

IvanIvanoff (approved):

tspenov (approved):

PR #2552, opened 2021-02-24
Transfer subscriptions trial

tspenov:
## Changes

1. Refactoring subscriptions API.
2. Transfer trial

## Checklist:


lib/sanbase/billing/billing.ex
```diff
@@ -70,4 +72,27 @@ defmodule Sanbase.Billing do
       {:error, _} -> true
     end)
   end
+
+  @spec create_or_update_stripe_customer(%User{}, String.t() | nil) ::
+          {:ok, %User{}} | {:error, %Stripe.Error{}}
+  def create_or_update_stripe_customer(_, _card_token \\ nil)
```
IvanIvanoff, line 78: Here you do not need to underscore vars. Let's keep the proper names for better documentation:
```suggestion
  def create_or_update_stripe_customer(user, card_token \\ nil)
```

lib/sanbase/billing/billing.ex
```diff
@@ -70,4 +72,27 @@ defmodule Sanbase.Billing do
...
+
+  def create_or_update_stripe_customer(%User{stripe_customer_id: stripe_id} = user, card_token)
+      when is_nil(stripe_id) do
```
IvanIvanoff, line 81:
```suggestion
  def create_or_update_stripe_customer(%User{stripe_customer_id: nil} = user, card_token) do
```

lib/sanbase/billing/billing.ex
```diff
@@ -70,4 +72,27 @@ defmodule Sanbase.Billing do
...
+
+  def create_or_update_stripe_customer(%User{stripe_customer_id: stripe_id} = user, nil)
+      when is_binary(stripe_id) do
+    {:ok, user}
+  end
+
+  def create_or_update_stripe_customer(%User{stripe_customer_id: stripe_id} = user, card_token)
+      when is_binary(stripe_id) do
+    with {:ok, _} <- StripeApi.update_customer(user, card_token) do
+      {:ok, user}
```
IvanIvanoff, line 95: Should we force preload something here?

tspenov: No, here we update customer in Stripe if we have new card but the database user stays the same and we return it.

PR #2555, opened 2021-02-24
Improve selector slug validations

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/metric/selector.ex
+++ b/lib/sanbase/metric/selector.ex
@@ -26,11 +26,28 @@ defmodule Sanbase.Metric.Selector do
 
   # Private functions
 
+  defp valid_selector?(%{slug: nil}) do
+    {:error, "The slug or slugs arguments must not be null."}
+  end
+
   defp valid_selector?(%{slug: slug}) when is_binary(slug) do
     case @available_slugs_module.valid_slug?(slug) do
       true -> true
       false -> {:error, "The slug #{inspect(slug)} is not an existing slug."}
     end
+  end
+
+  # The check will make one ETS call per slug. It will be executed only for small
+  # list of slugs (less than 10 slugs)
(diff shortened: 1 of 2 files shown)


lib/sanbase/metric/selector.ex
```diff
@@ -26,13 +26,30 @@ defmodule Sanbase.Metric.Selector do
...
+
   defp valid_selector?(%{slug: slug}) when is_binary(slug) do
     case @available_slugs_module.valid_slug?(slug) do
       true -> true
       false -> {:error, "The slug #{inspect(slug)} is not an existing slug."}
     end
   end
 
+  # The check will make one ETS call per slug. It will be executed only for small
+  # lits of slugs (less than 10 slugs)
```
tspenov, line 41:
```suggestion
  # list of slugs (less than 10 slugs)
```

lib/sanbase/metric/selector.ex
```diff
@@ -26,13 +26,30 @@ defmodule Sanbase.Metric.Selector do
...
+  defp valid_selector?(%{slug: slugs}) when is_list(slugs) and length(slugs) < 10 do
+    # If all of the slugs are valid return `true`. Return {:error, error} otherwise.
+    slugs
+    |> Enum.find(true, fn slug ->
+      case @available_slugs_module.valid_slug?(slug) do
+        true -> false
+        false -> {:error, "The slug #{inspect(slug)} is not an existing slug."}
+      end
+    end)
```
tspenov, line 50: wouldn't find return the non-existing slug only not `{:error, ....}` ?

IvanIvanoff: True. I'll cover this with a test as well

tspenov (approved):

PR #2564, opened 2021-02-26
Rename and fix available_anomalies project field

atanasormanow:
## Checklist:

### DIFF
--- a/lib/sanbase/signal/signal_adapter.ex
+++ b/lib/sanbase/signal/signal_adapter.ex
@@ -28,7 +28,7 @@ defmodule Sanbase.Signal.SignalAdapter do
   def available_signals(), do: @signals
 
   @impl Sanbase.Signal.Behaviour
-  def available_signals(%{slug: slug}) when is_binary(slug) do
+  def available_signals(slug) when is_binary(slug) do
     {query, args} = available_signals_query(slug)
 
     ClickhouseRepo.query_transform(query, args, fn [signal] -> signal end)
--- a/lib/sanbase/signal/signal_adapter.ex
+++ b/lib/sanbase/signal/signal_adapter.ex
@@ -28,7 +28,7 @@ defmodule Sanbase.Signal.SignalAdapter do
   def available_signals(), do: @signals
 
   @impl Sanbase.Signal.Behaviour
-  def available_signals(slug) when is_binary(slug) do
+  def available_signals(%{slug: slug}) when is_binary(slug) do
     {query, args} = available_signals_query(slug)
 
     ClickhouseRepo.query_transform(query, args, fn [signal] -> signal end)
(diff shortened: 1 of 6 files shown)


lib/sanbase/signal/sql_query/signal_sql_query.ex
```diff
@@ -25,9 +25,13 @@ defmodule Sanbase.Signal.SqlQuery do
     query = """
     SELECT name
     FROM #{@metadata_table}
-    PREWHERE
-      name = ?1
-    GROUP BY name
+    PREWHERE signal_id in (
+      SELECT DISTINCT(signal_id)
+      from #{@table}
```
IvanIvanoff, line 30: nitpick: upcase all the SQL keywords

lib/sanbase/signal/sql_query/signal_sql_query.ex
```diff
@@ -27,9 +27,9 @@ defmodule Sanbase.Signal.SqlQuery do
     FROM #{@metadata_table}
     PREWHERE signal_id in (
       SELECT DISTINCT(signal_id)
-      from #{@table}
+      FROM #{@table}
       INNER JOIN (
-        SELECT * FROM asset_metadata WHERE name = ?1
+        SELECT * FROM asset_metadata PREWHERE name = ?1
```
IvanIvanoff, line 32:
```suggestion
        SELECT * FROM asset_metadata FINAL PREWHERE name = ?1
```

lib/sanbase/signal/sql_query/signal_sql_query.ex
```diff
@@ -45,11 +45,11 @@ defmodule Sanbase.Signal.SqlQuery do
     query = """
     SELECT DISTINCT(name)
     FROM asset_metadata
-    WHERE asset_id in (
+    PREWHERE asset_id in (
       SELECT DISTINCT(asset_id)
       FROM #{@table}
       INNER JOIN (
-        SELECT * FROM #{@metadata_table} WHERE name = ?1
+        SELECT * FROM #{@metadata_table} PREWHERE name = ?1
```
IvanIvanoff, line 52:
```suggestion
        SELECT * FROM #{@metadata_table} FINAL PREWHERE name = ?1
```

lib/sanbase/signal/signal_adapter.ex
```diff
@@ -28,7 +28,7 @@ defmodule Sanbase.Signal.SignalAdapter do
   def available_signals(), do: @signals
 
   @impl Sanbase.Signal.Behaviour
-  def available_signals(%{slug: slug}) when is_binary(slug) do
```
IvanIvanoff, line 31: The fix should be the other way around - pass the selector %{slug: slug} instead of expecting a slug here. This is in line with the other functions that accept a selector instead of a binary slug.

IvanIvanoff (approved):
Just add 1 test that uses this API

PR #2571, opened 2021-03-02
Error when trying to subscribe for the same plan

tspenov:
## Changes

* Error message when trying to subscribe to the same plan
* Some subscription queries refactoring

## Checklist:

### DIFF
--- a/lib/sanbase/billing/subscription.ex
+++ b/lib/sanbase/billing/subscription.ex
@@ -7,7 +7,6 @@ defmodule Sanbase.Billing.Subscription do
   use Ecto.Schema
 
   import Ecto.Changeset
-  import Ecto.Query
 
   alias Sanbase.Billing
   alias Sanbase.Billing.Plan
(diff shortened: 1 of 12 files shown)


lib/sanbase/billing/subscription.ex
```diff
@@ -218,21 +220,23 @@ defmodule Sanbase.Billing.Subscription do
...
   def user_subscriptions(%User{id: user_id}) do
-    user_id
-    |> user_subscriptions_query()
-    |> active_subscriptions_query()
-    |> join_plan_and_product_query()
+    __MODULE__
+    |> Query.filter_user_query(user_id)
+    |> Query.all_active_and_trialing_subscriptions_query()
+    |> Query.join_plan_and_product_query()
+    |> Query.order_by_query()
```
IvanIvanoff, line 227: With time I start to dislike the `_query` suffix more and more. For example, here it's pretty clear what this function is as it is placed inside the `Query` module. Also, in most of the cases it's clear from the name if this is something that extends a query or if it is something that is "final" (returns result). `sort`, `order_by`, `filter` sound like adding something to the query, while `user_subscriptions` sounds like returning a result.

lib/sanbase/billing/subscription.ex
```diff
@@ -250,20 +254,28 @@ defmodule Sanbase.Billing.Subscription do
...
-  end
-
   # Private functions
 
+  defp active_subscriptions_for_this_plan(user, plan) do
+    __MODULE__
+    |> Query.all_active_subscriptions_for_plan_query(plan.id)
+    |> Query.filter_user_query(user.id)
+    |> Repo.all()
+    |> Enum.any?()
```
IvanIvanoff, line 264: Maybe `Enum.empty?` is a little bit more clear here.

lib/sanbase/billing/subscription.ex
tspenov: Removed the `_query` suffix. Here the usage is redundant because I use it `Query` module. Maybe it will be more clear in case of import but I plan to use it with the module name so it should be fine.

PR #2590, opened 2021-03-10
Event bus

IvanIvanoff:
## Changes

> Note: This PR is introducing the core Event Bus functionality. It does not include all places that should emit events and does not include all the subscribers. These things will be added with other PRs so the PRs are not betting too big.

### Overview
The Event Bus is a mechanism that allows different components to communicate with each other
without knowing about each other.
The Event Bus allows to decouple the creation of an event from the processing  of that same event.
A component sends events to the Event Bus without knowing what component will process these event
or how many components will process the event. The components processing the events do not know and
do not need to  know who emitted the event.

### Refactoring
As part of the introduction of the Event Bus, a few reworks were required so the emitting of events can be done better. A few issue were fixed as well (see create_chart_event, etc.)

### Introduce an Event Bus

Introduce an Event Bus that will be the central places for handling different type of events, emitted by the three different sanbase types (alerts, scrapers, web). The implementation is done by using the [otobus/event_bus](https://github.com/otobus/event_bus) library.

There are different types of subscribers that listen for events and act accordingly:
- **KafkaExporterSubscriber** - Export all of the generated events in a Kafka topic for persistence.
- **UserEventsSubscriber** - Act on the user events - user registration, updating of email/name, etc.
- **NotificationSubscriber** - Generate notifications based on events

The emitting of events must be transparent and must not lead to any performance issues. There is an `EventEmitter` behavior that defines the `emit_event/3` callback function. All `emit_event/3` function must return their first argument back, so the function can be inserted in pipelines.

### Event emitter example

If a new event type needs to be emitted, the following steps must be done:
- Handle the new event type in the `Sanbase.EventBus.Event.valid?/1` function. For example validating an event that is emitted when email address is changed:
```elixir
def valid?(%{event_type: :update_email, user_id: id, old_email: old, new_email: new}),
  do: valid_integer_id?(id) and valid_string_field_change?(old, new)
```
- Add an `emit_event/3` function that properly emits this event:
```elixir
  use Sanbase.EventBus.EventEmitter
  @topic :user_events

  def handle_event({:error, _} = result, _event_type, _args), do: result
 
  def handle_event({:ok, user}, :update_email, %{old_email: _, new_email: _} = args) do
    Map.merge(%{event_type: :update_email, user_id: user.id}, args)
    |> notify()

    {:ok, user}
  end
  
  defp notify(data) do
    Sanbase.EventBus.notify(%{topic: @topic, data: data})
  end
  ```
- Emit the event from the proper place in the code:

```elixir
user
|> User.changeset(%{
  email: user.email_candidate,
  email_candidate: nil,
  email_candidate_token_validated_at: validated_at
})
|> Repo.update()
|> emit_event(:update_email, %{old_email: user.email, new_email: user.email_candidate})
```

### Event processing example

Events are processed by modules that are subscribed to one or more of the event bus topics.
Most of the topic subscriptions are done in `application.ex` when the app is started.
Subscribing to a topic is done this way:
```elixir
EventBus.subscribe({Sanbase.EventBus.KafkaExporterSubscriber, [".*"]})
```

The subscriber module must implement a `process/1` function which accepts `event_shadow`
tuple which holds the topic and the event id by which the whole event can be obtained.

In most of the cases, the subscriber is a GenServer so the events can be processed asynchronously.
Because of that, the GenServer must be started in the supervisor tree as well. In this case, the subscribers
are defined in the `common_children/0` function which defines the common children for all sanbase pod types (web, scrapers, signals):
```elixir
 def common_children() do
    [
      ...

      Sanbase.EventBus.KafkaExporterSubscriber,

      ...
    ]
  end
  ```
This is the `KafkaExporterSubscriber` (omitting the GenServer related boilerplate):
```elixir


  def process({_topic, _id} = event_shadow) do
    GenServer.cast(__MODULE__, event_shadow)
    :ok
  end

  def handle_cast({topic, id} = event_shadow, state) do
    case EventBus.fetch_event(event_shadow) do
      %{} = event ->
        kv_tuple = {event.id, event |> Map.from_struct() |> Jason.encode!()}

        :ok =
          Sanbase.KafkaExporter.persist_async(
            kv_tuple,
            :sanbase_event_bus_kafka_exporter
          )

      nil ->
        {:noreply, state}
    end

    :ok = EventBus.mark_as_completed({__MODULE__, topic, id})

    {:noreply, state}
  end
  ``` 

## Checklist:

### CODE REVIEW

File: lib/sanbase/accounts/accounts_event_emitter.ex [hunk 1]
```diff
@@ -0,0 +1,58 @@
+defmodule Sanbase.Accounts.EventEmitter do
+  @behaviour Sanbase.EventBus.EventEmitter.Behaviour
+
+  @topic :user_events
+
+  def emit_event({:error, _} = result, _event, _args), do: result
+
+  def emit_event({:ok, user}, :register_user, %{login_origin: _} = args) do
```
tspenov, line 8: maybe `{:ok, user} = result` and then return `result` to be clear that this function only returns what is received as first arg :?

IvanIvanoff: Even if it is bound to `result` there is no guarantee that the first argument is returned. In most of the cases I encountered an error, the error was that I was ending the function with `notify/1` without returning the first arg in any form.

Second, I had this but in this case it was enough to go over the line limit and the formatter made the function take much more lines, not that this is a big argument.

I have a task (not written down) to try and come up with some nice way to enforce these requirements. I think an approach similar to https://hexdocs.pm/ex_contract/readme.html. An example from the docs is:
```elixir
  ensures result == x * y
  @spec test_ensures(x :: integer, y :: integer) :: integer | no_return
  def test_ensures(x, y) do
    x * y
  end
```

which we can make :
```elixir
  ensures result === first_arg
  @spec test_ensures(x :: integer, y :: integer) :: integer | no_return
  def emit_event(first_arg, event_type, args) do
    // ...
    
    first_arg
  end
```

And, if possible, enable these checks only in test/dev and also make sure that every event is emitted at least once during testing.

tspenov: Sounds good. It would be best if it can be done only in the behaviour callback and not touching implementation functions.

lib/sanbase/event_bus/event_bus.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Sanbase.EventBus do
+  @moduledoc """
+  The Event Bus is a mechanism that allows different components to communicate
+  with each other without knowing about eachother.
```
tspenov, line 4:
```suggestion
  with each other without knowing about each other.
```

lib/sanbase/event_bus/event_bus.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  The Event Bus allows to decouple the creation of an event from the processing
+  of that same event. A component sends events to the Event Bus without knowing
+  what component will process these event or how many components will process
```
tspenov, line 8:
```suggestion
  what component will process these events or how many components will process
```

lib/sanbase/event_bus/event_emitter_behaviour.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Sanbase.EventBus.EventEmitter.Behaviour do
+  @type entity :: any()
+  @type event_type :: atom()
+  @type extra_args :: map()
+
+  @doc """
+  Invoked in order to emit and event and push it to the Event Bus.
```
tspenov, line 7:
```suggestion
  Invoked in order to emit and event an push it to the Event Bus.
```

lib/sanbase/accounts/accounts_event_emitter.ex
IvanIvanoff: I have thought about this, but I am not sure it is possible this way. What we can do is to have a behavior, but also implement a `__using__` macro that will expect the module to implement not `emit_event`, but something else (`do_emit_event`, `handle_event`, etc.) and it will define the `emit_event` function which wraps this, but adds additional checks. Now that I write this, I think we can do this without any dependencies but with some conditional compilations (and if we want to keep the checks in prod, too, then we can avoid conditional compilations at all).

IvanIvanoff: With the new code, the following change will lead to the error:
image
image

tspenov: 🎉

tspenov (approved):

lib/sanbase/event_bus/event_bus.ex
```diff
@@ -0,0 +1,71 @@
...
+
+    EventSource.notify params do
+      data = Map.fetch!(params, :data)
+
+      case Sanbase.EventBus.EventValidation.valid?(data) do
+        true ->
+          data
+
+        false ->
+          raise(
```
tspenov, line 64: I think we should not raise when validation fails. Maybe emit another raw event or log but not fail the action.

IvanIvanoff: I did this in order to easily catch errors. Maybe we can conditionally compile the function that handles the error - raise in dev/test and emit an error event to a special topic in prod?

tspenov: yes, raise in dev/test is good but not on prod where it might fail a user triggered request.

IvanIvanoff: Done.

PR #2594, opened 2021-03-11
Expose last triggered

IvanIvanoff:
## Changes

Get the last datetime an alert has been triggered. One alert can have many `last_triggered` values - if the `target` is a watchlist, the `last_triggered` is a map where the key is a slug and the value is when this slug was triggered. The exposed last triggered datetime fetches the latest such value.

> Note: The raw datetimes are not stored, but instead they are rounded to 30 seconds because they are used as part of the cache key.
```graphql
{
  getTriggerById(id: 1){
    trigger{
      lastTriggeredDatetime
    }
  }
}
```

## Checklist:


test/sanbase_web/graphql/alerts/triggers_api_test.exs
```diff
@@ -333,18 +345,55 @@ defmodule SanbaseWeb.Graphql.TriggersApiTest do
...
+
+      Sanbase.Alert.Scheduler.run_alert(Sanbase.Alert.Trigger.MetricTriggerSettings)
+
+      last_triggered_datetime =
+        get_trigger_by_id(context.conn, trigger_id)
+        |> get_in(["data", "getTriggerById", "trigger", "lastTriggeredDatetime"])
+
+      datetime = Sanbase.DateTimeUtils.from_iso8601!(last_triggered_datetime)
+
+      # Check the difference to be at most 11 seconds as the last_triggered_datetime
```
tspenov, line 388:
```suggestion
      # Check the difference to be at most 31 seconds as the last_triggered_datetime
```

test/sanbase_web/graphql/alerts/triggers_api_test.exs
```diff
@@ -333,18 +345,55 @@ defmodule SanbaseWeb.Graphql.TriggersApiTest do
...
+      # is used as part of a cache key and because of that it is rounded per 10 seconds.
```
tspenov, line 389:
```suggestion
      # is used as part of a cache key and because of that it is rounded per 30 seconds.
```

PR #2604, opened 2021-03-17
Add timeseries ohlc get metric

tspenov:
## Checklist:

 ## Screenshots

image


lib/sanbase/clickhouse/metric/metric_adapter.ex
```diff
@@ -81,6 +83,10 @@ defmodule Sanbase.Clickhouse.MetricAdapter do
     end)
   end
 
+  @impl Sanbase.Metric.Behaviour
+  def timeseries_ohlc_data(_metric, _selector, _from, _to, _interval, _opts),
+    do: {:error, "Not implemented"}
```
IvanIvanoff, line 88: Let's implement it here as this will make it available for a lot of metrics. It is a just a copy of the timeseries data with a change `{aggregation(aggregation, "value", "dt")}` => multiple aggregations.

I am wondering if we can actually just add an aggregation definition `:ohlc` that will expand to the whole string needed. This way we could actually reuse the timeseries query and just extend the `aggregation/3` function in the `Sanbase.Metric.SqlQuery.Helper` module.

tspenov: I think with aggregation is better
image
image

tspenov:
close in favour of: https://github.com/santiment/sanbase2/pull/2612

PR #2609, opened 2021-03-18
Add test for updating notes, add max size of notes inserted through admin

tspenov:
## Checklist:

image

### DIFF
--- a/lib/sanbase/blockchain_address/blockchain_address.ex
+++ b/lib/sanbase/blockchain_address/blockchain_address.ex
@@ -56,9 +56,13 @@ defmodule Sanbase.BlockchainAddress do
       changesets,
       Ecto.Multi.new(),
       fn {changeset, offset}, multi ->
+        # notes is an optional field. It should be replaced only if it is in the changeset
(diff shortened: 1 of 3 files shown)


lib/sanbase/blockchain_address/blockchain_address.ex
```diff
@@ -16,6 +16,7 @@ defmodule Sanbase.BlockchainAddress do
     addr
     |> cast(attrs, [:address, :infrastructure_id, :notes])
     |> validate_required([:address])
+    |> validate_length(:notes, max: 45)
```
IvanIvanoff, line 19: i think this is too short

tspenov: that is the limit in frontend
https://www.notion.so/santiment/9b1939e0ccb9494db3b074a6b74787e0?v=e9a9914a6279477981f54dcb3f011320&p=e4b09d70db474e1d9a0f2ef2a3eb108c

haritonasty: > i think this is too short

your suggestion? 60. don't want to allow a lot because it will be not pretty in table ui

PR #2612, opened 2021-03-19
Implement ohlc aggregation for get_metric timeseries_data

tspenov:
## Checklist:

## Usage

```graphql
      {
        getMetric(metric: "price_usd"){
          timeseriesData(
            slug: "ethereum"
            from: "utc_now-5d"
            to: "utc_now"
            interval: "1d"
            aggregation: OHLC
            ){
              datetime
              valueOhlc {
                open
                high
                low
                close
              }
            }
        }
      }
```

```graphql
      {
        getMetric(metric: "daily_active_addresses"){
          timeseriesData(
            slug: "ethereum"
            from: "utc_now-30d"
            to: "utc_now"
            interval: "1w"
            aggregation: OHLC
            ){
              datetime
              valueOhlc {
                open
                high
                low
                close
              }
            }
        }
      }
```

### DIFF
--- a/lib/sanbase/prices/price.ex
+++ b/lib/sanbase/prices/price.ex
@@ -5,7 +5,13 @@ defmodule Sanbase.Price do
   import Sanbase.Price.SqlQuery
 
   import Sanbase.Utils.Transform,
-    only: [maybe_unwrap_ok_value: 1, maybe_apply_function: 2, wrap_ok: 1]
+    only: [
+      maybe_unwrap_ok_value: 1,
+      maybe_apply_function: 2,
+      wrap_ok: 1,
+      transform_metric_result_func: 0,
+      transform_metric_ohlc_result_func: 0
+    ]
 
   import Sanbase.Metric.Transform, only: [maybe_nullify_values: 1, remove_missing_values: 1]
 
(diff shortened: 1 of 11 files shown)


lib/sanbase/prices/price.ex
```diff
@@ -198,17 +204,11 @@ defmodule Sanbase.Price do
...
-            datetime: DateTime.from_unix!(timestamp),
-            value: value
-          }
-      end
-    )
+    if aggregation == :ohlc do
+      ClickhouseRepo.query_transform(query, args, transform_metric_ohlc_result_func())
+    else
+      ClickhouseRepo.query_transform(query, args, transform_metric_result_func())
+    end
```
IvanIvanoff, line 211: I think it would be better to extract the whole thing in a single function like this:
```elixir

def exec_timeseries_data_query(query, args) do
  Sanbae.ClickhouseRepo.query_transform(query, args, fn
    [unix, value] ->
      %{datetime: DateTime.from_unix!(unix), value: value}

    [unix, open, high, low, close] ->
      %{
        datetime: DateTime.from_unix!(unix),
        value_ohlc: %{
          open: open,
          high: high,
          low: low,
          close: close
        }
      }
  end)
end
```
and just call it here (and in other places where ohlc is needed)

lib/sanbase_web/graphql/resolvers/metric_resolver.ex
```diff
@@ -291,4 +299,21 @@ defmodule SanbaseWeb.Graphql.Resolvers.MetricResolver do
       {:error, "Missing required `from` argument"}
     end
   end
+
+  defp valid_timeseries_selection?(requested_fields, args) do
+    aggregation = Map.get(args, :aggregation, nil)
+    value_requested? = MapSet.member?(requested_fields, "value")
+    value_ohlc_requested? = MapSet.member?(requested_fields, "valueOhlc")
+
+    cond do
```
IvanIvanoff, line 308: maybe add a check
```elixir
value_requested? and value_ohlc_requested? ->
  {:error, "Cannot select both value and valueOhlc fields at the same time"}
```

PR #2620, opened 2021-03-24
Add wallet hunters create/fetch proposals APIs

tspenov:
## Checklist:

## Usage

```graphql
{
  walletHunterProposals(
    selector: {
      filter: [{field: "state", value: "active"}],
      sort_by: {field: "created_at", direction: DESC},
      page: 1
      page_size: 10
    }
  ) {
        proposalId
        user {
          email
        }
        title
        text
        address
        labels {
          name
        }
        reward
        state
        isRewardClaimed
        createdAt
        finishAt
        votesFor
        votesAgainst
        sheriffsRewardShare
        fixedSheriffReward
  }
}
```

```
{
  "data": {
    "walletHunterProposals": [
      {
        "claimedReward": false,
        "createdAt": "2021-03-24T09:03:19Z",
        "finishAt": "2021-03-25T09:03:19Z",
        "fixedSheriffReward": 10.0,
        "hunterAddress": "0xcb8c7409fe98a396f32d6cff4736bedc7b60008c",
        "proposalId": "1",
        "reward": 110.0,
        "sheriffsRewardShare": 2.0e3,
        "state": "ACTIVE",
        "text": null,
        "title": null,
        "user": null,
        "votesAgainst": 0.0,
        "votesFor": 0.0
      },
      {
        "claimedReward": false,
        "createdAt": "2021-03-23T20:34:16Z",
        "finishAt": "2021-03-24T20:34:16Z",
        "fixedSheriffReward": 10.0,
        "hunterAddress": "0xcb8c7409fe98a396f32d6cff4736bedc7b60008c",
        "proposalId": "0",
        "reward": 100.0,
        "sheriffsRewardShare": 2.0e3,
        "state": "ACTIVE",
        "text": "t",
        "title": "t2",
        "user": {
          "email": "<email hidden>"
        },
        "votesAgainst": 0.0,
        "votesFor": 0.0
      }
    ]
  }
}
```

```graphql
    mutation {
      createWalletHunterProposal(
        proposalId:1,
        text:"t",
        title:"t2",
        signature:"test"
        messageHash:"alabala",
        hunterAddress:"0xcb8c7409fe98a396f32d6cff4736bedc7b60008c",
      ) {
        proposalId
        user {
          email
        }
        title
        text
        address
        labels {
          name
        }
        reward
        state
        isRewardClaimed
        createdAt
        finishAt
        votesFor
        votesAgainst
        sheriffsRewardShare
        fixedSheriffReward
      }
    }
```


lib/sanbase/smart_contracts/utils.ex
```diff
@@ -1,4 +1,24 @@
 defmodule Sanbase.SmartContracts.Utils do
+  @type address :: String.t()
+
+  @spec call_contract(address, String.t(), list(), list()) :: any()
+  def call_contract(contract, call, args, return_types) do
```
IvanIvanoff, line 5: Can you add some description of the arguments here - what are `call`, `args` and `return_types`?

lib/sanbase/smart_contracts/utils.ex
```diff
@@ -1,4 +1,24 @@
...
+    abi = ABI.encode(call, args) |> Base.encode16(case: :lower)
+
+    {:ok, res_enc} =
```
IvanIvanoff, line 8: What is `res_enc`?  Can you replace with a name without abbreviation or add a comment. These on-chain specific parts are hard to understand if you don't know the details.

lib/sanbase/wallet_hunters/proposal.ex
```diff
@@ -0,0 +1,221 @@
...
+      |> Repo.insert()
+      |> case do
+        {:ok, proposal} ->
+          proposal = Repo.preload(proposal, :user)
+
+          response =
+            wallet_proposal(proposal.proposal_id)
+            |> map_response()
+            |> hd()
+            |> Map.merge(proposal)
```
IvanIvanoff, line 59: Is it guaranteed that the `map_response` is not empty?

lib/sanbase_web/graphql/schema/types/wallet_hunter_types.ex
```diff
@@ -0,0 +1,49 @@
...
+
+  object :wallet_hunter_proposal do
+    field(:proposal_id, non_null(:id))
+    field(:user, :public_user)
+    field(:title, :string)
+    field(:text, :string)
+    field(:hunter_address, :string)
+    field(:reward, :float)
+    field(:state, :wallet_hunter_proposal_states)
+    field(:claimed_reward, :boolean)
```
IvanIvanoff, line 41: `is_reward_claimed` :?

test/sanbase/billing/query_access_level_test.exs
```diff
@@ -15,6 +15,7 @@ defmodule Sanbase.Billing.QueryAccessLevelTest do
       expected_free_queries =
         [
           :active_widgets,
+          :all_wallet_hunter_proposals,
```
IvanIvanoff, line 18: Lately I find it much more convenient to use 1 API that allows fetching all entities as well as filter them if you provide the proper arguments (like comments, chart_configurations, table_configurations, etc.). If you rename to `wallet_hunter_proposals` without any args this means `all`, if you provide `user_id` it might mean the proposals of a given user, etc.

lib/sanbase/wallet_hunters/proposal.ex
tspenov: yes

lib/sanbase/wallet_hunters/proposal.ex
```diff
@@ -0,0 +1,166 @@
...
+  defp filter_response(response, filter) do
+    response
+    |> Enum.filter(fn proposal ->
+      Enum.map(filter, fn %{field: field, value: value} ->
+        field = String.to_existing_atom(field)
+        to_string(proposal[field]) == value
+      end)
+      |> Enum.all?()
+    end)
+  end
```
IvanIvanoff, line 110: Do we need to apply many filters? If we would need many filters in the future, the API will change (from `filter: filter`, to `filter: list_of(filter)`), so I think we could make the filters a list from the beginning.

lib/sanbase/wallet_hunters/proposal.ex
```diff
@@ -0,0 +1,166 @@
...
+  defp sort_response(response, nil), do: response
+
+  defp sort_response(response, %{field: field, direction: direction}) do
+    field = String.to_existing_atom(field)
+
+    if field in [:created_at, :finish_at] do
+      Enum.sort_by(response, & &1[field], {direction, DateTime})
+    else
+      Enum.sort_by(response, & &1[field], direction)
+    end
```
IvanIvanoff, line 121:
```suggestion
    direction = if field in [:created_at, :finish_at], do: {direction, DateTime}, else: direction
    Enum.sort_by(response, & &1[field], direction)
```

DmitriVanGuard (approved):

lib/sanbase/wallet_hunters/proposal.ex
```diff
@@ -0,0 +1,166 @@
...
+  end
+
+  defp paginate(response, page, page_size) do
+    page = page || 1
+    page_size = page_size || 10
+    start_index = (page - 1) * page_size
+    Enum.slice(response, start_index, page_size)
```
IvanIvanoff, line 128: In case of missing args I think we should not paginate? Or if we do, this is better expressed as default values in the API. At least this is the behaviour in the other APIs.

lib/sanbase/wallet_hunters/proposal.ex
tspenov: It is already a list but only with `[%{field: field, value: value}]`. I can't imagine using anything else for filtering and it will be an internal API so it is easier to change later if we need to.

lib/sanbase/wallet_hunters/proposal.ex
tspenov: We have default values for pagination in API which are the same as the ones here.

PR #2634, opened 2021-03-29
Add update_blockchain_address_user_pair mutation API

IvanIvanoff:
## Changes

When a user adds notes/labels to a blockchain address, these values are added not on the blockchain address itself, but on a (blockchain_address, user) pair. The `blockchain address` entity is similar to the `project` entity - external users cannot update any of its fields.

The blockchain address notes and labels can be changed by the Santiment team only.
 
This separation is needed because of 2 main reasons:
- Any wrong/spam labels are not shown to all users.
- Users might want their notes and labels to be private and this is the default behavior.

The (blockchain_addres, user) pair can be created and updated. Creation can be done by invoking the `blockchainAddress` API with a selector containing address and infrastructure. If no such address exists (for the current user) it is created. If such address exists - it is returned.

The pair notes and labels can be updated by the `updateBlockchainAddressUserPair` mutation. In this case if a selector with address and infrastructure is provided and it does not point to an existing pair, an error is returned and no pair is created. 

> Note: The id of pairs and blockchain addresses are not connected. The blockchain address with id 1 and the pair with id 1 do not have anything in common.

Get a pair by id:
```graphql
{
  blockchainAddressUserPair(selector: {id: 1}) {
    id
    labels {
      name
    }
    blockchainAddress {
      address
    }
  }
}

```

Get a pair by address/infrastructure or create it if it does not exist:
```graphql
{
  blockchainAddressUserPair(selector: {address:"0x8644e45ac771c34b6e4e855B6cfa0BB672f89310" infrastructure: "ETH"}) {
    id
    labels {
      name
    }
    blockchainAddress {
      address
    }
  }
}
```

Update an existing blockchain address user pair:
```graphql
mutation{
  updateBlockchainAddressUserPair(
    selector: { id: 1 }
    notes:"new_notes"
    labels: ["label1", "label2"]
  ){
    id
    notes
    labels { name }
    blockchainAddress{ address }
  }
}
```

Update an existing blockchain address user pair by using address/infrastructure selector. If the pair does not exists this will fail.
```graphql
mutation{
  updateBlockchainAddressUserPair(
    selector: { address: "0x123" infrastructure: "ETH" }
    notes:"new_notes"
    labels: ["label1", "label2"]
  ){
    id
    notes
    labels { name }
    blockchainAddress{ address }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
+++ b/lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
@@ -46,6 +46,28 @@ defmodule Sanbase.BlockchainAddress.BlockchainAddressUserPair do
     end
   end
 
+  def by_selector(%{address: address, infrastructure: infrastructure} = selector, user_id) do
+    from(
+      pair in __MODULE__,
+      join: blockchain_address in assoc(pair, :blockchain_address),
+      on: pair.blockchain_address_id == blockchain_address.id,
+      join: infrastructure in assoc(blockchain_address, :infrastructure),
+      on: blockchain_address.infrastructure_id == infrastructure.id,
+      where:
+        blockchain_address.address == ^address and infrastructure.code == ^infrastructure and
+          pair.user_id == ^user_id,
+      preload: ^@preloads
+    )
+    |> Sanbase.Repo.one()
+    |> case do
+      %__MODULE__{user_id: _} = pair ->
+        {:ok, pair}
+
+      _ ->
+        {:error, "Blockchain address user pair with selector #{inspect(selector)} does not exist"}
+    end
+  end
+
   def by_selector(%{blockchain_address_id: blockchain_address_id}, user_id) do
     from(pair in __MODULE__,
       where: pair.user_id == ^user_id and pair.blockchain_address_id == ^blockchain_address_id,
@@ -101,6 +123,19 @@ defmodule Sanbase.BlockchainAddress.BlockchainAddressUserPair do
     |> case do
       {:ok, result} -> {:ok, Map.values(result)}
       {:error, _name, error, _changes_so_far} -> {:error, error}
+    end
+  end
+
+  def create(address, infrastructure, user_id) do
+    {:ok, blockchain_address} =
+      Sanbase.BlockchainAddress.by_selector(%{address: address, infrastructure: infrastructure})
+
+    %__MODULE__{}
+    |> changeset(%{blockchain_address_id: blockchain_address.id, user_id: user_id})
+    |> Sanbase.Repo.insert()
+    |> case do
+      {:ok, struct} -> {:ok, Sanbase.Repo.preload(struct, @preloads)}
+      {:error, error} -> {:error, error}
     end
   end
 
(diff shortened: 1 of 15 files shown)


tspenov:
If we have an `update`, do we also need a `create` mutation ?

lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
```diff
@@ -23,13 +23,60 @@ defmodule Sanbase.BlockchainAddress.BlockchainAddressUserPair do
...
+  def by_selector(%{blockchain_address_id: blockchain_address_id}, user_id) do
+    from(pair in __MODULE__,
+      where: pair.user_id == ^user_id and pair.blockchain_address_id == ^blockchain_address_id
+    )
+    |> Sanbase.Repo.one()
+    |> case do
+      %__MODULE__{user_id: ^user_id} = pair ->
+        {:ok, pair}
+
+      %__MODULE__{user_id: _} ->
```
tspenov, line 57: we have `where: pair.user_id == ^user_id` in the condition so this case is not possible I think

lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
```diff
@@ -23,13 +23,60 @@ defmodule Sanbase.BlockchainAddress.BlockchainAddressUserPair do
...
+        {:error, "Blockchain address pair with id #{id} does not belong to the querying user."}
+
+      _ ->
+        {:error, "Blockchain address pair with #{id} does not exist"}
+    end
+  end
+
+  def by_selector(%{blockchain_address_id: blockchain_address_id}, user_id) do
```
tspenov, line 48: I think it could be split in 2 - `by_selector` and `filter_by_user`  and these reused in different combinations

lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
```diff
@@ -23,13 +23,60 @@ defmodule Sanbase.BlockchainAddress.BlockchainAddressUserPair do
...
+      _ ->
+        {:error,
+         """
+         Blockchain address for user id #{user_id} with blockchain_address pair\
+         with #{blockchain_address_id} does not exist
+         """}
+    end
+  end
+
+  def update(%__MODULE__{} = pair, attrs) do
```
tspenov, line 73: You can add separate `update_changeset` :?

IvanIvanoff: I've just read https://medium.com/very-big-things/towards-maintainable-elixir-the-core-and-the-interface-c267f0da43 and I was trying the approach where do not have a separate changeset function that is called from a single place. We need a `changeset` so it is used by ex_admin (for marking required fields), but `update_changeset` won't be used anywhere else.

lib/sanbase/blockchain_address/blockchain_address_user_pair.ex
IvanIvanoff: The two selector queries does not have anything in common (except for the preload), so these functions won't be reused.
One is:
```elixir
from(pair in __MODULE__, where: pair.id == ^id, preload: ^@preloads)
```
and the other is
```elixir
from(pair in __MODULE__,
      where: pair.user_id == ^user_id and pair.blockchain_address_id == ^blockchain_address_id,
      preload: ^@preloads
    )
```

This is done like this in order to improve the error message. In the first case a wrong `id` can point to a single record of another user. In the second case, a single blockchain_address_id can be present in many other records and we can't return the error "belongs to another user" because the field is not unique and can be duplicated.

PR #2613, opened 2021-03-19
Adding better social volume query generation

spiderjako:
## Changes

Results with this PR on stage:
```
{
  "data": {
    "allProjects": [
      {
        "socialVolumeQuery": "btc OR btcsome OR btcan OR btcl OR xyzbtc OR bitcoin OR bitcon OR bicoin OR bitoin OR bitcion OR bitcoincurr NOT gold NOT cash NOT classic"
      },
      {
        "socialVolumeQuery": "eth OR ethwtc OR ether OR btcethereum OR ethfct OR ethl OR ethereum OR etheruem OR ethereumvm OR ethereuum OR ethereumit OR ethereumi NOT cash NOT gold NOT classic"
      },
      {
        "socialVolumeQuery": "tether OR usdt OR tetherusdt OR tetherfiat"
      },
      {
        "socialVolumeQuery": "binance AND (coin OR token) OR bnb"
      },
      {
        "socialVolumeQuery": "\"bnbn\" OR \"binance native\" OR \"binance-coin-native\""
      },
      {
        "socialVolumeQuery": "ada OR cardano OR cardanoeo OR cardanoada OR adacardano OR cardanoeo OR ada OR adacardano OR cardanoada OR cardanopric"
      },
      {
        "socialVolumeQuery": "\"dot\" OR \"polkadot\" OR \"polkadot-new\""
      },
      {
        "socialVolumeQuery": "xrp OR rippl OR xrpp OR xrpa OR ripplesxrp OR xrpt OR ripple OR ripplei OR ripplea OR rippley OR ripplet OR ripplesxrp OR cripple OR crippl"
      },
      {
        "socialVolumeQuery": "\"uni\" OR \"uniswap\""
      },
      {
        "socialVolumeQuery": "\"theta\" NOT \"Theta Token\" NOT \"Theta Fuel\""
      },
      {
        "socialVolumeQuery": "ltc OR ltcltc OR ltcxlm OR ltch OR btcltc OR litecoin OR litecoin OR litcoin OR litecoim OR litecoinlitecoin OR litecoi OR llitecoin"
      },
      {
        "socialVolumeQuery": "chainlink OR (chain AND link) OR blockchainlink"
      },
      {
        "socialVolumeQuery": "usdcoin OR usdc OR (usd AND coin)"
      },
      {
[...]

PR #2658, opened 2021-04-07
Revert "Adding better social volume query generation"

IvanIvanoff:
Reverts santiment/sanbase2#2613


spiderjako (approved):

tspenov:
What is the problem ?

IvanIvanoff:
@tspenov I don't understand what is the root of https://sentry.production.internal.santiment.net/sentry/sanbase-backend/issues/100588/

It fails with:
```

(RuntimeError) Source does not exist: SanbaseWeb.Graphql.SanbaseDataloader

Registered sources are:

[SanbaseWeb.Graphql.SanbaseRepo, :error]
```

while all of the others that use the dataloader work fine.

PR #2662, opened 2021-04-09
Precompute the autogenerated social volume query.

IvanIvanoff:
## Changes

The autogenerated social volume query computation has O(n^2) complexity
which results in ~4 million operations in order to compute them for all
projects. This causes timeouts when the value is not cached. In order to
avoid computing it on the fly, it is computed in the background every 15
minutes and stored in the database.

This PR is based on https://github.com/santiment/sanbase2/pull/2613 but adds the database storage.

## Checklist:


lib/sanbase/clickhouse/metric/metric_files/change_metrics.json
```diff
@@ -3228,8 +3228,7 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
-  }
-,
+  },
```
IvanIvanoff, line 3231: These are fixed by a formatter.

spiderjako (approved):

PR #2667, opened 2021-04-12
Add timebound signal and tweak interval_to_str

atanasormanow:
## Changes
Add support for timebound field in signals.
When timebound field is present, multiple signals are generated based on the given intervals. The templates in the other fields are required in order for the template engine to do the substitution.

Note that the `..._20y` signal is added separately in the JSON file. This is because the restriction is different compared to the same signals with other intervals.

```json
  {
    "human_readable_name": "Circulation Anomaly {{timebound_human_readable}}",
    "name": "anomaly_circulation_{{timebound}}",
    "signal": "anomalies_stack_circulation_{{timebound}}",
    "timebound": [
      "1d",
      "7d",
      "30d",
      "60d",
      "90d",
      "180d",
      "365d",
      "2y",
      "3y",
      "5y",
      "10y"
    ],
    "access": "free",
    "selectors": ["slug"],
    "aggregation": "count",
    "min_interval": "1m",
    "table": "signals",
    "data_type": "timeseries"
  }
```

## Checklist:


lib/sanbase/signal/file_handler.ex
```diff
@@ -2,6 +2,9 @@ defmodule Sanbase.Signal.FileHandler do
   @moduledoc false
 
   defmodule Helper do
+    alias Sanbase.TemplateEngine
+    alias Sanbase.DateTimeUtils
```
IvanIvanoff, line 6: In most places we have `import Sanbase.DateTimeUtils, only: [interval_to_str: 1]` as there is very low chance of name collisions with that name.

lib/sanbase/signal/file_handler.ex
```diff
@@ -24,6 +27,22 @@ defmodule Sanbase.Signal.FileHandler do
         end
       )
     end
+
+    def resolve_timebound_signals(signal_map, timebound_values) do
+      timebound_values
```
IvanIvanoff, line 32: You can add here 
```elixir
%{"name" => name, "signal" => signal, "human_readable_name" => human_readable_name} = signal_map
```
and reuse them below. This will result in a better error when one of the fields is missing as otherwise it will fail inside the `TemplateEngine` function with a not clear error:
```elixir
** (FunctionClauseError) no function clause matching in String.replace/4    
```

atanasormanow: Yeah, that's way cleaner.

lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -53,5 +53,239 @@
     "min_interval": "1m",
     "table": "signals",
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Anomalies MVRV USD {{timebound_human_readable}}",
+    "name": "anomalies_mvrv_usd_{{timebound}}",
```
IvanIvanoff, line 59: I think `anomaly` is better here instead of `anomalies`, what do you think?

atanasormanow: I think so as well tbh. In all the signals prefixed that way or only in this one tho?

lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -53,5 +53,239 @@
...
+    "aggregation": "count",
+    "min_interval": "1m",
+    "table": "signals",
+    "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Anomaly MVRV USD",
+    "name": "anomaly_mvrv_usd",
+    "signal": "anomalies_mvrv_usd_20y",
+    "access": "restricted",
```
IvanIvanoff, line 85: Let's make all `free` for now

lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -53,5 +53,239 @@
...
+    "access": "free",
+    "selectors": ["slug"],
+    "aggregation": "count",
+    "min_interval": "1m",
+    "table": "signals",
+    "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Anomaly Token Velocity",
+    "name": "anomaly_token_velocity",
```
IvanIvanoff, line 127: Let's match the name of the metric to the name of the signal here. A few example:
- The `token_velocity` metric is exposed as `velocity`
- The `stack_age_consumed` metric is exposed as `age_consumed`/`age_destroyed`.
- Check also token supply, cumulative age consumed

lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -53,5 +53,239 @@
...
+    "selectors": ["slug"],
+    "aggregation": "count",
+    "min_interval": "1m",
+    "table": "signals",
+    "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Ath",
+    "name": "ath",
+    "signal": "ath",
```
IvanIvanoff, line 271: This does not look very serious, it looks more like a typo 😃 
Let's properly expose as `Price USD All Time High`/`price_usd_all_time_high`

lib/sanbase/signal/file_handler.ex
```diff
@@ -39,9 +65,21 @@ defmodule Sanbase.Signal.FileHandler do
...
+                  fn signal ->
+                    case Map.get(signal, "timebound") do
+                      nil ->
+                        [signal]
 
+                      timebound_values ->
+                        Helper.resolve_timebound_signals(signal, timebound_values)
+                    end
+                  end
+                )
```
IvanIvanoff, line 80: I think this better be a function and just call it to assign the result to the module attribute. You can do it like this:
```elixir
@signals_json expand_timebound_signals(@signals_json_pre_timebound_expand)
```

IvanIvanoff (approved):

PR #2685, opened 2021-04-19
Add access restriction handling for signals

atanasormanow:
## Changes
- Signal access restriction now works the same way as with metrics.
- The signals and metrics will be checked for missing required fields when read from the json files.

## Checklist:


lib/sanbase/billing/graphql_schema.ex
```diff
@@ -91,6 +92,13 @@ defmodule Sanbase.Billing.GraphqlSchema do
     |> Enum.map(fn {metric, _access} -> metric end)
   end
 
+  def get_signals_with_access_level(level) do
+    Enum.filter(Sanbase.Signal.access_map(), fn {_signal, signal_level} ->
+      level == signal_level
+    end)
+    |> Enum.map(fn {signal, _access} -> signal end)
```
IvanIvanoff, line 99: I suggest we rewrite this to avoid 2 traversals of the list in the following way:
```elixir
Sanbase.Signal.access_map()
|> Enum.reduce([], fn
  {signal, ^level}, acc -> [signal | acc]
  _, acc -> acc
end)
```

It can be done as a separate function that accepts the access_map as argument and it can be reused bot for signals and metrics.

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -53,15 +53,15 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
+  @free_queries GraphqlSchema.get_all_with_access_level(:free)
+  @free_queries_mapset MapSet.new(@free_queries)
+  def free_queries_mapset(), do: @free_queries_mapset
 
   @restricted_metrics GraphqlSchema.get_all_with_access_level(:restricted)
   @restricted_metrics_mapset MapSet.new(@restricted_metrics)
   def restricted_metrics_mapset(), do: @restricted_metrics_mapset
 
-  @all_metrics @free_metrics ++ @restricted_metrics
+  @all_metrics @free_queries ++ @restricted_metrics
```
IvanIvanoff, line 64: It's confusing to rename only one of the attributes. Now metrics = queries + metrics. We can use the `query_or_argument` name used in some of the places.

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -127,7 +127,7 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
         when plan: atom(), product: binary(), restriction_type: atom()
   def get_available_metrics_for_plan(product, plan, restriction_type \\ :all) do
     case restriction_type do
-      :free -> @free_metrics
+      :free -> @free_queries
```
IvanIvanoff, line 130: Here the queries/metrics naming mixture is confusing. See the comment above.

lib/sanbase_web/graphql/middlewares/access_control.ex
```diff
@@ -55,9 +55,9 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
       definition.name
       |> Macro.underscore()
       |> String.to_existing_atom()
-      |> get_query_or_metric(source, arguments)
+      |> get_query_argument(source, arguments)
```
IvanIvanoff, line 58: Maybe  "query_**or**_argument" (add `or`) so we know it's either the query name or the argument.

lib/sanbase/billing/graphql_schema.ex
atanasormanow: Yeah, thanks,  i was thinking about an alternative solution as well, cuz it looked a bit weird.

lib/break.ex
```diff
@@ -8,4 +8,10 @@ defmodule Sanbase.Break do
       raise(CompileError, unquote(value))
     end
   end
+
+  def if_kw_invalid?(keyword_list, valid_keys: valid_keys) do
+    for {k, _} <- keyword_list,
```
IvanIvanoff, line 13: Keyword lists, unlike maps, cannot be partially pattern matched (pattern match one of the keys if there are more) and they have order. So, if you add a new key, or if you already have keys and swap them, the pattern match will no longer work. So you better do:
```elixir
def if_kw_invalid?(keyword_list, opts) do
  valid_keys = Keyword.fetch!(opts, :valid_keys) # Use fetch! instead of get to enforce the key
  ...
```

lib/sanbase/billing/graphql_schema.ex
```diff
@@ -47,13 +47,23 @@ defmodule Sanbase.Billing.GraphqlSchema do
     # meta(access: :restricted, min_plan: [sanapi: :pro, sanbase: :free])
     query_min_plan_map = get_query_min_plan_map()
 
-    Sanbase.Metric.min_plan_map()
-    |> Enum.into(query_min_plan_map, fn
-      {metric, product_plan_map} when is_map(product_plan_map) ->
-        {{:metric, metric}, product_plan_map}
+    metrics_and_queries_min_plan_map =
```
IvanIvanoff, line 50: It would be better to assign every `min_plan_map` to a variable holding only it (query/metric/signal) and end the function with a `Map.merge` pipeline merging all of them only once. Now, if we need to add a new type here, the next variable would be named `metrics_and_queries_and_signals_min_plan_map`

lib/sanbase/billing/plan/access_checker.ex
```diff
@@ -118,16 +103,16 @@ defmodule Sanbase.Billing.Plan.AccessChecker do
...
 
   @spec get_available_metrics_for_plan(product, plan, restriction_type) :: list(binary())
         when plan: atom(), product: binary(), restriction_type: atom()
   def get_available_metrics_for_plan(product, plan, restriction_type \\ :all) do
     case restriction_type do
-      :free -> @free_metrics
+      :free -> @free_query_or_argument
       :restricted -> @restricted_metrics
       :custom -> @custom_access_queries
       :all -> @all_metrics
```
IvanIvanoff, line 118: Something here is not right. For `:free` it uses `query_or_argument`, for `:restricted` - `metrics` and for `:custom` - `queries`.

IvanIvanoff: I see that `get_available_metrics_for_plan` is used in the context of metrics only, so all results should be metrics only.

lib/sanbase/clickhouse/metric/file_handler.ex
```diff
@@ -4,11 +4,23 @@ defmodule Sanbase.Clickhouse.MetricAdapter.FileHandler do
...
       map
       |> Enum.into(%{}, fn
-        %{"name" => name, ^field => value} -> {name, transform_fn.(value)}
-        %{"name" => name} -> {name, nil}
+        %{"name" => name, ^field => value} ->
+          {name, transform_fn.(value)}
+
+        %{"name" => name} ->
+          if required? do
+            Break.break("The field \"#{field}\" in the #{Jason.encode!(name)} metric is required")
```
IvanIvanoff, line 20: Why `Jason.encode!/1` here?

IvanIvanoff: Also, this functionality is exactly the same as the signal file handler with the except of the `metric`/`signal` word in the error. You can add a TODO comment here to extract this common logic in another PR after that.

lib/sanbase/clickhouse/metric/file_handler.ex
```diff
@@ -4,11 +4,26 @@ defmodule Sanbase.Clickhouse.MetricAdapter.FileHandler do
   require Sanbase.Break, as: Break
 
   defmodule Helper do
-    def name_to_field_map(map, field, transform_fn \\ fn x -> x end) do
+    def name_to_field_map(map, field, opts \\ []) do
+      Sanbase.Break.if_kw_invalid?(opts, valid_keys: [:transform_fn, :required?])
```
tspenov, line 8: You can use just `Break` since it is required above.

tspenov (approved):

IvanIvanoff (approved):

PR #2705, opened 2021-04-29
Evaluate the alerts in smaller batches

IvanIvanoff:
## Changes

Up until now, the alerts of a given type were all evaluated in a stream and only after all of them finish evaluating they are sent to the users. If an error stops the evaluation, none of the signals are sent.

With this rework the whole cycle of evaluate-send-update_db of the alerts is broken into smaller isolated batches that do all the work for a small number of alerts. This way errors are located only to a small batch of alerts instead of breaking all of them.

Also, as part of the rework, the logging of the alert evaluation progress is improved.

## Checklist:


lib/sanbase/alerts/evaluator/scheduler.ex
```diff
@@ -42,11 +42,78 @@ defmodule Sanbase.Alert.Scheduler do
 
   # Private functions
 
+  @batch_size 20
   defp run(type) do
-    {updated_user_triggers, sent_list_results} =
+    Logger.info("Schedule evaluation for the alerts of type #{type}")
+
+    run_uuid = "[" <> (:crypto.strong_rand_bytes(5) |> Base.hex_encode32()) <> "]"
```
tspenov, line 49: we can use: `UUID.uuid1()` since we have `uuid` lib already :?

lib/sanbase/alerts/evaluator/evaluator.ex
```diff
@@ -102,7 +100,20 @@ defmodule Sanbase.Alert.Evaluator do
...
+            e ->
+              Logger.error("""
+              Error in populate_payload/1 for alert with id #{user_trigger.id}
+
+              Template: #{inspect(template)}
+
+              KV: #{inspect(kv)}
+              """)
+
+              reraise e, __STACKTRACE__
```
tspenov, line 115: first time I see there is such thing as `reraise`

IvanIvanoff: This preserves the stacktrace. If it was just `raise` it would be rewritten.

lib/sanbase/alerts/evaluator/scheduler.ex
tspenov: In that case I think it can stay as is now but maybe remove the `[]` from `run_uuid` and put them in the interpolated log message :?

IvanIvanoff: It is not used outside the logs, the brackets are here so they are not missed in some log. I can do this change though

tspenov (approved):

PR #2721, opened 2021-05-14
Support charge.failed stripe event

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/billing/subscription/event_handling/discord_notification.ex
+++ b/lib/sanbase/billing/subscription/event_handling/discord_notification.ex
@@ -20,7 +20,7 @@ defmodule Sanbase.Billing.DiscordNotification do
   end
 
   def handle_event(
-        fail_type,
+        :charge_fail,
         %{
           data: %{
             amount: amount,
@@ -29,7 +29,7 @@ defmodule Sanbase.Billing.DiscordNotification do
           }
         }
       )
-      when fail_type in [:payment_fail, :charge_fail] and amount > 1 do
+      when amount > 1 do
     seller_message = stripe_event["data"]["outcome"]["seller_message"]
     failure_message = stripe_event["data"]["failure_message"]
     formatted_amount = format_cents_amount(amount)
@@ -40,13 +40,7 @@ defmodule Sanbase.Billing.DiscordNotification do
     Event: https://dashboard.stripe.com/events/#{stripe_event_id}
     """
 
-    title =
-      case fail_type do
-        :payment_fail -> "Stripe Payment"
-        :charge_fail -> "Stripe Charge"
-      end
-
-    do_send_to_discord(message, title)
+    do_send_to_discord(message, "Stripe Charge")
   end
 
   def handle_event(:cancel_subscription, %{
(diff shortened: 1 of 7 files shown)


lib/sanbase/billing/subscription/event_handling/discord_notification.ex
```diff
@@ -29,18 +29,26 @@ defmodule Sanbase.Billing.DiscordNotification do
...
-    """
-
-    do_send_to_discord(message, "Stripe Payment")
+    message =
+      """
+      ⛔ Failed card charge for #{formatted_amount}.
+      Details: #{failure_message}. #{seller_message}
+      Event: https://dashboard.stripe.com/events/#{stripe_event_id}
+      """
+      |> IO.inspect(label: "42", limit: :infinity)
```
tspenov, line 43: inspect left

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
```diff
@@ -29,18 +29,26 @@ defmodule Sanbase.Billing.DiscordNotification do
           }
         }
       )
-      when amount > 1 do
+      when fail_type in [:payment_fail, :charge_fail] and amount > 1 do
     seller_message = stripe_event["data"]["outcome"]["seller_message"]
```
tspenov, line 33: event_data map pattern match in function head is on atoms but access with strings `stripe_event["data"]["outcome"]["seller_message"]`. One of them is not correct.

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
```diff
@@ -20,7 +20,7 @@ defmodule Sanbase.Billing.DiscordNotification do
   end
 
   def handle_event(
-        :payment_fail,
+        fail_type,
         %{
           data: %{
             amount: amount,
```
tspenov, line 26: amount in `charge.failed` should be taken from `data["amount"]` but  in`payment_failed` from `data["total"]`

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
IvanIvanoff: `stripe_event` is the raw stripe event and it is placed in `extra_in_memory_data`. The rest of the keys are predefined and are atoms.

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
IvanIvanoff: `data["total"]` is for the success case only. Here is the old implementation: https://github.com/santiment/sanbase2/blob/9fc753fc8a3b29e0f862ddea1661f3a5400ef93d/lib/sanbase/billing/stripe_event.ex#L101-L117

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
IvanIvanoff: fixed

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
IvanIvanoff: Fixed

PR #2751, opened 2021-06-03
Add explanation to trending words alerts

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/alerts/trigger/settings/trending_words_trigger_settings.ex
+++ b/lib/sanbase/alerts/trigger/settings/trending_words_trigger_settings.ex
@@ -38,8 +38,6 @@ defmodule Sanbase.Alert.Trigger.TrendingWordsTriggerSettings do
             template_kv: %{},
             extra_explanation: nil,
             include_default_explanation: false,
-            default_explanation:
-              "A coin's appearance in trending words may suggest an increased risk of local tops and short-term price correction.",
             template: nil
 
   @type t :: %__MODULE__{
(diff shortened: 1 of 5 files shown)


lib/sanbase/alerts/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -35,7 +35,12 @@ defmodule Sanbase.Alert.Trigger.TrendingWordsTriggerSettings do
             filtered_target: %{list: []},
             triggered?: false,
             payload: %{},
-            template_kv: %{}
+            template_kv: %{},
+            extra_explanation: nil,
+            include_default_explanation: false,
+            default_explanation:
+              "A coin's appearance in trending words may suggest an increased risk of local tops and short-term price correction.",
```
IvanIvanoff, line 42: Two birds with 1 stone idea: Use a module attribute with this and place it at the top. This way we simplify the settings here (no long string) and this module attribute can serve as documentation for the reader of the code.

lib/sanbase/alerts/trigger/settings/trending_words_trigger_settings.ex
```diff
@@ -202,13 +210,13 @@ defmodule Sanbase.Alert.Trigger.TrendingWordsTriggerSettings do
...
-      {template, kv}
+      {template, kv} |> maybe_extend_with_explanation(settings)
     end
 
-    defp template_kv(%{operation: %{trending_word: true}} = operation, [word]) do
+    defp template_kv(%{operation: %{trending_word: true}} = settings, [word]) do
       kv = %{
         type: TrendingWordsTriggerSettings.type(),
-        operation: operation,
+        operation: settings,
```
IvanIvanoff, line 219: I think that was a mistake in the first place - it needs to be `operation`, it was just not bound to the correct value. This should be `settings.operation`

PR #2757, opened 2021-06-07
Extend top transfers

IvanIvanoff:
## Changes
Extend the result to include user specific data for the addresses in the top transfers API.
```graphql
{
  topTransfers(
    slug: "santiment"
    from: "utc_now-7d"
    to: "utc_now"
    page: 1
    pageSize: 5
  ){
    trxHash
    trxHash
    datetime
    
    fromAddress{
      address
      infrastructure
      # santiment-origin labels, they are the same for all callers
      labels{ name }
      currentUserAddressDetails{
        # user-origin labels, they are defined by the user and seen only by the user in this API
        labels{ name }
	watchlists { id name }
        notes
      }
    }
    
    toAddress{
      address
      infrastructure
      labels{ name }
      currentUserAddressDetails{
        labels{ name }
	watchlists { id name }
        notes
      }
    }
  }
}
```
The result contains json objects like this:
```json
      {
        "datetime": "2021-06-09T02:14:37Z",
        "fromAddress": {
          "address": "0x7c66550c9c730b6fdd4c03bc2e73c5462c5f7acc",
          "currentUserAddressDetails": {
            "labels": [],
            "notes": null,
            "watchlists": [
              {
                "id": 1485,
                "name": "AnotherBlockchainAddress"
              },
              {
                "id": 1484,
                "name": "BlockchainAddressWatchlis"
              }
            ]
          },
          "infrastructure": "ETH",
          "labels": [
            {
              "name": "CEX Trader"
            },
            {
              "name": "DEX"
            }
          ]
        },
        "toAddress": {
          "address": "0xbadffcb79efe57e954fd79621b29b0c2dc170171",
          "currentUserAddressDetails": {
            "labels": [
              {
                "name": "L1"
              },
              {
                "name": "exchange"
              },
              {
                "name": "Something"
              }
            ],
            "notes": null,
            "watchlists": []
          },
          "infrastructure": "ETH",
          "labels": [
            {
              "name": "CEX & DEX Trader"
            }
          ]
        },
        "trxHash": "0x196ed4cadbab1f49cf877409123a538dcb45492ce8fa5b9c627ce862eb2c69b5"
      }
```

## Checklist:


lib/sanbase_web/graphql/dataloader/postgres_dataloader.ex
```diff
@@ -149,10 +149,69 @@ defmodule SanbaseWeb.Graphql.MetricPostgresDataloader do
...
+      query =
+        from(
+          baup in Sanbase.BlockchainAddress.BlockchainAddressUserPair,
+          where: baup.user_id == ^user_id,
+          preload: [:labels],
+          inner_join: ba in Sanbase.BlockchainAddress,
+          on: baup.blockchain_address_id == ba.id and ba.address in ^addresses,
+          left_join: li in Sanbase.UserList.ListItem,
+          on: li.blockchain_address_user_pair_id == baup.id,
+          left_join: ul in Sanbase.UserList,
```
tspenov, line 164: these watchlists are only of the current user, right ?

lib/sanbase_web/graphql/dataloader/postgres_dataloader.ex
```diff
@@ -149,10 +149,69 @@ defmodule SanbaseWeb.Graphql.MetricPostgresDataloader do
...
+      elem =
+        row
+        |> Map.merge(%{
+          notes: row.blockchain_address_user_pair.notes,
+          watchlists: watchlist,
+          labels: labels
+        })
+
+      Map.update(acc, key, elem, fn user_address_pair ->
+        Map.update!(user_address_pair, :watchlists, &sort_watchlists(watchlist ++ &1))
```
tspenov, line 208: I know we'll probably have only a few watchlists but it is better to sort the watchlists after outside the reduce

lib/sanbase_web/graphql/dataloader/postgres_dataloader.ex
IvanIvanoff: The logic goes like this:
- Select pairs (blockchain address user pair) where the user_id matches
- Inner join with blockchain addresses on the pair id
- Left join with the list items on the pair id
- Left join with the watchlists on the list item id

If I am not missing something, this is enough to guarantee that the data is only for the current user because one list item can be contained in one watchlist. And from the pair matching, we're selecting only list items that are created by that user, which by extent means that only their watchlists are taken.

lib/sanbase_web/graphql/dataloader/postgres_dataloader.ex
IvanIvanoff: Done.

lib/sanbase_web/graphql/dataloader/postgres_dataloader.ex
tspenov: You can also add `ul.user_id == ^user_id` and that will be enough to guarantee it.

IvanIvanoff: Done

PR #2759, opened 2021-06-09
Add webinars models and APIs

tspenov:
## Ticket

1. Admin webinars
2. List webinars API
3. Register user for webinar

## Checklist:

## Usage

```graphql
    {
      getWebinars {
        url
        title
        description
        isPro
        startTime
        endTime
        imageUrl
      }
    }
```

```graphql
    mutation {
      registerForWebinar(webinar_id: #{webinar_id})
    }
```


lib/sanbase/webinars/registration.ex
```diff
@@ -0,0 +1,45 @@
...
+  def create(params) do
+    %__MODULE__{}
+    |> changeset(params)
+    |> Repo.insert()
+  end
+
+  def list_users_in_webinar(webinar_id) do
+    from(wr in __MODULE__, where: wr.webinar_id == ^webinar_id, preload: [:user, :webinar])
+    |> Repo.all()
+    |> Enum.map(& &1.user)
```
IvanIvanoff, line 37: You don't need to preload the webinar here (and the user in the other query).

lib/sanbase/webinars/webinar.ex
```diff
@@ -0,0 +1,80 @@
...
+      %__MODULE__{is_pro: true} = webinar ->
+        %{webinar | url: nil}
+
+      %__MODULE__{is_pro: false} = webinar ->
+        webinar
+    end)
+  end
+
+  defp show_only_preview_fields?(webinars, %{is_logged_in: true, plan_atom_name: plan})
+       when plan == :pro do
```
IvanIvanoff, line 77: pattern match on `:pro` in the map directly instead of using `when`

PR #2765, opened 2021-06-14
Metric name aliases mapping

IvanIvanoff:
## Changes

Add a REST endpoint that returns a list of json lines mapping between the public metric names and their internal counterpart used. This endpoint is to be used to create a clickhouse dictionary. The format in the dictionary is `JSONEachRow`, so every row is a separate JSON map and they are not combined into a list.

## Checklist:

### DIFF
--- a/lib/sanbase_web/controllers/metric_name_controller.ex
+++ b/lib/sanbase_web/controllers/metric_name_controller.ex
@@ -4,13 +4,15 @@ defmodule SanbaseWeb.MetricNameController do
   alias Sanbase.Model.Project
   require Logger
 
-  def clickhouse_metric_aliases(conn, _params) do
+  def api_metric_name_mapping(conn, _params) do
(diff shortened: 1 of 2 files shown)


tspenov (approved):

lib/sanbase_web/controllers/metric_name_controller.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  def api_metric_name_mapping(conn, _params) do
+    map = Sanbase.Clickhouse.MetricAdapter.FileHandler.name_to_metric_map()
+
+    data =
+      Enum.map(map, fn {k, v} ->
+        %{public_name: k, internal_name: v}
+        |> Jason.encode!()
+      end)
+      |> Enum.join("\n")
```
tspenov, line 15: Is that valid json with joining with newline in between ?

PR #2767, opened 2021-06-15
Apply restrictions to get_raw_signals

IvanIvanoff:
## Changes

Apply the access restrictions to the result of `get_raw_signals`. The restrictions are applied by removing all the data except the signal name - the rest of the fields are set to null and the `is_hidden` flag is set to true.

## Checklist:


tspenov (approved):
LGTM, but lets add a test case

atanasormanow (approved):

IvanIvanoff:
@tspenov I've added tests

IvanIvanoff:
This can still not be merged as it would break the frontend.

PR #2783, opened 2021-06-21
Fix wallet hunters contract execution error handling

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/wallet_hunters/contract.ex
+++ b/lib/sanbase/wallet_hunters/contract.ex
@@ -123,8 +123,11 @@ defmodule Sanbase.WalletHunters.Contract do
   def wallet_proposals() do
     count = wallet_proposals_count()
 
-    if count > 0 do
-      contract_execute("walletProposals", [0, count])
+    if is_integer(count) and count > 0 do
+      case contract_execute("walletProposals", [0, count]) do
+        {:error, _} -> []
+        result -> result
+      end
     else
       []
     end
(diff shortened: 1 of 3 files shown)


lib/sanbase/wallet_hunters/contract.ex
```diff
@@ -123,8 +123,12 @@ defmodule Sanbase.WalletHunters.Contract do
   def wallet_proposals() do
     count = wallet_proposals_count()
 
-    if count > 0 do
+    if is_integer(count) and count > 0 do
       contract_execute("walletProposals", [0, count])
+      |> case do
```
IvanIvanoff, line 128: one small nitpick: when you have a single expression it looks better to not pipe it:
```elixir
case  contract_execute("walletProposals", [0, count]) do
```

lib/sanbase_web/graphql/resolvers/wallet_hunters_resolver.ex
```diff
@@ -71,4 +75,7 @@ defmodule SanbaseWeb.Graphql.Resolvers.WalletHuntersResolver do
          0}
     end)
   end
+
+  defp wrap_ok([]), do: {:nocache, {:ok, []}}
```
IvanIvanoff, line 79: Could this `[]` be a valid response instead of an error converted to empty string?
If you do this wrap here I think you can pass the error go through instead of transforming it to [] and possible not cache here a correct result.

tspenov: It could have been when there were no proposals. Now it can't. I see your point with passing the error tuple but it would require one more condition in the `fetch_all` function. Also other way - moving `:nocache` outside of resolver I don't like because it is specifically for `cache_resolve`

PR #2788, opened 2021-06-23
Fix wallet hunters tests

tspenov:
## Checklist:


test/sanbase_web/graphql/wallet_hunters/wallet_hunters_api_test.exs
```diff
@@ -121,32 +121,37 @@ defmodule SanbaseWeb.Graphql.WalletHuntersApiTest do
...
-      :ok
+
+      mock_fun =
+        [
+          fn -> {:ok, "0x0000000000000000000000000000000000000000000000000000000000000003"} end,
+          fn -> all_proposals_resp() end
+        ]
+        |> Sanbase.Mock.wrap_consecutives(arity: 1)
+
+      global_mock =
```
IvanIvanoff, line 132: Maybe call it `prepared_mock` :?
I think only `Tesla` uses `global` vs `local` mock terminology and has options to enable/disable it.

PR #2800, opened 2021-06-25
Cryptocompare historical scraper

IvanIvanoff:
## Changes

- Add `Oban`
- Introduce `Sanbase.Cryptocompare.HistoricalWorker` that fetches minute historical data for a given date and represents an oban job.
  - The job is unique for 60 days. This means that if you rerun the job adding a few times in one day it won't add these jobs again. This is done so when we run the `schedule_oban_jobs` function in the span of a few days, the already finished jobs are not retried as this will consume a lot of API calls.
- Run `Sanbase.Cryptocompare.CCCAGGPairData.schedule_oban_jobs()` in order to schedule all historical jobs.

## Checklist:


lib/sanbase/cryptocompare/historical_scheduler.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    {:ok, %{}}
+  end
+
+  def enabled?(), do: Config.get(:enabled?) |> String.to_existing_atom()
+
+  def add_jobs(base_asset, quote_asset, from, to) do
+    Ecto.Multi.new()
+    |> add_jobs_to_multi(base_asset, quote_asset, from, to)
+    |> Sanbase.Repo.transaction()
```
tspenov, line 49: Why do we use multi instead of just `insert_all` ?

IvanIvanoff: I haven't checked if there was `Oban.insert_all`. There is.

tspenov (approved):

PR #2808, opened 2021-07-01
Scrape Cryptocompare websocket data

IvanIvanoff:
## Changes

Subscribe to the CCCAGG Aggregate Index websocket endpoint of Cryptocompare and export the data to Kafka. In order to be able to use this subscription, the API key used must have an Enterprise subscription.

The data is exported into 2 different tables:
- The USD and BTC prices are exported to the currently used asset_prices_topic and they will end up in our existing database.
- The full data is exported to a new `asset_price_pairs` topic. It is called like this because now every coin will have a lot more trading pairs compared to just USD and BTC. Every record will contain data only for a given base/quote asset pair.
- The data in the new topic will be stored with its cryptocompare name instead of our slug. This is so we can export all the data we need without losing the data for which we don't have an asset. In this PR https://github.com/santiment/sanbase2/pull/2811 I've added the mapping of the known assets. We can add a CH dictionary that will be used to fetch the proper name in the queries.

## Checklist:

## Deployment steps
The Kafka topic has been created and the API keys are added here: https://github.com/santiment/devops/pull/2362

### DIFF
--- a/lib/sanbase/cryptocompare/websocket_scraper.ex
+++ b/lib/sanbase/cryptocompare/websocket_scraper.ex
@@ -1,0 +1,242 @@
+defmodule Sanbase.Cryptocompare.WebsocketScraper do
+  @moduledoc ~s"""
+  Scrape the prices from Cryptocompare websocket API
+  https://min-api.cryptocompare.com/documentation/websockets
+
+  Use the cryptocompare API to fetch prices aggregated across many exchanges
+  in near-realtime. For every base/quote asset pairs fetch:
+    - price
+    - volume 24h (sliding window) - in number of tokens and in the quote asset currency
+    - top tier exchanges volume 24h (sliding window) - in number of tokens and
+      in the quote asset currency
+  """
+  use WebSockex
+
+  alias Sanbase.ExternalServices.Coinmarketcap.PricePoint, as: AssetPricesPoint
+  alias Sanbase.Cryptocompare.PricePoint, as: CryptocompareAssetPricesPoint
+
+  require Logger
+  require Sanbase.Utils.Config, as: Config
+
+  # giving it a name makes sure only one instance of the scraper is alive at a time
+  @name :cryptocompare_websocket_scraper
+
+  @asset_price_pairs_exporter :asset_price_pairs_exporter
+  @asset_prices_exporter :prices_exporter
+
+  def child_spec(_opts \\ []) do
+    %{
+      id: :__cryptocompare_websocket_price_scraper__,
+      start: {__MODULE__, :start_link, []}
+    }
+  end
+
+  def start_link() do
+    state = %{
+      last_points: %{},
+      subscriptions: MapSet.new(),
+      msg_number: 0
+    }
+
+    extra_headers = [{"authorization", "Apikey #{api_key()}"}]
+
+    WebSockex.start_link(websocket_url(), __MODULE__, state,
+      name: @name,
+      extra_headers: extra_headers
+    )
+  end
+
+  def enabled?(), do: Config.get(:enabled?) |> String.to_existing_atom()
+
+  def terminate(reason, state) do
+    if reason != :normal do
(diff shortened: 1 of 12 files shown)


lib/sanbase/prices/point.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Sanbase.Price.Point do
```
tspenov, line 1: This looks like duplicate to `Sanbase.Cryptocompare.PricePoint`

mix.exs
```diff
@@ -145,8 +148,8 @@ defmodule Sanbase.Mixfile do
       {:ueberauth_twitter, github: "santiment/ueberauth_twitter"},
       {:uuid, "~> 1.1"},
       {:vex, "~> 0.8.0", override: true},
-      {:ethereumex, "~> 0.7.0"},
-      {:ex_abi, "~> 0.5.4"}
+      {:waffle, "~> 1.1"},
```
tspenov, line 151: is waffle used ?

IvanIvanoff: Not sure. I just sort the deps in alphabetical order from time to time. Waffle was between 2 deps starting with `c`

lib/sanbase/prices/point.ex
IvanIvanoff: Yes, it is. First I though of putting all files in `sanbase/price` but then decided to go with a separate directory. This is a leftover.

lib/sanbase/cryptocompare/websocket_scraper.ex
```diff
@@ -0,0 +1,235 @@
...
+    export_data_point(point, last_points)
+
+    {:ok, %{state | last_points: last_points, msg_number: state.msg_number + 1}}
+  end
+
+  def handle_frame(%{"MESSAGE" => "FORCE_DISCONNECT"} = msg, _frame, state) do
+    Logger.warn("""
+    [CryptocompareWS] Received FORCE_DISCONNET for socket #{state.socket_id}. Reason: #{msg["INFO"]}
+    """)
+
```
tspenov, line 120: shouldn't `handle_disconnect` be called here ?

lib/sanbase/cryptocompare/websocket_scraper.ex
```diff
@@ -0,0 +1,235 @@
...
+  defp get_slug_data_map() do
+    result =
+      Sanbase.Model.Project.SourceSlugMapping.get_source_slug_mappings("cryptocompare")
+      |> Map.new()
+
+    {:ok, result}
+  end
+
+  defp api_key(), do: Config.module_get(Sanbase.Cryptocompare, :api_key)
+  defp websocket_url(), do: "wss://streamer.cryptocompare.com/v2?api_key=#{api_key()}"
```
tspenov, line 234: we can maybe use the api key in header version of authentication ? secrets in query string looks more prone to be visible in logs and other places

lib/sanbase/cryptocompare/websocket_scraper.ex
IvanIvanoff: `handle_disconnect/1` is a callback that is called from the library code. It is called in different scenarios: `:tcp_closed`, `:"$websockex_close_timeout"`, `:remote_close`, etc.

If we reach here we have received a :text message. This is properly pushed from the streamer (server) and it is followed by the server stopping the connection. This connection stop causes the `handle_disconnect` to be invoked.

I have tested this scenario the following way: Introduce a `Process.sleep(80000)` in the websocket library for the `:ping` frames and when the server does not receive 2 consecutive heartbeats this triggers the `handle_disconnect` callback.

lib/sanbase/cryptocompare/websocket_scraper.ex
IvanIvanoff: Done

tspenov (approved):

PR #2813, opened 2021-07-06
Add limit for login attempts

atanasormanow:
## Checklist:


lib/sanbase/accounts/email_login_attempts.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  def login_attempts_count(_user), do: 2
+
+  def check_login_attempts(user) do
+    login_attempts = login_attempts_count(user)
+
+    if login_attempts >= @allowed_login_attempts do
+      {:error, "Too many login attempts, please try again later."}
+    else
+      {:ok, user}
```
IvanIvanoff, line 13: Return `true` here. This function is not intended to do any transformation on the user struct

priv/repo/migrations/20210705104243_add_email_login_attempts_table.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Sanbase.Repo.Migrations.AddEmailLoginAttemptsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:email_login_attempts) do
+      add(:user_id, references(:users), on_delete: :delete_all, null: false)
+      add(:email, :string, null: false)
```
IvanIvanoff, line 7: No need to store emails in my opinion. All the email data can be fetched through the users association. Furthermore, you need to apply all the email normalizations here (strip whispaces, lowercase, etc.) before inserting here and this can cause issues if something changes.

lib/sanbase/accounts/user.ex
```diff
@@ -90,6 +90,7 @@ defmodule Sanbase.Accounts.User do
     has_many(:chart_configurations, Sanbase.Chart.Configuration, on_delete: :delete_all)
     has_many(:user_attributes, Sanbase.Intercom.UserAttributes, on_delete: :delete_all)
     has_many(:user_events, Sanbase.Intercom.UserEvent, on_delete: :delete_all)
+    has_many(:email_login_attempts, Sanbase.Accounts.EmailLoginAttempt, on_delete: :delete_all)
```
IvanIvanoff, line 93: More and more I believe that we should add these if we're going to use them. We won't use the attempts through the user struct, so to keep it simpler we can leave this out.

lib/sanbase_web/graphql/resolvers/user/auth_resolver.ex
```diff
@@ -77,8 +77,10 @@ defmodule SanbaseWeb.Graphql.Resolvers.AuthResolver do
         context: %{origin_url: origin_url}
       }) do
     with {:ok, user} <- User.find_or_insert_by(:email, email, %{username: args[:username]}),
+         true <- EmailLoginAttempt.has_allowed_login_attempts(user),
          {:ok, user} <- User.update_email_token(user, args[:consent]),
          {:ok, _user} <- User.send_login_email(user, origin_url, args) do
+      EmailLoginAttempt.record_login_attempt(user)
```
IvanIvanoff, line 83: Shouldn't this be in the `with` pipeline?

priv/repo/migrations/20210705104243_add_email_login_attempts_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Sanbase.Repo.Migrations.AddEmailLoginAttemptsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:email_login_attempts) do
+      add(:user_id, references(:users), on_delete: :delete_all, null: false)
```
IvanIvanoff, line 6: Maybe you can add the IP address here also and do both checks?

test/sanbase_web/graphql/user/user_api_test.exs
```diff
@@ -482,6 +482,22 @@ defmodule SanbaseWeb.Graphql.UserApiTest do
       assert result["firstLogin"]
     end
 
+    test "fails if the user has no allowed attempts left", context do
+      Sanbase.Mock.prepare_mock2(
+        &Sanbase.Accounts.EmailLoginAttempt.has_allowed_login_attempts/1,
+        false
+      )
```
IvanIvanoff, line 489: Add the login attempts struct in the factory `factory.ex`. Then you can insert a few attempts in the DB and do the actual test. As we spoke earlier - prefer to execute the graphql query without relying on too many external strange functions.
```elixir
for _ <- 1..5, do: insert(:email_login_attempt, user_id: user.id)
result =
conn
|> post("/graphql", mutation_skeleton(mutation))
|>  json_response(200)
// |> get the error somehow

```
and let the code do the actual DB call so you can also test the SQL query.

test/sanbase_web/graphql/user/user_api_test.exs
```diff
@@ -482,6 +482,22 @@ defmodule SanbaseWeb.Graphql.UserApiTest do
...
+      |> Sanbase.Mock.run_with_mocks(fn ->
+        msg =
+          execute_mutation_with_error(
+            context.conn,
+            context.mutation_func.(%{email: "john@example.com"})
+          )
+
+        assert msg == "Can't login"
```
IvanIvanoff, line 497: Try to improve the message when the limit is hit, so the user can be shown something like `Too many login attempts, try again after a few minutes` for example.

test/sanbase_web/graphql/user/user_api_test.exs
IvanIvanoff: I added a PR where I add some more login tests and refactor these here. Here is how I did it: https://github.com/santiment/sanbase2/pull/2824/files#diff-4fb7ad575541742b43b4b15046bd625c4e6a9d24f19602f8791b92e8be4cc4f9R243-R256
By the time this PR is ready I will have merged my changes so you can rebase to master then and start from there.

lib/sanbase/accounts/email_login_attempt.ex
```diff
@@ -0,0 +1,45 @@
...
+  import Ecto.Query
+
+  alias Sanbase.Repo
+
+  @interval_in_minutes 15
+  @allowed_login_attempts 5
+
+  schema "email_login_attempts" do
+    belongs_to(:user, Sanbase.Accounts.User)
+    field(:ip_address, :string, size: 15)
```
IvanIvanoff, line 13: I am not convinced this is the best approach. This way if the client changes the IP address they still can send hundreds of login emails to the same address.

IvanIvanoff: Maybe this format can work but if you use 2 separate checks:
- Check for `user_id` count only
- Check for `remote_ip` count only
And if any of the 2 checks fails - fail the whole login process.

lib/sanbase/accounts/email_login_attempt.ex
```diff
@@ -0,0 +1,53 @@
...
+
+    if too_many_login_attempts? or too_many_ip_attempts? do
+      {:error, :too_many_login_attempts}
+    else
+      :ok
+    end
+  end
+
+  def record_login_attempt(%{id: user_id}, remote_ip) do
+    %__MODULE__{user_id: user_id, ip_address: remote_ip}
```
IvanIvanoff, line 31: You should do this through a changeset

lib/sanbase_web/graphql/resolvers/user/auth_resolver.ex
```diff
@@ -74,14 +74,25 @@ defmodule SanbaseWeb.Graphql.Resolvers.AuthResolver do
...
+        }
       }) do
+    remote_ip = Sanbase.Utils.IP.ip_tuple_to_string(remote_ip)
+
     with {:ok, user} <- User.find_or_insert_by(:email, email, %{username: args[:username]}),
+         :ok <- EmailLoginAttempt.has_allowed_login_attempts(user, remote_ip),
          {:ok, user} <- User.update_email_token(user, args[:consent]),
-         {:ok, _user} <- User.send_login_email(user, origin_url, args) do
+         {:ok, _user} <- User.send_login_email(user, origin_url, args),
+         %EmailLoginAttempt{} <- EmailLoginAttempt.record_login_attempt(user, remote_ip) do
```
IvanIvanoff, line 88: Make this an :ok|:error tuple. This will require you to rework the `Repo.insert!` to `Repo.insert` inside this function.

test/support/factory/factory.ex
```diff
@@ -713,6 +713,10 @@ defmodule Sanbase.Factory do
     %Proposal{}
   end
 
+  def email_login_attempt_factory() do
+    %EmailLoginAttempt{}
```
IvanIvanoff, line 717: Add some default data here. For the user you can use `build(:user)` to create a new user in case nothing is provided.

priv/repo/migrations/20210705104243_add_email_login_attempts_table.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Sanbase.Repo.Migrations.AddEmailLoginAttemptsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:email_login_attempts) do
```
IvanIvanoff, line 5: You can 2 indexes - on on the ip and one on the user_id so the searches can be done faster.

lib/sanbase/accounts/email_login_attempt.ex
```diff
@@ -0,0 +1,62 @@
...
+    %__MODULE__{}
+    |> changeset(%{user_id: user_id, ip_address: remote_ip})
+    |> Repo.insert()
+  end
+
+  def changeset(%__MODULE__{} = attempt, attrs \\ %{}) do
+    attempt
+    |> cast(attrs, [:user_id, :ip_address])
+    |> validate_required([:user_id, :ip_address])
+    |> validate_length(:ip_address, min: 7, max: 15)
```
tspenov, line 45: We should also support ipv6 addresses

atanasormanow: Fair, i wonder if it's safe to leave it without any validation, as `Sanbase.Utils.IP.ip_tuple_to_string/1` will throw an error in case it fails to parse it anyways

tspenov: yes, removing the validation seems ok to me

IvanIvanoff: This `ip_tuple_to_string` is called from outside this module, so it's not a 100% guarantee

lib/sanbase/accounts/email_login_attempt.ex
```diff
@@ -0,0 +1,65 @@
+defmodule Sanbase.Accounts.EmailLoginAttempt do
+  use Ecto.Schema
+
+  import Ecto.{Query, Changeset}
+
+  alias Sanbase.Repo
+
+  @interval_in_minutes 15
```
IvanIvanoff, line 8: Let's make this interval smaller, say 5 minutes

PR #2828, opened 2021-07-14
Improve human readable values in alerts

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/template_engine/template_engine.ex
+++ b/lib/sanbase/template_engine/template_engine.ex
@@ -52,9 +52,20 @@ defmodule Sanbase.TemplateEngine do
   # Number bigger than 1000000 are made human readable: 1.54 Million, 85.00 Billion
   defp human_readable(data) do
     case data do
-      num when is_number(num) and num >= 1_000_000 -> Number.Human.number_to_human(num)
-      num when is_number(num) and num >= 1000 -> Number.Delimit.number_to_delimited(num)
-      _ -> data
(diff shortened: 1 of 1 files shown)


lib/sanbase/template_engine/template_engine.ex
```diff
@@ -52,9 +52,11 @@ defmodule Sanbase.TemplateEngine do
   # Number bigger than 1000000 are made human readable: 1.54 Million, 85.00 Billion
   defp human_readable(data) do
     case data do
-      num when is_number(num) and num >= 1_000_000 -> Number.Human.number_to_human(num)
-      num when is_number(num) and num >= 1000 -> Number.Delimit.number_to_delimited(num)
-      _ -> data
+      num when is_number(num) and num >= 1_000_000 or num <= -1_000_000 -> Number.Human.number_to_human(num)
+      num when is_number(num) and num >= 1000 or num <= -1000 -> Number.Delimit.number_to_delimited(num)
+      num when is_number(num) and num > -1 and num < 1 -> Number.Delimit.number_to_delimited(precision: 8)
+      num when is_float(num) -> Float.round(num, 2)
```
tspenov, line 58: We can use again `Number.Delimit.number_to_delimited(precision: 2)`. It also returns a string

IvanIvanoff: It returns the same result, but for sake of consistency we should make all the fields strings indeed.

PR #2848, opened 2021-07-23
Switch the postgres migrations to use a separate pool

IvanIvanoff:
## Changes
The migrations sometimes need privileges that are not granted to the user that queries the database. These privileges include creation of new schemas and creation of new types. To address that introduce a new repo that uses a separate user, with more privileges.

The DATABASE_URL_MIGRATION_USER URL should never be used in local development as this would allow it to run migrations against stage/prod databases from your  local machine. It should not be unsafe as we have the `database_safety.ex` task that guards this.

Introduce the new repo with `pool_size: 0` and `max_overflow: 5` which would make it not keep any connections alive but can create up to 5 when needed. This is important as the repo would be used only when running the migration in the start process and won't be used after that.

## Checklist:


config/config.exs
```diff
@@ -12,7 +12,7 @@ config :tzdata, :autoupdate, :disabled
 # General application configuration
 config :sanbase,
   env: Mix.env(),
-  ecto_repos: [Sanbase.Repo],
+  ecto_repos: [Sanbase.MigrationRepo],
```
tspenov, line 15: why `Sanbase.Repo` is removed ?

IvanIvanoff: So the migrations are run using the `MigrationRepo`, but now I think this is not the way to do it.

IvanIvanoff: In order to fix that I had to change the ecto command aliases in mix.exs and change `priv/repo` to `priv/migration_repo`

PR #2850, opened 2021-07-23
Migrate user attributes to kafka

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/intercom/user_attributes.ex
+++ b/lib/sanbase/intercom/user_attributes.ex
@@ -35,3 +35,46 @@ defmodule Sanbase.Intercom.UserAttributes do
     |> Repo.all()
   end
 end
+
+defmodule Sanbase.Intercom.UserAttributes.KafkaDump do
+  import Ecto.Query
+  alias Sanbase.Intercom.UserAttributes
+  alias Sanbase.Repo
+
+  @topic "sanbase_user_intercom_attributes"
+
+  def run_async() do
+    Task.Supervisor.async_nolink(Sanbase.TaskSupervisor, fn ->
+      run()
+    end)
+  end
+
+  def run() do
+    query =
+      from(ua in UserAttributes,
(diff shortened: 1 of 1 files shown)


lib/sanbase/intercom/user_attributes.ex
```diff
@@ -35,3 +35,46 @@ defmodule Sanbase.Intercom.UserAttributes do
...
+              attributes |> Map.delete("email") |> Map.delete("name") |> Jason.encode!(),
+            timestamp: timestamp
+          }
+
+          SanExporterEx.Producer.send_data(@topic, [{key, Jason.encode!(data)}])
+        end)
+      end,
+      timeout: :infinity
+    )
+  end
```
IvanIvanoff, line 79: I think it would be better to fetch the user attributes (they won't be that many) and don't use a stream and transaction here. In this case there's nothing to be rolled back in case of a failure, which defies the usage of the transaction.

tspenov: The data is like timeseries, every day dump of the attributes. They are over 60 Mil so stream is needed. And transaction is used because repo stream  does not work outside of transaction.

lib/sanbase/intercom/user_attributes.ex
```diff
@@ -35,3 +35,46 @@ defmodule Sanbase.Intercom.UserAttributes do
...
+      )
+
+    stream = Repo.stream(query, timeout: :infinity)
+
+    Repo.transaction(
+      fn ->
+        stream
+        |> Enum.each(fn %{user_id: user_id, attributes: attributes, timestamp: timestamp} ->
+          timestamp = DateTime.to_unix(timestamp)
+          key = "#{user_id}_#{timestamp}" |> IO.inspect()
```
IvanIvanoff, line 65: Is this guaranteed to be unique?

tspenov:
not needed

PR #2858, opened 2021-07-26
Use MigrationRepo

IvanIvanoff:
Reverts santiment/sanbase2#2857


IvanIvanoff:
Close in favor of #2860

PR #2859, opened 2021-07-26
Migrate `twitter_followers` from influxDB to CH

atanasormanow:
## Checklist:


lib/sanbase/twitter/followers_worker.ex
```diff
@@ -0,0 +1,57 @@
...
+    result =
+      generate_dates_inclusive(from, to)
+      |> Enum.chunk_every(30)
+      |> Enum.map(fn list -> {List.first(list), List.last(list)} end)
+      |> Enum.flat_map_reduce(:ok, fn {interval_from, interval_to}, acc ->
+        case TwitterFollowers.timeseries_data(
+               "twitter_followers",
+               %{slug: slug},
+               date_to_datetime(interval_from),
+               date_to_datetime(interval_to),
```
IvanIvanoff, line 37: You would lose whole days this way.
After the generate/chunk/map you'll get data like this:
```elixir
[
  {~D[2021-01-01], ~D[2021-01-30]},
  {~D[2021-01-31], ~D[2021-03-01]},
  ...
]
```

With `date_to_datetime` you'll transform every date to a datetime with 00:00:00 time.
The first call will be with: `~U[2021-01-01 00:00:00Z]` and `~U[2021-01-30 00:00:00Z]`
The second call will be with: `~U[2021-01-31 00:00:00Z]` and `~U[2021-03-01 00:00:00Z]`

The problem is that no call includes the data in the interval `~U[2021-01-30 00:00:01Z]` to `~U[2021-01-30 23:59:59Z`.

To fix this you can extend the `date_to_datetime` to accept an optional second parameter `opts` where you can control the time. So your call would look like:

```elixir
date_to_datetime(interval_from, time: ~T[00:00:00]),
date_to_datetime(interval_to, time, ~T[23:59:59.999]),
```

lib/sanbase/twitter/followers_scheduler.ex
```diff
@@ -0,0 +1,64 @@
...
+  def get_recorded_slugs() do
+    query = """
+    SELECT args->>'slug' FROM oban_jobs
+    """
+
+    {:ok, %{rows: recorded_slugs}} = Ecto.Adapters.SQL.query(Sanbase.Repo, query)
+
+    recorded_slugs
+    |> Enum.reject(&(&1 == [nil]))
+    |> Enum.concat()
```
IvanIvanoff, line 53: I think it is more readable this way:
```elixir
recorded_slugs
    |> List.flatten()
    |> Enum.reject(&is_nil/1)
```

lib/sanbase/twitter/followers_scheduler.ex
```diff
@@ -0,0 +1,64 @@
...
+      resume()
+    end
+
+    {:ok, %{}}
+  end
+
+  def enabled?(), do: Config.get(:enabled?) |> String.to_existing_atom()
+
+  def add_jobs() do
+    {:ok, projects_with_twitter_link} = TwitterFollowers.available_slugs()
```
tspenov, line 36: shouldn't that be slugs or the name is misleading ?

lib/sanbase/twitter/followers_scheduler.ex
```diff
@@ -0,0 +1,64 @@
...
+    slugs = Enum.map(projects_with_twitter_link, & &1.slug)
+
+    (slugs -- get_recorded_slugs())
+    |> Enum.map(&create_oban_job(&1))
+    |> Oban.insert_all()
+  end
+
+  def get_recorded_slugs() do
+    query = """
+    SELECT args->>'slug' FROM oban_jobs
```
tspenov, line 46: shouldn't we specify also the `queue` in the select query because here we select from all oban jobs? Also I think there should be a `distinct` otherwise there will be duplicates?

lib/sanbase/twitter/followers_scheduler.ex
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    {:ok, %{rows: recorded_slugs}} = Ecto.Adapters.SQL.query(Sanbase.Repo, query)
```
tspenov, line 49: That might be outside the scope of this PR but we can add a schema module on oban jobs and utilize more ecto instead of raw sql?

lib/sanbase/twitter/followers_scheduler.ex
atanasormanow: I suppose it should be slugs, but it's not the name. It seems that the function returns projects atm.

atanasormanow: This metric adapter indeed returns projects instead of slugs, will fix it here.

tspenov: actually I think we should fix this earlier in separate PR cause it might cause subtle bug

lib/sanbase/twitter/followers_scheduler.ex
IvanIvanoff: There is such schema defined in the `Oban.Job` module. If we want to use it we can - `from(job in Oban.Job, select: ..., where: ...)`. We should not define it ourselves as we'll need to keep it in sync one very update, which we'll quickly stop doing.

lib/sanbase/twitter/followers_scheduler.ex
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def create_oban_job(slug) do
+    {:ok, first_datetime} =
+      Twitter.MetricAdapter.first_datetime("twitter_followers", %{slug: slug})
+
+    {:ok, last_datetime} =
+      Twitter.MetricAdapter.last_datetime_computed_at("twitter_followers", %{slug: slug})
+
+    FollowersWorker.new(%{slug: slug, from: first_datetime, to: last_datetime})
```
IvanIvanoff, line 62: I think you should not include the `to` here, but use `now` in the job. Otherwise some time will have passed between you defining the job and the actual job execution and you might miss some of the data.

lib/sanbase/twitter/followers_worker.ex
```diff
@@ -0,0 +1,69 @@
...
+        end
+      end)
+
+    case result do
+      {data, :ok} -> {:ok, data}
+      {_, error} -> error
+    end
+  end
+
+  defp export_data(data) do
```
tspenov, line 61: you should pass also `slug` here

lib/sanbase/twitter/timeseries_point.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Sanbase.Twitter.TimeseriesPoint do
+  defstruct [
+    :datetime,
+    :value
+  ]
```
tspenov, line 5: that should be extended with `:slug` also

PR #2876, opened 2021-07-30
Add alerts stats per user

tspenov:
## Changes

```
    {
      alertsStats {
        totalFired
        totalFiredWeeklyAvg
        totalFiredPercentChange
        data {
          slug
          project { slug }
          count
          percentChange
          alertTypes
        }
      }
    }
```

## Checklist:


atanasormanow (approved):

lib/sanbase/alerts/stats.ex
```diff
@@ -0,0 +1,124 @@
...
+    end)
+    |> Enum.sort_by(fn {_slug, data} -> data.count end, :desc)
+  end
+
+  defp percent_change(old, new) do
+    if new - old > 0 do
+      {(new - old) / old * 100, "more"}
+    else
+      {(old - new) / old * 100, "less"}
+    end
```
IvanIvanoff, line 122: we have this in `Sanbase.Math`. It also has some guards that here would fail - if `old` is `0` or one of the arguments is not a number.
```elixir
case Sanbase.Math.percent_change(old,new) do
  n when is_number(n) and n > 0 -> "more"
  n when is_number(n). and n <= 0 -> "less"
  nil -> "error"
  end
  ```

File: lib/sanbase/alerts/stats.ex [hunk 2]
```diff
@@ -0,0 +1,124 @@
...
+      "signal_data" -> alert.signal
+      _ -> alert.type
+    end
+    |> String.replace("_", " ")
+    |> String.capitalize()
+  end
+
+  defp slug_ticker_map(slugs) do
+    slugs
+    |> Sanbase.Model.Project.tickers_by_slug_list()
```
IvanIvanoff, line 90: I think we should deprecate these functions in the `Project` module and use the `Project.List` module instead. I think you can add a function similar to `field_slug_map/3` that would return a map from slug to a field (and probably don't take a list of values/slugs as argument). Or better, I think you can make it so that it takes a list of slugs that defaults to `:all`

IvanIvanoff: So it would be something like:
```elixir
def slug_field_map(field, opts \\ []) do
  opts = Keyword.put(opts, preload?: false)
  projects_query(opts)
  |> select([p], {p.slug, field(p, ^field))
  |> Repo.all()
  |> Map.new()
end
```

This way we can leverage all the supports common opts and make all fetching of projects lists/maps very uniform.

IvanIvanoff: And use it just like `slug_field_map(:ticker)`

PR #2892, opened 2021-08-02
Do not use a common process to send data to Kafka in CPC historical

IvanIvanoff:
## Changes

Add a function to KafkaExporter that sends the data to a topic
immediately. Prefer using this new function instead of calling
SanExporterEx directly as the latter does not have a in-memory producer
to be used in dev/test.

This new function is used when exporting data in sync from the caller process,
so the errors can be handled and the job can be properly stored along with
the errors. This is an improvement as the Oban queue is running with a
concurrency of 25, so not sequencing everyhing through the same process
makes the processes not wait for the exporter's gen_server to handle their export
so not sequencing everyhing through the same process makes the
processes. This also makes the oban job performs more isolated as now
they do not rely on a KafkaExporter gen_server exporter running.

## Checklist:

### DIFF
--- a/lib/sanbase/cryptocompare/supervisor.ex
+++ b/lib/sanbase/cryptocompare/supervisor.ex
@@ -63,21 +63,6 @@ defmodule Sanbase.Cryptocompare.Supervisor do
         start_if(
           fn -> HistoricalScheduler end,
           fn -> HistoricalScheduler.enabled?() end
-        ),
-
-        # Kafka exporter for the historical OHLCV exporter
-        start_if(
-          fn ->
-            Sanbase.KafkaExporter.child_spec(
-              id: :asset_ohlcv_price_pairs_exporter,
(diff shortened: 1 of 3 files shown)


lib/sanbase/cryptocompare/supervisor.ex
```diff
@@ -63,21 +63,6 @@ defmodule Sanbase.Cryptocompare.Supervisor do
         start_if(
           fn -> HistoricalScheduler end,
           fn -> HistoricalScheduler.enabled?() end
-        ),
-
-        # Kafka exporter for the historical OHLCV exporter
-        start_if(
```
IvanIvanoff, line 69: This is removed while the price_pairs_only exporter is not because:
- ohlcv exporter is used only in 1 place, which is now removed in this PR
- the price_pairs_only exporter is used in the websocket exporter, so it cannot be removed

tspenov (approved):

PR #2895, opened 2021-08-03
Use Clickhouse instead of latest cmc data

IvanIvanoff:
## Checklist:


lib/sanbase_web/graphql/dataloader/clickhouse_dataloader.ex
```diff
@@ -12,6 +12,8 @@ defmodule SanbaseWeb.Graphql.ClickhouseDataloader do
     |> Enum.group_by(fn %{selector: selector} -> selector end)
     |> Sanbase.Parallel.map(fn {selector, group} ->
       {metric, from, to, opts} = selector
+      {:ok, from} = Sanbase.DateTimeUtils.to_datetime(from)
+      {:ok, to} = Sanbase.DateTimeUtils.to_datetime(to)
```
tspenov, line 16: maybe we should change it `to_datetime_interval` passing `{form, to}` and returning `{from, to}` using the same `DateTime.utc_now()` ?

PR #2912, opened 2021-08-13
Add labelled balances metrics

IvanIvanoff:
## Changes
- Add 2 labelled balance metrics:
  - `labelled_historical_balance`
  - `labelled_historical_balance_changes`
- Add `required_selectors` to the `getMetric` API. In case of the new label metrics, the `label_fqn` or `label_fqns` field must be provided. As this is not immediately obvious and the required selectors differ from metric to metric, now these fields are enforced and a nice error is returned when they are missing.
- Add some helper functions that simplify the json files handling
- Add `cumulative_sum` transform to the `getMetric` API
- Fix the `fit_from_datetime` to first apply the datetime rounding before comparison. This fixes issues where the first data point is inappropriately excluded
- Rework how `additional_filters` is implemented - use positional params instead of string interpolation
image

To get the labels that can be used in `labelFqn` and `labelFqns` fields:
```graphql
{
  getBlockchainAddressLabels{
    name
    origin
    humanReadableName
  }
}
```

```json
{
  "data": {
    "getBlockchainAddressLabels": [
      {
        "humanReadableName": "owner->KuCoin",
        "name": "santiment/owner->KuCoin:v1",
        "origin": "santiment"
      },
      {
        "humanReadableName": "owner->Poloniex",
        "name": "santiment/owner->Poloniex:v1",
        "origin": "santiment"
      }
    ]
  }
}
```

```graphql
{
  getMetric(metric: "labelled_historical_balance_changes"){
     timeseriesData(
        selector: {slug: "bitcoin" label_fqn: "santiment/owner->Coinbase:v1"}
        from: "utc_now-8d"
        to: "utc_now-5d"
        interval: "1d"){
           datetime
           value
        }
    }
}
```

```json
{
  "data": {
    "getMetric": {
      "timeseriesData": [
        {
          "datetime": "2021-08-26T00:00:00Z",
          "value": -174.4806595099975
        },
        {
          "datetime": "2021-08-27T00:00:00Z",
          "value": 165.83934908000023
        },
        {
          "datetime": "2021-08-28T00:00:00Z",
          "value": -856.1770302899997
        }
      ]
    }
  }
}
```

## Checklist:


tspenov:
`blockchainAddressLabels` is an existing query, so I propose instead of changing it, to create another API query for v2 labels.
Also the new query should return both `human label` and `fqn` and frontend should keep a mapping showing to user the `human readable` but using `fqn` when reading. Another way is that mapping to be in the backend.

IvanIvanoff:
> `blockchainAddressLabels` is an existing query, so I propose instead of changing it, to create another API query for v2 labels.
> Also the new query should return both `human label` and `fqn` and frontend should keep a mapping showing to user the `human readable` but using `fqn` when reading. Another way is that mapping to be in the backend.

The `fqn` is necessary to be exposed and used for API users that want the data to not automatically update so it can be used for ML. The format of `fqn` is very straightforward so I think the FE can just decode it and use the parts it needs (the key and value). I don't want to provide a human readable name directly as it will be less flexible. For example the version can be parsed and shown next to the key/value and you can choose from v1, v2, v3, etc. (group these together). Other features are also possible by letting the FE decode the `fqn` and not doing it for them.

`blockchainAddressLabels` is called ~8 times in the last 90 days, more than half of those would be me testing it. I much prefer to remove unused APIs so we don't maintain dead code. Instead of removing it, I am just reusing it here.
One more reason for removing/reusing dead APIs is that API users don't accidentally start using it instead of the newer one, which would make it only more difficult to remove later.

image

lib/sanbase/blockchain_address/label_change/blockchain_adderss_label_change.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Sanbase.BlockchainAddress.BlockchainAddressLabelChange do
+  import __MODULE__.SqlQuery
+  alias Repo
```
tspenov, line 3: leftover

lib/sanbase/clickhouse/exchanges/exchange_metric.ex
```diff
@@ -1,12 +1,38 @@
...
+  end
+
+  # Private functions
+
+  defp top_exchanges_by_balance_query(slug_or_slugs, limit, filters) do
+    args = [slug_or_slugs, limit]
+
+    {additional_filters_str, args} = additional_filters(filters, args, trailing_and: false)
+
+    IO.inspect({filters, additional_filters_str}, label: "SK:DJLASHDKJSLKDHAFS")
```
tspenov, line 35: inspect left

lib/sanbase/metric/sql_query_helper.ex
```diff
@@ -72,34 +72,73 @@ defmodule Sanbase.Metric.SqlQuery.Helper do
 
   # Add additional `=`/`in` filters to the query. This is mostly used with labeled
   # metrics where additional column filters must be applied.
-  def additional_filters([], _opts), do: []
+  def additional_filters([], args, _opts), do: {"", args}
+
+  def additional_filters(filters, args, opts) do
```
tspenov, line 77: I think this needs some unit tests cause it can break existing queries

tspenov (approved):
please, verify that `additional_filters` generates the correct sql, otherwise 👍

lib/sanbase/metric/sql_query_helper.ex
IvanIvanoff: We can't really test that as it requires you to execute the code against a CH database :(

PR #2915, opened 2021-08-16
Add finished oban jobs table

IvanIvanoff:
## Changes

Having millions of rows in the oban_jobs table causes performance issues. We need to keep some of the jobs in order to do uniqueness checks, so moving the jobs to another table that won't be used for anything else would improve the situation.

## Checklist:


priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs
```diff
@@ -0,0 +1,17 @@
...
+      add(:queue, :string)
+      add(:worker, :string)
+      add(:args, :map)
+      add(:inserted_at, :naive_datetime)
+      add(:completed_at, :naive_datetime)
+    end
+
+    create_if_not_exists(index(:finished_oban_jobs, [:queue]))
+    create_if_not_exists(index(:finished_oban_jobs, [:inserted_at]))
+    create_if_not_exists(index(:finished_oban_jobs, [:args], using: :gin, prefix: prefix))
```
tspenov, line 15: is `prefix` defined somewhere ?

priv/repo/structure.sql
```diff
@@ -178,6 +177,8 @@ $$;
 
 SET default_tablespace = '';
 
+SET default_table_access_method = heap;
```
tspenov, line 180: That needs to be removed because it breaks CI

priv/repo/structure.sql
```diff
@@ -6567,10 +6637,13 @@ INSERT INTO public."schema_migrations" (version) VALUES (20210616123403);
...
+INSERT INTO public."schema_migrations" (version) VALUES (20210708111602);
 INSERT INTO public."schema_migrations" (version) VALUES (20210712125345);
 INSERT INTO public."schema_migrations" (version) VALUES (20210716075649);
 INSERT INTO public."schema_migrations" (version) VALUES (20210721140912);
 INSERT INTO public."schema_migrations" (version) VALUES (20210722141655);
 INSERT INTO public."schema_migrations" (version) VALUES (20210727124520);
 INSERT INTO public."schema_migrations" (version) VALUES (20210727124524);
 INSERT INTO public."schema_migrations" (version) VALUES (20210728101938);
+INSERT INTO public."schema_migrations" (version) VALUES (20210803092012);
+INSERT INTO public."schema_migrations" (version) VALUES (20210816150914);
```
tspenov, line 6649: why 3 new schema versions are present ? I see one migration only

IvanIvanoff: These are some missing old migrations

priv/repo/structure.sql
IvanIvanoff: fixed

lib/sanbase/cryptocompare/jobs.ex
```diff
@@ -0,0 +1,51 @@
...
+      WHERE queue = '#{queue}' AND completed_at IS NOT NULL
+      LIMIT #{limit}
+    ),
+    moved_rows AS (
+      DELETE FROM oban_jobs a
+      WHERE a.id IN (SELECT id FROM finished_job_ids_cte)
+      RETURNING a.queue, a.worker, a.args, a.inserted_at, a.completed_at
+    )
+    INSERT INTO finished_oban_jobs(queue, worker, args, inserted_at, completed_at)
+    SELECT * FROM moved_rows;
```
tspenov, line 41: `count(*)` ?

lib/sanbase/cryptocompare/jobs.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  defp do_move_completed_jobs(queue, limit) do
+    # Instead of deleting the records directly from the oban_jobs table, define
+    # a CTE that selects the needed jobs first so we can put a `limit` on how
+    # many can be done at once.
+    # In a second CTE delete those records from the oban_jobs and return them,
+    # so they can be used to be inserted into the `finished_oban_jobs` table.
+    # Return the number of affected rows so when they become 0 we can
+    query = """
+    WITH finished_job_ids_cte AS (
```
tspenov, line 30: I think that should be in transaction - either with `begin .. end` or `Repo.transaction` - since we have delete and then insert and we want both or nothing

lib/sanbase/cryptocompare/jobs.ex
```diff
@@ -0,0 +1,51 @@
...
+  def move_finished_jobs(opts \\ []) do
+    iterations = Keyword.get(opts, :iterations, 200)
+    limit = Keyword.get(opts, :limit, 10_000)
+
+    count =
+      1..iterations
+      |> Enum.reduce_while(0, fn _, rows_count_acc ->
+        case do_move_completed_jobs("cryptocompare_historical_jobs_queue", limit) do
+          {:ok, 0} -> {:cont, rows_count_acc}
+          {:ok, rows_count} -> {:halt, rows_count + rows_count_acc}
```
tspenov, line 15: why we continue when there are no moved rows and halt when there are moved rows ? Shouldn't it be the opposite

lib/sanbase/cryptocompare/jobs.ex
IvanIvanoff: I tried doing that, but I could not succeed. If you add `begin; ... commit;` you get the error:
image

lib/sanbase/cryptocompare/jobs.ex
IvanIvanoff: I need to swap those 2, it's an error.

lib/sanbase/cryptocompare/jobs.ex
IvanIvanoff: The `select *` is used as input for `insert into`, so it cannot be changed. 
You are probably mistaking this with `RETURNING *`, which would actually return the actual data. The result when executing this is something like this (note the `rows: nil`):
```elixir
{:ok,
 %Postgrex.Result{
   columns: nil,
   command: :insert,
   connection_id: 90234,
   messages: [],
   num_rows: 5,
   rows: nil
 }}
 ```
 
 If you instead have `SELECT * FROM moved_rows RETURNING *` only then the result is:
 ```elixir
 {:ok,
 %Postgrex.Result{
   columns: ["id", "queue", "worker", "args", "inserted_at", "completed_at"],
   command: :insert,
   connection_id: 91600,
   messages: [],
   num_rows: 5,
   rows: [
     [
       16,
       "cryptocompare_historical_jobs_queue",
       "Sanbase.Cryptocompare.HistoricalWorker",
       %{"base_asset" => "2X3F", "date" => "2021-01-02", "quote_asset" => "USD"},
       ~N[2021-08-17 09:04:09.000000],
       ~N[2021-08-17 09:04:09.000000]
     ],
     [
       17,
       "cryptocompare_historical_jobs_queue",
       "Sanbase.Cryptocompare.HistoricalWorker",
       %{"base_asset" => "2X3F", "date" => "2021-01-03", "quote_asset" => "USD"},
       ~N[2021-08-17 09:04:09.000000],
       ~N[2021-08-17 09:04:09.000000]
     ],
     [
       18,
       "cryptocompare_historical_jobs_queue",
       "Sanbase.Cryptocompare.HistoricalWorker",
       %{"base_asset" => "2X3F", "date" => "2021-01-04", "quote_asset" => "USD"},
       ~N[2021-08-17 09:04:09.000000],
       ~N[2021-08-17 09:04:09.000000]
     ],
     [
       19,
       "cryptocompare_historical_jobs_queue",
       "Sanbase.Cryptocompare.HistoricalWorker",
       %{"base_asset" => "2X3F", "date" => "2021-01-05", "quote_asset" => "USD"},
       ~N[2021-08-17 09:04:09.000000],
       ~N[2021-08-17 09:04:09.000000]
     ],
     [
       20,
       "cryptocompare_historical_jobs_queue",
       "Sanbase.Cryptocompare.HistoricalWorker",
       %{"base_asset" => "2X3F", "date" => "2021-01-01", "quote_asset" => "USD"},
       ~N[2021-08-17 09:04:09.000000],
       ~N[2021-08-17 09:04:09.000000]
     ]
   ]
 }}
 ```

File: lib/sanbase/cryptocompare/jobs.ex (hunk 6 above)
Comment (IvanIvanoff) [reply lib/sanbase/cryptocompare/jobs.ex:15]: I'll rewrite the test so it uses 2 iterations with limit 3 and expect the result to be 6. This way it will test that this :cont is proper here.

File: priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs (hunk 1 above)
Comment (IvanIvanoff) [reply priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs:15]: Fixed

File: lib/sanbase/cryptocompare/jobs.ex (hunk 4 above)
Comment (tspenov) [reply lib/sanbase/cryptocompare/jobs.ex:41]: ups, my bad - I thought it is 2 statements insert and then select for some reason

File: lib/sanbase/cryptocompare/jobs.ex [hunk 7]
```diff
@@ -0,0 +1,51 @@
...
+    # Return the number of affected rows so when they become 0 we can
+    query = """
+    WITH finished_job_ids_cte AS (
+      SELECT id from oban_jobs
+      WHERE queue = '#{queue}' AND completed_at IS NOT NULL
+      LIMIT #{limit}
+    ),
+    moved_rows AS (
+      DELETE FROM oban_jobs a
+      WHERE a.id IN (SELECT id FROM finished_job_ids_cte)
```
tspenov, line 37: will it work if you leave only `WITH moved_rows AS` and put `finished_job_ids_cte` directly in it ?

priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs
```diff
@@ -0,0 +1,60 @@
...
+      add(:args, :map)
+      add(:inserted_at, :naive_datetime)
+      add(:completed_at, :naive_datetime)
+    end
+
+    prefix = get_prefix()
+    create_if_not_exists(index(:finished_oban_jobs, [:queue]))
+    create_if_not_exists(index(:finished_oban_jobs, [:inserted_at]))
+    create_if_not_exists(index(:finished_oban_jobs, [:args], using: :gin, prefix: prefix))
+
```
tspenov, line 18: why only this one needs `prefix` ?

tspenov (approved):

atanasormanow (approved):

lib/sanbase/cryptocompare/jobs.ex
IvanIvanoff: I've switched to using a postgres function, which would have the transactional behavior

priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs
IvanIvanoff: I think neither of those needs the prefix

lib/sanbase/cryptocompare/jobs.ex
IvanIvanoff: Reworked to use a function

PR #2916, opened 2021-08-17
Remove oban jobs for unsupported assets

IvanIvanoff:
## Changes

This task is not defined as a cron job as it should be executed once.
Log into the pod and start the task in a separate process (so it's not killed if you exit the pod)

## Checklist:

### DIFF
--- a/lib/sanbase/cryptocompare/jobs.ex
+++ b/lib/sanbase/cryptocompare/jobs.ex
@@ -1,8 +1,12 @@
 defmodule Sanbase.Cryptocompare.Jobs do
+  alias Sanbase.Model.Project
+
   # Execute the function until the moved rows are 0 or up to 100 iterations.
   # The iterations are needed to avoid an infinite loop. If there is a task that
   # finishes one job every second we risk to always return 1 row and never finish
   # the job.
+  @queue Sanbase.Cryptocompare.HistoricalScheduler.queue() |> to_string()
+
   def move_finished_jobs(opts \\ []) do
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20210816150914_add_finished_oban_jobs_table.exs
```diff
@@ -57,4 +57,11 @@ defmodule Sanbase.Repo.Migrations.AddFinishedObanJobsTable do
       _ -> "public"
     end
   end
+
+  defp get_prefix() do
```
tspenov, line 61: leftover

lib/sanbase/cryptocompare/jobs.ex
```diff
@@ -19,6 +23,32 @@ defmodule Sanbase.Cryptocompare.Jobs do
     {:ok, count}
   end
 
+  def remove_oban_jobs_unsupported_assets() do
```
tspenov, line 26: this one is not used except in tests

IvanIvanoff: I've made it public for that reason

IvanIvanoff: Check the PR description - as this is going to be executed once (but it will take a huge amount of time to run, maybe reruns will be required in case of timeouts) it is not defined as a cron job, but will need to be run manually.

tspenov: yep, missed the description

tspenov (approved):

PR #2923, opened 2021-08-27
remove trials on sign up, add trials with CC

tspenov:
## Changes

* Remove trials on sign up
* Add trails with CC
* Add API for trial eligibility for current user
* Create the liquidity sub on login async
* Send emails for trials with CC

```json
{ 
    currentUser {
        isEligibleForSanbaseTrial
    }
}
```

## Checklist:


lib/sanbase/accounts/user.ex
```diff
@@ -254,14 +260,11 @@ defmodule Sanbase.Accounts.User do
             %{field => value, salt: User.generate_salt(), first_login: true}
           )
 
-        %User{}
-        |> User.changeset(user_create_attrs)
-        |> Repo.insert()
-        |> maybe_create_liquidity_or_trial_subscription(attrs)
+        create(user_create_attrs)
+        |> maybe_create_liquidity_subscription_async(attrs)
```
IvanIvanoff, line 264: I think you better use the event bus for this. Its idea is exactly to decouple such side actions from the real action. Creating users should do only that - create users and possible emit a create user event.
BTW, we now emit the `register_user`, which I think is the better place to do these things. Imagine someone enters your email in the login form - it will pass through here and this code will be executed. But this does not mean that you did this. But if you do it on `register_user` (the first successful login) then it's ok - it's the first time the account is used.

lib/sanbase/accounts/user.ex
```diff
@@ -254,14 +260,11 @@ defmodule Sanbase.Accounts.User do
...
 
       user ->
-        maybe_create_free_subscription(user, attrs)
-        {:ok, user}
+        maybe_create_liquidity_subscription_async({:ok, user}, attrs)
```
IvanIvanoff, line 267: Why should you run this code in `find_or_insert_by` when this user already exists? Isn't this executed on every login?

lib/sanbase/billing/billing.ex
```diff
@@ -44,6 +44,18 @@ defmodule Sanbase.Billing do
...
+  def eligible_for_sanbase_trial?(user_id) do
+    sanbase_subscriptions =
+      Subscription
+      |> Subscription.Query.user_has_any_subscriptions_for_product?(
+        user_id,
+        Product.product_sanbase()
+      )
+      |> Repo.all()
+
+    sanbase_subscriptions == []
```
IvanIvanoff, line 56: |> Repo.all()
|> Enum.empty?()

lib/sanbase/billing/subscription.ex
```diff
@@ -124,12 +124,22 @@ defmodule Sanbase.Billing.Subscription do
   @spec subscribe(%User{}, %Plan{}, string_or_nil, string_or_nil) ::
           {:ok, %__MODULE__{}} | {:error, %Stripe.Error{} | String.t()}
   def subscribe(user, plan, card_token \\ nil, coupon \\ nil) do
-    with :ok <- active_subscriptions_for_this_plan(user, plan),
+    with :ok <- has_active_subscriptions_for_this_plan(user, plan),
```
IvanIvanoff, line 127: `has_active_subscriptions` name would be enough, I don't really like `for_this_plan` as the argument makes that clear.

lib/sanbase/billing/billing.ex
```diff
@@ -44,6 +44,18 @@ defmodule Sanbase.Billing do
...
   end
 
+  def eligible_for_sanbase_trial?(user_id) do
+    sanbase_subscriptions =
+      Subscription
+      |> Subscription.Query.user_has_any_subscriptions_for_product?(
+        user_id,
+        Product.product_sanbase()
+      )
+      |> Repo.all()
```
IvanIvanoff, line 54: I think the `Subscription` module should execute this and return you the result instead of you building the query and executing it here. This is part of the attempt to reduce the `Repo` calls on modules outside of said modules.

lib/sanbase/billing/subscription.ex
```diff
@@ -124,12 +124,22 @@ defmodule Sanbase.Billing.Subscription do
...
+
+      if plan.product_id == @product_sanbase do
+        SignUpTrial.maybe_remove_sign_up_trial(user)
+
+        eligible_for_sanbase_trial? &&
+          TrialEmail.create_and_send_welcome_email(%{
+            user_id: user.id,
+            subscription_id: db_subscription.id
+          })
+      end
```
IvanIvanoff, line 142: This should be a separate function `maybe_send_welcome_email` and put in the `with` pipeline

lib/sanbase/accounts/user.ex
tspenov: The idea is to check on every login. Maybe we should emit event on every login and then execute this.

lib/sanbase/billing/subscription.ex
```diff
@@ -356,20 +366,28 @@ defmodule Sanbase.Billing.Subscription do
...
+
+        _ ->
+          defaults
+      end
+
+    if Billing.eligible_for_sanbase_trial?(user.id) do
+      Map.put(defaults, :trial_end, DateTime.utc_now() |> DateTime.to_unix())
     end
+
+    defaults
```
IvanIvanoff, line 390: This `if` has no effect - you need to assign the result returned to it back to `defaults`.
To be even more clear I would do:
```elixir
case Billing.eligible_for_sanbase_trial?(user.id) do
    true ->  Map.put(defaults, :trial_end, DateTime.utc_now() |> DateTime.to_unix())
    false -> defaults
end

Comment (tspenov) [reply lib/sanbase/billing/subscription.ex]: niice catch!

File: lib/sanbase/accounts/user.ex (hunk 1 above)
Comment (tspenov) [reply lib/sanbase/accounts/user.ex]: rewritten with event bus

File: lib/sanbase/accounts/accounts_event_emitter.ex [hunk 8]
```diff
@@ -1,16 +1,26 @@
...
   use Sanbase.EventBus.EventEmitter
+  alias Sanbase.Billing
 
   @topic :user_events
   def topic(), do: @topic
 
   def handle_event({:error, _}, _event, _args), do: :ok
 
   def handle_event({:ok, user}, :register_user, %{login_origin: _} = args) do
+    Sanbase.Accounts.Email.schedule_emails_after_sign_up(user)
```
IvanIvanoff, line 11: You should assert that the event is handled so it can be retried if something fails:
```elixir
{:ok, _} = Sanbase.Accounts.Email.schedule_emails_after_sign_up(user)
```
or something similar that would fail could be used.

lib/sanbase/accounts/accounts_event_emitter.ex
```diff
@@ -1,16 +1,26 @@
...
+
     Map.merge(%{event_type: :register_user, user_id: user.id}, args)
     |> notify()
   end
 
+  def handle_event({:ok, user}, :login_user, %{login_origin: _} = args) do
+    Billing.maybe_create_liquidity_subscription_async(user.id)
```
IvanIvanoff, line 18: It would be better to handle this synchronously so it can be retried if it fails. This way you'll start an async task that, if fails, won't be retried and you lose some of the nicer benefits of the event bus.

lib/sanbase/accounts/accounts_event_emitter.ex
IvanIvanoff: Actually, you are using the event bus incorrectly. In the `event_emitter` you must not handle any event, only emit them. You handle the events in the subscribers.

lib/sanbase/accounts/accounts_event_emitter.ex
IvanIvanoff: Same comment as above - you cannot handle the events in the emitter, they must be handled in the subscriber modules.

lib/sanbase/accounts/email.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Sanbase.Accounts.Email do
```
IvanIvanoff, line 1: We already have `Sanbase.Accounts.User.Email` module. By looking at the names only, it is confusing which is doing what.
You can rename this to `EmailJobs` and let it handle only such background scheduled jobs.

lib/sanbase/accounts/user.ex
```diff
@@ -181,6 +181,12 @@ defmodule Sanbase.Accounts.User do
   defdelegate update_all_uniswap_san_staked_users(), to: __MODULE__.UniswapStaking
   defdelegate fetch_uniswap_san_staked_user(user), to: __MODULE__.UniswapStaking
 
+  def create(attrs) do
+    %__MODULE__{}
+    |> changeset(attrs)
+    |> Repo.insert()
+  end
```
IvanIvanoff, line 188: for completeness, you can emit an event here

lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -57,6 +59,8 @@ defmodule SanbaseWeb.AccountsController do
              client: device_data.client
            ),
          {:ok, _} <- User.mark_as_registered(user, %{login_origin: :twitter}) do
+      User.emit_event_on_login(user, %{login_origin: :google})
```
IvanIvanoff, line 62: Take a look at the rest of the event emitter use cases - all just call `emit_event` directly. I think it would be cleaner to just call `emit_event` instead of going through a helper function. This helper function does not make anything shorter. Compare:
```elixir
User.emit_event_on_login(user, %{login_origin: :email})
emit_event({:ok, user}, :login_event, %{login_origin: :email})
```

lib/sanbase/billing/subscription.ex
```diff
@@ -275,6 +275,12 @@ defmodule Sanbase.Billing.Subscription do
     |> Repo.all()
   end
 
+  def all_user_subscriptions_for_product(user_id, product_id) do
+    __MODULE__
+    |> __MODULE__.Query.user_has_any_subscriptions_for_product?(user_id, product_id)
+    |> Repo.all()
```
IvanIvanoff, line 281: When I see something ending with `?` I automatically assume it's a true/false.
Also I don't see this query reused anywhere - I think it would be better for this function to directly call `Repo.all` inside of it.

lib/sanbase/email/mailer.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"user_id" => user_id, "template" => template} = args}) do
+    user = User.by_id!(user_id)
+    vars = args["vars"] || %{}
+    opts = args["opts"] || %{}
+
+    Sanbase.MandrillApi.send(template, user.email, vars, opts)
+
+    :ok
```
IvanIvanoff, line 14: Do not ignore the result of the `send/4` function so in case of failure the job will be retried again.

IvanIvanoff: Successful jobs should return :ok or an {:ok, value} tuple. The `send/4` function returns either {:ok, value} or {:error, reason} which is perfect for here - in case of :ok it will finish successfully and in case of {:error, reason} the error will be stored in the `errors` jsonb[] field in the DB so we can see why it failed.

lib/sanbase/event_bus/user_events_subscriber.ex
```diff
@@ -26,6 +26,20 @@ defmodule Sanbase.EventBus.UserEventsSubscriber do
...
 
+  defp handle_event(%{data: %{event_type: :register_user, user_id: user_id}}, event_shadow, state) do
+    {:ok, _} = Sanbase.Accounts.EmailJobs.schedule_emails_after_sign_up(user_id)
+
+    EventBus.mark_as_completed({__MODULE__, event_shadow})
+    state
+  end
+
+  defp handle_event(%{data: %{event_type: :login_user, user_id: user_id}}, event_shadow, state) do
+    Sanbase.Billing.maybe_create_liquidity_subscription(user_id)
```
IvanIvanoff, line 37: Should we check here for :ok return before marking the event as completed?

lib/sanbase/web.ex
```diff
@@ -35,6 +35,7 @@ defmodule Sanbase.Application.Web do
       # Transform a list of transactions into a list of transactions
       # where addresses are marked whether or not they are an exchange address
       Sanbase.Clickhouse.MarkExchanges,
+      {Oban, oban_config()},
```
IvanIvanoff, line 38: nitpick: maybe we can rename these functions to be `oban_web_config` and `oban_scrapers_config`

lib/sanbase/event_bus/user_events_subscriber.ex
tspenov: I think it is fine here. it returns `{:ok, } | {:error, _} | false`

PR #2925, opened 2021-08-30
Add funding rates metric name aliases

spiderjako:
## Checklist:


lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -1400,7 +1400,7 @@
   },
   {
     "human_readable_name": "Binance Exchange Funding Rates (USDT)",
-    "name": "usdt_bnb_funding_rates",
```
IvanIvanoff, line 1403: The API is being used, if you rename it you'll break these API calls:
image


What you can do is the following:
Add an alias with the new name and switch all the clients to use it, only then drop the old name. This is how the age consumed/destroyed is exposed:
```json
{
  "human_readable_name": "Age Consumed",
  "name": "age_destroyed",
  "aliases": ["age_consumed"],
  ...
}
```

Removing APIs can/should be done only in a way that existing clients code won't break.

IvanIvanoff (approved):

PR #2929, opened 2021-09-01
Improve cache key generation. Avoid cache key collisions

IvanIvanoff:
## Changes

A few cache key collisions were found in metric.ex. A few different functions were all using the `{__MODULE__, :available_metrics_for_slug, selector}` key

## Checklist:

### DIFF
--- a/lib/sanbase/metric/metric.ex
+++ b/lib/sanbase/metric/metric.ex
@@ -489,7 +489,8 @@ defmodule Sanbase.Metric do
 
     parallel_fun = fn module ->
       cache_key =
-        {__MODULE__, :available_metrics_for_slug, module, selector} |> Sanbase.Cache.hash()
+        {__MODULE__, :available_metrics_for_slug_in_module, module, selector}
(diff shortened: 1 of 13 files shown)


lib/sanbase/metric/metric.ex
```diff
@@ -528,7 +529,7 @@ defmodule Sanbase.Metric do
   def available_histogram_metrics_for_slug(selector) do
     available_metrics =
       Sanbase.Cache.get_or_store(
-        {__MODULE__, :available_metrics_for_slug, selector} |> Sanbase.Cache.hash(),
+        {__MODULE__, :available_histogram_metrics_for_slug, selector} |> Sanbase.Cache.hash(),
         fn -> available_metrics_for_slug(selector) end
```
tspenov, line 533: shouldn't we also change that `available_metrics_for_slug(selector)` is called for all but they are saved with different cache key?

IvanIvanoff: fixed

PR #2935, opened 2021-09-03
export metrics csv

tspenov:
## Checklist:


lib/sanbase/metric_csv_exporter.ex
```diff
@@ -0,0 +1,326 @@
...
+
+    write_to_file(filename, data)
+  end
+
+  defp run(%{metrics: metrics, interval: interval} = args, from, to) do
+    for slug <- args[:slugs] do
+      lists =
+        for metric <- metrics do
+          {:ok, data} = Metric.timeseries_data(metric, %{slug: slug}, from, to, interval)
+          data
```
IvanIvanoff, line 264: This should be wrapped in a `get_timeseries_data` call that retries N times (probably with some exponential backoff*) so we don't fail the whole export because of some timeouts.

*the exponential backoff can be very simply implemented as `get_timeseries_data` having a last parameter `attempt` and use it like:
```elixir
if attempt > @max_attempts, do: raise("eror")
first_attempt = if attempt == 0, do: 0, else: 1
:math.pow(3, 2) + 15*first_attempt  + :rand.uniform(30) * attempt
```

tspenov: sounds good, will add it

PR #2946, opened 2021-09-13
Add rate limits to insights and comments creation

IvanIvanoff:
## Changes

Add limits to how many insights and comments (of any type) a user can create in a time window. 
This is to avoid spam by malicious parties.

## Checklist:

### DIFF
--- a/lib/sanbase/insight/post.ex
+++ b/lib/sanbase/insight/post.ex
@@ -17,6 +17,7 @@ defmodule Sanbase.Insight.Post do
   alias Sanbase.Chart.Configuration
 
   require Logger
+  require Sanbase.Utils.Config, as: Config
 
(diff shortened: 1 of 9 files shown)


lib/sanbase/insight/post.ex
```diff
@@ -94,6 +95,50 @@ defmodule Sanbase.Insight.Post do
     {:ok, map}
   end
 
+  def can_create?(user_id) do
```
tspenov, line 98: I think insights and comments (and maybe more places after) can use same function `can_create?` - the only difference is the config with rate limits and the entity in the error message.

IvanIvanoff: Done.

tspenov (approved):

PR #2971, opened 2021-09-28
Fix warnings

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/accounts/email_jobs.ex
+++ b/lib/sanbase/accounts/email_jobs.ex
@@ -72,7 +72,6 @@ defmodule Sanbase.Accounts.EmailJobs do
         :first_education_email -> days_after(4)
         :trial_suggestion -> days_after(6)
         :second_education_email -> days_after(7)
-        _ -> nil
       end
 
     Sanbase.Mailer.new(
--- a/lib/sanbase/mandrill_api.ex
+++ b/lib/sanbase/mandrill_api.ex
@@ -1,5 +1,6 @@
 defmodule Sanbase.MandrillApi do
   require Sanbase.Utils.Config, as: Config
+  require Application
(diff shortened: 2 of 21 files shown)


lib/sanbase/mandrill_api.ex
```diff
@@ -1,5 +1,6 @@
 defmodule Sanbase.MandrillApi do
   require Sanbase.Utils.Config, as: Config
+  require Application
```
tspenov, line 3: it should work without this require, right ?

lib/sanbase/accounts/email_jobs.ex
```diff
@@ -72,7 +72,6 @@ defmodule Sanbase.Accounts.EmailJobs do
         :first_education_email -> days_after(4)
         :trial_suggestion -> days_after(6)
         :second_education_email -> days_after(7)
-        _ -> nil
```
tspenov, line 75: Here the default is `nil` which means don't schedule for later but send immediately after insert. I don't see a reason for removing it but also it is not reached now, so I am ok both ways

tspenov (approved):

lib/sanbase/accounts/email_jobs.ex
IvanIvanoff: It's not used and emits warnings. But my bigger issue is that such catch all cases can bite us back in the future - one might forget to specify when an email must be scheduled or mistypes the name - both these would result in an error if we remove it.

lib/sanbase/mandrill_api.ex
IvanIvanoff: It emits warnings with the latest elixir version. I think it will stop working in some of the next versions

PR #2983, opened 2021-10-11
Improve the insights searching capabilities

IvanIvanoff:
## Changes

Introduce a prefix searching for the whole document instead of using
ILIKE only for the title. If the input is a single alphanumeric word,
treat it as an incomplete word and append ':*' to the end, so it is
searching for prefixes. For example 'mvr' would yield results very
similar to 'mvrv'.

## Checklist:

### DIFF
--- a/lib/sanbase/insight/search.ex
+++ b/lib/sanbase/insight/search.ex
@@ -7,6 +7,51 @@ defmodule Sanbase.Insight.Search do
   """
 
   def run(query, search_term) do
+    case is_incomplete_word?(search_term) do
+      true ->
+        # Add `:*` to the end of the word so we can search for prefixes
+        to_tsquery_incomplete_search(query, search_term)
+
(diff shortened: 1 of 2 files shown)


lib/sanbase/insight/search.ex
```diff
@@ -7,6 +7,51 @@ defmodule Sanbase.Insight.Search do
   """
 
   def run(query, search_term) do
+    case is_incomplete_word?(search_term) do
```
tspenov, line 10: maybe we should skip search when `search_term` is 1 or 2 or 2 chars ?

IvanIvanoff: I've added pagination, so 1 letter searches should not be that expensive.
Looking at other solutions like the one we use for https://academy.santiment.net/ - they let you search with 1 letter.

PR #2990, opened 2021-10-14
Add ftx funding rates

spiderjako:
## Changes

This PR is for adding the FTX funding rates to the intraday metrics.

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
+++ b/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
@@ -1433,5 +1433,22 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "FTX Exchange Funding Rates",
+    "name": "ftx_perpetual_funding_rate",
+    "metric": "ftx_funding_rates",
+    "version": "2019-01-01",
+    "access": "restricted",
+    "selectors": ["slug"],
+    "min_plan": {
+      "SANAPI": "free",
+      "SANBASE": "free"
+    },
+    "aggregation": "avg",
+    "min_interval": "1h",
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -1433,5 +1433,23 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "FTX Exchange Funding Rates",
+    "name": "ftx_funding_rate",
+    "metric": "ftx_funding_rates",
+    "aliases": ["ftx_perpetual_funding_rate"],
```
IvanIvanoff, line 1441: Aliases are to be used only when an existing metric name needs to change. Adding one metric with 2 names from the start is just confusing.

lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json
```diff
@@ -1402,7 +1402,7 @@
     "human_readable_name": "Binance Exchange Funding Rates (USDT)",
     "name": "usdt_bnb_funding_rates",
     "metric": "usdt_bnb_funding_rates",
-    "aliases": ["usdt_binance_funding_rate"],
+    "aliases": ["usdt_binance_funding_rate", "usdt_binance_perpetual_funding_rate"],
```
IvanIvanoff, line 1405: Why do we need so many names for 1 metric? This would mean that the metric can be accessed with 3 different names, why?

spiderjako: I've made a mistake when adding the alias. It should've been the second name (the one just added). Is removing some of the names more viable than adding a new one?

PR #3005, opened 2021-11-02
Split ContextPlug into 3 separate plugs

IvanIvanoff:
## Changes

The ContextPlug no longer just builds the context but is overloaded with functionality. It is now split into 3 plugs, running in this order:
- AuthPlug - authenticate the request and put the result in the `conn.private[:san_authentication]` field
- ContextPlug - Get the data from the authentication and build the absinthe context (add additionally remote_ip, origin host parts, querying device platform (from the user agent), jwt tokens, etc.)
- RequestHaltPlug - decide whether or not to block the request from further processing based on:
  - Limited access - cannot make sansheets requests
  - Api call limit reached.
- The logic of the plug functions is not changed. A few small rewrites to use `with` instead of nested `case`s were done.
- As a side task, fix a some labels tests and labels handling as I got random test fails because of it.
 

## Checklist:


lib/sanbase/clickhouse/labels.ex
```diff
@@ -101,8 +103,12 @@ defmodule Sanbase.Clickhouse.Label do
...
     add_labels = fn
       # In this case the address type does not exist, so the result is not used
-      nil -> nil
-      map -> Map.put(map, :labels, Map.get(labels_map, map.address, []))
+      nil ->
+        nil
+
+      map ->
+        labels = Map.get(labels_map, map.address, []) |> Enum.sort_by(& &1.name)
+        Map.put(map, :labels, labels)
```
IvanIvanoff, line 111: This fixes some tests that were failing at random and making it hard to test my changes.

PR #3007, opened 2021-11-03
Automate free basic api sub for pro plus

tspenov:
## Checklist:


lib/sanbase/billing/subscription/pro_plus.ex
```diff
@@ -0,0 +1,85 @@
...
+    end)
+  end
+
+  def delete_free_basic_api do
+    all_free_basic_api_subs()
+    |> Enum.each(fn sub ->
+      if not user_has_pro_plus?(sub.user_id) do
+        StripeApi.delete_subscription(sub.stripe_id)
+      end
+    end)
```
IvanIvanoff, line 25: This will make a huge number of DB calls - every `user_has_pro_plus?` needs to call the DB.
Alternatively, you can drop the `user_id` check and fetch all needed data once outside the Enum call and just check against that result

IvanIvanoff: This logic can be applied to the other functions as well.

tspenov: yes, can optimize db calls and get only the intersection. The slow part here are the API calls for checking the coupon.

PR #3013, opened 2021-11-05
Fix 500 remove watchlist twice

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
+++ b/lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
@@ -272,7 +272,8 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
   end
 
   def update_watchlist(_root, %{id: id} = args, %{context: %{auth: %{current_user: current_user}}}) do
-    if has_permissions?(id, current_user) do
+    with %UserList{} = watchlist <- fetch_watchlist_by_id(id),
+         true <- has_permissions?(watchlist, current_user, "update") do
       case UserList.update_user_list(current_user, args) do
         {:ok, user_list} ->
           {:ok, user_list}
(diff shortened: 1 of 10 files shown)


lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex
```diff
@@ -342,26 +341,42 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
 
   # Private functions
 
-  defp has_permissions?(id, %User{id: user_id}) do
-    UserList.by_id(id).user_id == user_id
+  defp has_permissions?(watchlist, %User{id: user_id}, action) do
+    case watchlist.user_id == user_id do
+      true -> true
+      false -> {:error, "Cannot #{action} watchlist belonging to another user"}
+    end
```
IvanIvanoff, line 360: I think you can rework this to look like this, so you can support permissions on public watchlists, if needed (read operation for example):
```elixir
case watchlist do
  %UserList{user_id: ^user_id} -> true
  %UserList{is_public: true} when action in [:action_that_can_work_on_public_watchlist] -> true
  _ -> {:error, "Cannot #{action} watchlist belonging to another user"}
end
  ```

File: lib/sanbase_web/graphql/resolvers/user/user_list_resolver.ex [hunk 2]
```diff
@@ -342,26 +341,42 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserListResolver do
...
+  end
+
+  defp fetch_watchlist_by_id(watchlist_id) do
+    case UserList.by_id(watchlist_id) do
+      nil -> {:error, "Watchlist does not exist"}
+      watchlist -> watchlist
+    end
+  end
```
IvanIvanoff, line 368: I think it would be better to rework `by_id` to work the same across our codebase - return :ok or :error tuple and introduce by_id! which raises. Looking at the code, doing this would require only minor changes in some places.

PR #3020, opened 2021-11-09
Sharable chart layout

IvanIvanoff:
## Changes

The metrics and queries on a chart layout can be accessed by anonymous/non-pro users by including a special token in the request. The token can be obtained by the owner of the layout by executing the following graphql query:
```graphql
{
  getChartConfigurationSharedAccessToken(chartConfigurationId: 5){
      uuid
  }
}
```
> Note: Executing the request multiple times will result in returning the same token.

The obtained uuid must be provided back to the backend via the HTTP header:
```
X-SharedAccess-Authorization: SharedAccessToken <uuid>
```

Requests that contain this header will be granted access to the metrics and queries in the chart layout.

Example: You have a chart layout with `mvrv_usd` metric. Something interesting happened in the last 1 week. Anonymous users have the last 30 days cut-off, but you want to share this with your twitter audience. 
Option 1: Make a screenshot. This will capture the current state of things but you don't get updates if you visit the screenshot after 2 days.
Option 2: Generate a sharable link. For end users the mechanism will look just like clicking on a button. This will create a similar link to the short link but with a different behaviour - the viewers who use this link will have access to the restricted data as if they have sanbase pro account.

Notes for frontend devs: You can obtain the token and store it in the `data` field of the short url:
```graphql
{
  getFullUrl(shortUrl: "<short url>"){
     fullUrl
     data
   }
}
```

Then, when you use the short url you will have benefits from both:
- the short url will lead to a chart with loaded metrics
- the token stored in the `data` will be used and provided as header so you'll gain access to the data.

Anonymous user sees this **without** the token:
image

Anonymous user sees this **with** the token:
image

## Checklist:

### DIFF
--- a/lib/sanbase/chart/shared_access_token.ex
+++ b/lib/sanbase/chart/shared_access_token.ex
@@ -145,12 +145,23 @@ defmodule Sanbase.Chart.Configuration.SharedAccessToken do
 
   defp list_to_individual_entries(nil, _project), do: []
 
+  # Returns a list of %{metric: _, slug: _} maps where every element shows
+  # which metric and asset is shown on the chart.
   defp list_to_individual_entries(list, %Sanbase.Model.Project{slug: project_slug}) do
+    # The strings we split by are how the frontend combines multiple metrics
+    # or metric of another asset (not the chart configuration's one) into a
+    # single string.
     list
     |> Enum.flat_map(fn entry ->
+      # Multiple metrics can be separated by __MM__. This is used for example
+      # for combining some holders distributions metrics into a single one
+      # on sanbase's frontend.
       String.split(entry, "__MM__")
     end)
     |> Enum.map(fn entry ->
+      # The frontend uses the format `slug-CC-ticker-CC-metric` or
+      # `slug_MC_ticker_MC_metric` to represent that a metric of another
+      # asset should be displayed on the chart configuration.
       case String.split(entry, ["-CC-", "_MC_"]) do
         [slug, _ticker, metric] -> %{metric: Inflex.underscore(metric), slug: slug}
         [metric] -> %{metric: Inflex.underscore(metric), slug: project_slug}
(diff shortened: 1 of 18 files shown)


lib/sanbase/chart/shared_access_token.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  defp list_to_individual_entries(nil, _project), do: []
+
+  defp list_to_individual_entries(list, %Sanbase.Model.Project{slug: project_slug}) do
+    list
+    |> Enum.flat_map(fn entry ->
+      String.split(entry, "__MM__")
+    end)
+    |> Enum.map(fn entry ->
+      case String.split(entry, ["-CC-", "_MC_"]) do
```
tspenov, line 154: what are these `-CC-` , `_MC_` and `__MM__` ?

IvanIvanoff: This is how the frontend combines metrics into single string. I'll add description

tspenov: but why we deal with this on backend? I think it is better if frontend send us this parsed in json field

IvanIvanoff: This is currently used behavior across our chats and we need to support it. Take for example the featured chart layouts: https://tinyurl.com/ycnhpnuf. As of writing this we have these in the metrics for example: 
image

If we want the frontend to provide the json we need a big rework that includes:
- adding support for this on the backend (rework our DB schema)
- add support for old and new behavior to the frontend 
- run backend migrations that will transform the old format to the new one. This migration will in fact include exactly the same code that is written here.

Overall, it is a big refactor that should happen separately. The good thing is that the shared chart layout does only hold the chart layout ID, so once we refactor the code, the existing tokens will continue to work just fine.

PR #3027, opened 2021-11-16
Add broken_data field to the get_metric API

IvanIvanoff:
## Changes

Add `brokenData` field that will return metadata for periods of time where the data is broken. Currently we have only 1 such case we cannot easily fix, so for now that data is stored in a JSON file for the speed of development and performance.

```graphql
{
  getMetric(metric: "dev_activity") {
    brokenData(selector: {slug: "bitcoin"}, from: "utc_now-70d", to: "utc_now") {
      from
      to
      why
      why
      notes
      actionsToFix
    }
  }
}
```
```json
{
  "data": {
    "getMetric": {
      "brokenData": [
        {
          "actionsToFix": "The github events are not possible to be refetched so the gap cannot be filled.",
          "from": "2021-10-07T00:00:00Z",
          "notes": "Due to missing github data the development activity related metrics are lower than they should be.",
          "to": "2021-11-01T00:00:00Z",
          "why": "Third-party data provider outage."
        }
      ]
    }
  }
}
```

## Checklist:


lib/sanbase/metric/helper.ex
```diff
@@ -71,6 +72,9 @@ defmodule Sanbase.Metric.Helper do
 
     if function_exported?(module, :deprecated_metrics_map, 0),
       do: @deprecated_metrics_acc(module.deprecated_metrics_map)
+
+    # if function_exported?(module, :broken_data, 4),
```
tspenov, line 76: commented code left

lib/sanbase/clickhouse/github/broken_data.json
```diff
@@ -0,0 +1,10 @@
+[
+    {
+        "from": "2021-10-07T00:00:00Z",
+        "to": "2021-11-01T00:00:00Z",
+        "what": "Github events are partially or fully missing in the specified range.",
+        "why": "Third-party data provider outage.",
+        "notes": "Due to missing github data the development activity related metrics are lower than they should be.",
+        "actions_to_fix": "The github events are not possible to be refetched so the gap cannot be filled."
+    }
+]
```
tspenov, line 10: no newline at the end of file

lib/sanbase/clickhouse/github/broken_data.ex
```diff
@@ -0,0 +1,27 @@
...
+  @broken_data File.read!(json_file)
+               |> Jason.decode!()
+               |> Sanbase.MapUtils.atomize_keys()
+               |> Enum.map(fn elem ->
+                 elem
+                 |> Map.put(:from, Sanbase.DateTimeUtils.from_iso8601!(elem[:from]))
+                 |> Map.put(:to, Sanbase.DateTimeUtils.from_iso8601!(elem[:to]))
+               end)
+
+  def get(_metric, _slug, from, to) do
```
tspenov, line 14: maybe also one function without from and to, returning all gaps ?

lib/sanbase/metric/broken_data/broken_data.json
```diff
@@ -0,0 +1,3 @@
+[
+    {}
+]
```
tspenov, line 3: no newline

lib/sanbase/clickhouse/github/broken_data.ex
tspenov: metric and slug are unused here. What is the convention in json file?
I read it like if we don't have specified metric and slug in the json file - the broken data is for every metric and every slug. If we specify - it is concretely for this metric and for this slug.

IvanIvanoff: There is metric and slug as this function is added to all metric adapters. Once we need to use metric/slug it would be harder to add them and change all existing behavior rather than having them from the start.

tspenov (approved):

PR #3033, opened 2021-11-18
Switch from parity to erigon eth node

tspenov:
## Changes

* Seems that get transaction by hash now wants second param. With Parity it was `latest` by default or this code didn't work
* filters does not work with erigon, but they are substituted with `get_ethLogs`

## Checklist:


lib/sanbase/wallet_hunters/contract.ex
```diff
@@ -51,37 +51,13 @@ defmodule Sanbase.WalletHunters.Contract do
   def get_trx_receipt_by_id(trx_id) do
     maybe_replace_rinkeby(
       fn ->
-        Logger.info("[Parity] Get eth transaction receipt via Ethereumex client.")
+        Logger.info("[EthNode] Get eth transaction receipt via Ethereumex client.")
         Ethereumex.HttpClient.eth_get_transaction_receipt(trx_id)
```
IvanIvanoff, line 55: can this now be done without ethereumex?

tspenov: yes, but it was also possible before

PR #3035, opened 2021-11-18
Download invoices

tspenov:
## Checklist:


lib/sanbase/billing/invoices/download.ex
```diff
@@ -0,0 +1,94 @@
+defmodule Sanbase.Billing.Invoices.Download do
+  def run() do
+    for year <- [2020, 2021] do
+      for month <- 1..12 do
+        if year == 2021 and month > 11 do
```
IvanIvanoff, line 5: Why skip? You might want to write it with `Enum.reduce_while` and get the months until the current month with something like:
```elixir
now = DateTime.utc_now()
for year <- [2020..now.year], month <- 1..12 do
  if year == now.year and month > now.month do
    :ok
  else
    get({month, year})
  end
end
```

PR #3041, opened 2021-11-24
Add labels api

tspenov:
## Checklist:

```graphql
    mutation {
      addBlockchainAddressLabels(
        selector: {address: "0x1", infrastructure: "ETH"}
        labels: ['L1', 'L2']
      )
    }
```


lib/sanbase/labels/api.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Sanbase.Labels.Api do
```
IvanIvanoff, line 1: This should be under the `Sanbase.Clickhouse.Label` module I guess

PR #3042, opened 2021-11-24
Add name to the user GQL type and DB table

IvanIvanoff:
## Changes

- Add `name` to the user type
- Add `changeName` mutation that changes the current user `name`. The mutation returns the `user` type.

```graphql
mutation {
  updateName(name: "My New Name"){
      id
      name
      username
      email
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/accounts/user.ex
+++ b/lib/sanbase/accounts/user.ex
@@ -47,6 +47,7 @@ defmodule Sanbase.Accounts.User do
   schema "users" do
     field(:email, :string)
     field(:email_candidate, :string)
+    field(:name, :string)
     field(:username, :string)
     field(:salt, :string)
     field(:san_balance, :decimal)
@@ -141,11 +142,13 @@ defmodule Sanbase.Accounts.User do
       :stripe_customer_id,
       :test_san_balance,
       :twitter_id,
-      :username
+      :username,
+      :name
     ])
     |> normalize_user_identificator(:username, attrs[:username])
     |> normalize_user_identificator(:email, attrs[:email])
     |> normalize_user_identificator(:email_candidate, attrs[:email_candidate])
+    |> validate_change(:name, &validate_name_change/2)
     |> validate_change(:username, &validate_username_change/2)
     |> validate_change(:email_candidate, &validate_email_candidate_change/2)
     |> validate_change(:avatar_url, &validate_url_change/2)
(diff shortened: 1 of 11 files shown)


lib/sanbase/accounts/user.ex
```diff
@@ -294,11 +297,19 @@ defmodule Sanbase.Accounts.User do
     |> String.trim()
   end
 
+  defp validate_name_change(_, name) do
```
tspenov, line 300: I think username is fine to restrict to only ascii but name should be without such restrictions.

tspenov: And validation function can be reused - first param is the field as atom so we can use in return keyword list.

lib/sanbase_web/graphql/resolvers/user/user_resolver.ex
```diff
@@ -144,11 +144,22 @@ defmodule SanbaseWeb.Graphql.Resolvers.UserResolver do
     end
   end
 
-  def change_username(_root, %{username: new_username}, %{
-        context: %{auth: %{auth_method: :user_token, current_user: user}}
-      }) do
-    User.change_username(user, new_username)
-    |> case do
+  def change_name(_root, %{name: new_name}, %{context: %{auth: %{current_user: user}}}) do
```
tspenov, line 147: We can have single endpoint for updating user data and update based on args ?

IvanIvanoff: We can, but this would be part of a major rework, mostly on the FE side. It's out of the scope of this PR

lib/sanbase/accounts/user.ex
IvanIvanoff: > I think username is fine to restrict to only ascii but name should be without such restrictions.
I actually think even more restrictions are actually needed, ASCII is too permissive - it allows quotes, parentheses, etc. so people are tempted to try XSS - you can find such cases in our DB.

IvanIvanoff: > And validation function can be reused - first param is the field as atom so we can use in return keyword list.

I tried to do that aggressively, but the result code is kinda ugly. It would look like:
```elixir
    |> validate_change(:username, &validate_ascii_or_nil_field(&1, &2, "Username")
```

I've extracted the core checks into separate functions, that is called by the other functions.

tspenov: why not just: `|> validate_change(:username, &validate_ascii/2)` and then

```elixir
defp validate_ascii(field, value) do
...
[{field, "#{field|> Atom.to_string |> String.capitalize} ..." }]
```

tspenov (approved):
There is one relevant test fail, though

PR #3044, opened 2021-11-25
Add not implemented error for timeseries data per slug

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/github/metric_adapter.ex
+++ b/lib/sanbase/clickhouse/github/metric_adapter.ex
@@ -83,6 +83,13 @@ defmodule Sanbase.Clickhouse.Github.MetricAdapter do
   end
 
   @impl Sanbase.Metric.Behaviour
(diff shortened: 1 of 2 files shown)


spiderjako (commented):
In general this is okay, but couldn't there be a more generalised way of noting which function is not in CH and is not implemented for this? Also, social metrics for random words are also not implemented.

IvanIvanoff (approved):

tspenov:
> In general this is okay, but couldn't there be a more generalised way of noting which function is not in CH and is not implemented for this? Also, social metrics for random words are also not implemented.

It is good point but to be honest I am not aware of good way to do this. One way is to define the default implementation in the `__using__` macro and use `use Sanbase.Metric.Behavour` instead of `@behaviour Sanbase.Metric.Behavour`. But that comes at other costs.

spiderjako (approved):

PR #3105, opened 2022-01-20
Add freezing to alerts

IvanIvanoff:
## Changes

Replace the existing restrictions of 10 alerts per user for free
accounts with the freezing mechanism. Users can now have an unlimited
number of alerts but they will get frozen after a number of predefined days.
Frozen alerts cannot run and are locked for updates. They can only be
deleted. Frozen alerts get unfrozen when the user updates to the Sanbase PRO
plan.

To see if an alert is frozen, check the the `isFrozen` field
```graphql
{
  getTriggerById(id: 5) {
    trigger {
      isFrozen
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/alerts/alert_job.ex
+++ b/lib/sanbase/alerts/alert_job.ex
@@ -14,23 +14,27 @@ defmodule Sanbase.Alert.Job do
       |> Sanbase.Repo.all()
 
     alerts
-    |> Enum.map(fn alert ->
-      UserTrigger.update_changeset(alert, %{trigger: %{is_frozen: true}})
-    end)
-    |> Enum.with_index()
-    |> Enum.reduce(
-      Ecto.Multi.new(),
-      fn {changeset, offset}, multi ->
-        Ecto.Multi.update(multi, offset, changeset)
-      end
-    )
-    |> Sanbase.Repo.transaction()
-    |> case do
-      {:ok, _} ->
-        :ok
+    |> Enum.chunk_every(300)
+    |> Enum.each(fn alerts_chunk ->
+      multi_update_result =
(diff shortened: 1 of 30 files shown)


lib/sanbase/alerts/alert_job.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Sanbase.Alert.Job do
+  @days 0
```
tspenov, line 2: What we decided 30 days, right ?

tspenov:
We should expose `is_frozen` to graphql API in order frontend to show these differently :?

lib/sanbase/alerts/alert_job.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  import Ecto.Query
+  import Sanbase.Alert.TriggerQuery, only: [trigger_is_not_frozen: 0]
+
+  alias Sanbase.Alert.UserTrigger
+
+  def freeze_alerts() do
+    alerts =
+      from(ut in UserTrigger,
+        where: ut.inserted_at >= ago(@days, "day") and trigger_is_not_frozen()
```
tspenov, line 12: we should not freeze the alerts of pro users, right ?

lib/sanbase/alerts/alert_job.ex
IvanIvanoff: this is leftover of testing

lib/sanbase/alerts/alert_job.ex
IvanIvanoff: It is now done. Note that the check is done after the fetching of the alerts.

PR #3109, opened 2022-01-25
Add realtime price validator

IvanIvanoff:
## Changes

Run a swarm of GenServers that will keep in memory the last N (currently 6) prices of a given slug/currency pair (SAN/USD, SAN/BTC, etc.). Run more than 1 for two main reasons:
- increased throughput
- easier handling of restarts - only part of the slugs need to be re-initialized.

Initialize every GenServer with the latest prices found in the Clickhouse database. Every GenServer handles those assets, whose hash falls in the interval 0..N-1

Every price validation returns `true` or `{:error, reason}`. In case the price is validated correctly, it replaces the oldest price in the list of prices.

The

## Checklist:

### DIFF
--- a/lib/sanbase/prices/validator/validator_node.ex
+++ b/lib/sanbase/prices/validator/validator_node.ex
@@ -1,0 +1,98 @@
+defmodule Sanbase.Price.Validator.Node do
+  use GenServer
+  @max_prices 6
+
+  alias Sanbase.Model.Project
+  alias Sanbase.Price
+
+  def child_spec(opts) do
+    name = Keyword.fetch!(opts, :name)
+
+    %{
+      id: name,
+      start: {__MODULE__, :start_link, [opts]}
+    }
+  end
+
+  def start_link(opts) do
+    name = Keyword.fetch!(opts, :name)
+
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  def init(opts) do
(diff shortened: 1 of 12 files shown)


lib/sanbase/prices/validator/validator_node.ex
```diff
@@ -0,0 +1,108 @@
...
+  # If there are less than @max_prices prices, return as is
+  # If there are @max_prices in the list, drop the oldest price as we'll add the
+  # newest price to the list. This is executed only if the new price that is
+  # validated is accepted
+  defp maybe_drop_oldest_price(prices) when length(prices) < @max_prices, do: prices
+  defp maybe_drop_oldest_price([_ | prices]), do: prices
+
+  defp price_outlier?(_slug, _quote_asset, _price, [], _opts), do: false
+
+  defp price_outlier?(slug, quote_asset, price, prices, opts) do
```
tzanko-matev, line 83: Please add documentation

tzanko-matev (commented):
What do we do if the price is not valid? It's not clear from the code to me

tspenov:
Maybe as a first step before removing outlier prices we run it only in debug mode for some time with different coefficients and only log the results ?

lib/sanbase/prices/validator/validator_node.ex
IvanIvanoff: Added

PR #3113, opened 2022-01-27
Add contract metrics

IvanIvanoff:
## Changes

Add 2 metrics that are working with raw contract addresses that might not be associated with any slug on Santiment. It works on the raw data tables, so it works with every contract address that exists.

These new APIs work with the following selector: `selector: {contractAddress: "0x00000000219ab540356cBB839Cbe05303d7705Fa"}`. This reuses the `contractAddress` field that has been used before. This does not change the old behavior - if another metric like `daily_active_addresses` gets a `contractAddress` argument it will be resolved to a `slug` and will use it. This is done by extending the selector with the slug of that contract address if it exists, and then it's up to the metric to choose which argument to look for an use.

```graphql

{
  getMetric(metric:"contract_transactions_count") {
    timeseriesData(
      selector: {contractAddress: "0x00000000219ab540356cBB839Cbe05303d7705Fa"},
      from:"utc_now-60d",
      to:"utc_now-50d",
      interval: "1d",
    ) {
      value
      datetime
    }
  }
}
```

```graphql

{
  getMetric(metric:"contract_interacting_addresses_count") {
    timeseriesData(
      selector: {contractAddress: "0x00000000219ab540356cBB839Cbe05303d7705Fa"},
      from:"utc_now-60d",
      to:"utc_now-50d",
      interval: "1d",
    ) {
      value
      datetime
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/prices/validator/validator_node.ex
+++ b/lib/sanbase/prices/validator/validator_node.ex
@@ -1,8 +1,8 @@
 defmodule Sanbase.Price.Validator.Node do
   @moduledoc ~s"""
-  Check if a new realtime price is valid or is and outlier.any()
-
+  Check if a new realtime price is valid or is and outlier.
   """
+
   use GenServer
   @max_prices 6
 
--- a/lib/sanbase/clickhouse/uniswap/metric_adapter.ex
+++ b/lib/sanbase/clickhouse/uniswap/metric_adapter.ex
@@ -1,5 +1,6 @@
 defmodule Sanbase.Clickhouse.Uniswap.MetricAdapter do
   @behaviour Sanbase.Metric.Behaviour
+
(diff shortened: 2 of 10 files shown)


lib/sanbase/prices/validator/validator_node.ex
```diff
@@ -1,8 +1,8 @@
 defmodule Sanbase.Price.Validator.Node do
   @moduledoc ~s"""
-  Check if a new realtime price is valid or is and outlier.any()
-
+  Check if a new realtime price is valid or is and outlier.
```
tspenov, line 3:
```suggestion
  Check if a new realtime price is valid or is an outlier.
```

tspenov (approved):

PR #3118, opened 2022-02-01
Implement opts filter/order slugs in Price module

IvanIvanoff:
## Changes

Add `opts` to 2 functions in the Price module. This is needed otherwise fields that are not available across all sources (like marketcap_usd is not available in cryptocomapre data) will cause issues when we mix the sources.

## Checklist:

### DIFF
--- a/lib/sanbase/prices/price_sql_query.ex
+++ b/lib/sanbase/prices/price_sql_query.ex
@@ -214,8 +214,8 @@ defmodule Sanbase.Price.SqlQuery do
     {query, args}
   end
 
-  def slugs_by_filter_query(metric, from, to, operation, threshold, aggregation) do
-    {query, args} = filter_order_base_query(metric, from, to, aggregation)
+  def slugs_by_filter_query(metric, from, to, operation, threshold, aggregation, source) do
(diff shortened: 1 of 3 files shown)


lib/sanbase/prices/price_sql_query.ex
```diff
@@ -238,7 +238,7 @@ defmodule Sanbase.Price.SqlQuery do
     {query, args}
   end
 
-  defp filter_order_base_query(metric, from, to, aggregation) do
+  defp filter_order_base_query(metric, from, to, aggregation, source) do
```
tspenov, line 241: I think you can pass `opts` (currently with aggregation and source) to the base query. That way if we need to add another option we can just change the base query without touching all other functions.

IvanIvanoff: The approach in the whole module is:
```elixir
with {:ok, source} <- opts_to_source(opts) do
   # code goes here
end
```
so none of the other `*_query` functions accepts `opts`. Furthermore, the SQL query modules inside themselves sometimes build an opts keyword list that passes things like argument position.

tspenov (approved):

PR #3119, opened 2022-02-01
Allow update username before privacy settings accepted

tspenov:
## Checklist:


IvanIvanoff (approved):

lib/sanbase_web/graphql/schema/queries/user_queries.ex
```diff
@@ -138,7 +138,7 @@ defmodule SanbaseWeb.Graphql.Schema.UserQueries do
     field :change_username, :user do
       arg(:username, non_null(:string))
 
-      middleware(JWTAuth)
+      middleware(JWTAuth, allow_access: true)
```
IvanIvanoff, line 141: we might also rename this param to have a meaningful name - `allow_access_without_terms_accepted: true` maybe?

PR #3120, opened 2022-02-01
Add metrics_json to the chart configuration type

IvanIvanoff:
## Changes

This field will replace the metrics field, which is just an array of
string and this way the FE is forced to put all the information about
the metric in a single string - the metric name, the slug name, etc.

The `chartConfiguration` GQL type now has an extra field `metricsJson` which accepts JSON maps encoded as a string.

## Checklist:


priv/repo/migrations/20220201122953_add_metrics_json_field_chart_configuration.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.AddMetricsJsonFieldChartConfiguration do
+  use Ecto.Migration
+
+  def change do
+    alter table("chart_configurations") do
+      add(:metrics_json, :jsonb)
```
tspenov, line 6: maybe default value in db ?

tspenov (approved):

priv/repo/migrations/20220201122953_add_metrics_json_field_chart_configuration.exs
IvanIvanoff: Done.

PR #3147, opened 2022-02-22
Send trial will end reminder

tspenov:
## Changes

Send reminder email that trial will end in a few days.

## Checklist:

### DIFF
--- a/lib/sanbase/accounts/user.ex
+++ b/lib/sanbase/accounts/user.ex
@@ -97,6 +97,10 @@ defmodule Sanbase.Accounts.User do
     timestamps()
   end
 
+  def get_name(%__MODULE__{} = user) do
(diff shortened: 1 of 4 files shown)


lib/sanbase/accounts/user.ex
```diff
@@ -97,6 +97,10 @@ defmodule Sanbase.Accounts.User do
     timestamps()
   end
 
+  def get_name(%__MODULE__{} = user) do
+    user.name || user.username || user.email || "Anon"
+  end
```
IvanIvanoff, line 102: Judging by the name I expected to be just `user.name`. As we have functions like `change_name` this can be confused with it. Should be renamed to something like `get_human_readable_name` :?

PR #3155, opened 2022-03-04
Mask the apikey in the errors

IvanIvanoff:
## Changes
This change is **NOT** due to any security reasons. If an apikey is not
valid and is then shown in an error message this does not pose any risk
as this apikey is no longer valid and thus cannot be used.
This change addresses issues that at first glance look like a security
risk (but are not). In case a no longer valid apikey is used in a google
sheet, then people who open that sheet might see the invalid apikey.

## Checklist:


spiderjako (approved):
lgtm

tspenov (approved):

PR #3160, opened 2022-03-08
Allow conditional start of the kafka consumer

IvanIvanoff:
## Changes

Check the `KAFKA_CONSUMER_ENABLED` env var when deciding if the kafka consumer should be started. This is useful in dev environment - it can be disabled when working on other parts of the system. To disable the kafka consumer add `KAFKA_CONSUMER_ENABLED=false` to your `.env.dev` file

## Checklist:

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -244,6 +244,7 @@ config :kaffy,
   router: SanbaseWeb.Router
 
 config :sanbase, Sanbase.Kafka.Consumer,
+  enabled?: {:system, "KAFKA_CONSUMER_ENABLED", true},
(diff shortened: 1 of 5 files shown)


config/config.exs
```diff
@@ -244,6 +244,7 @@ config :kaffy,
   router: SanbaseWeb.Router
 
 config :sanbase, Sanbase.Kafka.Consumer,
+  enabled?: {:system, "KAFKA_CONSUMER_ENABLED", true},
```
tspenov, line 247: I think it should be false by default

IvanIvanoff: I've made it false here and made it true on prod.

tspenov (approved):

PR #3166, opened 2022-03-11
Add address USD valuation balance functions

IvanIvanoff:
## Checklist:


docs/user-triggers-api.md
```diff
@@ -284,6 +284,36 @@ these are the tickers of the projects. Supported currencies are: `XRP`, `BTC`, `
...
+address in the `target`
+```json
+// The USD valuation of the null address increased by $1 million in the past 24 hours
+{
+  "type": "wallet_usd_valuation",
+  "channel": "telegram",
+  "target": { "address": "0x0000000000000000000000000000000000000000" },
+  "time_window": "1d",
+  "selector": { "infrastructure": "ETH"},
+  "operation": { "amount_down": 1000000 }
```
tspenov, line 313: `amount_up` if it is increased

lib/sanbase/alerts/trigger/settings/wallet_usd_valuation_trigger_settings.ex
```diff
@@ -0,0 +1,197 @@
...
+
+      with {:ok, %{} = result} <- usd_value_change(selector, from, to) do
+        {address,
+         [
+           %{datetime: from, usd_value: result.previous_usd_value},
+           %{datetime: to, usd_value: result.current_usd_value}
+         ]}
+      else
+        result ->
+          raise("""
```
tspenov, line 93: Is it ok to raise here, would it break other things ?

IvanIvanoff: This is more of a code invariant that I've left because of where this code is used.

The code is scheduled and executed by the alerts scheduler. This is less interactive compared to a graphql API. If there's a graphql API error we'll see reports for it, it will appear in sentry, etc. But if something changes in the code related to alerts it would be very easy to miss if we don't raise - it would log an error or something like this, but this is not actionable.

docs/user-triggers-api.md
IvanIvanoff: Fixed.

lib/sanbase/blockchain_address/blockchain_address.ex
```diff
@@ -43,21 +43,39 @@ defmodule Sanbase.BlockchainAddress do
...
-      # BTC and BTC forks
-      Regex.match?(~r/^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,39}$/, address) -> address
-      true -> address
+      Regex.match?(~r/^0x([A-Fa-f0-9]{40})$/, address) ->
+        "ETH"
+
+      Regex.match?(~r/^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,39}$/, address) ->
+        "BTC"
+
+      Regex.match?(~r/r[0-9a-zA-Z^[0OlI]]{23,33}$/, address) and
```
tspenov, line 73: Is that working as expected `Regex.match?(~r/r[0-9a-zA-Z^[0OlI]]{23,33}$/, address)` ?

IvanIvanoff: it's a leftover, I'll write the way we discussed it in chat.

IvanIvanoff: Fixed.

tspenov (approved):

PR #3198, opened 2022-04-13
Add nft social volume metric

tspenov:
## Changes

```graphql
    {
      getMetric(metric: "nft_social_volume"){
        timeseriesData(
          selector: {contractAddressRaw:"0x..."}
          from: "utc_now-50d"
          to: "utc_now"
          interval: "1d"
      	){
            datetime
            value
          }
      }
    }
```

## Checklist:


lib/sanbase/clickhouse/metric/metric_adapter.ex
```diff
@@ -77,19 +77,20 @@ defmodule Sanbase.Clickhouse.MetricAdapter do
 
   def timeseries_data(_metric, %{slug: []}, _from, _to, _interval, _opts), do: {:ok, []}
 
-  def timeseries_data(metric, %{slug: slug}, from, to, interval, opts) do
+  def timeseries_data(metric, selector, from, to, interval, opts) do
```
IvanIvanoff, line 80: Here you can define a guard like `defguard is_supported_selector(s) when is_map(s) and (is_map_key(s, :slug) or is_map_key(s, :contract_address_raw))`

lib/sanbase/clickhouse/metric/metric_adapter.ex
```diff
@@ -77,19 +77,20 @@ defmodule Sanbase.Clickhouse.MetricAdapter do
...
     aggregation = Keyword.get(opts, :aggregation, nil) || Map.get(@aggregation_map, metric)
 
     # FIXME: Some of the `nft` metrics need additional filter for `owner=opensea`
     # to show correct values. Remove after fixed by bigdata.
     filters =
-      if String.starts_with?(metric, "nft_") do
+      if String.starts_with?(metric, "nft_") and metric not in ["nft_social_volume"] do
         [owner: "opensea"]
```
IvanIvanoff, line 87: I think this will continue to cause more trouble. Do we still need this filter?

lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -35,13 +35,17 @@ defmodule Sanbase.Clickhouse.MetricAdapter.SqlQuery do
     field(:computed_at, :utc_datetime)
   end
 
-  def timeseries_data_query(metric, slug_or_slugs, from, to, interval, aggregation, filters) do
+  def timeseries_data_query(metric, selector, from, to, interval, aggregation, filters) do
+    asset_filter =
+      selector[:slug] ||
+        Sanbase.BlockchainAddress.to_internal_format(selector[:contract_address_raw])
```
IvanIvanoff, line 41: Shouldn't we also handle this in other places, not only in timeseries_data? 
Also, this should better happen in a separate function that is something like:
```elixir
case selector do
  %{slug: slug_or_slugs} -> slug_or_slugs
  %{contract_address_raw: address} -> Sanbase.BlockchainAddress.to_internal_format(address)
  unsupported -> raise("Got unsupported selector: #{inspect(unsupported)}")
 end
 ```
Also, `asset_filter` is not very clear to me, maybe `asset_filtering_entities` or something like this?

tspenov: for now implemented - first and last datetime for contract address selector. `Available metrics` now is called within the project query. Let's see how it will be used from frontend and update. Aggregated timeseries also can be implemented later.

PR #3211, opened 2022-04-21
Alerts bugfix and improve scheduling

IvanIvanoff:
## Changes

1. Handle a bug that I currently don't know how/why is happening, but quite some alerts crash because of it (in trigger_metric_helper.ex).
2. Improve the alerts scheduling. Now much less alerts will be properly detected as not receivable (if the user has hit their daily limits) and won't take resources computing.

## Checklist:

### DIFF
--- a/lib/sanbase/accounts/user/alert.ex
+++ b/lib/sanbase/accounts/user/alert.ex
@@ -1,16 +1,31 @@
 defmodule Sanbase.Accounts.User.Alert do
+  alias Sanbase.Accounts.UserSettings
+  alias Sanbase.Accounts.Settings
+
   def can_receive_telegram_alert?(user) do
     user = Sanbase.Repo.preload(user, :user_settings)
 
-    settings = Sanbase.Accounts.UserSettings.settings_for(user)
+    settings = UserSettings.settings_for(user)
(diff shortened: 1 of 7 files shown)


lib/sanbase/accounts/user/alert.ex
```diff
@@ -1,16 +1,31 @@
...
-    settings = Sanbase.Accounts.UserSettings.settings_for(user)
 
-    is_binary(user.email) and settings.alert_notify_email
+    settings = UserSettings.settings_for(user)
+
+    limit = Settings.get_alerts_limit_per_day(settings, "email")
+    fired_today = Settings.get_alerts_fired_today(settings, "email")
+
+    alert_receivable? = is_binary(user.email) and settings.alert_notify_email
+    alert_limit_reached? = fired_today >= limit
```
tspenov, line 27: maybe extract

```elixir
defp alert_limit_reached?(settings, channel) do
  limit = Settings.get_alerts_limit_per_day(settings, channel)
  fired_today = Settings.get_alerts_fired_today(settings, channel)

  fired_today >= limit
end
```

IvanIvanoff: Done.

tspenov (approved):

PR #3212, opened 2022-04-21
Remove hydra oauth related code

tspenov:
## Changes

Hydra oauth no longer used - so remove

## Checklist:


IvanIvanoff:
BTW, maybe we can add a doc in /docs like `removed_features` and link to PRs that remove no longer needed features. In the future, if we need to reintroduce this it would be nice to have something to remind us that we had it. Link to PRs removing the code would do.

PR #3226, opened 2022-05-02
Add social metrics sources: youtube, twitter

volynvlad:
## Checklist:


apozdeyev (approved):

lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
```diff
@@ -670,6 +670,52 @@
...
+    "name": "social_volume_twitter_crypto",
+    "metric": "social_volume_twitter_crypto",
+    "version": "2020-07-31",
+    "access": {
+      "historical": "restricted",
+      "realtime": "free"
+    },
+    "selectors": [
+      "slug",
+      "text"
```
tspenov, line 684: let's remove `text` selector from all for now because currently it won't work out of the box. When fetching with `text`selector we should implement some logic to fetch via `metrics-hub-server->metrics-hub->ES` path.

tspenov: Ok, after whitelisting the metrics that won't be needed

tspenov (approved):

PR #3241, opened 2022-05-18
Adding trending metric to API

filips121:
## Changes

Adding trending metric to API

## Ticket

New metric - Project position in Top Trending Words
https://www.notion.so/santiment/New-metric-Project-position-in-Top-Trending-Words-7578cb16e4ff484493884c080f055115

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
+++ b/lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
@@ -1604,5 +1604,27 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Project rank based on the scores in the top trending words",
+    "name": "trending_rank",
+    "metric": "trending_rank",
+    "version": "2019-01-01",
+    "access": {
+      "historical": "restricted",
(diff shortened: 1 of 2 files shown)


lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
```diff
@@ -1604,5 +1604,27 @@
     "table": "intraday_metrics",
     "has_incomplete_data": false,
     "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Project rank based on the scores in the top trending words",
+    "name": "trending_rank",
+    "metric": "trending_rank",
```
IvanIvanoff, line 1611: We already have the `rank` metric and it is connected with marketcap. Just by `trending_rank` it's not immediately obvious (as it's not widespread) to rank them based on trending words. Maybe we can extend the name to `trending_words_rank` so it's much clearer what it is.

Note that the public name (`name`) and the internal name (`metric`) do not need to match and the public name can differ.

filips121: I agree, fixed

IvanIvanoff (approved):

PR #3252, opened 2022-05-25
Add extra clearing of ETS records before updating the quota

IvanIvanoff:
## Changes

Adding clearing of the ETS record before fetching a new quota and
putting it in the ETS help alleviate the situation where the quota
fetching fails. This way the ETS record is cleared and the usage
cannot be recorded twice in the database.

## Checklist:


tspenov (approved):
Let's try it out.
There are some connected tests that fail.

PR #3283, opened 2022-06-15
Limit dashboard stored size

IvanIvanoff:
## Changes

Allow storing a cached version of a panel only if it's gzip compressed size is less than 500KB.

## Checklist:

### DIFF
--- a/test/sanbase_web/graphql/dashboard/dashboard_api_test.exs
+++ b/test/sanbase_web/graphql/dashboard/dashboard_api_test.exs
@@ -448,6 +448,58 @@ defmodule SanbaseWeb.Graphql.DashboardApiTest do
                } = dashboard_cache
       end)
     end
+
+    test "cannot store result bigger than the allowed limit", context do
+      dashboard =
+        execute_dashboard_mutation(context.conn, :create_dashboard)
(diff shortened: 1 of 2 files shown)


test/sanbase_web/graphql/dashboard/dashboard_api_test.exs
```diff
@@ -448,6 +448,58 @@ defmodule SanbaseWeb.Graphql.DashboardApiTest do
...
+            :compute_and_store_dashboard_panel,
+            %{
+              dashboard_id: dashboard["id"],
+              panel_id: panel["id"]
+            }
+          )
+          |> get_in(["errors", Access.at(0), "message"])
+
+        assert error_msg =~
+                 "Cannot cache the panel because its compressed size is 517.88KB which is over the limit of 500KB"
```
tspenov, line 500: is it always 517.88KB even when using random for generating data ?

IvanIvanoff: Yes, I am seeding the rand generator with the same number:
```elixir
# seed the rand generator so it gives the same result every time
:rand.seed(:default, 42)
```

PR #3311, opened 2022-07-01
Add sanitized string types and use them in insights

IvanIvanoff:
## Checklist:


lib/sanbase_web/graphql/schema/custom_types/sanitized_string.ex
```diff
@@ -0,0 +1,41 @@
...
+    serialize(&serialize_sanitized_markdown_string/1)
+    parse(&parse_sanitized_markdown_string/1)
+  end
+
+  defp serialize_sanitized_string(value) when is_binary(value) do
+    HtmlSanitizeEx.strip_tags(value)
+  end
+
+  defp serialize_sanitized_string(nil), do: {:ok, nil}
+  defp serialize_sanitized_string(_), do: :error
```
tspenov, line 21: wouldn't that return :error for all non-null strings of type `sanitized_string` ?

lib/sanbase_web/graphql/schema/custom_types/sanitized_string.ex
```diff
@@ -0,0 +1,41 @@
...
+  alias Absinthe.Blueprint.Input
+
+  scalar :sanitized_string, name: "SanitizedString" do
+    serialize(&serialize_sanitized_string/1)
+    parse(&parse_sanitized_string/1)
+  end
+
+  scalar :sanitized_markdown_string, name: "SanitizedMarkdownString" do
+    serialize(&serialize_sanitized_markdown_string/1)
+    parse(&parse_sanitized_markdown_string/1)
```
tspenov, line 13: we keep the "> " but then we call `strip_tags` - wouldn't that remove `>` again ?

lib/sanbase_web/graphql/schema/custom_types/sanitized_string.ex
IvanIvanoff: We have 3 functions:
- serialize_sanitized_string(value) when is_binary(value)
- serialize_sanitized_string(nil)
- serialize_sanitized_string(_)

This would return error for all non-string values. I guess it's better to straight raise an error here.

lib/sanbase_web/graphql/schema/custom_types/sanitized_string.ex
IvanIvanoff: These are 2 separate types. In the markdown case we don't call strip_tags

lib/sanbase_web/graphql/schema/custom_types/sanitized_string.ex
tspenov: wow, didn't see the first one

tspenov (approved):

tspenov:
a test will be useful but it could be added later if locally tested

IvanIvanoff:
@tspenov I want to see how it behaves on sanbase frontend. If we decide we're going to keep it I'll add tests

PR #3318, opened 2022-07-07
Switch twitter followers from influx to CH

tspenov:
## Checklist:


lib/sanbase/twitter/twitter.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp last_record_query(twitter_handle) do
+    query = """
+      SELECT
+        dt,
+        followers_count
+      FROM #{@table}
+      PREWHERE twitter_handle = ?1
+      ORDER BY dt DESC
+      LIMIT 1
```
IvanIvanoff, line 67: Isn't this quite slow? In other places we do some artificial filter like `AND dt >= now() - INTERVAL 14 DAY` so all assets that don't have their data updated for 2+ weeks are considered deprecated.

PR #3324, opened 2022-07-11
Remove influxdb

tspenov:
Fix first and last datetime functions

## Checklist:


IvanIvanoff:
You should also remove the `instream` dependency from the mix .exs and .lock files. There could be other dependencies that will no longer be needed,  you can run `mix deps.unlock --unused` after removing the dependency from the .exs file

IvanIvanoff (approved):

PR #3386, opened 2022-08-02
Add README for SQL Editor basics

IvanIvanoff:
## Checklist:


docs/clickhouse-user-defined-functions.md
```diff
@@ -0,0 +1,6 @@
+```sql
+CREATE FUNCTION get_asset_id AS (__slug_internal_arg__) -> (SELECT asset_id FROM asset_metadata FINAL where name = lower(__slug_internal_arg__) LIMIT 1);
```
apozdeyev, line 2: I also created dictionaries to get asset/metric metadata by name, let's use them 
https://github.com/santiment/devops/blob/fe843a45c94822239cb1a7f2261db244e69a853c/stage/k8s-apps/sanbase/clickhouse_tables/dictionaries.yaml#L85

docs/clickhouse/clickhouse-exploration.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `%` in the beginning means that there could be other characters to the left.
+The `%` in the end means that there could be other characters to the right.
+
+## Get information about a table
+
+In order to inspect the structure of a given table one can execute the `DESCRIBE` statement:
+
+```sql
+DESCRIBE intraday_metrics
```
filips121, line 33: There is also a shorter version: DESC intraday_metrics (maybe put a note so users don't have to write DESCRIBE all the time, but this way is also good).

docs/clickhouse/clickhouse-exploration.md
```diff
@@ -0,0 +1,70 @@
...
+│ contract_addresses │ Array(String)    │              │                                │         │                  │                │
+│ specification      │ Nullable(String) │              │                                │         │                  │                │
+└────────────────────┴──────────────────┴──────────────┴────────────────────────────────┴─────────┴──────────────────┴────────────────┘
+```
+
+In order to see how a table was created one can execute the `SHOW CREATE TABLE` statement. This includes information
+about the partitioning, ordering, table engine and other settings. Knowing the `ORDER BY` helps creating better and faster queries.
+
+```sql
+SHOW CREATE TABLE intraday_metrics
```
filips121, line 53: There is also a shorter version: SHOW CREATE intraday_metrics (no need for TABLE word)

docs/clickhouse/clickhouse-exploration.md
```diff
@@ -0,0 +1,70 @@
...
+┌─statement──────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ CREATE TABLE default.intraday_metrics
+(
+    `asset_id` UInt64 CODEC(DoubleDelta, LZ4),
+    `metric_id` UInt64 CODEC(DoubleDelta, LZ4),
+    `dt` DateTime CODEC(DoubleDelta, LZ4),
+    `value` Float64,
+    `computed_at` DateTime
+)
+ENGINE = ReplicatedReplacingMergeTree('/clickhouse/tables/global/intraday_metrics_v2', '{hostname}', computed_at)
```
filips121, line 65: Also, one note we can put somewhere (maybe not for the exploration part). But when users work with ReplicatedReplacingMergeTree tables, they can use the keyword FINAL after the table_name to select the values with last computed_at DateTime.
For example:
without FINAL - 3 records with different computed_at
`
SELECT *
FROM asset_metadata
WHERE asset_id = dictGet('assets_by_name', 'asset_id', 'bitcoin')
`
with FINAL - 1 record with last computed_at
`
SELECT *
FROM asset_metadata
FINAL
WHERE asset_id = dictGet('assets_by_name', 'asset_id', 'bitcoin')
`

IvanIvanoff: I've thought about this, too.
In the queries that power the API we don't use `FINAL` for performance reasons - it is much faster to group by and aggregate. But while it's faster, it's also much less readable and you should know the primary key.
I guess we can split it in 2 parts - use `FINAL` by default, fallback to `GROUP BY` + aggregate when speed is needed.

docs/clickhouse/clickhouse-exploration.md
IvanIvanoff: In this case I prefer not to mention `DESC` as it's the same as `ASC` and `DESC` directions used in `ORDER BY`. Using the whole word will lead to less confusion.

docs/clickhouse/clickhouse-exploration.md
filips121: I agree, this 2 part solution seems like a good solution.

docs/clickhouse-user-defined-functions.md
IvanIvanoff: done

IvanIvanoff:
Close in favor of https://github.com/santiment/academy/pull/224

docs/clickhouse/sql-editor-readme.md
```diff
@@ -0,0 +1,545 @@
...
+- `asset_id` - An `UInt64` unique identifier for an asset. The data for that id
+  is stored in the `asset_metadata` table.
+- `metric_id` - An `UInt64` unique identifier for metric. The data for that id
+  is stored in the `metric_metadata` table.
+- `value` - A `Float` column holding the metric's value for the given
+  asset/metric pair.
+- `computed_at` - A `DateTime` column storing the date and time when the
+  given row was computed.
+
+### The FINAL keyword
```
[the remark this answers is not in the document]
IvanIvanoff: The documentation reflects the current version of how things work, so it is unavoidable without DB changes.

PR #3390, opened 2022-08-05
Support plans with custom restrictions

IvanIvanoff:
## Changes

The custom API plans will give flexibility when configuring restrictions:
- Configure which metrics are accessible
- Configure the number of API calls
- Configure the free/restricted metrics access. This includes:
  - Configure how much historical/real-time data is available, 2 years of historical data with no real-time restriction
  - Configure which metrics the restrictions are applied to. This is easy to configure at the expense of flexibility. This is handled by using one of the FREE/BASIC/PRO/PREMIUM restrictions.
 
### Implementation details

The incorporation of custom plans should happen painlessly and transparently.
For this reason, the only changes (excluding improvements) to existing code are done in the `AccessChecker` module. Now it serves as a dispatch module. The old AccessChecker code is moved to `StandardAccessChecker`. The dispatch module, upon stumbling upon a plan whose name starts with `CUSTOM_` will dispatch to `CustomAccessChecker`. If the name is not starting with `CUSTOM_` - it dispatches to `StandardAccessChecker`.

The same logic has been added to API calls.

### Creation of new plans

Basic UI has been added to create/update new plans:
image
image
image

## Checklist:


lib/sanbase/billing/plan.ex
```diff
@@ -27,23 +28,57 @@ defmodule Sanbase.Billing.Plan do
     # interval is one of `month` or `year`
     field(:interval, :string)
     field(:stripe_id, :string)
-
     # there might be still customers on this plan, but new subscriptions should be disabled.
-    field(:is_deprecated, :boolean, default: false)
-
+    field(:is_deprecated, :boolean)
```
tspenov, line 32: why remove the default ?

IvanIvanoff: I'll return it

PR #3418, opened 2022-08-31
Add views count for moderators in explorer

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/entity/entity.ex
+++ b/lib/sanbase/entity/entity.ex
@@ -32,6 +32,7 @@ defmodule Sanbase.Entity do
   alias Sanbase.UserList
   alias Sanbase.Dashboard
   alias Sanbase.Alert.UserTrigger
+  alias Sanbase.Accounts.Interaction
 
   # The list of supported entitiy types. In order to add a new entity type, the
   # following steps must be taken:
@@ -209,6 +210,38 @@ defmodule Sanbase.Entity do
     query
     |> limit(^limit)
     |> offset(^offset)
+  end
+
+  def extend_with_views_count(result) do
+    concat_entities =
+      Enum.map(result, fn elem ->
+        [{type, entity}] = Map.to_list(elem)
+        "#{Interaction.deduce_entity_column_name(type)}#{entity.id}"
+      end)
+
+    query = """
+    WITH concat_entities AS (select concat(entity_type, entity_id) AS entity, interaction_type FROM user_entity_interactions)
+    SELECT entity, count(interaction_type)
+    FROM concat_entities
+    WHERE entity IN (#{concat_entities |> Enum.map(fn elem -> "'#{elem}'" end) |> Enum.join(",")}) AND interaction_type = 'view'
+    GROUP BY entity;
+    """
+
+    query_res = Sanbase.Repo.query!(query)
+
+    entity_count_map =
+      query_res.rows
+      |> Enum.into(%{}, fn [entity, count] ->
+        {entity, count}
+      end)
+
+    Enum.map(result, fn elem ->
+      [{type, entity}] = Map.to_list(elem)
+      key = "#{Interaction.deduce_entity_column_name(type)}#{entity.id}"
+
+      entity = %{entity | views: entity_count_map[key] || 0}
+      [{type, entity}] |> Enum.into(%{})
+    end)
   end
 
   # Private functions
--- a/lib/sanbase/entity/entity.ex
+++ b/lib/sanbase/entity/entity.ex
@@ -213,26 +213,31 @@ defmodule Sanbase.Entity do
   end
 
   def extend_with_views_count(result) do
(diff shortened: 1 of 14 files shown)


lib/sanbase/entity/entity.ex
```diff
@@ -211,6 +212,38 @@ defmodule Sanbase.Entity do
...
+        [{type, entity}] = Map.to_list(elem)
+        "#{Interaction.deduce_entity_column_name(type)}#{entity.id}"
+      end)
+
+    query = """
+    WITH concat_entities AS (select concat(entity_type, entity_id) AS entity, interaction_type FROM user_entity_interactions)
+    SELECT entity, count(interaction_type)
+    FROM concat_entities
+    WHERE entity IN (#{concat_entities |> Enum.map(fn elem -> "'#{elem}'" end) |> Enum.join(",")}) AND interaction_type = 'view'
+    GROUP BY entity;
```
IvanIvanoff, line 227: You can `GROUP BY entity_type, entity_id` instead of concat + group by the result.
I suspect the following code might be more performant, as with the `concat` you'll scan and transform the whole table before applying the filter (no index can be used).

```elixir
entity_type_id_conditions = Enum.reduce(result, dynamic(true), fn elem, query ->
[{type, %{id: entity_id}}] = Map.to_list(elem)
dynamic([row], ^dynamic or (row.entity_type == ^type and row.entity_id == ^entity_id))

query = """
SELECT entity_id, entity_type, count(*) AS total_views
FROM user_entity_interactions
WHERE interaction_type = 'view' AND ^entity_type_id_conditions
GROUP BY entity_id, entity_type
"""
```

lib/sanbase/entity/entity.ex
```diff
@@ -211,6 +212,43 @@ defmodule Sanbase.Entity do
...
+          ^dynamic_query or (row.entity_type == ^type and row.entity_id == ^entity_id)
+        )
+      end)
+
+    entity_type_id_conditions =
+      dynamic([row], row.interaction_type == "view" and ^entity_type_id_conditions)
+
+    query =
+      from(row in Interaction,
+        where: ^entity_type_id_conditions,
```
IvanIvanoff, line 232: wouldn't it be much simpler to do just:
```elixir
where: row.interaction_type == "view" and entity_type_id_conditions
```

lib/sanbase/entity/entity.ex
```diff
@@ -211,6 +212,43 @@ defmodule Sanbase.Entity do
...
+      |> Enum.into(%{}, fn [entity_id, entity_type, count] ->
+        {"#{entity_type}#{entity_id}", count}
+      end)
+
+    Enum.map(result, fn elem ->
+      [{type, entity}] = Map.to_list(elem)
+      key = "#{Interaction.deduce_entity_column_name(type)}#{entity.id}"
+
+      entity = %{entity | views: entity_count_map[key] || 0}
+      [{type, entity}] |> Enum.into(%{})
```
IvanIvanoff, line 248: That's just `%{type => entity}`

lib/sanbase/entity/entity.ex
```diff
@@ -211,6 +212,43 @@ defmodule Sanbase.Entity do
...
+      from(row in Interaction,
+        where: ^entity_type_id_conditions,
+        select: [row.entity_id, row.entity_type, fragment("COUNT(*)")],
+        group_by: [row.entity_id, row.entity_type]
+      )
+
+    entity_count_map =
+      Sanbase.Repo.all(query)
+      |> Enum.into(%{}, fn [entity_id, entity_type, count] ->
+        {"#{entity_type}#{entity_id}", count}
```
IvanIvanoff, line 240: There's no need to create strings here, the key can be just the tuple `{entity_type, entity_id}`

lib/sanbase/entity/entity.ex
```diff
@@ -211,6 +212,43 @@ defmodule Sanbase.Entity do
...
+      end)
+
+    Enum.map(result, fn elem ->
```
IvanIvanoff, line 243: nitpick: I was confused by the `result` name here as I expected `result` is the result of the database call here. Maybe call it `type_entity_list` and elem can be `type_entity` :?

lib/sanbase/entity/entity.ex
tspenov: I think that `where: row.interaction_type == "view" and entity_type_id_conditions` is not supported - you can check here: https://hexdocs.pm/ecto/Ecto.Query.html#dynamic/2-where-having-and-a-join-s-on

IvanIvanoff: I saw that in the docs, but I haven't seen the comment above it that this is forbidden, so I assumed it's valid code.

tspenov:
comments addressed in: https://github.com/santiment/sanbase2/pull/3418

PR #3441, opened 2022-09-13
Social dominance of top10 trending words

tspenov:
## Checklist:


lib/sanbase/social_data/social_dominance.ex
```diff
@@ -61,6 +63,34 @@ defmodule Sanbase.SocialData.SocialDominance do
     end
   end
 
+  def social_dominance_trending_words() do
+    now = Timex.now()
+    from = Timex.shift(now, hours: -@hours_back_ensure_has_data)
+    to = now
```
IvanIvanoff, line 69:
```suggestion
    to = Timex.now()
    from = Timex.shift(to, hours: -@hours_back_ensure_has_data)
```

lib/sanbase/social_data/social_dominance.ex
```diff
@@ -61,6 +63,34 @@ defmodule Sanbase.SocialData.SocialDominance do
...
+    interval = "1h"
+    source = :total
+
+    with {:ok, trending_words} <-
+           Sanbase.SocialData.TrendingWords.get_currently_trending_words(@trending_words_size),
+         words <- Enum.map(trending_words, & &1.word),
+         {:ok, words_volume} <-
+           Sanbase.SocialData.social_volume(%{words: words}, from, to, interval, source),
+         {:ok, total_volume} <-
+           Sanbase.SocialData.social_volume(%{text: "*"}, from, to, interval, source) do
```
IvanIvanoff, line 79: You can alias the `SocialData` module and these would probably fit 1 line.

lib/sanbase/social_data/social_dominance.ex
```diff
@@ -61,6 +63,34 @@ defmodule Sanbase.SocialData.SocialDominance do
...
+      words_mentions_sum =
+        words_volume
+        |> Enum.reduce(0.0, fn word, acc ->
+          acc + List.last(word.timeseries_data).mentions_count
+        end)
+
+      total_mentions = List.last(total_volume).mentions_count
+
+      dominance = Sanbase.Math.percent_of(words_mentions_sum, total_mentions) || 0.0
```
IvanIvanoff, line 88: Maybe keep `nil` instead of 0.0 here. It would return `nil` in case any of the provided values is not a number.

lib/sanbase/social_data/social_dominance.ex
```diff
@@ -61,6 +63,34 @@ defmodule Sanbase.SocialData.SocialDominance do
...
+         words <- Enum.map(trending_words, & &1.word),
+         {:ok, words_volume} <-
+           Sanbase.SocialData.social_volume(%{words: words}, from, to, interval, source),
+         {:ok, total_volume} <-
+           Sanbase.SocialData.social_volume(%{text: "*"}, from, to, interval, source) do
+      words_mentions_sum =
+        words_volume
+        |> Enum.reduce(0.0, fn word, acc ->
+          acc + List.last(word.timeseries_data).mentions_count
+        end)
```
IvanIvanoff, line 84: In this case we know the list has at most 3 elements, so `reduce` would not bring any performance benefits. I think with 3 functions it would be easier to read.
```suggestion
      words_mentions_sum =
        Enum.map(words_volume, &List.last(&1).mentions_count)
        |> Enum.sum()
```

lib/sanbase/social_data/social_dominance.ex
tspenov: I think 0.0 is a good fallback value for the float type result here

PR #3456, opened 2022-09-23
Improve CH query speed when asset is not present in CH

tspenov:
## Changes

The change seems to behave better when asset is not present in the asset_metadata in CH. 

The query with `asset_id = ` takes a lot of time to complete when asset is not present while the one with `asset_id IN ` finishes fast. There is no difference in the execution plan between both.
Here is example of the slow query:
```sql
SELECT
    toUnixTimestamp(intDiv(toUInt32(toDateTime(dt)), 600) * 600) AS t,
    sumKahan(value)
FROM
(
    SELECT
        dt,
        argMax(value, computed_at) AS value
    FROM intraday_metrics
    PREWHERE (dt >= toDateTime(1651598800)) AND (dt < toDateTime(1663928371)) AND (asset_id = (
        SELECT asset_id
        FROM asset_metadata
        FINAL
        PREWHERE name IN ('diamond-launch')
        LIMIT 1
    )) AND (metric_id = (
        SELECT metric_id
        FROM metric_metadata
        FINAL
        PREWHERE name = 'social_volume'
        LIMIT 1
    ))
    GROUP BY
        asset_id,
        dt
)
WHERE (value IS NOT NULL) AND (NOT isNaN(value))
GROUP BY t
ORDER BY t ASC
FORMAT JSONCompact
```

## Checklist:

### DIFF
--- a/lib/sanbase/metric/sql_query_helper.ex
+++ b/lib/sanbase/metric/sql_query_helper.ex
@@ -119,7 +119,7 @@ defmodule Sanbase.Metric.SqlQuery.Helper do
   def asset_id_filter(%{slug: slug}, opts) when is_binary(slug) do
     arg_position = Keyword.fetch!(opts, :argument_position)
 
-    "asset_id = ( SELECT asset_id FROM asset_metadata FINAL PREWHERE name = ?#{arg_position} LIMIT 1 )"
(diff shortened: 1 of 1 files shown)


IvanIvanoff (approved):

tspenov:
Same issue reported here: https://github.com/ClickHouse/ClickHouse/issues/41814

PR #3471, opened 2022-10-04
Add GraphiQL advanced interface on /graphiql_advanced endpoint

IvanIvanoff:
## Changes

Add `/advanced_graphiql` endpoint.

Keep the original `/graphiql` endpoint as-is due to backwards compatibility reasons. The new endpoint allows users to set custom headers. This is needed when API-only clients want to explore the API via graphiql. In order to do so, the Authorization header needs to be set:

image

## Checklist:


tspenov:
I think we had some problems with the advanced interface in the past but can't remember what were they 🤷‍♂️

tspenov (approved):

PR #3483, opened 2022-10-06
Send a discord msg on payment action required

tspenov:
## Checklist:


lib/sanbase/billing/subscription/event_handling/discord_notification.ex
```diff
@@ -43,6 +43,30 @@ defmodule Sanbase.Billing.DiscordNotification do
...
+    customer = stripe_event["data"]["object"]["customer"]
+    customer = "https://dashboard.stripe.com/customers/#{customer}"
+
+    message = """
+    ⛔ A payment for #{formatted_amount} requires action.
+    Customer: #{customer}
+    Event: https://dashboard.stripe.com/events/#{stripe_event_id}
+    """
+
+    do_send_to_discord(message, "Stripe Payment", webhook: "bug_report")
```
IvanIvanoff, line 67: is this always happening because of bugs?

IvanIvanoff (approved):

lib/sanbase/billing/subscription/event_handling/discord_notification.ex
tspenov: It is not a bug but it needs an action from our side so there will be more visible. Other option is to tag someone.

IvanIvanoff: I was confused by the `bug_report` webhook name. Maybe `payment_action_required` would be better?

tspenov: changed

PR #3493, opened 2022-10-11
Start implementing generic admin

tspenov:
## Checklist:


lib/sanbase_web/controllers/generic_controller.ex
```diff
@@ -0,0 +1,103 @@
...
+  defp search_by_text(module, text) do
+    module.by_search_text(text)
+  end
+
+  defp search_by_id(module, id) do
+    case Repo.get(module, id) do
+      nil -> []
+      result -> [result]
+    end
+  end
```
IvanIvanoff, line 94: It could be better or worse, but also possible to check if the module has `by_id` function and if it does - use it instead. 
Another option is to look for a generic way to preload all associations or at least point to their page.

PR #3494, opened 2022-10-12
Add active holders distribution to API

filips121:
## Checklist:


IvanIvanoff:
I think it would be better to add something similar to how timebound is handled: https://github.com/santiment/sanbase2/blob/master/lib/sanbase/clickhouse/metric/metric_files/available_v2_metrics.json#L291-L303

Otherwise we end up with 1000 lines of JSON where 3 json maps would suffice, the rest would be the values that will fill the pattern.

tspenov (approved):

PR #3516, opened 2022-10-21
Add special handling of rewriting modules for price_eth

IvanIvanoff:
## Changes
Short and dirty fix of which metric adapter module is used for `price_eth`.
The issue is that the default module for `price_usd`/`price_btc` is not the proper default module for `price_eth`.

The current fix is a temporary workaround until a proper, generic way, to handle this is implemented.

## Checklist:

### DIFF
--- a/lib/sanbase/metric/metric.ex
+++ b/lib/sanbase/metric/metric.ex
@@ -795,8 +795,15 @@ defmodule Sanbase.Metric do
   defp maybe_change_module(module, metric, selector, opts)
        when metric in @price_pair_metrics do
     case Keyword.get(opts, :source) || Map.get(selector, :source) do
-      "cryptocompare" -> Sanbase.PricePair.MetricAdapter
-      _ -> module
+      "cryptocompare" ->
+        Sanbase.PricePair.MetricAdapter
+
+      # NOTE: Temporary. This will be reworked and handled in a generic way
+      source when metric == "price_eth" and source != "cryptocompare" ->
+        Sanbase.Clickhouse.MetricAdapter
+
+      _ ->
+        module
     end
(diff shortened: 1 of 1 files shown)


antisaa (approved):

lib/sanbase/metric/metric.ex
```diff
@@ -146,6 +146,7 @@ defmodule Sanbase.Metric do
         metric_not_available_error(metric, type: :timeseries)
 
       module when is_atom(module) ->
+        IO.inspect(module)
```
tspenov, line 149: inspect left

lib/sanbase/metric/metric.ex
```diff
@@ -795,8 +796,15 @@ defmodule Sanbase.Metric do
   defp maybe_change_module(module, metric, selector, opts)
        when metric in @price_pair_metrics do
     case Keyword.get(opts, :source) || Map.get(selector, :source) do
-      "cryptocompare" -> Sanbase.PricePair.MetricAdapter
-      _ -> module
+      "cryptocompare" ->
+        Sanbase.PricePair.MetricAdapter
+
+      # NOTE: Temporary. This will be reworked and handled in a generic way
+      source when metric == "price_eth" and source != "cryptocompare" ->
```
tspenov, line 803: source != "cryptocompare" is not needed

tspenov (approved):

lib/sanbase/metric/metric.ex
IvanIvanoff: I know, I just left this as a comment/clarification that we need to change only in this case and to guard against reordering of the clauses.

PR #3546, opened 2022-11-07
Read projects related data from github repository

IvanIvanoff:
## Checklist:


lib/sanbase_web/controllers/repo_reader_controller.ex
```diff
@@ -0,0 +1,60 @@
...
+    case endpoint_secret() == secret do
+      true ->
+        changed_files = Map.get(params, "changed_files", [])
+        branch = Map.get(params, "branch", "main")
+
+        case Sanbase.RepoReader.validate_changes(branch, changed_files) do
+          :ok ->
+            conn
+            |> put_resp_header("content-type", "application/json; charset=utf-8")
+            |> Plug.Conn.send_resp(200, %{result: :ok} |> Jason.encode!())
```
tspenov, line 18: can be simplified with `json(conn, result)`: https://hexdocs.pm/phoenix/Phoenix.Controller.html#json/2

lib/sanbase/repo_reader/repo_reader.ex
```diff
@@ -0,0 +1,136 @@
...
+      error -> error
+    end
+  end
+
+  # Private functions
+
+  defp do_update_projects(path, changed_directories) do
+    with {:ok, %Repository{} = repo} <- clone_repo(path),
+         {:ok, projects_map} = read_files(repo, directories_to_read: changed_directories) do
+      slugs = Map.keys(projects_map)
```
tspenov, line 72: we use the user-made directories as project slugs - do we validate this somewhere ?

IvanIvanoff: I added and then removed that check at some point. I'm going to think about this and either add a check or just use the slug - I'm more in favour in the second one as the directory can easily be capitalized - `Bitcoin` vs `bitcoin`

IvanIvanoff: Fixed.

lib/sanbase_web/controllers/repo_reader_controller.ex
IvanIvanoff: done.

PR #3562, opened 2022-11-16
Propagate changed email to Stripe

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/accounts/accounts_event_emitter.ex
+++ b/lib/sanbase/accounts/accounts_event_emitter.ex
@@ -36,7 +36,11 @@ defmodule Sanbase.Accounts.EventEmitter do
     |> notify()
   end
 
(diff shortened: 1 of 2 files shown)


lib/sanbase/accounts/accounts_event_emitter.ex
```diff
@@ -36,7 +36,11 @@ defmodule Sanbase.Accounts.EventEmitter do
     |> notify()
   end
 
-  def handle_event({:ok, user}, :update_email, %{old_email: _, new_email: _} = args) do
+  def handle_event({:ok, user}, :update_email, %{old_email: _, new_email: new_email} = args) do
+    if user.stripe_customer_id do
+      Stripe.Customer.update(user.stripe_customer_id, %{email: new_email})
```
IvanIvanoff, line 41: you can add `{:ok, _} = ...` to confirm the action succeeded so it can be retried if it fails.

PR #3609, opened 2022-12-07
Refactor the User module. Move functions to Accounts context

IvanIvanoff:
## Changes

There are a lot of things in the User module that don't belong there. Remove delegation functions and just use the proper module. Move functions that interact with multiple module from the User module to the Accounts context

## Checklist:

### DIFF
--- a/lib/sanbase/accounts/accounts.ex
+++ b/lib/sanbase/accounts/accounts.ex
@@ -22,8 +22,8 @@ defmodule Sanbase.Accounts do
     multi_result =
       Ecto.Multi.new()
       |> Ecto.Multi.run(:add_user, fn _, _ -> User.create(%{username: address}) end)
-      |> Ecto.Multi.run(:add_eth_account, fn _repo, %{add_user: %User{id: id}} ->
-        EthAccount.create(%{user_id: id, address: address})
+      |> Ecto.Multi.run(:add_eth_account, fn _repo, %{add_user: %User{} = user} ->
+        EthAccount.create(user.id, address)
       end)
       |> Repo.transaction()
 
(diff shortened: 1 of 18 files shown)


lib/sanbase/accounts/accounts.ex
```diff
@@ -41,4 +35,41 @@ defmodule Sanbase.Accounts do
...
+      end
+    else
+      {:error,
+       "Cannot remove ethereum address #{address}. There must be an email or other ethereum address set."}
+    end
+  end
+
+  # Helpers
+
+  defp can_remove_eth_account?(%User{id: user_id, email: email}, address) do
```
tspenov, line 65: should these couple of functions that work with `eth_account` also go  `eth_account.ex`

IvanIvanoff: I'm not sure.
I've moved big parts of the internals to `EthAccount` - previously this function constructed the SQL query itself and executed it. Now it only calls the function from that module and adds some logic on top of it.

What belongs to such modules are functions that call functions from multiple modules - `User` and `EthAccount` in this case. In such cases, I can argue it can belong to either of those 2, but here it looks more appropriate. With time we can move more functions here.

IvanIvanoff: On a second thought, I think I can do the following:
- leave `create_user_with_eth_address` in this module
- move `add_eth_account` and `remove_eth_account` to EthAccount

This is most in line with what I described above and makes most sense to me.

tspenov: Sounds good to me 👍

PR #3611, opened 2022-12-08
Rework login/registration state and events

IvanIvanoff:
## Changes

Currently, the registration progress is tracked by a single `is_registered` boolean field in the `users` table. It is set to `true` when the registration is finished - email verified, google oauth successfully passed, etc.

There is a virtual field `first_login` which is set to `true` only when the user is inserted. This makes it very hard to propagate to the API result. For example, if we have code like:
```elixir
{:ok, user} = User.insert_or_find_by(...)
{:ok, user} = User.mark_as_registered(...)
```

The second call to the `User` module will set the `first_login` field to false, as we're getting the user not for the first time.
Furthermore, when authentication via `/auth/google`, this endpoint only sets the cookies but does not return the user. The first call to `currentUser` after that has `first_login: false` as this user is already seen as existing and registered, so it is not known if this is the first, second, or 100th login.

This PR reworks this process by introducing a `registration_state` JSON field that holds a `state` and additional `data` (like `origin_url`, `login_origin`, etc.). The state is evolved as a state machine - using the current state and an action (`send_login_email`, `google_oauth`, `login`, etc.) allows for multiple states to exist, also making it possible to forward the state from the `currentUser` process, too.

This state evolving is done by calling `Sanbase.Accounts.forward_registration(user, action, data)`, which is doing the state changing and events emitting. It is now responsible for emitting the `register_user` event. The additional data includes the details for the login/registration and is used for the events.

The code now will look like this:
```elixir
 with {:ok, user} <- twitter_login(email, twitter_id),
      {:ok, _, user} <- Accounts.forward_registration(user, "twitter_oauth", args),
      {:ok, %{} = jwt_tokens_map} <- SanbaseWeb.Guardian.get_jwt_tokens(user, device_data) do
  conn
  |> SanbaseWeb.Guardian.add_jwt_tokens_to_conn_session(jwt_tokens_map)
  |> redirect(external: redirect_urls.success)
else
  _ ->
    conn
    |> redirect(external: redirect_urls.fail)
end
```

`forward_registration` returns returns `{:error, error}` or a 3-element :ok tuple `{:ok, state_action, user}` where `state_action` is one of `:keep_state` or `:evolve_state`, with the following meaning:
- `:keep_state` - the state did not change or it was updated by some concurrent process on the same node or by other node.
- `:evolve_state` - the state changed. This can be used when emitting events, setting `firstLogin` or doing some other action that requires `only once` semantics. 
This is ensured by using an atomic `where ... update .. select` SQL query that will update only once, no matter the concurrency.

This can be illustrated by the following example:
```sql
{
  "data": {
    "first": {
      "email": "test@santiment.net",
      "firstLogin": true,
      "id": "16"
    },
    "second": {
      "email": "test@santiment.net",
      "firstLogin": false,
      "id": "16"
    },
    "third": {
      "email": "test@santiment.net",
      "firstLogin": false,
      "id": "16"
    }
  }
}
```

The `registration` state will also keep the `data` field even after finishing. Example:
image

## Checklist:

### DIFF
--- a/lib/sanbase/accounts/user/user_registration_state.ex
+++ b/lib/sanbase/accounts/user/user_registration_state.ex
@@ -1,0 +1,89 @@
+defmodule Sanbase.Accounts.User.RegistrationState do
+  @moduledoc ~s"""
+  The Registration State is a JSON field of the User that holds the current
+  state of the registration process along with some metadata.
+
+  The state is evolved like a state machine - given the current state and some
+  action, the state can either evolve to a new state or be kept the same.
+  """
+
+  alias Sanbase.Accounts.EventEmitter
+  alias Sanbase.Accounts.User
+  alias __MODULE__.StateMachine
+
+  @doc ~s"""
+  Take a step forward in the registration progress by executing the given `action`.
+
+  The new data is merged with the old data and this is stored in the map as well,
+  so it can be used later.
+
+  In some of the cases emit events like:
+  If the state changes from not finished to finish, emit :register_user event
+  """
+  def forward(%User{registration_state: registration_state} = user, action, data) do
+    current_state = Map.fetch!(registration_state, "state")
+
+    old_data =
+      Map.get(registration_state, "data", %{})
+      |> Map.new(fn {k, v} -> {String.to_existing_atom(k), v} end)
+
+    merged_data = Map.merge(old_data, data)
+
+    case StateMachine.forward(current_state, action) do
+      :keep_state ->
+        :keep_state
+
+      {:next_state, next_state} ->
+        {:next_state, next_state, merged_data}
+    end
+    |> emit_event(user)
+  end
+
+  def is_registered(%User{registration_state: registration_state}) do
+    registration_state["state"] == "finished"
+  end
+
+  def is_first_login(%User{registration_state: registration_state}, action)
+      when action in ["eth_login", "email_login_verify"] do
+    %{"state" => current_state} = registration_state
+
+    # This is going to be used in case of eth_login and email_login_verify actions
+    # In these cases, if this action is performed and the state is not finished, but
+    # moves to `finished` after performing the action, then this can be treated
+    # as first_login: true`
+    current_state != "finished" and
+      match?({:next_state, "finished"}, StateMachine.forward(current_state, action))
+  end
+
+  def login_to_finish_registration?(%User{registration_state: registration_state}) do
+    registration_state["state"] == "wait_login_to_finish_registration"
+  end
+
+  defp emit_event(:keep_state, _), do: :keep_state
+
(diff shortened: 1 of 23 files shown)


priv/repo/migrations/20221215103058_add_registration_state_field.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Sanbase.Repo.Migrations.AddRegistrationStateField do
+  use Ecto.Migration
+
+  def change do
+    # alter table(:users) do
+    #   add(:registration_state, :jsonb, default: "{\"state\": \"init\"}")
+    # end
+  end
+end
```
tspenov, line 9: maybe remove if unused

IvanIvanoff: oops, it's not unused, it's wrongly commented 😬

IvanIvanoff: fixed

lib/sanbase_web/controllers/auth_controller.ex
```diff
@@ -26,17 +27,13 @@ defmodule SanbaseWeb.AccountsController do
     redirect_urls = get_redirect_urls(params)
     device_data = SanbaseWeb.Guardian.device_data(conn)
     email = auth.info.email
-    origin_url = Plug.Conn.get_req_header(conn, "host")
+    origin_url = Plug.Conn.get_req_header(conn, "host") |> List.first()
```
IvanIvanoff, line 30: [get_req_header/2](https://hexdocs.pm/plug/Plug.Conn.html#get_req_header/2) returns `[binary()]`, so it is strange to me that this worked before, maybe there was no proper validation for the events that include it.

lib/sanbase/accounts/user/user_registration_state.ex
```diff
@@ -0,0 +1,89 @@
...
+    result
+  end
+
+  defp emit_event(result, _user), do: result
+
+  defmodule StateMachine do
+    # In case the user is already registered, do nothing.
+    def forward("finished", _), do: :keep_state
+
+    def forward(_, "send_login_email"), do: {:next_state, "login_email_sent"}
```
tspenov, line 82: Is it possible to be in `wait_login_to_finish_registration` and then the new state is `send_login_email`, that way it will not go to `finished` ?

IvanIvanoff: Yes, but `wait_login_to_finish_registration` is put by `/auth/google`, which immediately redirects to sanbase, which uses `currentUser` and fires the `login` action. In the end, this state is being used for half a second.

PR #3689, opened 2023-02-14
Remove mining_pools_distribution query and code

IvanIvanoff:
## Checklist:


lib/sanbase/project/available_queries.ex
```diff
@@ -90,7 +90,7 @@ defmodule Sanbase.Project.AvailableQueries do
   end
 
   @mineable_specific_queries ["exchangeWallets", "allExchanges"]
-  @ethereum_specific_queries ["miningPoolsDistribution", "minersBalance", "gasUsed"]
```
IvanIvanoff, line 93: This is intentional. `minersBalance` query that is just redirecting to the getMetric API of the same name. It can be discouraged to be used by removing it from the list here.

tspenov (approved):

PR #3693, opened 2023-02-15
Add sql generator command

tspenov:
## Checklist:


lib/sanbase/discord/command_handler.ex
```diff
@@ -262,7 +262,10 @@ defmodule Sanbase.Discord.CommandHandler do
 
     prompt = String.trim(msg.content, "!ai")
 
-    case Sanbase.OpenAI.complete(prompt) do
+    case Sanbase.OpenAI.complete(
+           prompt,
+           to_string(msg.author.username <> msg.author.discriminator)
```
IvanIvanoff, line 267: <> is binary concatenation, so it already produces a binary, hence no need of calling `to_string/1`

PR #3699, opened 2023-02-21
Cryptocompare open interest data export

IvanIvanoff:
## Changes

Export open interest by using [this API](https://min-api.cryptocompare.com/documentation?key=Futures&cat=futures_v1_historical_hours)

It fetches historical data by providing arguments `to_ts` and `limit` and fetches the previous `limit` number of data points before the timestamp `to_ts`.

There Oban job defined can be provided with different set of arguments in order to:
- Scrape 2000 (max limit) data points and after finishing, schedule the job for the 2000 points before it.
- Run in *daily* mode - scrape 24-30 data points and **do not** schedule a new job for the days before that.

### Historical scraping
In order to implement historical scraping, we need to define 1 job for each market/instrument pair that has arguments:
```elixir
args = %{
  "market" => "binance",
  "instrument" => "ETH-USDT-VANILLA-PERPETUAL",
  "schedule_next_job" => true,
  "limit" => 2000 
}
```
This set of arguments will make it so that every job defines a new job where `to_ts` is the *earliest* timestamp it has scraped. This way it goes back in history, scraping all the data.

### Daily scraping
A cron job uses schedules a job every day with the following set of arguments:
```elixir
%{
args = %{
  "market" => "binance",
  "instrument" => "ETH-USDT-VANILLA-PERPETUAL",
  "schedule_next_job" => false,
  "limit" => 30
}
```
This way it fetches the previous day of data and does not schedule a job for the days before that. It has `limit=30` so it can get some missing data. There is a postgres table that follows the progress and will reject all timestamps that were previously already exported.

## Checklist:


lib/sanbase/billing/plan/standard_access_checker/api_access_checker.ex
```diff
@@ -10,23 +10,17 @@ defmodule Sanbase.Billing.Plan.ApiAccessChecker do
...
     realtime_data_cut_off_in_days: 30
   }
 
   @basic_plan_stats %{
-    api_calls_month: 100_000,
     historical_data_in_days: 2 * 365
   }
 
-  @pro_plan_stats %{
-    api_calls_month: 300_000
```
tspenov, line 24: these are not used, right ?

lib/sanbase_web/graphql/plugs/auth_plug.ex
```diff
@@ -403,6 +403,12 @@ defmodule SanbaseWeb.Graphql.AuthPlug do
     |> List.first()
     |> case do
       origin_url when is_binary(origin_url) ->
+        # WIP
+        # NOTE: Do not commit
+        if not String.contains?(origin_url, "santiment.net") do
```
tspenov, line 408: do we need this ?

lib/sanbase/cryptocompare/add_historical_jobs_worker.ex
```diff
@@ -0,0 +1,55 @@
...
+    Logger.info("[Cryptocompare Price.AddHistoricalJobsWorker] Finished adding historical jobs.")
+
+    :ok
+  end
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"type" => "schedule_historical_price_jobs"}}) do
+    Logger.info("[Cryptocompare Price.AddHistoricalJobsWorker] Start adding historical jobs.")
+
+    Sanbase.Cryptocompare.Price.CCCAGGPairData.schedule_previous_day_jobs()
```
tspenov, line 21: Shouldn't we scrape same day data ? Like every hour - fetch the newest data points ?

IvanIvanoff: I'll start with daily until we have a stable working exporter. After that, we can very easily change the cron job to run every hour.

I am also not sure how many API calls it will make if it runs every hour. At least during the initial scrape of historical data it will do a lot of calls.

lib/sanbase/billing/plan/standard_access_checker/api_access_checker.ex
IvanIvanoff: not used

PR #3721, opened 2023-03-17
Add commands for indexing and asking specific branch

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/discord/command_handler.ex
+++ b/lib/sanbase/discord/command_handler.ex
@@ -13,6 +13,8 @@ defmodule Sanbase.Discord.CommandHandler do
   @prefix "!q"
   @ai_prefix "!ai"
   @docs_prefix "!docs"
+  @index_prefix "!gi"
+  @ask_prefix "!ga"
   @mock_role_id 1
   @max_size 1800
   @ephemeral_message_flags 64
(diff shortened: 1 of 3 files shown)


lib/sanbase/discord/command_handler.ex
```diff
@@ -13,6 +13,8 @@ defmodule Sanbase.Discord.CommandHandler do
   @prefix "!q"
   @ai_prefix "!ai"
   @docs_prefix "!docs"
+  @index_prefix "!gi"
+  @ask_prefix "!ga"
```
IvanIvanoff, line 17: `!ai` and `!docs` are self-explanatory, but I see `!gi` and `!ga` is confusing.
Who's the main user of these commands? It would be better to be just `!git_index` and `!git_ask`?

tspenov: It is for internal use - the idea is when you write documentation in academy to check whether it answer and how some questions you want to be answered with it. The index probably won't stay long - because it can be done with trigger on commit. 
You are right they are not self-explanatory but for internal use I prefer the shorter names.

lib/sanbase/discord/command_handler.ex
```diff
@@ -366,6 +376,76 @@ defmodule Sanbase.Discord.CommandHandler do
...
+    end
+  end
+
+  def extract_branch_name_and_question(message) do
+    regex = ~r/^!ga\s+(?<branch>[a-zA-Z0-9\-\_\/\.]+)\s+(?<question>.+)/
+
+    case Regex.named_captures(regex, message) do
+      %{"branch" => branch, "question" => question} -> {:ok, branch, question}
+      _ -> {:error, "Invalid input format"}
+    end
```
IvanIvanoff, line 446: It would be simpler if you don't use regexes:
```elixir
[_, branch, question] = "!ga my_branch What is the answer to this question?" |> String.split(" ", parts: 3)
["!ga", "my_branch", "What is the answer to this question?"]
```

tspenov:
Addressed comments here: https://github.com/santiment/sanbase2/pull/3723

PR #3779, opened 2023-05-04
Add summaries to getTrendingWords GraphQL query

IvanIvanoff:
The summaries are generated from the documents that mention the trending words. The summary is a type of context that gives better insight why the word is trending. An AI generated summary is used instead of the raw documents so we don't expose the real documents and also to narrow down a lot of documents to a short text that can be read fast.

## Checklist:


lib/sanbase/social_data/trending_words.ex
```diff
@@ -218,20 +229,36 @@ defmodule Sanbase.SocialData.TrendingWords do
       t,
       word,
       any(project) AS project,
-      SUM(score) / #{length(sources)} AS total_score
+      SUM(score) / #{length(sources)} AS total_score,
+      groupArray(summary2) AS summaries
```
tspenov, line 233: Why groupArray again? Would just any work?

IvanIvanoff: The first `groupArray` (the inner one) gathers the data for the same word, source and time into a list.
This `groupArray` (the outer one) gathers all the summaries for the word and time (combining all sources).
If this is `any` we'll keep the summary for one of the sources only, at random.

tspenov (approved):

PR #3787, opened 2023-05-05
Restart ratelimiter on timeout

tspenov:
## Changes

That will fix the case where `{:error, :timeout}` comes for some reason from the RateLimiter gen server due to some congestion. Or will rule it out.

## Checklist:

### DIFF
--- a/lib/sanbase/discord/discord_consumer.ex
+++ b/lib/sanbase/discord/discord_consumer.ex
@@ -307,13 +307,35 @@ defmodule Sanbase.DiscordConsumer do
   end
 
   defp warm_up(msg) do
+    t1 = System.monotonic_time(:millisecond)
+
     Nostrum.Api.get_current_user()
     |> case do
       {:ok, _} ->
         log(msg, "WARM UP SUCCESS")
 
+      {:error, :timeout} ->
+        log(msg, "WARM UP TIMEOUT. Restart Ratelimiter", type: :error)
+        {:ok, pid} = restart_ratelimiter()
+        log(msg, "WARM UP TIMEOUT. Ratelimiter pid: #{pid}", type: :error)
+
       error ->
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -122,7 +122,18 @@ config :sasl, sasl_error_logger: false
...
+    :request_id,
+    :api_token,
+    :user_id,
+    :remote_ip,
+    :complexity,
+    :query,
+    :san_balance,
+    :shard,
+    :guild,
+    :channel
```
IvanIvanoff, line 135: Does adding this affect the rest of the logs?

lib/sanbase/discord/discord_consumer.ex
```diff
@@ -312,8 +312,25 @@ defmodule Sanbase.DiscordConsumer do
...
+        log(msg, "WARM UP TIMEOUT. Ratelimiter pid: #{pid}", type: :error)
+
       error ->
         log(msg, "WARM UP ERROR #{inspect(error)}", type: :error)
     end
   end
+
+  def restart_ratelimiter() do
+    supervisor_pid =
+      Process.whereis(Ratelimiter)
```
IvanIvanoff, line 327: You do not have an alias for this, shouldn't this be `Nostrum.Api.Ratelimiter` instead?

tspenov: It has: `@registered_name Ratelimiter`: https://github.com/Kraigie/nostrum/blob/master/lib/nostrum/api/ratelimiter.ex#L26

config/config.exs
tspenov: It does not, it just not shows them

tspenov: though removed them, because they are not very helpful

PR #3841, opened 2023-06-16
Make scrapers work when Clickhouse is down

IvanIvanoff:
## Checklist:


lib/sanbase/prices/validator/validator_node.ex
```diff
@@ -55,7 +63,7 @@ defmodule Sanbase.Price.Validator.Node do
 
         state = latest_prices |> Map.put(:number, number)
 
-        {:ok, state}
+        {:norepl, state}
```
tspenov, line 66: should be `noreply`

IvanIvanoff: fixed in #3842

PR #3899, opened 2023-08-25
Add sentiment ratios to getTrendingWords

IvanIvanoff:
## Changes

- Add `positiveSentimentRatio`, `negativeSentimentRatio`, and `neutralPositiveRatio` fields to the `getTrendingWords` API. These values show the % of documents that mention the given word that have positive/negative/neutral sentiment. The values are in the range [0,1] and add up to 1.
  - As of now, the ratios are available **only** for words that refer to projects.
- Conditionally choose the source that constitutes `ALL` so the API does not break on stage.

```graphql
{
  getTrendingWords(size: 10, from: "utc_now-1d", to: "utc_now", interval: "12h", wordTypeFilter: PROJECT) {
    datetime
    topWords {
      word
      project {
        slug
      }
      score
      positiveSentimentRatio
      negativeSentimentRatio
      neutralSentimentRatio
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/social_data/trending_words.ex
+++ b/lib/sanbase/social_data/trending_words.ex
@@ -50,16 +50,12 @@ defmodule Sanbase.SocialData.TrendingWords do
           datetme: DateTime.t(),
           position: position
         }
-
-  @default_source "reddit,telegram,twitter_crypto"
+  @table "trending_words_docs_v3"
 
   # When calculating the trending now words fetch the data for the last
   # N hours to ensure that there is some data and we're not in the middle
   # of computing the latest data
-
   @hours_back_ensure_has_data 3
-
-  @table "trending_words_docs_v2"
 
   @spec get_trending_words(
           DateTime.t(),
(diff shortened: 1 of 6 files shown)


volynvlad (approved):

test/sanbase/queries/query_resolve_parameters_test.exs
```diff
@@ -0,0 +1,68 @@
+defmodule Sanbase.Queries.ResolveParametersTest do
```
tspenov, line 1: This test seems to be from another branch

lib/sanbase/social_data/trending_words.ex
```diff
@@ -315,4 +320,11 @@ defmodule Sanbase.SocialData.TrendingWords do
     |> Sanbase.Clickhouse.Query.put_sql(sql)
     |> Sanbase.Clickhouse.Query.add_parameter(:ticker_slug, ticker_slug)
   end
+
+  defp default_source do
+    case Sanbase.Utils.Config.module_get(Sanbase, :deployment_env) do
+      "prod" -> "reddit,telegram,twitter_crypto"
+      _ -> "4chan,bitcointalk,reddit"
```
tspenov, line 327: Why different sources for prod and other envs ?

IvanIvanoff: This is a deployment env (prod, stage). On stage the data currently includes other sources, so if we leave it as-is, it will return an empty result.

test/sanbase/queries/query_resolve_parameters_test.exs
IvanIvanoff: Fixed.

tspenov (approved):

PR #3949, opened 2023-10-23
Migrate balances labels

tspenov:
## Changes

Tested manually 

```sql
SELECT
    address,
    balance,
    label_id
FROM
(
    SELECT
        address,
        argMax(balance, dt) / 18 AS balance
    FROM eth_balances_realtime
    PREWHERE addressType = 'normal'
    GROUP BY address
)
GLOBAL ANY INNER JOIN
(
    SELECT
        address,
        label_id
    FROM current_label_addresses
    WHERE (blockchain = 'ethereum') AND (label_id IN (
        SELECT label_id
        FROM label_metadata
        WHERE key IN ('whale_usd_balance')
    ))
) USING (address)
WHERE (address = '0x00000000219ab540356cbb839cbe05303d7705fa') AND (balance > 1e-10)
ORDER BY balance DESC
LIMIT 1

Query id: ee31cce9-c2f0-4c96-8aaa-a797162363af

┌─address────────────────────────────────────┬───────────────balance─┬────────────label_id─┐
│ 0x00000000219ab540356cbb839cbe05303d7705fa │ 1.7506586258333243e24 │ 2798264455269738766 │
└────────────────────────────────────────────┴───────────────────────┴─────────────────────┘

1 rows in set. Elapsed: 0.541 sec. Processed 7.03 million rows, 344.87 MB (13.00 million rows/s., 637.46 MB/s.)

clickhouse-facing-1.clickhouse-facing-headless.default.svc.cluster.local :) select * from label_metadata where label_id = 2798264455269738766;

SELECT *
FROM label_metadata
WHERE label_id = 2798264455269738766

Query id: ac933e07-34d2-4b61-a5ec-6a2257fa9b60

┌─owner_id─┬─owner─────┬─key───────────────┬─value─┬─version─┬──────────created_at─┬─latest─┬─description─┬─asset_name─┬─display_name────────────────┬─has_balances─┬─search_text─┬─deprecated─┐
│        0 │ santiment │ whale_usd_balance │       │ v1      │ 2023-08-17 11:41:41 │      1 │             │ ethereum   │ whale_usd_balance(ethereum) │            1 │ ᴺᵁᴸᴸ        │          0 │
└──────────┴───────────┴───────────────────┴───────┴─────────┴─────────────────────┴────────┴─────────────┴────────────┴─────────────────────────────┴──────────────┴─────────────┴────────────┘

```

## Checklist:


lib/sanbase/balances/balance_sql_query.ex
```diff
@@ -420,8 +420,11 @@ defmodule Sanbase.Balance.SqlQuery do
     GLOBAL ANY INNER JOIN
     (
       SELECT address
-      FROM blockchain_address_labels
-      PREWHERE blockchain = {{#{blockchain_key}}} AND label IN ({{#{labels_key}}})})
```
tspenov, line 424: Not sure if this has ever worked. `({{#{labels_key}}})})` - this should produce syntax error

IvanIvanoff: There's an extra pair of `})`, so this should have failed always.
You can add some real examples in the `Sanbase.RunExamples` module so we can track for regressions in the SQL 
 in the future

alexander-gr:
The query looks fine:) I'm wondering weather it's suitable to use only key: 
```
SELECT label_id
FROM label_metadata
WHERE key IN ('whale_usd_balance')
```
Usually the safest way is to use label fqn (e.g. `santiment/whale_usd_balance(ethereum):v1`). But is some cases using key could be more suitable.
Could you provide what is this query serves for and where that key comes from?

tspenov:
> The query looks fine:) I'm wondering weather it's suitable to use only key:
> 
> ```
> SELECT label_id
> FROM label_metadata
> WHERE key IN ('whale_usd_balance')
> ```
> 
> Usually the safest way is to use label fqn (e.g. `santiment/whale_usd_balance(ethereum):v1`). But is some cases using key could be more suitable. Could you provide what is this query serves for and where that key comes from?

It is top addresses (by balance) filtered by labels (in this case label keys) but it could be by label fqns.
https://github.com/santiment/sanbase2/blob/master/lib/sanbase/blockchain_address/list/blockchain_address_list_selector.ex#L84

alexander-gr (approved):

lib/sanbase/balances/balance_sql_query.ex
tspenov: Added

PR #4010, opened 2023-12-01
Add sentiment bal vc v2 metrics

volynvlad:
## Changes

Add sentiment balance and sentiment volume consumed v2 metrics (calculated with CH instead of ES)

## Checklist:


test/sanbase/billing/metric_access_level_test.exs
```diff
@@ -1434,6 +1434,56 @@ defmodule Sanbase.Billing.MetricAccessLevelTest do
         "total_trustlines_count",
         "dex_volume_in_xrp_5m",
         "dex_volume_in_usd_5m",
+        # sentiment balance v2 metrics
```
apozdeyev, line 1437: seems forgot to add sentiment_volume_consumed_5m

lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
```diff
@@ -1735,5 +1735,719 @@
...
+    },
+    "aggregation": "avg",
+    "min_interval": "1h",
+    "table": "intraday_metrics",
+    "has_incomplete_data": true,
+    "data_type": "timeseries"
+  },
+  {
+    "human_readable_name": "Sentiment Volume Consumed Daily 4chan v2",
+    "name": "weighted_sentiment_4chan_v2_1d",
```
IvanIvanoff, line 2183: it would be better to always have the `_v2` as a prefix, so we'll have:
- `sentiment_weighted_4chan_1d` (old)
- `sentiment_weighted_4chan_1d_v2`
so, if you do a text search for `sentiment_weighted_4chan_1d` you'll be able to find both the old and the new metric, i.e. not infixing the version between some of the other arguments.

volynvlad:
Close and open new PR https://github.com/santiment/sanbase2/pull/4260

PR #4025, opened 2023-12-11
Add purchasing power parity pricing for Turkey

tspenov:
## Changes

Add purchasing power parity pricing (PPP) API.

```graphql
    {
      pppSettings {
        isEligibleForPPP
        country
        percentOff
        plans {
          id
          name
          interval
          amount
          product {
            name
          }
        }
      }
    }
```

## Checklist:


lib/sanbase/billing/plan.ex
```diff
@@ -16,9 +16,34 @@ defmodule Sanbase.Billing.Plan do
...
+
+  def ppp_plans_for_country(nil), do: []
+
+  def ppp_plans_for_country(country_code) do
+    case Map.get(@country_ppp_plan_map, country_code) do
+      nil -> []
+      key -> @ppp_plans[key] |> by_ids()
+    end
+  end
+
```
IvanIvanoff, line 46: What if we put this into `Billing.Plan.PurchasingPower` module? We can avoid using `ppp` naming here.

lib/sanbase/billing/plan.ex
```diff
@@ -35,6 +60,10 @@ defmodule Sanbase.Billing.Plan do
     field(:is_deprecated, :boolean, default: false)
     # plans that customers can't subscribe on their own
     field(:is_private, :boolean)
+
+    # is plan for parity purchasing power
+    field(:is_ppp, :boolean, default: false)
```
IvanIvanoff, line 65:
```suggestion
    @desc ~s"""
    Mark if this is a parity purchasing power power.
    In some countries our pricing is reduced in order to better match the purchasing power of the population.
    """
    field(:is_ppp, :boolean, default: false)
```

You can use `@desc` attribute so this comment will be seen while exploring the API via GraphiQL, making the doc visible outside of our codebase.

lib/sanbase/billing/plan.ex
tspenov: extracted into own module

PR #4038, opened 2023-12-18
Rework Alert internals

IvanIvanoff:
## Changes

- Make each alert `evaluate/2` function return ok or error tuple.
- Handle a special `evaluate/2` error type that has the form `{:error, {:disable_alert, reason}}` that is used when an alert is using an already deprecated metric, or somehow else ended up with a metric that is not valid (mistyped, unsupported, deprecated). Currently these alerts are evaluated every 5 minutes, non-stop.

## Checklist:


IvanIvanoff:
test

PR #4053, opened 2024-01-12
Update Ecto from 3.9 to 3.11

IvanIvanoff:
## Changes

Updating it requries an update to the Clickhousex driver. The reason is
that Ecto changed how the scheme argument is handled, so now we need to
explicitly convert it from string to atom before passing it to
Mint.HTTP.connect/4

## Checklist:


mix.exs
```diff
@@ -84,9 +84,9 @@ defmodule Sanbase.Mixfile do
       {:earmark, "~> 1.2"},
       {:ecto_enum, "~> 1.4"},
       {:ecto_psql_extras, "~> 0.3"},
-      {:ecto_sql, "== 3.9.0"},
+      {:ecto_sql, "~> 3.10"},
       # Our config breaks with the newest ecto version
-      {:ecto, "== 3.9.1"},
+      {:ecto, "~> 3.10"},
```
tspenov, line 89: it is actually to 3.10 but in description says 3.11

tspenov (approved):

PR #4061, opened 2024-01-19
Fix warnings. Fix getMostRecent

IvanIvanoff:
## Changes

- Fix warnings
- Fix ordering of recent entities

## Checklist:

### DIFF
--- a/lib/sanbase/entity/entity.ex
+++ b/lib/sanbase/entity/entity.ex
@@ -331,9 +331,17 @@ defmodule Sanbase.Entity do
           # In all cases the fields are the same except for insights. When
           # fetching user own insights, some of them might be drafts so they
           # won't have :published_at field and then :inserted_at shall be used.
-          Map.get(entity, creation_time_field) || Map.get(entity, creation_time_field_backup)
+          creation_time =
+            Map.get(entity, creation_time_field) || Map.get(entity, creation_time_field_backup)
+
+          creation_time_unix =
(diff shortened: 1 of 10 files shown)


lib/sanbase/entity/entity.ex
```diff
@@ -331,9 +331,17 @@ defmodule Sanbase.Entity do
...
-          Map.get(entity, creation_time_field) || Map.get(entity, creation_time_field_backup)
+          creation_time =
+            Map.get(entity, creation_time_field) || Map.get(entity, creation_time_field_backup)
+
+          creation_time_unix =
+            DateTime.from_naive!(creation_time, "Etc/UTC") |> DateTime.to_unix()
+
+          # Transform to unix timestamp so we can compare the tuples. Add the id as the secon
+          # element so in case of conflicts, we put the entity with higher id first (created later)
+          {creation_time_unix, Map.get(entity, :id)}
```
tspenov, line 342: is that correct ? Now returns tuple, before `Map.get(entity, creation_time_field) || Map.get(entity, creation_time_field_backup)`

IvanIvanoff: This does not return anything. This only controls how the list is sorted, the function argument `result` is what is actually returned.
This was breaking some tests where a few entities are inserted with the same timestamp. In such case we should also order them in descending order of their ID, as the creation timestamp is the same.

tspenov (approved):

PR #4062, opened 2024-01-22
Add pagination

tspenov:
## Changes

* Adding pagination to generic admin - both index and filter table
* Improve search functionality - you can search `field=part_of_field_value`

image
image

## Checklist:

### DIFF
--- a/lib/sanbase_web/controllers/generic_controller.ex
+++ b/lib/sanbase_web/controllers/generic_controller.ex
@@ -147,9 +147,11 @@ defmodule SanbaseWeb.GenericController do
     fetched_rows =
       case rows do
         nil ->
+          sort_field = sort_field(module)
+
           Repo.all(
             from(m in module,
-              order_by: [desc: m.id],
+              order_by: [desc: field(m, ^sort_field)],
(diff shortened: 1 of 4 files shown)


lib/sanbase_web/controllers/generic_controller.ex
```diff
@@ -108,13 +130,48 @@ defmodule SanbaseWeb.GenericController do
...
+      case rows do
+        nil ->
+          Repo.all(
+            from(m in module,
+              order_by: [desc: m.id],
+              preload: ^preloads,
+              limit: ^page_size,
+              offset: ^offset
+            )
+          )
```
IvanIvanoff, line 157: Not all tables have id, for example `UserFollower`.
I suggest you move this to a separate function, where you do something like this:
```elixir
fields = module.__schema__(:fields)

cond do
  :id in fields -> # ...
  :inserted_at in fields -> # ...
  true -> raise("Ecto Module #{module} does not have :id or :inserted_at fields"
  end
  ```

tspenov: Added

PR #4072, opened 2024-01-25
Add user affiliate details API

tspenov:
## Changes

Check if user has already submitted the form

```graphql
{
  currentUser{
    areAffiliateDetailsSubmitted
   }
}
```

```graphql
mutation {
  addUserAffiliateDetails(telegramHandle: "@exampleHandle", marketingChannels: "Social Media")
}
```

## Checklist:


lib/sanbase/affiliate/user_affiliate_details.ex
```diff
@@ -0,0 +1,36 @@
...
+    belongs_to(:user, User)
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(user_affiliate_details, attrs) do
+    user_affiliate_details
+    |> cast(attrs, [:telegram_handle, :marketing_channels, :user_id])
+    |> validate_required([:telegram_handle, :user_id])
```
IvanIvanoff, line 20: Why do we require a telegram handle and what is it? If it's required, you should also add `null: false` to the migration.


The current user telegram data can be obtained via the user settings. This way of adding telegram handles also verifies that the account belongs to you?

tspenov: telegram handle is a string used for contacting the user for the payout - I don't think there is need to verify it.
You can check the form in the mockup.
image

livebooks/user_affiliate_details_test.livemd
```diff
@@ -0,0 +1,56 @@
...
+    # Execute the query with the context
+    case Absinthe.run(query, SanbaseWeb.Graphql.Schema, opts) do
+      {:ok, result} ->
+        {:ok, result}
+
+      {:error, errors} ->
+        {:error, errors}
+    end
+  end
+end
```
IvanIvanoff, line 27: I suggest the following:
- Move this to a compiled module, so we can reuse it in other livebooks;
- Improve the way the context is constructed by calling `AuthPlug` and `ContextPlug` -- in some queries we look at fields other than the `current_user` like `product_id`, `plan`, etc. You can search for `AuthPlug.call` and you'll see such examples in the tests. It might require building a connection though.

lib/sanbase_web/graphql/schema/queries/user_affiliate_details_queries.ex
```diff
@@ -0,0 +1,29 @@
...
+  alias SanbaseWeb.Graphql.Middlewares.UserAuth
+  alias SanbaseWeb.Graphql.Resolvers.UserAffiliateDetailsResolver
+
+  object :user_affiliate_details_queries do
+    field :are_user_affiliate_datails_submitted, :boolean do
+      meta(access: :free)
+
+      middleware(UserAuth)
+
+      resolve(&UserAffiliateDetailsResolver.are_user_affiliate_datails_submitted/3)
```
IvanIvanoff, line 13: Shouldn't this be a field of `currentUser`?
```graphql
{
  currentUser{
    areAffiliateDetailsSubmitted
   }
}
```

vekkele (approved):

livebooks/user_affiliate_details_test.livemd
tspenov: Extracted and passed through AuthPlug and  ContextPlug

PR #4075, opened 2024-01-26
Fix the unique index for dashboard_id in the votes table

IvanIvanoff:
## Changes

The unique index must be on [:dashboard_id, :user_id] and not on
:dashboard_id only. Each user can have a single record for an entity.
Multiple votes are recored by bumping the count column.

## Checklist:

### DIFF
--- a/priv/repo/migrations/20240126133441_fix_dashboard_votes_unique_index.exs
+++ b/priv/repo/migrations/20240126133441_fix_dashboard_votes_unique_index.exs
@@ -1,0 +1,8 @@
+defmodule Sanbase.Repo.Migrations.FixDashboardVotesUniqueIndex do
+  use Ecto.Migration
+
+  def change do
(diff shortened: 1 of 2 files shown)


priv/repo/migrations/20240126133441_fix_dashboard_votes_unique_index.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Sanbase.Repo.Migrations.FixDashboardVotesUniqueIndex do
+  use Ecto.Migration
+
+  def up do
```
tspenov, line 4: you can put it in change since it is reversible

tspenov (approved):

priv/repo/migrations/20240126133441_fix_dashboard_votes_unique_index.exs
IvanIvanoff: done

PR #4093, opened 2024-02-01
Support ecosystem-aggregated metrics

IvanIvanoff:
## Changes

We need to be able to do a few things:
- Get a list of ecosystems (similar to how we get projects by `allProjects`)
- Get Github activity metrics computed for each ecosystem

The metrics that can be used in these APIs are:
- `ecosystem_dev_activity`
- `ecosystem_github_activity`
- `ecosystem_dev_activity_contributors_count`
- `ecosystem_github_activity_contributors_count`

These metrics are computed on top of all the github organizations that are included in an ecosystem.

These two things cannot be achieved with our existing APIs. That's why we add the following API:
```graphql
{
  getEcosystems {
     name
     projects { id slug }
     aggregatedTimeseriesData(metric: "dev_activity" from: "utc_now-30d" to: "utc_now")
     timeseriesData(metric: "dev_activity" from: "utc_now-365d" to: "utc_now" interval: "7d"){
       datetime
       value
     }
   }
}
```
The API also allows fetching a subset of the ecosystems:
```graphql
{
  getEcosystems(ecosystems: ["ethereum", "energi"]){
    name
    aggregatedTimeseriesData(metric: "ecosystem_dev_activity" from: "utc_now-7d" to: "utc_now")
  }
}
```

Why adding `timeseriesData` to the `getEcosystems` instead of using `getMetric`?
- We'll support only a limited number of metrics, starting with 4 dev activity related metrics
- `getMetric` will need 1 call per ecosystem, which is not pratical
- `getMetric` already has a `timeseriesDataPerSlug` which won't fit the `ecosystem` identifier, so we need to extend it with `timeseriesDataPerEcosystem` or introduce a more generic `timeseriesDataPerIdentifier`. We can do that in the future after we see what needs arise when we use this API.

### Filter ecosystems

One possible flow would be the following:
1. Use aggregatedTimeseriesData to get the top 10 ecosystems by dev activity
2. Use these 10 ecosystems (out of 125) and get the timeseries data only for them. Showing 120+ charts will be not comprehensible.

To do this:
```graphql
{
  getEcosystems(ecosystems: ["ethereum", "bitcoin", ...]) {
     ecosystem
     timeseriesData(...){...}
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/clickhouse_repo.ex
+++ b/lib/sanbase/clickhouse_repo.ex
@@ -123,6 +123,7 @@ defmodule Sanbase.ClickhouseRepo do
     sanitized_query = sanitize_query(query)
 
     maybe_store_executed_clickhouse_sql(sanitized_query, ordered_params)
+    maybe_print_interpolated_query(sanitized_query, ordered_params)
 
     case __MODULE__.query(sanitized_query, ordered_params) do
       {:ok, result} ->
@@ -141,6 +142,7 @@ defmodule Sanbase.ClickhouseRepo do
     sanitized_query = sanitize_query(query)
 
     maybe_store_executed_clickhouse_sql(sanitized_query, ordered_params)
+    maybe_print_interpolated_query(sanitized_query, ordered_params)
 
     case __MODULE__.query(sanitized_query, ordered_params) do
       {:ok, result} ->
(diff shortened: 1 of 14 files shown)


lib/sanbase/clickhouse_repo.ex
```diff
@@ -286,19 +288,46 @@ defmodule Sanbase.ClickhouseRepo do
...
 
       :ok
     end
   rescue
     _ -> :ok
   end
+
+  case Mix.env() do
+    :dev ->
+      defp maybe_print_interpolated_query(query, params) do
```
tspenov, line 302: Nice addition

PR #4115, opened 2024-02-17
Add alpine js. Add sidebar in the layout

tspenov:
## Checklist:


assets/js/app.js
```diff
@@ -3,6 +3,10 @@ import "phoenix_html"
 import {Socket} from "phoenix"
 import {LiveSocket} from "phoenix_live_view"
 import topbar from "../vendor/topbar"
+import Alpine from 'alpinejs'
+
+window.Alpine = Alpine
+Alpine.start()
```
IvanIvanoff, line 9: I have not used Alpine in a Phoenix project, but I've checked some tutorials.

They mention that this won't work well with LiveView as everytime LV updates the DOM, Alpine will lose track of the element it is attached to.

You need to add this to the livesocket definition:
```js
let liveSocket = new LiveSocket("/live", Socket, {
  // ...
  dom: {
    onBeforeElUpdated(from, to) {
      if (from._x_dataStack) {
        window.Alpine.clone(from, to)
      }
    }
  }
})
```

References:
- https://medium.com/@nnissen/petal-to-the-metal-d5d0b18d2af
- https://elixirforum.com/t/proper-alpinejs-integration-with-liveview/53409
- https://fullstackphoenix.com/tutorials/combine-phoenix-liveview-with-alpine-js

tspenov: Nice, I'll check this out.

lib/sanbase_web/components/layouts/admin2.html.heex
```diff
@@ -1,7 +1,38 @@
...
+      <.icon name="hero-home" class="mr-2" />
+      Admin Home
+    <% end %>
+    <button @click="isOpen = !isOpen" class="top-0 left-0">
+      <SanbaseWeb.CoreComponents.icon name="hero-bars-3" />
+    </button>
+  </nav>
+  
+  <div class="flex flex-1 min-h-0">
+    <div id="sidebar" phx-update="ignore" class="flex-shrink-0 w-64 overflow-y-auto bg-gray-900 transition duration-300 transform lg:static" x-show="isOpen" x-transition:enter="transition ease-out duration-300" x-transition:enter-start="-translate-x-full" x-transition:enter-end="translate-x-0" x-transition:leave="transition ease-in duration-300" x-transition:leave-start="translate-x-0" x-transition:leave-end="-translate-x-full">
```
IvanIvanoff, line 13: This does not look pretty -- we're missing the `heex` filetype in our `.formatter.exs`. If you add it, then it is formatted like this, which makes it way more readable. It also formats a few other files:
image

lib/sanbase_web/components/layouts/admin2.html.heex
```diff
@@ -1,7 +1,38 @@
...
+      <div class="flex flex-col items-center justify-center mt-8">
+        <div class="flex items-center">
+          <img src="/images/emails/san-logo.svg" />
+          <span class="mx-2 text-2xl font-semibold text-white">Sanbase Admin</span>
+        </div>
+        <%= live_render(@conn, SanbaseWeb.LiveSearch) %>
+      </div>
+      <%= for {name, path} <- @routes do %>
```
IvanIvanoff, line 21: The preferred way is to use the special `:for` attribute like this:
```elixir
<nav class="mt-1" :for={{name, path} <- @routes>
...
</nav>
```

On a side note, isn't it more proper to have 1 enclosing `nav` tag and use div/a/ul/etc. inside it? Instead of having multiple `nav` tags.

lib/sanbase_web/components/layouts/admin2.html.heex
```diff
@@ -1,7 +1,38 @@
...
+        <nav class="mt-1">
+          <span>
+            <%= link(name,
+              to: path,
+              class:
+                "flex items-center px-4 mt-1 duration-200 border-l-4 border-gray-900 text-gray-500 hover:bg-gray-600 hover:bg-opacity-25 hover:text-gray-100"
+            ) %>
```
IvanIvanoff, line 28: Why interpolate instead of using the `<.link> ... </.link>` component?

lib/sanbase_web/components/layouts/admin2.html.heex
```diff
@@ -1,7 +1,38 @@
-<nav class="bg-blue-500 p-4">
-  <%= link to: Routes.generic_path(@conn, :home), class: "text-white hover:text-gray-300" do %>
-    <.icon name="hero-home" class="mr-2" />
-    Admin Home
-  <% end %>
-</nav>
-<%= @inner_content %>
+<div x-data="{ isOpen: true }" class="min-h-screen flex flex-col">
+  <nav class="bg-blue-500 p-4">
+    <%= link to: Routes.generic_path(@conn, :home), class: "text-white hover:text-gray-300" do %>
```
IvanIvanoff, line 3: We should migrate to using [Verified Routes](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html) exclusively. This should be: `<.link href={~p"/admin2/"}`. We added support for them with the Phoenix 1.7 update: https://github.com/santiment/sanbase2/pull/3956/files#diff-0581ce56cbfe498898aa2685ff6f7541f1f11558982a52cfb966a3971b81db66R133
If they don't work somewhere, then probably some part of the migration was missed.

PR #4117, opened 2024-02-19
Improve table search

tspenov:
## Changes

image
image

Component borrowed from: https://flowbite.com/docs/forms/search-input/#search-with-dropdown

## Checklist:


lib/sanbase_web/components/table.ex
```diff
@@ -394,34 +389,100 @@ defmodule SanbaseWeb.TableComponent do
...
+                viewBox="0 0 20 20"
+              >
+                <path
+                  stroke="currentColor"
+                  stroke-linecap="round"
+                  stroke-linejoin="round"
+                  stroke-width="2"
+                  d="m19 19-4-4m0-7A7 7 0 1 1 1 8a7 7 0 0 1 14 0Z"
+                />
+              </svg>
```
IvanIvanoff, line 481: For simplicity, we can replace custom SVGs with a call to `<.icon name="hero-magnifying-glass" />`
```suggestion
              <.icon name="hero-magnifying-glass" />
```

lib/sanbase_web/components/table.ex
```diff
@@ -394,34 +389,100 @@ defmodule SanbaseWeb.TableComponent do
...
+                viewBox="0 0 10 6"
+              >
+                <path
+                  stroke="currentColor"
+                  stroke-linecap="round"
+                  stroke-linejoin="round"
+                  stroke-width="2"
+                  d="m1 1 4 4 4-4"
+                />
+              </svg>
```
IvanIvanoff, line 427: And this can be replaced with `<.icon name="hero-chevron-down" />`.
You should try to see if the size/position does not break though.

lib/sanbase_web/components/table.ex
```diff
@@ -394,34 +389,100 @@ defmodule SanbaseWeb.TableComponent do
     """
   end
 
-  attr(:placeholder, :string, default: "Search...")
-  attr(:resource, :string, required: true)
-  attr(:search_value, :string, required: true)
-  attr(:text_input_title, :string, default: "")
-
   def search(assigns) do
```
IvanIvanoff, line 392: Maybe add a comment with link to https://flowbite.com/docs/forms/search-input/, so we can document where this came from.

PR #4131, opened 2024-02-23
Change style of the index table in admin

tspenov:
## Changes

image

## Checklist:


IvanIvanoff:
I suggest a few other improvements:
- Replace `edit`link with a button
- Add also `show` in actions in all cases and don't force the user to click on the underlined id (like in projects).

PR #4151, opened 2024-03-07
extend wallet alert

IvanIvanoff:
## Changes

- **Make use of multiple addresses in wallet_movement test**
- **Add maybe_combine_balances function to wallet trigger module**

To use the combined balance of multiple addresses, pass a list of addresses to the `"address"` param, as well as `"use_combined_balance": true` flag.

```json
{
  "type": "wallet_movement",
  "channel": "telegram",
  "target": { "address": ["px1234", "px5678"], "use_combined_balance": true },
  "selector": { "infrastructure": "BTC" },
  "operation": { "above": 1000 }
}
```

## Checklist:


docs/user-triggers-api.md
```diff
@@ -210,6 +210,29 @@ these are the tickers of the projects. Supported currencies are: `XRP`, `BTC`, `
...
+
+```json
+// The bitcoin balances of any of the addresses px1234 and px5678 is above 1000.
+// If both addresses have balance over 1000, two sepparate alert messages are received.
+{
+  "type": "wallet_movement",
+  "channel": "telegram",
+  "target": { "address": ["px1234", "px5678"], "use_combined_balance": false },
+  "selector": { "infrastructure": "BTC" },
+  "operation": { "above": 1000 }
```
tspenov, line 232: shouldn't we do it more generic and put it like filter in the `operation` like so:
`"operation": { "above": 1000, "type": "any|sum|all" }` - any of the targets is above 1000, the sum of all the targets is above 1000, all of the targets are above 1000 (not needed in this case)

IvanIvanoff: I can argue that the flag fits both the target and the operation, although I slightly more lean towards the target. This flag mutates the target's value before it goes to the operation, i.e. it is applied in the `get_data` (target + metric) phase and not in the `evaluate` (operation) phase.

tspenov (approved):

PR #4164, opened 2024-03-19
Start removing APIs

IvanIvanoff:
## Changes
Remove old, deprecated in favor of getMetric, APIs:
- burnRate
- transactionVolume
- mvrvRatio
- socialVolume
- socialDominance
- topicSearch
- historyTwitterData
- twitterData
- dailyActiveAddresses
- exchangeWallets
- githubAvailableRepos
- averageTokenAgeConsumedInDays
- networkGrowth
- nvtRatio
- realizedValue
- tokenCirculation
- tokenVelocity
- topSocialGainersLosers
- socialGainersLosersStatus

## Checklist:


tspenov:
Maybe we can return an error message with suggestion to use the `getMetric` equivalent if any of of the removed metrics is being accessed. What do you think ?

IvanIvanoff:
@tspenov if we want to return errors it means that we need to leave the queries to be available.
They don't have real usage (noone is actively using them) and they don't have documentation, so there's not really a way someone will start using them.

PR #4185, opened 2024-04-01
add get metric rounding

IvanIvanoff:
## Changes

- **Add rounding to Sanbase.Metric**
  - Add rounding to 6 decimals as it turns out we have metrics stored with up to 18 decimals
- **Unsupported selectors in getMetric do not cause Internal Server Error**
  - Querying `dev_activity_1d` while providing `owner`/`label` causes an internal server error. This PR reworks it to have no error. We should improve the `available_selectors` and reject anything that is not fit for this metric.
- **Add required_selectors to some exchange label metrics**

## Checklist:


IvanIvanoff:
@tspenov tests added

tspenov (approved):

PR #4216, opened 2024-04-12
Add /forms/suggest_github_organizations LiveView

IvanIvanoff:
## Changes

image
image
image
image
image

## Checklist:


tspenov:
In terms of UX - should we notify the user that submitted a suggestion with the moderated status?

IvanIvanoff:
@tspenov we don't store the user that made the suggestion, as the form does not require any authentication

I was thinking about exposing the suggestions for a given coin when you select that coin, so you can avoid re-suggesting something that was already suggested.

PR #4231, opened 2024-04-22
Add selfResetApiRateLimits query

IvanIvanoff:
## Changes

Allow users to reset the API call limits once every 90 days by themselves.
This will help in cases where the user has made a mistake (infinite loop), had too many tests during development or they just need a higher plan. Instead of waiting for santiment support that will reset your limits, you can do that once per 90 days yourself.

The `selfApiRateLimitsResetAt` field is exposed in the user settings which points to the last datetime when the user has reset the limits:
```graphql
{
  currentUser{
    settings{
      selfApiRateLimitsResetAt
      canSelfResetApiRateLimits
      canSelfResetApiRateLimitsAt
    }
  }
}
```
```json
{
  "data": {
    "currentUser": {
      "settings": {
        "canSelfResetApiRateLimits": false,
        "canSelfResetApiRateLimitsAt": "2024-07-21T12:24:55Z",
        "selfApiRateLimitsResetAt": "2024-04-22T12:24:55Z"
      }
    }
  }
}
```

### API for self-resetting the limits

Success:
```graphql
mutation {
  selfResetApiRateLimits {
    id
    settings {
      selfApiRateLimitsResetAt
    }
  }
}
```
```json
{
  "data": {
    "selfResetApiRateLimits": {
      "id": "7",
      "settings": {
        "selfApiRateLimitsResetAt": "2024-04-22T12:24:55Z"
      }
    }
  }
}
```

---

Fail because the last reset was less than 90 days ago:
```graphql
mutation {
  selfResetApiRateLimits {
    id
    settings {
      selfApiRateLimitsResetAt
    }
  }
}
```
```json
{
  "data": {
    "selfResetApiRateLimits": null
  },
  "errors": [
    {
      "message": "Cannot self reset the API calls rate limits.\nThe last reset was less than 90 days ago on 2024-04-22 12:24:55Z\n",
      "path": [
        "selfResetApiRateLimits"
      ],
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

## Checklist:

### DIFF
--- a/lib/sanbase_web/graphql/schema/queries/user_queries.ex
+++ b/lib/sanbase_web/graphql/schema/queries/user_queries.ex
@@ -14,6 +14,7 @@ defmodule SanbaseWeb.Graphql.Schema.UserQueries do
   }
 
   alias SanbaseWeb.Graphql.Middlewares.JWTAuth
+  alias SanbaseWeb.Graphql.Middlewares.UserAuth
 
   object :user_queries do
     @desc "Returns the user currently logged in."
@@ -201,6 +202,11 @@ defmodule SanbaseWeb.Graphql.Schema.UserQueries do
 
       middleware(JWTAuth)
       resolve(&UserResolver.change_avatar/3)
+    end
+
+    field :self_reset_api_rate_limits, :user do
+      middleware(UserAuth)
+      resolve(&UserResolver.self_reset_api_rate_limits/3)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


lib/sanbase_web/graphql/schema/queries/user_queries.ex
```diff
@@ -202,5 +203,10 @@ defmodule SanbaseWeb.Graphql.Schema.UserQueries do
       middleware(JWTAuth)
       resolve(&UserResolver.change_avatar/3)
     end
+
+    field :self_reset_api_rate_limits, :user do
+      middleware(UserAuth)
+      resolve(&UserResolver.self_reset_api_rate_limits/3)
+    end
```
tspenov, line 210: Maybe also expose `can_self_reset_api_rate_limits` ? That way the frontend can call this first and place a disabled btn and hint why user can't reset and when user will be able to reset.

IvanIvanoff: done

PR #4274, opened 2024-05-22
Auto refresh queries

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/queries/refresh/worker.ex
+++ b/lib/sanbase/queries/refresh/worker.ex
@@ -1,0 +1,25 @@
+defmodule Sanbase.Queries.RefreshWorker do
+  use Oban.Worker,
+    queue: :refresh_queries,
+    max_attempts: 3
+
+  alias Sanbase.Queries.Refresh
+
+  @oban_conf_name :oban_web
+  @one_day 24 * 60 * 60
+
+  @impl Oban.Worker
+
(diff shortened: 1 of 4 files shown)


lib/sanbase/queries/refresh/worker.ex
```diff
@@ -0,0 +1,25 @@
...
+    next_refresh_in_seconds = args["next_refresh_in_seconds"] || @one_day
+    data = new(args, schedule_in: next_refresh_in_seconds)
+    Oban.insert!(@oban_conf_name, data)
+
+    Refresh.refresh_query(query_id, user_id)
+  end
+
+  def perform(%Oban.Job{args: %{"user_id" => user_id, "query_id" => query_id}}) do
+    Refresh.refresh_query(query_id, user_id)
+  end
```
IvanIvanoff, line 24: Maybe something like this better conveys the intent:
```suggestion
  def perform(%Oban.Job{args: %{"user_id" => user_id, "query_id" => query_id}} = job) do
    # Schedule a new job to refresh the query on the first attempt
    if job.attempt == 1, do: schedule_next_refresh(job)
    Refresh.refresh_query(query_id, user_id)
  end

  def schedule_next_refresh(%Oban.Job{args: args}) do
      next_refresh_in_seconds = args["next_refresh_in_seconds"] || @one_day
      data = new(args, schedule_in: next_refresh_in_seconds)
      Oban.insert!(@oban_conf_name, data)
  end
```

PR #4281, opened 2024-05-29
Add label fqn metrics support

IvanIvanoff:
## Changes

The `intraday_labeled_metrics_v2` table exposes the metric `combined_labeled_balance`, which is computed on a wide variety of labels.

The reason is the labels directly can cause some confusion. For example, the labels `santiment/owner-><VALUE>:v1` can represent an address of a CEX, of a fund, or of some other entity.

The metrics will include:
- `combined_historical_balance_centralized_exchanges`
  - Accepts only a slug as an argument and gives you the total number of coins/tokens that CEXes hold together
- `historical_balance_centralized_exchanges`
  - Accepts slug and labelFqn which identifies the exchange in question.

Additionally, the getMetric's metadata is extended so the available labelFqns can be obtained.

To make it easier for clients, several "fake" metrics are created (the same internal metric with a different public name) are created. Each metric refers to a given group of labels - funds, (de)centralized exchanges, whales, etc.

A special table is created which combines different labels into groups, making it easy to create metrics on top of these groups. Each metric defines a set of values that point to some of the records in this table, and these records.

Table rows in the table look like this:
```sql
Row 1:
──────
asset_name:       bancor
asset_id:         309
display_name:     santiment/whale_usd_balance(bancor):v1
fqn:              santiment/whale_usd_balance(bancor):v1
label_key:        whale_usd_balance
owner:            santiment
version:          v1
metric_id:        624
parent_label_key: ᴺᵁᴸᴸ
group:            whales

Row 2:
──────
asset_name:       status
asset_id:         2576
display_name:     santiment/owner->binance:v1
fqn:              santiment/owner->binance:v1
label_key:        owner
owner:            santiment
version:          v1
metric_id:        624
parent_label_key: centralized_exchange
group:            exchanges

Row 3:
──────
asset_name:       omisego
asset_id:         1950
display_name:     santiment/centralized_exchange:v1
fqn:              santiment/centralized_exchange:v1
label_key:        centralized_exchange
owner:            santiment
version:          v1
metric_id:        624
parent_label_key: ᴺᵁᴸᴸ
group:            exchanges
```

## API Examples
```graphql
{
  getMetric(metric: "historical_balance_centralized_exchanges") {
    timeseriesData(selector: {slug: "bitcoin", labelFqn: "santiment/owner->binance:v1"}, from: "utc_now-45d", to: "utc_now-43d", interval: "1d") {
      value
      datetime
    }
    metadata {
      isLabelFqnMetric
      availableLabelFqns
      availableSlugs
    }
  }
}
```
```json
{
  "data": {
    "getMetric": {
      "metadata": {
        "availableLabelFqns": [
          "santiment/owner->binance:v1"
        ],
        "availableSlugs": [
          "bitcoin"
        ],
        "isLabelFqnMetric": true
      },
      "timeseriesData": [
        {
          "datetime": "2024-04-26T00:00:00Z",
          "value": 486080.3572366099
        },
        {
          "datetime": "2024-04-27T00:00:00Z",
          "value": 487229.5979543099
        },
        {
          "datetime": "2024-04-28T00:00:00Z",
          "value": 487967.3102217299
        }
      ]
    }
  }
}
```

Some labels have an asset embedded in them. At the moment, the only such labels are those who refer to whales.
One address can have `santiment/whale_usd_balance(ethereum):v1` and `santiment/whale_usd_balance(santiment):v1`, where the asset name points for which asset the address is a whale.

The metadata's `availableLabelFqns` can also accept a `slug` parameter, so the assets can be filtered, i.e. only those labels who refer to a given asset are returned:

```graphql
{
  getMetric(metric: "historical_balance_whales_usd"){
    metadata{
      availableLabelFqns(slug: "ethereum")
    }
  }
}
```
```json
{
  "data": {
    "getMetric": {
      "metadata": {
        "availableLabelFqns": [
          "santiment/whale_usd_balance(ethereum):v1"
        ]
      }
    }
  }
}
```

## Implementation details

Each of the metrics' JSON file looks like this:
```json
{
  "human_readable_name": "Combined Historical Balance of Funds",
  "name": "combined_historical_balance_funds",
  "metric": "combined_labeled_balance",
  "fixed_parameters": {
    "labels": {
      "group": "funds",
      "label_key": "fund",
      "__parametrized__": ["asset_name"]
    }
  }
...
}
```
The important parts are two:
- Usually, if there are more than one metric referring to the same internal `metric`, the compilation would fail. The check has been upgraded so it looks at the combination of `metric` + `fixed_parameters`. In case of no `fixed_parameters`, it would behave as it did previously.
- The `fixed_parameters` contain details which rows of the labels' group table to use when fetching the available label FQNs.

The `fixed_parameters` have the following fields (some are optional):
- `group` - the name of the group of the metric. Can be used on the FE to build the nesting of metrics (Labeled Balance -> Funds, Labeled Balance -> Centralized Exchanges, etc.)
- label_key - the label key part of the label FQN. Exposed for easier filtering
- parent_label_key - for some labels it is not clear what they point to. For example, when `label_key` is `owner` it is not very clear what entity is owning, but then we can add `parent_label_key: "centralized_exchange", `parent_label_key: "decentralized_exchange"`, etc., to add some additional information to the label.
- `__parametrized__` points to the table column that needs to be filled using some argument passed from the client.


lib/sanbase/clickhouse/metric/sql_query/metric_sql_query.ex
```diff
@@ -49,31 +50,96 @@ defmodule Sanbase.Clickhouse.MetricAdapter.SqlQuery do
...
+          end
+
+        sql = """
+        label_id = (
+          SELECT label_id
+          FROM label_metadata
+          WHERE
+            fqn = (
+              SELECT DISTINCT(fqn)
+              FROM test_anatolii_labeled_balances_filtered_2
```
tspenov, line 123: Some test table leftover

IvanIvanoff: The table is done. I've replaced it now.

PR #4283, opened 2024-05-31
Fix warnings

tspenov:
## Checklist:


Dockerfile
```diff
@@ -82,7 +82,7 @@ RUN mix assets.deploy
 RUN mix sentry.package_source_code
 
 # Compile the release
-RUN mix compile
+RUN mix compile --warnings-as-errors
```
IvanIvanoff, line 85: We should also add a `RUN mix compile --warnings-as-errors` in `Dockerfile-test` before the test command, so we can catch this much earlier.

PR #4289, opened 2024-06-07
Disable Oban.Web jobs locally

IvanIvanoff:
## Checklist:

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -35,6 +35,14 @@ config :phoenix,
   stacktrace_depth: 20,
   plug_init_mode: :runtime
 
+# Disable the Oban.Web jobs in local env.
+# When testing Oban.Web jobs locally comment out these 2 lines
+config :sanbase, Oban.Web, queues: false
+
+# Disable the Oban.Scrapers jobs in local env.
+# When testing Oban.Scrapers jobs locally comment out these 2 lines
+config :sanbase, Oban.Scrapers, queues: false
+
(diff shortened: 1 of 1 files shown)


config/scrapers_config.exs
```diff
@@ -22,6 +22,11 @@ config :sanbase, Sanbase.Twitter.Worker,
   update_interval: 1000 * 60 * 60 * 6,
   sync_enabled: {:system, "TWITTER_SCRAPER_ENABLED", false}
 
+config :sanbase, Oban.Web,
+  repo: Sanbase.Repo,
+  queues: [email_queue: 5, refresh_queries: 1],
```
tspenov, line 27: Do we need this on scrapers ? Also not 100% percent sure if they will work - maybe they need some part of the supervision tree that might not be started

config/scrapers_config.exs
```diff
@@ -22,6 +22,11 @@ config :sanbase, Sanbase.Twitter.Worker,
...
+  name: :oban_web
```
tspenov, line 28: if it is on scrapers - shouldn't it be :oban_scrapers

config/scrapers_config.exs
IvanIvanoff: Locally we disable the scheduling of new jobs, but it could be the case that we could probably run some job. I'll add it for Oban.Scrapers just to be sure.

tspenov (approved):

PR #4317, opened 2024-07-10
update kafka setup

IvanIvanoff:
- **Removed kafka consumer and websocket API. Remove old kafka deps**
- **Start implementing kafka exported directly with brod**

## Checklist:


docs/websocket.md
```diff
@@ -1,4 +1,6 @@
-# Sanbase WebSocket APIs
+#  DEPRECATED
```
tspenov, line 1: do we deprecate all or only the metrics ?

mix.exs
```diff
@@ -63,7 +63,7 @@ defmodule Sanbase.Mixfile do
       {:absinthe_phoenix, "~> 2.0"},
       {:absinthe_plug, "~> 1.5"},
       {:absinthe, "~> 1.5"},
-      {:brod, "== 3.8.1", manager: :rebar3, override: true, runtime: false},
+      {:brod, "~> 3.19"},
```
tspenov, line 66: there is also 4.0 version - can we use it ?

IvanIvanoff: Will update. It's 4.0 as the breaking change is making snappyer as optional dep I guess.

docs/websocket.md
IvanIvanoff: Only metrics, will fix it.

mix.exs
IvanIvanoff: done

docs/websocket.md
IvanIvanoff: done

tspenov (approved):

PR #4372, opened 2024-08-07
All accounts social metrics

volynvlad:
## Changes

Add the 

- mentions_percentage_twitter
- mentions_percentage_total
- mentions_count_twitter
- mentions_count_total

These metrics are to be used only on the Social Context page and don't given to API users.
To achieve this, the JSON spec of the metric now supports the `"is_hidden": true` flag, which hides
the metric from the `{ getAvailableMetrics }` and `{ projectBySlug(slug: "bitcoin"){ availableMetrics }}` GraphQL queries.

## Ticket

https://www.notion.so/santiment/New-Social-Volume-metric-for-Social-Trends-27200f937c9646f686bd3b4ee6e358ee

## Checklist:


lib/sanbase/clickhouse/metric/metric_files/social_metrics.json
```diff
@@ -2143,5 +2143,73 @@
     "has_incomplete_data": true,
     "data_type": "timeseries",
     "docs_links": ["https://academy.santiment.net/metrics/unique-social-volume"]
+  },
+  {
+    "human_readable_name": "Mentions Count Total",
+    "name": "mentions_count_total",
+    "metric": "social_volume_all_accounts",
```
IvanIvanoff, line 2150: What requires us introducing a new name for social volume - `mentions count`?

volynvlad: We want to distinguish SanBase&SanAPI from [Social Context](https://app.santiment.net/labs/trends/explore/*?slug=bitcoin&ticker=BTC&from=2024-02-06T23:00:01.960Z). In Sanbase and API show and give stable metrics, but in Social Context show all the data that we have. This data will be different, so we will introduce new metric name for `social volume` - `mentions count`, for `social dominance` - `mentions percentage`

PR #4378, opened 2024-08-13
Update stripe lib to latest

tspenov:
## Checklist:


lib/sanbase/billing/subscription/subscription.ex
```diff
@@ -647,7 +647,9 @@ defmodule Sanbase.Billing.Subscription do
   def add_payment_intent(result, _), do: result
 
   defp fetch_plan_id(db_subscription, stripe_subscription) do
-    case Plan.by_stripe_id(stripe_subscription.plan.id) do
+    stripe_plan_id = (stripe_subscription.items.data |> hd).plan.id
```
IvanIvanoff, line 650: From writing GraphQL tests I found using the `Access` protocol quite nice for accessing structures with list somewhere inside.
You could write:
```elixir
stripe_plan_id = get_in(stripe_subscription, [:items, :data, Access.at(0), :plan, :id])
```
It will also return `nil` if`data` is empty, so you can check that and return a better error.

lib/sanbase/stripe/stripe_api.ex
```diff
@@ -172,7 +168,7 @@ defmodule Sanbase.StripeApi do
   end
 
   def delete_subscription(stripe_id) do
-    Stripe.Subscription.delete(stripe_id)
+    Stripe.Subscription.cancel(stripe_id)
```
IvanIvanoff, line 171: is it the same behavior? Did they just rename the function?

IvanIvanoff (approved):

PR #4388, opened 2024-09-02
rework dashboards cache

IvanIvanoff:
## Changes

Rework how the Dashboards caching works
- Allow multiple caches of a dashboard -- each with different set of parameters overrides.
  - This allows sharing a dashboard page with changed parameters.
- Allow the dashboard cache to be refreshed by anyone. 
  - Previously the users needed to re-submit the result back with a different GraphQL call. This allows users to submit random garbage.
  - Now when any user can refresh the cache, to mitigate this problem the `runDashboardSqlQuery` GraphQL query accepts the `storeExecution` boolean flag. When it is set to `true`, the result will directly update the cache.  
  - Still, the dashboard needs to be public so other users can update its cache.
- The argument `parametersOverride` is now accepted in `runDashboardSqlQuery` and `getCachedDashboardQueriesExecutions`.

### Run & Cache dashboard queries
**Successfully updating the dashboard cache by non-owner**:
```graphql
{
  runDashboardSqlQuery(
      dashboardId: 3
      dashboardQueryMappingId: "9d7931fd-44a1-45d6-b633-ec13886e16a4"
      storeExecution: true
     ){
      clickhouseQueryId
      columnTypes
      columns
      rows
  }
}
```
```json
{
  "data": {
    "runDashboardSqlQuery": {
      "clickhouseQueryId": "3e0496bd-26ca-46d2-9fb3-20ead7719bec",
      "columnTypes": [
        "String",
        "UInt64",
        "String"
      ],
      "columns": [
        "big_num",
        "num",
        "slug"
      ],
      "rows": [
        [
          "2.12 Trillion",
          2123801239123,
          "ethereum"
        ]
      ]
    }
  }
}
```

---

**Unsuccessful cache update due to the dashboard being private and not owned by the user**:
```graphql
{
  runDashboardSqlQuery(
      dashboardId: 3
      dashboardQueryMappingId: "9d7931fd-44a1-45d6-b633-ec13886e16a4"
      storeExecution: true
     ){
      clickhouseQueryId
      columnTypes
      columns
      rows
  }
}
```
```json
{
  "data": {
    "runDashboardSqlQuery": null
  },
  "errors": [
    {
      "message": "Dashboard query mapping with id 63812020-3cf4-4115-a8ef-4e36403042ee does not exist,\nit is not part of dashboard 3, or the dashboard is not public.\n",
      "path": [
        "runDashboardSqlQuery"
      ],
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

---

**Execute a dashboard query and store it with overridden parameters**
```graphql
{
  runDashboardSqlQuery(
    	dashboardId: 3
    	dashboardQueryMappingId: "9d7931fd-44a1-45d6-b633-ec13886e16a4"
    	storeExecution: true
    	parametersOverride: "{\"slug\": \"a_completely_new_parameter_from_params\"}"
  	){
      clickhouseQueryId
      columnTypes
      columns
      rows
  }
}
```
```json
{
  "data": {
    "runDashboardSqlQuery": {
      "clickhouseQueryId": "4f4dce59-904f-4733-bd1a-db59e0a742a3",
      "columnTypes": [
        "String",
        "UInt64",
        "String"
      ],
      "columns": [
        "big_num",
        "num",
        "slug"
      ],
      "rows": [
        [
          "2.12 Trillion",
          2123801239123,
          "a_completely_new_parameter_from_params"
        ]
      ]
    }
  }
}
```

### Get cached dashboard queries

**Get cached dashboard with the original parameters**: 
```graphql
{
  getCachedDashboardQueriesExecutions(dashboardId:3){
    queries{
      dashboardQueryMappingId
      queryId
      columns
      columnTypes
      rows
    }
  }
}
```
```json
{
  "data": {
    "getCachedDashboardQueriesExecutions": {
      "queries": [
        {
          "columnTypes": [
            "String",
            "UInt64",
            "String"
          ],
          "columns": [
            "big_num",
            "num",
            "slug"
          ],
          "dashboardQueryMappingId": "9d7931fd-44a1-45d6-b633-ec13886e16a4",
          "queryId": 4,
          "rows": [
            [
              "2.12 Trillion",
              2123801239123,
              "ethereum"
            ]
          ]
        }
      ]
    }
  }
}
```

---

**Get cached dashboard with the overriden parameters**: 
```graphql
{
  getCachedDashboardQueriesExecutions(
    dashboardId:3 
    parametersOverride: "{\"slug\": \"a_completely_new_parameter_from_params\"}"
){
    queries{
      dashboardQueryMappingId
      queryId
      columns
      columnTypes
      rows
    }
  }
}
```
```json
{
  "data": {
    "getCachedDashboardQueriesExecutions": {
      "queries": [
        {
          "columnTypes": [
            "String",
            "UInt64",
            "String"
          ],
          "columns": [
            "big_num",
            "num",
            "slug"
          ],
          "dashboardQueryMappingId": "9d7931fd-44a1-45d6-b633-ec13886e16a4",
          "queryId": 4,
          "rows": [
            [
              "2.12 Trillion",
              2123801239123,
              "a_completely_new_parameter_from_params"
            ]
          ]
        }
      ]
    }
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/dashboards/dashboard/dashboard_cache.ex
+++ b/lib/sanbase/dashboards/dashboard/dashboard_cache.ex
@@ -167,8 +167,8 @@ defmodule Sanbase.Dashboards.DashboardCache do
         |> maybe_transform_error()
 
       _ ->
-        {:error, "Dashboard with id #{dashboard_id} does not exist or the dashboard\
-         is private and the user with id #{querying_user_id} is not the owner of it."}
+        {:error,
+         "Dashboard with id #{dashboard_id} does not exist or the dashboard is private and the user with id #{querying_user_id} is not the owner of it."}
     end
   end
 
--- a/lib/sanbase/dashboards/dashboard/dashboard_cache.ex
+++ b/lib/sanbase/dashboards/dashboard/dashboard_cache.ex
@@ -21,6 +21,7 @@ defmodule Sanbase.Dashboards.DashboardCache do
 
   @type user_id :: Sanbase.Accounts.User.user_id()
   @type dashboard_id :: Dashboard.dashboard_id()
+  @type parameters_override :: map()
   @type dashboard_query_mapping_id :: Dashboard.dashboard_query_mapping_id()
 
   @type query_cache :: %{
@@ -38,6 +39,7 @@ defmodule Sanbase.Dashboards.DashboardCache do
 
   schema "dashboards_cache" do
     field(:dashboard_id, :integer)
+    field(:parameters_override_hash, :string, default: "none")
     field(:queries, :map, default: %{})
 
     timestamps()
@@ -55,6 +57,23 @@ defmodule Sanbase.Dashboards.DashboardCache do
   # end
 
   @doc ~s"""
+  Hash the parameters map to a string.
+  The hash will be stored in the database.
+  """
+  @spec hash_parameters(map()) :: String.t()
+  def hash_parameters(%{} = params) when map_size(params) == 0, do: "none"
+
+  def hash_parameters(%{} = params) do
+    binary = :erlang.term_to_binary(params)
+
+    # Base64 encodes 6 bits of information per character, so 16 characters is 96 bits
+    # 1.3x10^13 attempts are needed (13 trillion) to have 0.1% chance of collisions
+    :crypto.hash(:sha256, binary)
+    |> Base.encode64(padding: false)
+    |> :erlang.binary_part(0, 16)
+  end
+
+  @doc ~s"""
   Fetch the latest cache values for the given dashboard.
 
   The second `opts` argument can contain the following options:
@@ -62,7 +81,7 @@ defmodule Sanbase.Dashboards.DashboardCache do
   - transform_loaded_queries - If set to true, the loaded query caches are
   transformed from having `compressed_rows` to `rows`. Defaults to true.
   """
-  @spec by_dashboard_id(dashboard_id(), user_id, Keyword.t()) :: {:ok, t()} | {:error, String.t()}
+
   # def by_dashboard_id(dashboard_id, querying_user_id, opts) do
   #   Ecto.Multi.new()
   #   |> Ecto.Multi.run(:get_dashboard_cache, fn _ ->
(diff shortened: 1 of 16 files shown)


lib/sanbase/dashboards/dashboard/dashboard_cache.ex
```diff
@@ -62,16 +81,23 @@ defmodule Sanbase.Dashboards.DashboardCache do
...
-  def by_dashboard_id(dashboard_id, querying_user_id, opts \\ []) do
-    query = from(d in __MODULE__, where: d.dashboard_id == ^dashboard_id)
+  @spec by_dashboard_id(dashboard_id, parameters_override, user_id, Keyword.t()) ::
+          {:ok, t()} | {:error, String.t()}
+  def by_dashboard_id(dashboard_id, parameters_override, querying_user_id, opts \\ []) do
+    hash = hash_parameters(parameters_override)
+
+    query =
+      from(d in __MODULE__,
+        where: d.dashboard_id == ^dashboard_id and d.parameters_override_hash == ^hash
```
tspenov, line 99: why we filter by `parameters_override_hash` - isn't `dashboard_id` enough ?

lib/sanbase/dashboards/dashboard/dashboard_cache.ex
```diff
@@ -125,32 +151,46 @@ defmodule Sanbase.Dashboards.DashboardCache do
...
+          dashboard_id: dashboard_id,
+          parameters_override_hash: hash
+        })
         |> Repo.insert()
         |> maybe_transform_error()
 
       _ ->
-        {:error,
-         "Dashboard with id #{dashboard_id} does not exist or the user with id #{querying_user_id} is not the owner of it."}
+        {:error, "Dashboard with id #{dashboard_id} does not exist or the dashboard\
```
tspenov, line 170: maybe separate case for `is_public == false and user_id != querying_user_id` so we have separate error message ?

IvanIvanoff: We have the same behavior in most other places -- not owner or does not exist. For users, they cannot do anything about neither of those, so why bother split it. Also, users will not see private dashboards, so actually hitting this error should be close to never.

My approach to this is same as with passwords -- you just report that something is not ok, you never tell "email is wrong, but the password matches one" (while here we have nowhere the level of security issues this can cause with passwords)

lib/sanbase/dashboards/dashboard/dashboard_cache.ex
IvanIvanoff: We now allow storing multiple caches of a dashboard, one per set of extra override parameters.
If you store the cache without overriding the dashboard, the hash here will be the string `none`.
But if you want to share my dashboard with someone else and change the parameters, you provide `parameters_override: %{slug: "new_slug"}` and cache this. So this way we have 2 records in the DB for that dashboard.

tspenov: makes sense now

tspenov (approved):

PR #4397, opened 2024-09-26
Disallow comments on private entities

IvanIvanoff:
## Checklist:


lib/sanbase/queries/query/query.ex
```diff
@@ -200,14 +201,32 @@ defmodule Sanbase.Queries.Query do
 
   # Entity-based queries
 
+  @impl Sanbase.Entity.Behaviour
+  def get_visibility_data(id) do
```
tspenov, line 205: we can extract the common functionality like this:
```elixir
@impl Sanbase.Entity.Behaviour
def get_visibility_data(id) do
  get_entity_visibility_data(base_query(), "query", id)
end

defp get_entity_visibility_data(base_query, entity_name, entity_id) do
  query =
    from(entity in base_query,
      where: entity.id == ^entity_id,
      select: %{
        is_public: entity.is_public,
        is_hidden: entity.is_hidden,
        user_id: entity.user_id
      }
    )

  case Sanbase.Repo.one(query) do
    %{} = map -> {:ok, map}
    nil -> {:error, "The #{entity_name} with id #{entity_id} does not exist"}
  end
end
```

priv/repo/structure.sql
```diff
@@ -9564,3 +9560,5 @@ INSERT INTO public."schema_migrations" (version) VALUES (20240725122924);
 INSERT INTO public."schema_migrations" (version) VALUES (20240805115620);
 INSERT INTO public."schema_migrations" (version) VALUES (20240809122904);
 INSERT INTO public."schema_migrations" (version) VALUES (20240904135651);
+INSERT INTO public."schema_migrations" (version) VALUES (20240917083648);
+INSERT INTO public."schema_migrations" (version) VALUES (20240926135951);
```
tspenov, line 9564: there are 2 migrations

lib/sanbase/queries/query/query.ex
IvanIvanoff: Only `post` has some different logic, but the rest can probably be reused.

IvanIvanoff: Done.

priv/repo/structure.sql
IvanIvanoff: fixed.

lib/sanbase/insight/post.ex
```diff
@@ -120,6 +120,57 @@ defmodule Sanbase.Insight.Post do
...
+        select: %{
+          is_public: entity.state == ^@approved and entity.ready_state == ^@published,
+          is_hidden: entity.is_hidden,
+          user_id: entity.user_id
+        }
+      )
+
+    case Repo.one(query) do
+      %{} = map -> {:ok, map}
+      nil -> {:error, "The dashboard with id #{id} does not exist"}
```
tspenov, line 137: that should be for insight

tspenov (approved):

lib/sanbase/insight/post.ex
IvanIvanoff: fixed.

PR #4400, opened 2024-09-27
Use glitchtip instead of sentry

tspenov:
## Checklist:


lib/sanbase/application/application.ex
```diff
@@ -46,6 +46,10 @@ defmodule Sanbase.Application do
     # Note: Sentry.LoggerBackend now allows to send non-exception errors to sentry, too.
     # This is not enabled yet.
     {:ok, _} = Logger.add_backend(Sentry.LoggerBackend)
```
tspenov, line 48: Not sure if this should be removed

IvanIvanoff: `Logger`'s backend is fulfilling the same purpose as Erlang's `:logger` handlers, so we need to leave only the `:logger.add_handler(...)` part.

Also, in the Sentry dependency docs we see this:
image

IvanIvanoff (approved):

PR #4402, opened 2024-09-30
Add treatWordAsLuceneQuery to wordsSocialVolume

IvanIvanoff:
## Changes

Add a boolean parameter named `treatWordAsLuceneQuery` to the `wordsSocialVolume` GraphQL query.
If the parameter is set to `true`, the words are passed to metricshub as they are, so `eth AND nft` becomes a valid input.
If the parameter is set to `false`, the words are lowercased before being sent. This was required because otherwise words that are valid Lucene keywords broke the query. For example, the ticker of notcoin is `NOT`, which is also a keyword, so it needed to be lowercased otherwise it caused a syntax error.
Defaults to `false`.

If a query like `eth OR nft` is queried without setting the flag to `true`, it is interpreted as `eth or nft`, which matches if any of those words is seen in a message, essentially turning it into `eth OR nft OR or`

```graphql
{
  wordsSocialVolume(
    from: "utc_now-2d"
    to: "utc_now"
    interval: "1d"
    selector: {words: ["eth AND nft", "btc"]}
    treatWordAsLuceneQuery: true
  ) {
      word
      timeseriesData {
        mentionsCount
        datetime
      }
  }
}
```
```json
{
  "data": {
    "wordsSocialVolume": [
      {
        "timeseriesData": [
          {
            "datetime": "2024-09-28T00:00:00Z",
            "mentionsCount": 420
          },
          {
            "datetime": "2024-09-29T00:00:00Z",
            "mentionsCount": 487
          },
          {
            "datetime": "2024-09-30T00:00:00Z",
            "mentionsCount": 281
          }
        ],
        "word": "btc"
      },
      {
        "timeseriesData": [
          {
            "datetime": "2024-09-28T00:00:00Z",
            "mentionsCount": 1
          },
          {
            "datetime": "2024-09-29T00:00:00Z",
            "mentionsCount": 2
          },
          {
            "datetime": "2024-09-30T00:00:00Z",
            "mentionsCount": 1
          }
        ],
        "word": "eth AND nft"
      }
    ]
  }
}
```

## Checklist:

### DIFF
--- a/lib/sanbase/social_data/social_helper.ex
+++ b/lib/sanbase/social_data/social_helper.ex
@@ -16,17 +16,12 @@ defmodule Sanbase.SocialData.SocialHelper do
 
   def sources_total_string(), do: @sources |> Enum.join(",")
 
-  def replace_words_with_original_casing({:error, error}, _words), do: {:error, error}
+  def replace_words_with_original_casing(result, words) do
+    Enum.map(result, fn %{word: lowercase_word} = map ->
+      original_word = Enum.find(words, fn w -> String.downcase(w) == lowercase_word end)
 
-  def replace_words_with_original_casing({:ok, result}, words) do
-    result =
-      Enum.map(result, fn %{word: lowercase_word} = map ->
-        original_word = Enum.find(words, fn w -> String.downcase(w) == lowercase_word end)
-
-        Map.put(map, :word, original_word)
-      end)
-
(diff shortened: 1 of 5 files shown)


volynvlad (approved):

tspenov:
@IvanIvanoff can you add a test

PR #4417, opened 2024-10-09
Add versioning to open interest exporter

IvanIvanoff:
## Changes
We need to re-export open_interest data into a new kafka topic - `open_interest_cryptocompare_v2` while also exporting data to the old topic.

To do this, introduce `version` to the Oban job.
If no version is provided, `v1` is considered the version. 
In case of `v1` the exporter works the same.
In case the job has version `v2` - it exports to the new kafka topic. Jobs for the same args with different version can work in parallel.

To start an export, we need to schedule manually the historical scrape once:
```elixir
Sanbase.Cryptocompare.OpenInterest.HistoricalScheduler.schedule_jobs(
  limit: 2000,
  version: "v2",
  schedule_next_job: true
)
```

## Checklist:

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -87,6 +87,8 @@ config :sanbase, Sanbase.KafkaExporter,
   asset_ohlcv_price_pairs_topic:
     {:system, "KAFKA_ASSET_OHLCV_PRICE_POINTS_TOPIC", "asset_ohlcv_price_pairs"},
   open_interest_topic: {:system, "KAFKA_OPEN_INTEREST_TOPIC", "open_interest_cryptocompare"},
+  open_interest_topic_v2:
+    {:system, "KAFKA_OPEN_INTEREST_TOPIC_V2", "open_interest_cryptocompare_v2"},
   funding_rate_topic: {:system, "KAFKA_FUNDING_RATE_TOPIC", "funding_rate_cryptocompare"},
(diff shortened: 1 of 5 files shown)


config/config.exs
```diff
@@ -87,6 +87,8 @@ config :sanbase, Sanbase.KafkaExporter,
   asset_ohlcv_price_pairs_topic:
     {:system, "KAFKA_ASSET_OHLCV_PRICE_POINTS_TOPIC", "asset_ohlcv_price_pairs"},
   open_interest_topic: {:system, "KAFKA_OPEN_INTEREST_TOPIC", "open_interest_cryptocompare"},
+  open_interest_topic_v2:
```
tspenov, line 90: that is not used

IvanIvanoff: It is used when exporting the data:
```elixir
    topic =
      case args["version"] do
        "v1" -> Config.module_get!(Sanbase.KafkaExporter, @topic)
        "v2" -> Config.module_get!(Sanbase.KafkaExporter, @topic_v2)
      end
```
where `@topic2` refers to this.

IvanIvanoff: The idea is that we'll re-export the data in parallel in a new topic, then create a new clickhouse table from it and when everything is ready -- switch the Airflow jobs. 
The biggest issue with the data is from 2022, but after that there are again some gaps for a few months in 2023. The data today looks good, so this approach will allow us to continue producing metrics while the re-export is in progress.

tspenov (approved):

PR #4429, opened 2024-10-16
Better handle Oban max_attempts

IvanIvanoff:
## Changes

- **Add explicit max_attempts to some Oban workers**
- **Avoid snoozing Oban jobs too many times**
  - There are cases on prod where a job has been snoozed over 500 times. Put a limit to that.
  - Snoozed jobs are kept in the `oban_jobs` table, so it has a lot of records.

## Checklist:

### DIFF
--- a/lib/sanbase/cryptocompare/price/price_historical_worker.ex
+++ b/lib/sanbase/cryptocompare/price/price_historical_worker.ex
@@ -31,16 +31,25 @@ defmodule Sanbase.Cryptocompare.Price.HistoricalWorker do
   def conf_name(), do: @oban_conf_name
 
   @impl Oban.Worker
-  def perform(%Oban.Job{args: args}) do
+  def perform(%Oban.Job{args: args, attempt: attempt}) do
     %{"base_asset" => base_asset, "quote_asset" => quote_asset, "date" => date} = args
     t1 = System.monotonic_time(:millisecond)
     should_snooze? = base_asset not in available_base_assets()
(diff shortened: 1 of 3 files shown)


lib/sanbase/cryptocompare/price/price_historical_worker.ex
```diff
@@ -49,7 +58,16 @@ defmodule Sanbase.Cryptocompare.Price.HistoricalWorker do
...
 
           :snooze ->
-            {:snooze, 86_400}
+            if attempt > 30 do
+              Logger.info(
+                "[Cryptocompare Historical] The job for #{base_asset}/#{quote_asset} and date #{date} has been snoozed too many times due to errors in the response format. \
+          Marking it as complete so it's no longer scheduled."
+              )
+
+              :ok
```
tspenov, line 67: Should we finish in `completed` state . I think it is better to go in `discarded` state. I won't be retried but will be visible that smth happened.

IvanIvanoff: Hmm, maybe `{:cancel, reason}`. I don't see how to force it being `discarded` other than just making it `error` many times.

IvanIvanoff: Updated to `{:cancel, String.t()}`.
Cancelled jobs are also not retried.

tspenov (approved):

PR #4431, opened 2024-10-17
Add getMetricSpikeExplanations API

IvanIvanoff:
## Changes

- **Allow running locally and connect to stage DB**
- **Add getMetricSpikeExplanations API**

## Checklist:


volynvlad (approved):

config/runtime.exs
```diff
@@ -1,5 +1,18 @@
 import Config
 
+if config_env() == :dev do
```
tspenov, line 3: why do we have dev env in runtime.exs ?

tspenov (approved):

config/runtime.exs
IvanIvanoff: If you connect to local postgres and you use SSL it might error.
But if you are running the app locally and connect to stage DB - it needs SSL enabled.
The env vars are loaded at runtime, so if I just put the same code in `dev.exs` it won't work. But in `runtime.exs` it works.

PR #4433, opened 2024-10-18
Use empty map instead of nil for parameters_override

tspenov:
## Checklist:


lib/sanbase/queries/refresh/refresh.ex
```diff
@@ -61,7 +61,7 @@ defmodule Sanbase.Queries.Refresh do
     |> Enum.each(fn mapping ->
       Sanbase.Dashboards.cache_dashboard_query_execution(
         mapping.dashboard_id,
-        _parameters_override_hash = nil,
+        _parameters_override_hash = %{},
```
IvanIvanoff, line 64: you can also rename the param to just `_parameters_override`. `hash` is the old implementation that has been changed, but these places have remained.

IvanIvanoff (approved):

PR #4438, opened 2024-10-22
Encode/decode social volume words to/from www form

IvanIvanoff:
## Changes

Otherwise words like $43,000 are considered as 2 words as the comma is
used for a separator. Encoding on our end and metricshub decoding on
its end will fix this problem.

This needs to be deployed together with the metricshub changes.

## Checklist:

### DIFF
--- a/test/sanbase_web/graphql/social_data/words_social_volume_api_test.exs
+++ b/test/sanbase_web/graphql/social_data/words_social_volume_api_test.exs
@@ -17,12 +17,12 @@ defmodule SanbaseWeb.Graphql.WordsSocialVolumeApiTest do
     body =
       %{
         "data" => %{
-          "btc" => %{
+          URI.encode_www_form("btc") => %{
             "2024-09-28T00:00:00Z" => 373,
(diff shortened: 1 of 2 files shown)


test/sanbase_web/graphql/social_data/words_social_volume_api_test.exs
```diff
@@ -17,12 +17,12 @@ defmodule SanbaseWeb.Graphql.WordsSocialVolumeApiTest do
...
       %{
         "data" => %{
-          "btc" => %{
+          URI.encode_www_form("btc") => %{
             "2024-09-28T00:00:00Z" => 373,
             "2024-09-29T00:00:00Z" => 487,
             "2024-09-30T00:00:00Z" => 323
           },
-          "eth or nft" => %{
+          URI.encode_www_form("eth or nft") => %{
```
tspenov, line 25: We return from the API the keys as www-form-urlencoded strings ? Does this mean that the frontend should change their logic also ?

IvanIvanoff:
Close in favor of https://github.com/santiment/sanbase2/pull/4550

PR #4515, opened 2025-01-03
Update elixir to 1.18 and elixir deps

IvanIvanoff:
## Changes

- **Update Elixir to 1.18.1 and Erlang/OTP to 27.2**
- **Format <%= %> into {...} in HEEX. Fix issues with GraphQL queries**

## Checklist:


lib/sanbase_web/live/available_metrics/available_metrics_description.ex
```diff
@@ -9,14 +9,14 @@ defmodule SanbaseWeb.AvailableMetricsDescription do
 
     Example:
 
-      {
-        getMetric(<b>metric: "price_usd"</b>){
-          timeseriesData(asset: "ethereum" from: "utc_now-90d" to: "utc_now" interval: "1d"){
+      &lbrace;
```
tspenov, line 12: can we `\{` ?

IvanIvanoff: no. It needs to be either `<%= "{" %>` or `%lbrace;` If you try to use `\{` the compiler says:
```
In case you don't want `{` to begin a new interpolation, you may write it using `&lbrace;` or using `<%= "{" %>`
```

tspenov (approved):

PR #4521, opened 2025-01-15
Let metricshub decide what total source is

IvanIvanoff:
## Checklist:


volynvlad (approved):

lib/sanbase/social_data/sentiment.ex
```diff
@@ -12,7 +12,7 @@ defmodule Sanbase.SocialData.Sentiment do
   @recv_timeout 25_000
 
   def sentiment(selector, from, to, interval, source, type)
-      when source in [:all, "all", :total, "total"] do
+      when source in [:all, "all", :total] do
```
IvanIvanoff, line 15: Note: `"total"` as string is removed to avoid infinite loop.

PR #4532, opened 2025-01-21
Generalize attempts to use resources

tspenov:
# Rate Limiting System Refactor

## Changes
- Introduced a new generic `AccessAttempt` module to handle rate limiting across different features
- Created `AccessAttemptBehaviour` to define a common interface for rate-limited actions
- Migrated email login attempts to use the new system
- Added rate limiting for coupon code validation
- Added comprehensive tests for the new rate limiting system

## Technical Details
- Created new `access_attempts` table to track attempts across different features
- Each attempt type (email login, coupon validation) has its own configuration for:
  - Time interval
  - Maximum attempts per user
  - Maximum attempts per IP address
- Implemented proper error handling and user-friendly error messages

This refactor makes it easier to add rate limiting to new features while maintaining consistent behavior and configuration across the application.

## Example implementation of the behaviour

```elixir
defmodule Sanbase.Accounts.CouponAttempt do
  @behaviour Sanbase.Accounts.AccessAttemptBehaviour
  alias Sanbase.Accounts.AccessAttempt

  @impl true
  def type, do: "coupon"

  @impl true
  def config do
    %{
      interval_in_minutes: 10,
      allowed_user_attempts: 5,
      allowed_ip_attempts: 20
    }
  end

  @impl true
  def check_attempt_limit(user, remote_ip) do
    AccessAttempt.check_attempt_limit(type(), user, remote_ip)
  end

  @impl true
  def create(user, remote_ip) do
    AccessAttempt.create(type(), user, remote_ip)
  end
end
```

## Checklist:


lib/sanbase/accounts/access_attempt.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Sanbase.Accounts.AccessAttempt do
+  use Ecto.Schema
+  import Ecto.{Query, Changeset}
+  alias Sanbase.Repo
+
+  schema "access_attempts" do
```
IvanIvanoff, line 6: maybe `user_action_attempts` would fit better? Or just `action_attempts`?

priv/repo/migrations/20250121155544_add_access_attempts.exs
```diff
@@ -0,0 +1,18 @@
...
+      add(:ip_address, :string, null: false)
+      add(:type, :string, null: false)
+
+      timestamps()
+    end
+
+    create(index(:access_attempts, [:user_id]))
+    create(index(:access_attempts, [:ip_address]))
+    create(index(:access_attempts, [:type]))
+    create(index(:access_attempts, [:inserted_at]))
```
IvanIvanoff, line 16: I am not sure if this is an optimal set of indexes.
We'll never really query inserted_at or type on its own. It is always in conjunction with user_id or IP address.

tspenov: changed to use composite index on `user_id|ip address + type + inserted_at`

lib/sanbase/accounts/access_attempt.ex
tspenov: `access_attempts` still looks better to me.

PR #4546, opened 2025-02-07
Experimental metrics

tspenov:
## Checklist:


lib/sanbase_web/graphql/middlewares/access_control/access_control.ex
```diff
@@ -164,6 +165,50 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
     fun.(resolution)
   end
 
+  defp check_experimental_metric_access(
+         %Resolution{
+           context: %{__metric__: metric_name, current_user: current_user}
+         } = resolution
+       ) do
+    # Have to get from cache
+    with {:ok, metric} <- MetricRegistry.by_name(Atom.to_string(metric_name)),
```
IvanIvanoff, line 174: Isn't this `Atom.to_string` call causing an error? `metric_name` should be a string, and if it's not -- we need to investigate why it is an atom.

lib/sanbase/accounts/user.ex
```diff
@@ -86,6 +87,9 @@ defmodule Sanbase.Accounts.User do
     # GDPR related fields
     field(:privacy_policy_accepted, :boolean, default: false)
     field(:marketing_accepted, :boolean, default: false)
+
+    field(:status, :string, default: "regular")
```
IvanIvanoff, line 91: `status` is too vague for me in the context of User.
Maybe `metric_access_level`?

lib/sanbase_web/graphql/middlewares/access_control/access_control.ex
```diff
@@ -164,6 +165,50 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
...
+
+  defp do_check_experimental_metric(current_user, metric, resolution) do
+    if user_can_access_metric?(current_user, metric.status) do
+      resolution
+    else
+      case metric.status do
+        "alpha" ->
+          Resolution.put_result(
+            resolution,
+            {:error, "This metric is in alpha and is only available to alpha users."}
```
IvanIvanoff, line 192:
```suggestion
            {:error, "The metric #{metric.metric} is currently in alpha phase and is exclusively available to alpha users."}
```

lib/sanbase_web/graphql/middlewares/access_control/access_control.ex
```diff
@@ -164,6 +165,50 @@ defmodule SanbaseWeb.Graphql.Middlewares.AccessControl do
...
+          )
+
+        "beta" ->
+          Resolution.put_result(
+            resolution,
+            {:error, "This metric is in beta and is only available to alpha or beta users."}
```
IvanIvanoff, line 198:
```suggestion
            {:error, "The metric #{metric.metric} is currently in beta phase and is exclusively available to alpha and beta users."}
```

lib/sanbase/clickhouse/metric/registry/registry.ex
```diff
@@ -78,6 +78,11 @@ defmodule Sanbase.Clickhouse.MetricAdapter.Registry do
     {:timebound_flag_map, []}
   ]
 
+  def by_name(name) do
+    get_metrics([])
+    |> Enum.find(fn metric -> metric.metric == name end)
```
IvanIvanoff, line 83: In theory we can have more than 1 record with the same name.
In practice -- there are no such cases. Maybe I would revisit the unique index and make it so `metric` is guaranteed to be unique (before and after resolving parameters).

test/sanbase_web/graphql/middlewares/access_control_middleware_test.exs
```diff
@@ -441,6 +446,7 @@ defmodule SanbaseWeb.Graphql.AccessControlMiddlewareTest do
 
       {:ok, metric} = Sanbase.Metric.Registry.by_name("price_usd_5m", "timeseries")
       {:ok, _} = Sanbase.Metric.Registry.update(metric, %{status: "alpha"})
+      true = Sanbase.Clickhouse.MetricAdapter.Registry.refresh_stored_terms()
```
IvanIvanoff, line 449: better call `Sanbase.Metric.Registry.refresh_stored_terms` as it will call all the places (currently 3) where we need to refresh them.

PR #4556, opened 2025-02-13
Implement authentication and authorization for metric registry

IvanIvanoff:
## Checklist:


lib/sanbase_web/live/metric_registry/metric_registry_change_suggestions_live.ex
```diff
@@ -127,6 +127,10 @@ defmodule SanbaseWeb.MetricRegistryChangeSuggestionsLive do
 
   @impl true
   def handle_event("update_status", %{"status" => "undo", "record_id" => record_id}, socket) do
+    Permissions.raise_if_cannot(:apply_change_suggestions,
```
tspenov, line 130: could be even `socket = ensure_authorized(socket, :apply_change_suggestions)`
and extract 
```elixir
defp ensure_authorized(socket, permission) do
  Permissions.raise_if_cannot(permission, roles: socket.assigns.current_user_role_names)
  socket
end
```

IvanIvanoff: This function would not need to return the socket as it never changes it.
I can use that, but then I need to find a proper module (new one?) to put that function in it and then import it.

IvanIvanoff: Let's handle this in a separate clean-up PR, so we can move this to stage before tomorrow's meeting.

PR #4569, opened 2025-02-25
Try sending email on each node until success

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/email/template_mailer.ex
+++ b/lib/sanbase/email/template_mailer.ex
@@ -56,15 +56,24 @@ defmodule Sanbase.TemplateMailer do
           nil -> template[:subject]
         end
 
-      new()
-      |> to(rcpt_email)
-      |> from(from)
-      |> subject(subject)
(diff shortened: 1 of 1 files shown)


tspenov:
Maybe not needed after all :)

lib/sanbase/email/template_mailer.ex
```diff
@@ -75,6 +84,31 @@ defmodule Sanbase.TemplateMailer do
...
+      Task.Supervisor.async({Sanbase.TaskSupervisor, node}, fn ->
+        Sanbase.TemplateMailer.deliver(email)
+      end)
+      |> Task.await()
+
+    case result do
+      {:ok, _} = success -> success
+      _error -> try_nodes_sequentially(rest, email)
+    end
+  end
```
IvanIvanoff, line 110: Can this cause an infinite loop?
If the sending fails on node A it tries on node B, where if it fails in tries on node A, and so on and so on?

PR #4611, opened 2025-03-24
Use yearly snapshots in ethereum balance change computation

IvanIvanoff:
## Changes

Balance records are inserted only when a transfer/balance change happens.

Because of this when we want to check `by how much did the ETH balance of address X changed compared to 1 month ago` we **cannot** put `dt >= now() - interval 30 day`.
This is because the latest balance change can be 30 days ago, or it could be 3000 days ago.

To improve the situation here we use the yearly snapshot balances table.

## Checklist:

### DIFF
--- a/lib/sanbase/balances/balance_sql_query.ex
+++ b/lib/sanbase/balances/balance_sql_query.ex
@@ -39,6 +39,72 @@ defmodule Sanbase.Balance.SqlQuery do
   def decimals("dogecoin", _), do: 1
   def decimals(_, decimals), do: Integer.pow(10, decimals)
 
+  def balance_change_query(
+        addresses,
+        "ethereum" = slug,
+        decimals,
+        "ethereum" = blockchain,
+        from,
+        to
+      ) do
+    sql = """
+    WITH
+      toDateTime({{from}}) AS from,
+      toDateTime({{to}}) AS to,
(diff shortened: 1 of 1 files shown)


lib/sanbase/balances/balance_sql_query.ex
```diff
@@ -39,6 +39,72 @@ defmodule Sanbase.Balance.SqlQuery do
...
+      starting_balances AS
+      (
+          SELECT
+              address,
+              last_balance_at AS dt,
+              balance
+          FROM {{yearly_snapshot_table}}
+          WHERE (toYear(from) = year) AND (address IN (addresses_of_interest))
+      ),
+      ususal_balances AS
```
tspenov, line 64: typo: `ususal_balances`

lib/sanbase/balances/balance_sql_query.ex
```diff
@@ -39,6 +39,72 @@ defmodule Sanbase.Balance.SqlQuery do
...
+      merged AS
+      (
+          SELECT
+              *,
+              '' AS txID,
+              dt AS computedAt
+          FROM starting_balances
+          UNION ALL
+          SELECT *
+          FROM ususal_balances
```
tspenov, line 84: typo: `ususal_balances`

PR #4651, opened 2025-04-24
Stop using custom complexity weight for some metrics

IvanIvanoff:
## Changes

Improve the consistency of complexity computation.
The complexity weight was not working properly as well, as the metric
could not be extracted, so it was always 1 either way.

## Checklist:


IvanIvanoff:
Close in favor of https://github.com/santiment/sanbase2/pull/4652

Turns out that the complexity weight is used, but was not working properly in the new JSON apis.

PR #4775, opened 2025-08-13
Solve the PricePair renaming of slugs in result

IvanIvanoff:
## Changes

Cryptocompare data is stored with their cryptocompare base_asset name.
Santiment has multiple assets pointing to the same cryptocompare asset
like: arb-arbitrum -> ARB, arbitrum -> ARB, etc.

In the SQL we first map the san asset to the cryptocompare asset, and
then we map the cryptocompare asset to the san asset (in SQL). In this
second operation we lose track of the original asset, so we end up in a
situation where we map arb-arbitrum to ARB, but then we map ARB to
arbitrum, in practice renaming the slug in the result.

This commit fixes the issue in cases where only one such renaming
exists and the number of assets returned is the same as the number of
assets queried. This check is needed because if arb-arbitrum and
arbitrum are both provided, the SQL will return data only for 1 asset.

## Checklist:


lib/sanbase/prices/price_pair/metric_adapter.ex
```diff
@@ -202,4 +205,70 @@ defmodule Sanbase.PricePair.MetricAdapter do
   defp metric_to_quote_asset("price_usd"), do: "USD"
   defp metric_to_quote_asset("price_usdt"), do: "USDT"
   defp metric_to_quote_asset("price_eth"), do: "ETH"
+
+  defp rename_same_aliases({result, result_slugs_mapset}, slug_or_slugs) do
+    parameters_slugs_mapset = Enum.to_list(slug_or_slugs) |> MapSet.new()
```
tspenov, line 210: `List.wrap` instead of `Enum.to_list` ?

PR #4779, opened 2025-08-15
Add insights tools mcp

tspenov:
## Checklist:


IvanIvanoff:
@tspenov I've also renamed a little bit the tools so they have a more consistent naming.
The names can change in the future, too.
I've also added tests and did some refactoring of the code here and there.

PR #4791, opened 2025-08-24
Expose more metrics and assets via the MCP

IvanIvanoff:
## Checklist:

### DIFF
--- a/lib/sanbase/project/list/project_list.ex
+++ b/lib/sanbase/project/list/project_list.ex
@@ -700,7 +700,7 @@ defmodule Sanbase.Project.List do
     case Keyword.get(opts, :min_volume) do
       nil ->
         from(p in query,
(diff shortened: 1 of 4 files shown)


lib/sanbase/project/list/project_list.ex
```diff
@@ -700,7 +700,7 @@ defmodule Sanbase.Project.List do
     case Keyword.get(opts, :min_volume) do
       nil ->
         from(p in query,
-          inner_join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
+          left_join: latest_cmc in assoc(p, :latest_coinmarketcap_data),
```
tspenov, line 703: what will happen with those that don't have `latest_cmc_data`. Will they go last ?

tspenov: we can use `:asc_nulls_last`: https://hexdocs.pm/ecto/Ecto.Query.html#order_by/3

PR #4798, opened 2025-08-28
Add first mcp prompt

tspenov:
## Checklist:

### DIFF
--- a/lib/sanbase/mcp/market_pulse_check_prompt.ex
+++ b/lib/sanbase/mcp/market_pulse_check_prompt.ex
@@ -1,0 +1,69 @@
+defmodule Sanbase.MCP.MarketPulseCheckPrompt do
+  @moduledoc """
+  Market Pulse Check - A comprehensive market health analysis prompt
+
+  Performs a comprehensive market health check by:
+  1. Start with trending_stories_tool (time_period='1h') to capture immediate sentiment
+  2. Identify the top 3 mentioned tokens from related_tokens
+  3. For each token, fetch price_usd and volume_usd metrics
+  4. Compare current sentiment ratios against 30-day price trends
+  5. Summarize whether sentiment is leading or lagging price action
+
+  This gives a quick snapshot of market efficiency and potential opportunities.
+  """
+
+  use Anubis.Server.Component, type: :prompt
+
+  alias Anubis.Server.Response
+
+  schema do
+  end
+
+  @impl true
+  def get_messages(_params, frame) do
+    content = build_market_pulse_prompt()
+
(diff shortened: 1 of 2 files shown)


lib/sanbase/mcp/market_pulse_check_prompt.ex
```diff
@@ -0,0 +1,69 @@
+defmodule Sanbase.MCP.MarketPulseCheckPrompt do
+  @moduledoc """
+  Market Pulse Check - A comprehensive market health analysis prompt
+
+  Performs a comprehensive market health check by:
+  1. Start with trending_stories_tool (time_period='1h') to capture immediate sentiment
```
IvanIvanoff, line 6: Maybe 24h?
Not sure what is the lag, but it could be the case that there are no trending stories in the last 1h. The last 24h should better tackle the current hype.

IvanIvanoff, line 6: Hmm, should we also expose trending words/trending projects?

We could have a different prompt `Analyze projects that are trending in the social media now`?

lib/sanbase/mcp/market_pulse_check_prompt.ex
```diff
@@ -0,0 +1,69 @@
...
+    - volume_usd (30-day historical data)
+
+    ## Step 4: Cross-Reference Analysis
+    Compare current sentiment ratios vs recent price movements and volume patterns.
+
+    ## Step 5: Market Efficiency Summary
+    Provide insights on sentiment leading vs lagging price action, opportunities, and risks.
+
+    End your response with: "DYOR"
+    """
```
IvanIvanoff, line 67: We have 2 options to remove the duplication from the moduledoc and here. I think in the long term it would be much better to not maintain both places.

Option 1: Put the text in a module attribute **before** `@moduledoc` and interpolate that module attribute in the moduledoc and just return it here. Credo does complain about that.

Option 2: Create a separate module (could a nested module in here called `PromptText` which has a single function that returns the text. Then invoke this function in the moduledoc: `@moduledoc "#{PromptText.text}` and call it from here, too.

lib/sanbase/mcp/market_pulse_check_prompt.ex
tspenov: `1h` is what the frontend is using but yeah, 24h might be more stable.

lib/sanbase/mcp/market_pulse_check_prompt.ex
tspenov: I don't think we need the whole prompt in the description (not even the the list of steps, just description of the prompt does in a more general way) Also this prompt contains only the steps but maybe it should contain also other sections that cover how our tools can be used efficiently

PR #4865, opened 2025-09-30
Add tracing with langfuse

tspenov:
## Changes

image
image
image

## Checklist:


lib/openai/tracing.ex
```diff
@@ -0,0 +1,247 @@
...
+        :ok
+    end
+  end
+
+  defp update_trace_output(%Trace{} = trace, answer) do
+    Logger.debug("Updating trace #{trace.id} with output")
+
+    trace_event = %{
+      "type" => "trace-create",
+      "id" => UUID.uuid4(),
```
IvanIvanoff, line 144: Instead of adding new dependencies, we can use `Ecto.UUID.generate()`

PR #5120, opened 2026-04-16
Remove outdated signals, update project_in_trends, add total_liquidations

alresharin:
## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added anomaly signals for trending projects and total liquidations.

* **Updates**
  * Consolidated and simplified MVRV zone indicators, reducing supported time ranges and pruning legacy signals to streamline the catalog.

* **Tests**
  * Updated test fixtures and expectations to reflect the revised signal catalog and new anomaly signal names.


lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -329,8 +42,8 @@
     }
   },
   {
-    "human_readable_name": "Project In Trending Words",
-    "name": "project_in_trending_words",
+    "human_readable_name": "Anomaly: Project In Trending Words",
+    "name": "anomaly_project_in_trending_words",
```
IvanIvanoff, line 46: Note this is a breaking change -- we rename an existing signal.
The name of the signal in the DB is unchanged -- `project_in_trends` and we only change the public name, so all existing references to this will be broken. Is this expected?

alresharin: It is expected that we need to change public name for the project_in_trends signal that we had before, yes. As for references, which ones do you mean?

IvanIvanoff: By `reference` I mean existing code/scripts/etc. which have this name hardcoded.
But looking at the usage - it was not called by clients recently, so it should be fine renaming it.

IvanIvanoff (approved):

PR #5124, opened 2026-04-22
Allow admin-configurable metric availability lookback per user

IvanIvanoff:
## Changes

Add a nullable `available_metrics_lookback_days` integer column on users
(admin-editable only, not exposed via GraphQL). When set, it overrides
the hardcoded 14-day `end_dt` cutoff used by `available_metrics` and
`available_slugs_for_metric` ClickHouse queries, letting internal users
see metrics whose released version is stale but whose experimental
version is fresh.

The value is threaded through `Metric.available_metrics_for_selector/2`
as the `lookback_days` opt and included in the per-user cache keys
(both in the resolver and in the inner module-level cache) so different
lookback windows don't collide. Defaults to `@default_lookback_days 14`
when nil.

`get_available_metrics` is intentionally untouched — it returns the full
universe of metrics, not a per-user view.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Configurable lookback period for metric availability (default 14 days); per-user lookback influences metric results and caching.
  * Admin UI updated to manage users' lookback setting.

* **Bug Fixes**
  * Validation added to ensure lookback is a positive integer (<= 7300).
  * Cache keys and behavior adjusted so results vary correctly by user/lookback and handle slow upstream timeouts with clearer error text.

* **Tests**
  * Added tests for default/custom lookback propagation and selector/contract forwarding.

* **Chores**
  * Database schema updated to include new nullable lookback column.

### DIFF
--- a/lib/sanbase/accounts/user.ex
+++ b/lib/sanbase/accounts/user.ex
@@ -194,7 +194,10 @@ defmodule Sanbase.Accounts.User do
     |> unique_constraint(:twitter_id)
     |> validate_inclusion(:metric_access_level, @allowed_access_levels)
     |> validate_inclusion(:feature_access_level, @allowed_access_levels)
-    |> validate_number(:available_metrics_lookback_days, greater_than: 0)
+    |> validate_number(:available_metrics_lookback_days,
+      greater_than: 0,
+      less_than_or_equal_to: 7300
+    )
   end
 
   def san_balance(user), do: __MODULE__.SanBalance.san_balance(user)
--- a/lib/sanbase/accounts/user.ex
+++ b/lib/sanbase/accounts/user.ex
@@ -97,6 +97,7 @@ defmodule Sanbase.Accounts.User do
 
     field(:metric_access_level, :string, default: "released")
     field(:feature_access_level, :string, default: "released")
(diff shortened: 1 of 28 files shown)


tspenov (approved):

lib/sanbase/accounts/user.ex
```diff
@@ -97,6 +97,7 @@ defmodule Sanbase.Accounts.User do
 
     field(:metric_access_level, :string, default: "released")
     field(:feature_access_level, :string, default: "released")
+    field(:available_metrics_lookback_days, :integer)
```
YordanPavlov, line 100: Shouldn't the default of 14 live here as the ones above?

YordanPavlov (approved):

lib/sanbase/accounts/user.ex
IvanIvanoff: This is added as nullable column and the default is used when the field is null (default):
```elixir
lookback_days = Keyword.get(opts, :lookback_days) || @default_lookback_days
```
This allows us to change the default if we want to by changing 1 line of code instead of updating the records of thousands of users.

PR #5136, opened 2026-05-05
Fix projects by market segment count

tspenov:
## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Market segment results now properly filter out hidden projects and those missing required identifying information, ensuring only valid projects appear in the data.


IvanIvanoff (approved):
LGTM,
I don't know if we cannot preload only those projects by using a preload query? Can check later

PR #5146, opened 2026-05-11
Add eth_whale_dump, hyperliquid_avg_funding_rate signals

alresharin:
## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added "Anomaly: ETH Whale Dump" and "Anomaly: Hyperliquid AVG Funding Rate" signals — free, timeseries signals with 5‑minute minimum interval, count aggregation, and listed in the signals table

* **Tests**
  * Test expectations updated to include the two new anomaly signals

[Review Change Stack](https://app.coderabbit.ai/change-stack/santiment/sanbase2/pull/5146)

### DIFF
--- a/lib/sanbase/signal/signal_files/available_signals.json
+++ b/lib/sanbase/signal/signal_files/available_signals.json
@@ -70,5 +70,35 @@
       "SANAPI": "free",
       "SANBASE": "free"
     }
+  },
+  {
+    "human_readable_name": "Anomaly: ETH Whale Dump",
+    "name": "anomaly_eth_whale_dump",
+    "signal": "eth_whale_dump",
+    "access": "free",
+    "selectors": ["slug"],
+    "aggregation": "count",
+    "min_interval": "5m",
+    "table": "signals",
+    "data_type": "timeseries",
+    "min_plan": {
+      "SANAPI": "free",
+      "SANBASE": "free"
+    }
+  },
+  {
+    "human_readable_name": "Anomaly: Hyperliquid AVG Funding Rate",
+    "name": "anomaly_hyperliquid_avg_funding_rate",
+    "signal": "hyperliquid_avg_funding_rate",
+    "access": "free",
+    "selectors": ["slug"],
+    "aggregation": "count",
+    "min_interval": "5m",
+    "table": "signals",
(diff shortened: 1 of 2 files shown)


lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -70,5 +70,35 @@
...
+    "access": "free",
+    "selectors": ["slug"],
+    "aggregation": "count",
+    "min_interval": "5m",
+    "table": "signals",
+    "data_type": "timeseries",
+    "min_plan": {
+      "SANAPI": "free",
+      "SANBASE": "free"
+    }
```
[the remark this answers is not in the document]
IvanIvanoff: @coderabbitai This is not computed by the backend in this repo, but is populated by another team using AF/Clickhouse.
This is not a mistake,

lib/sanbase/signal/signal_files/available_signals.json
```diff
@@ -70,5 +70,35 @@
...
+  },
+  {
+    "human_readable_name": "Anomaly: Hyperliquid AVG Funding Rate",
+    "name": "anomaly_hyperliquid_avg_funding_rate",
+    "signal": "hyperliquid_avg_funding_rate",
+    "access": "free",
+    "selectors": ["slug"],
+    "aggregation": "count",
```
IvanIvanoff, line 95: The metric is `AVG funding rate`, so it does not seem appropriate to aggregate multiple such values with `count` - this will totally erase any data and will just return the number of aggregated rows.
Better use `avg` or `last` here

alresharin: Where the aggregation is used? For getSignal function?

alresharin: As far as the anomaly indicated both lower and higher outliers of the metric, i don't think it is worth to place avg aggregation instead of count. I think it is quite good to see amount of outliers per period of time, and avg aggregation may show senseless numbers for some periods of time. The exact numbers can be fetched by getRawSignals if necessary

IvanIvanoff: Ok, let's leave it with count then

